PR #1, opened 2018-08-31
Add SSH_AUTH_SOCK

mscherer:
As I sometime need to use ssh (either for a git clone or ansible), it is useful to have access to the agent.


debarshiray:
I ran into this in the past. Thanks for fixing it, misc!

Pushed as commit 0dedc0140d80ba9f after reformatting the commit message to fit within 72 characters.

debarshiray:
Oops, sorry, it got confused when I changed the commit message.

PR #24, opened 2018-12-07
add sudo to help txt

johrstrom:
Hi, while reading through the file and trying to understand sudo and sudoless I noticed the --help text doesn't mention the --sudo option.  I've also added the -s short option for --sudo.

### DIFF
--- a/fedora-toolbox
+++ b/fedora-toolbox
@@ -234,11 +234,13 @@ usage()
     echo "Usage: fedora-toolbox [--container <name>]"
     echo "                      [--release <release>]"
     echo "                      [-v | --verbose]"
+    echo "                      [-s | --sudo]"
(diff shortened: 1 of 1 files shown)


debarshiray:
Yes, it was deliberately left out of the `--help` output because (a) so far it has only been for debugging rootless `buildah` and `podman` for those will to dig into the code; and (b) one has to punch a hole through `sudoers` to avoid all the annoying password prompts which are not meant to be set up in production.

However, things are going to change as we converge with CoreOS.

debarshiray:
Sadly, I will have to turn this patch down, and suggest that we wait and see where CoreOS takes us.

johrstrom:
Fine by me, thanks for the explaination.


-------- Original message --------

Issue #28, opened 2019-01-06
Enter the toolbox as non root, self user


mildred:
Running fedora-toolbox 0.0.2-1.fc29, when I'm in the toolbox, I am root.

When I'm working, I'm mostly working as a simple user, and I'd like to enter the toolbox as my own user and not have to use `su` after entering the toolbox. Also, I had to manually set up my own user in /etc/passwd inside the toolbox on first use.

Additionally, it would be nice if I could run specific commands after `toolbox enter`. For example, being able to run `toolbox enter su $USER` would help my use case (I could just create an alias then)

debarshiray:
> Running fedora-toolbox 0.0.2-1.fc29, when I'm in the toolbox, I am root.

Oh, really? That's surprising. We definitely try to retain the same user inside the container as on the host. Grep for `$USER` and `$UID` inside the `fedora-toolbox` script. Also, see commit cfcf4eb31e14b3a

Do you have any fancy or non-standard set-up? For example, are you logging via LDAP or Kerberos, as opposed to having a simple local user?

debarshiray:
> Additionally, it would be nice if I could run specific commands after
> toolbox enter

Yes, this does sound like a good thing to have. I wonder if `run` would be a better name for the command in that case. It could also just be an alias for `enter`.

mildred:
The problem I had (when not specifying `--image`) is:

```
$ fedora-toolbox --verbose  --container test create
error looking up container "test": no container with name or ID test found: no such container
Trying to pull docker.io/fedora-toolbox-mildred:29...Failed
Trying to pull registry.fedoraproject.org/fedora-toolbox-mildred:29...Failed
Trying to pull quay.io/fedora-toolbox-mildred:29...Failed
Trying to pull registry.access.redhat.com/fedora-toolbox-mildred:29...Failed
Trying to pull registry.centos.org/fedora-toolbox-mildred:29...Failed
unable to pull fedora-toolbox-mildred:29: 5 errors occurred:

* Error determining manifest MIME type for docker://fedora-toolbox-mildred:29: Error reading manifest 29 in docker.io/library/fedora-toolbox-mildred: errors:
denied: requested access to the resource is denied
unauthorized: authentication required

* Error determining manifest MIME type for docker://registry.fedoraproject.org/fedora-toolbox-mildred:29: Error reading manifest 29 in registry.fedoraproject.org/fedora-toolbox-mildred: manifest unknown: manifest unknown
[...]

Issue #20, opened 2018-12-02
Make "enter" staying in current pwd


aengelke:
It'd be cool if `fedora-toolbox enter` would stay in the current working directory (if available in the container, e.g. `~/**`). Currently, `enter` spawns a new shell with the cwd being the home directory, requiring to `cd` again.

debarshiray:
Yes, I agree that we should do this. A few other people have asked for it too, and I personally feel the need for it while using `fedora-toolbox`.

debarshiray:
One option is to add a `--workdir` option to `podman exec` and have it override the image's default working directory before executing the shell.

I learnt in `#podman` that `docker` already has such an option, so adding it will also help with compatibility.

debarshiray:
`podman exec` [now has a --workdir option](https://github.com/containers/libpod/pull/2097) for specifying the working directory.

PR #30, opened 2019-01-07
Toolbox improvements

mildred:
Included in this PR:

- entering the same directory inside the toolbox than outside if possible, without podman support, #20
- allow to specify a command to execute on enter other than the shell
- Add `--user` CLI flag to enter the toolbox as the same user, #28

### DIFF
--- a/fedora-toolbox
+++ b/fedora-toolbox
@@ -209,6 +209,7 @@ enter()
             --tty \
             $user_flag \
             $toolbox_container \
+            /bin/sh -c 'cd "$0"; exec "$@"' "$PWD" \
             $shell_to_exec -l 2>&42
 )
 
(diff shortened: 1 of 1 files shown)


tosmi:
+1 for allowing to run a specific command with toolbox. another way to implement this could be a a `run` command. e.g. 

```
$ toolbox run emacs
```

starting emacs/vlc/you name it directly without entering first, would be a real plus. 

my personal workflow would be to create a container with all my dev tools installed and just starting emacs within the container (maybe via a systemd user service). 

if a `run` command is preferred, i could try to implement this and with some help try to get this merged.

thanks
toni

debarshiray:
Yeah, I am leaning towards using a separate `run` command for this because it seems odd to say *enter emacs*. If we go that way, `enter` can become a special-case of `run`!

Since #28 is now solved, the part about specifying the user in this pull request is no longer necessary. So let's close this and continue on #76

PR #34, opened 2019-01-10
Set up $HOME and /home to match the host

owtaylor:
Silverblue, and rpm-ostree more generally are moving to
HOME=/var/home/$USER and make the /home symlink just a compatibility
feature. See:

 https://github.com/projectatomic/rpm-ostree/pull/1726

Matching what the host does will reduce weird side-effects, so
propagate $HOME into the container, and if the host has /home as
a symlink to /var/home, do the same for the toolbox.

### DIFF
--- a/fedora-toolbox
+++ b/fedora-toolbox
@@ -38,6 +38,12 @@ create()
     tmpfs_size=$((64 * 1024 * 1024)) # 64 MiB
     working_container_name="fedora-toolbox-working-container-$(uuidgen --time)"
 
+    if [[ "$(readlink /home)" == var/home ]] ; then
+        need_home_link=true
+    else
+        need_home_link=false
+    fi
+
     if [ "$DBUS_SYSTEM_BUS_ADDRESS" != "" ]; then
         dbus_system_bus_address=$DBUS_SYSTEM_BUS_ADDRESS
     fi
@@ -65,7 +71,17 @@ create()
 
         echo "$0: trying to configure working container $working_container_name" >&42
 
+        if $need_home_link ; then
+            if ! $prefix_sudo buildah run $working_container_name -- \
+                    sh -c 'rmdir /home && mkdir -m 0755 -p /var/home && ln -s var/home /home' 2>&42; then
+                $prefix_sudo buildah rm $working_container_name >/dev/null 2>&42
+                echo "$0: failed to make /home a symlink"
+                exit 1
+            fi
+        fi
+
         if ! $prefix_sudo buildah run $working_container_name -- useradd \
+                     --home-dir $HOME \
                      --no-create-home \
                      --shell $SHELL \
                      --uid $UID \


fedora-toolbox
```diff
@@ -56,7 +64,17 @@ create()
             fi
         fi
 
+        if $need_home_link ; then
+            if ! $prefix_sudo buildah run $working_container_name -- \
+                    sh -c 'rmdir /home && mkdir -m 0755 -p /var/home && ln -s var/home /home' 2>&42; then
+                $prefix_sudo buildah rmi $working_container_name >/dev/null 2>&42
```
debarshiray: It should be `buildah rm`. It was a typo that was fixed in commit 62e20f9df3bea846. Sorry about that.

debarshiray:
I took the liberty to resolve the conflict, and changed `buildah rmi` to `buildah rm`.

fedora-toolbox
```diff
@@ -65,7 +73,17 @@ create()
...
+            if ! $prefix_sudo buildah run $working_container_name -- \
+                    sh -c 'rmdir /home && mkdir -m 0755 -p /var/home && ln -s var/home /home' 2>&42; then
+                $prefix_sudo buildah rm $working_container_name >/dev/null 2>&42
+                echo "$0: failed to make /home a symlink"
+                exit 1
+            fi
+        fi
+
         if ! $prefix_sudo buildah run $working_container_name -- useradd \
+                     --home-dir $HOME \
```
debarshiray: This changes the value of `$HOME` to be `/var/home/user` instead of `/home/user`. I see that until Silverblue 29, `$HOME` is still pointing to the `/home/user` symbolic link on the host. Has that changed with newer `rpm-ostree` or Silverblue?

debarshiray: Summarizing discussion with @owtaylor in #silverblue on Freenode:

It's better to retain the exact same `$HOME` string inside the toolbox as is on the host. Otherwise, the bash and zsh prompts get confused, because the current working directory no longer matches `$HOME` inside the toolbox. eg., `fedora-toolbox enter` inside the host's $HOME will lead to a confused bash prompt, and similarly $HOME/sub-directory for zsh.

This is relevant because even if `rpm-ostree` has switched the default in `/etc//default/useradd` to be `/var/home`, that change won't affect existing users, nor will it affect users created from Anaconda which might be using a different `/etc/default/useradd`.

debarshiray: Also, the actual relevant line is not this, but this one further above:
```
HOME="$(readlink $HOME --canonicalize)"
```

debarshiray:
Merged after testing on a few different set ups:

- /bin/bash as default shell, `/home` in  `/etc/default/useradd`, symlinked `/home`
- /bin/zsh as default shell, `/home` in  `/etc/default/useradd`, symlinked `/home`
- /bin/bash as default shell, `/var/home` in  `/etc/default/useradd`, symlinked `/home`
- /bin/bash as default shell, `/home` in  `/etc/default/useradd`, non-symlinked `/home`

I hope I got it right. Thanks for the help @owtaylor

PR #39, opened 2019-01-16
Add list command

ozoder:
Add the list command providing the possibility to list all images and containers created by fedora-toolbox (i.e those that have the "com.redhat.component=fedora-toolbox" label). Two flags can be used: --images and --containers, to list only images or containers. If none of these flags is used, the command lists both. Additionally, containers that are currently running are highlighted by green color.

### DIFF
--- a/fedora-toolbox
+++ b/fedora-toolbox
@@ -27,6 +27,11 @@ toolbox_prompt="🔹[\u@\h \W]\\$ "
 verbose=false
 
 
+LGC='\033[1;32m' # Light Green Color
+LBC='\033[1;34m' # Light Blue Color
+NC='\033[0m' # No Color
+
+
 has_prefix()
 (
     str="$1"
@@ -418,6 +423,54 @@ enter()
 )
 
 
+list_images()
+(
+    output=$($prefix_sudo podman images \
+                     --filter "label=com.redhat.component=fedora-toolbox" \
+                     --format "{{.ID}} {{.Repository}}:{{.Tag}} {{.Created}}" 2>&42 \
+             | sed "s/ \{2,\}/\t/g" 2>&42 \
+             | column --separator $'\t' --table --table-columns "IMAGE ID,IMAGE NAME,CREATED")
+
+    if [ "$output" != "" ]; then
+        echo -e "${LBC}Images created by fedora-toolbox${NC}"
+        echo "$output"
+    fi
+)
+
+
+list_containers()
+(
+    output=$($prefix_sudo podman ps \
+                     --all \
+                     --filter "label=com.redhat.component=fedora-toolbox" \
+                     --format "{{.ID}}  {{.Names}}  {{.Created}}  {{.Status}}  {{.Image}}" 2>&42 \
+             | sed "s/ \{2,\}/\t/g" 2>&42 \
+             | column \
+                       --separator $'\t' \
+                       --table \
+                       --table-columns "CONTAINER ID,CONTAINER NAME,CREATED,STATUS,IMAGE NAME" 2>&42)
+
+    if [ "$output" != "" ]; then
+        echo -e "${LBC}Containers created by fedora-toolbox${NC}"
+        echo "$output" | head --lines 1 2>&42
+        (
+            echo "$output" | tail --lines +2 2>&42
+        ) | (
+            unset IFS
+            while read container; do
+                container_id=$(echo $container | cut --delimiter " " --fields 1 2>&42)
+                is_running=$($prefix_sudo podman inspect $container_id --format "{{.State.Running}}" 2>&42)
+                if $is_running; then
+                    echo -e "${LGC}$container${NC}"
+                else
+                    echo "$container"
+                fi
+            done
+        )
+    fi
+)
+
+
 exit_if_extra_operand()
 {
     if [ "$1" != "" ]; then
@@ -447,6 +500,8 @@ usage()
     echo "                      [--release <release>]"
     echo "                      [-v | --verbose]"
     echo "                      enter"
+    echo "   or: fedora-toolbox list [-c | --containers]"
+    echo "                           [-i | --images]"
     echo "   or: fedora-toolbox --help"
 }
 
@@ -550,6 +605,33 @@ case $op in
         done
         exit_if_extra_operand $1
         enter
+        exit
+        ;;
+    list )
+        ls_images=false
+        ls_containers=false
+        while has_prefix "$1" -; do
+            case $1 in
+                -c | --containers )
+                    ls_containers=true
+                    ;;
+                -i | --images )
+                    ls_images=true
+                    ;;
+                * )
+                    exit_if_unrecognized_option $1
+            esac
+            shift
+        done
+        exit_if_extra_operand $1
+
+        if ! $ls_containers && ! $ls_images; then
+            ls_containers=true
+            ls_images=true
+        fi
+
+        $ls_images && list_images
+        $ls_containers && list_containers
         exit
         ;;
     * )


debarshiray:
@owtaylor do you have any comments about the formatting of the output, etc.?

fedora-toolbox
```diff
@@ -246,6 +250,47 @@ enter()
...
+        else
+            echo "$output"
+        fi
+    fi
+    if [ $2 = 1 ]; then
+        echo -e "${LBC}Containers created by fedora-toolbox (currently running containers in green)${NC}"
+        output=$($prefix_sudo podman ps \
+                --all \
+                --filter "label=com.redhat.component=fedora-toolbox" \
+                --format "table {{.ID}} {{.Names}} {{.Image}} {{.Created}} {{.Status}}")
```
debarshiray: Which version of `podman` were you developing against?

I tried `podman-0.12.1.2-1.git9551f6b` and `podman-1.0.0-1.git82e8011`, and `podman ps` is not handling `table` at all, unlike `podman images`.

For example:
```
$ podman ps --all --filter "label=com.redhat.component=fedora-toolbox" --format "table {{.ID}} {{.Names}} {{.Image}} {{.Created}} {{.Status}}"
table 21f509790423 fedora-toolbox-rishi:29 localhost/fedora-toolbox-rishi:29 46 hours ago Created
table c86c1d455331 fedora-toolbox-rishi:28 localhost/fedora-toolbox-rishi:28 9 days ago Created

$ podman images --filter "label=com.redhat.component=fedora-toolbox" --format "table {{.ID}} {{.Repository}} {{.Tag}} {{.Created}}"
IMAGE ID       REPOSITORY                                      TAG   CREATED
ec61775deac1   localhost/fedora-toolbox-rishi                  29    46 hours ago
45b540de989b   localhost/fedora-toolbox-rishi                  28    9 days ago
...
```

ozoder: I was using podman version 0.10.1, as I did not update my Silverblue in VM, my bad. However, when using the --format flag, podman ps omits table headings and I don't see any other flag that would print them anyway. For now, we might need to print out the output with the default columns.

debarshiray: Ok. I see that you already filed https://github.com/containers/libpod/issues/2221 for the `podman` regression.

fedora-toolbox
```diff
@@ -246,6 +250,49 @@ enter()
...
 
 
+list()
+(
+    to_print=""
+    if [ $1 = 1 ]; then
+        to_print="$to_print${LBC}Images created by fedora-toolbox${NC}\n"
+        output=$($prefix_sudo podman images \
+                --filter "label=com.redhat.component=fedora-toolbox")
+                #--format "table {{.ID}} {{.Repository}} {{.Tag}} {{.Created}}")
```
debarshiray: `podman image` is strange in that it doesn't put the name and the tag together separated by a colon. Every time I try to copy-paste the name of an image, I need to separately select the tag, which gets just a little bit annoying after some time. :)

It's also weird that it calls it "repository", not "name".

So, to have ultimate control over our table formatting we could do something like this instead:
```
podman images --filter "label=com.redhat.component=fedora-toolbox" --format "{{.ID}} {{.Repository}}:{{.Tag}} {{.Created}}" | sed 's/ \{2,\}/\t/g' | column --separator $'\t' --table --table-columns "IMAGE ID,IMAGE NAME,CREATED"
```

We need to closely look at what podman is doing to get the column separation right, but maybe it could work?

fedora-toolbox
```diff
@@ -246,6 +250,49 @@ enter()
...
+        if [[ $output = "" ]]; then
+            to_print="${to_print}No images available\n"
+        else
+            to_print="$to_print$output\n"
+        fi
+    fi
+    if [ $2 = 1 ]; then
+        to_print="$to_print${LBC}Containers created by fedora-toolbox (currently running containers in green)${NC}\n"
```
debarshiray: We should probably not say green, because the visible colour changes completely based on what terminal theme is being used. For example, with Solarized it turns grey.

fedora-toolbox
```diff
@@ -379,6 +428,30 @@ case $op in
...
+                    exit_if_unrecognized_option $1
+            esac
+            shift
+        done
+        exit_if_extra_operand $1
+        if [ $list_images = 0 ] && [ $list_containers = 0 ]; then
+            list_images=1
+            list_containers=1
+        fi
+        list "$list_images" "$list_containers"
```
debarshiray: I wonder if it will be more readable to have two separate `list_images` and `list_containers` functions.

fedora-toolbox
```diff
@@ -246,6 +250,49 @@ enter()
...
+        fi
+    fi
+    if [ $2 = 1 ]; then
+        to_print="$to_print${LBC}Containers created by fedora-toolbox (currently running containers in green)${NC}\n"
+        output=$($prefix_sudo podman ps \
+                --all \
+                --filter "label=com.redhat.component=fedora-toolbox")
+                #--format "table {{.ID}} {{.Names}} {{.Image}} {{.Created}} {{.Status}}")
+        if [[ $(echo "$output" | wc -l) -lt 2 ]]; then
+            to_print="${to_print}No containers available\n"
```
debarshiray: We should probably just not output anything if there is nothing there.

fedora-toolbox
ozoder: This does seem better than just the default output by podman, since for example podman rmi takes image name as an argument. I added this change, but leaving default output for containers until the fix for [podman#2221](https://github.com/containers/libpod/issues/2221) gets released.

debarshiray: Thanks!

We might just be able to avoid using `table` at all for `podman ps`, if use `column` to create our own table.

fedora-toolbox
```diff
@@ -246,6 +250,66 @@ enter()
...
 
 
+list_images()
+(
+    output=$($prefix_sudo podman images \
+            --filter "label=com.redhat.component=fedora-toolbox" \
+            --format "{{.ID}} {{.Repository}}:{{.Tag}} {{.Created}}" \
+            | sed 's/ \{2,\}/\t/g' | column --separator $'\t' --table \
+            --table-columns "IMAGE ID,IMAGE NAME,CREATED")
+    if ! [[ $output = "" ]]; then
```
debarshiray: Would be better to use single brackets and put quotes around `$output` instead for POSIX correctness.

fedora-toolbox
```diff
@@ -246,6 +250,66 @@ enter()
...
+            line_id="$(cut -d' ' -f1 <<< $line)"
+            if ! [[ $line_id = "CONTAINER" ]]; then
+                inspect_out=$($prefix_sudo podman inspect $line_id --format "{{.State.Running}}")
+            fi
+            if [[ $inspect_out = "true" ]]; then
+                to_print="$to_print${LGC}$line${NC}\n"
+            else
+                to_print="$to_print$line\n"
+            fi
+        done <<< "$output"
```
debarshiray: Sadly, the `<<<` construct isn't in POSIX, so we need to pipe the string into the `while` loop.

fedora-toolbox
```diff
@@ -246,6 +250,66 @@ enter()
...
+(
+    images_output=""
+    containers_output=""
+    if [ $1 = 1 ]; then
+        images_output="$(list_images)"
+    fi
+    if [ $2 = 1 ]; then
+        containers_output="$(list_containers)"
+    fi
+    if ! [[ $images_output = "" ]] && ! [[ $containers_output = "" ]]; then
```
debarshiray: Umm... can't we just directly call `list_images` and `list_containers`?

fedora-toolbox
```diff
@@ -246,6 +250,66 @@ enter()
...
+            --filter "label=com.redhat.component=fedora-toolbox")
+            #--format "table {{.ID}} {{.Names}} {{.Created}} {{.Status}} {{.Image}}")
+    if ! [[ $(echo "$output" | wc -l) -lt 2 ]]; then
+        to_print="${LBC}Containers created by fedora-toolbox${NC}\n"
+        while read -r line; do
+            if [[ $line = "" ]]; then
+                break
+            fi
+            line_id="$(cut -d' ' -f1 <<< $line)"
+            if ! [[ $line_id = "CONTAINER" ]]; then
```
debarshiray: It might slightly cleaner to use `tail` to skip the header. :)

fedora-toolbox
debarshiray: Changed.

fedora-toolbox
debarshiray: Done.

fedora-toolbox
debarshiray: Done.

fedora-toolbox
debarshiray: Done.

debarshiray:
I took the liberty to make the above changes for POSIX compatibility, and was able to use `column` to make `podman ps` work without `table`. Looks like it works with both `podman.git` master and `podman-1.0.0`.

PR #43, opened 2019-01-26
README.md: Rework in various ways

cgwalters:
- OSTree systems aren't fully immutable, you *can* install things
   we just discourage it
 - fedora-toolbox can (and should!) be used on non-ostree based
   systems as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,22 @@
-# Fedora Toolbox — Hacking on OSTree-based Fedoras
+# Fedora Toolbox — Unprivileged development environment
 
 [Fedora Toolbox](https://github.com/debarshiray/fedora-toolbox) is a tool that
 offers a familiar RPM based environment for developing and debugging software
-on locked down [OSTree](https://ostree.readthedocs.io/en/latest/) based Fedora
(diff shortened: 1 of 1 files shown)


debarshiray:
Thanks, @cgwalters 

Merged after sprinkling some Unicode em-dashes and hyperlinking to https://podman.io/.

README.md
```diff
@@ -1,15 +1,22 @@
...
-operating systems are shipped as *immutable* OSTree images, where it's
-difficult to setup a development environment with your favorite tools, editors
-and SDKs. A toolbox container solves that problem by providing a RPM based
-*mutable* container. You can tweak it to your heart's content and use DNF to
-install your favorite packages, all without worrying about breaking your
-operating system.
+that runs fully unprivileged using [Podman](https://podman.io/).
+
+The toolbox container is a fully *mutable* container; when you see
+`yum install ansible` for example, that's something you can do inside your
```
debarshiray: I wonder if we should be using `dnf` instead of `yum`, though. Especially for a new project that's barely at the pre-alpha stage. :)

debarshiray: Ah, ok. I didn't know about the RHEL 8 and Fedora bits.

PR #45, opened 2019-01-28
Add rm and rmi commands

ozoder:
Add rm and rmi commands for images/containers removal. Both commands take ID or name as an argument and if the specified image/container was created by fedora-toolbox, it's removed. 

Both commands have two flags: --all and --force. With the --all flag, the argument is not needed, as the command tries to remove all fedora-toolbox images/containers. The --force flag just runs force removal (i.e podman rm/rmi --force <ID>)

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -82,6 +82,26 @@ has_prefix()
 )
 
 
+has_substring()
+(
+    haystack="$1"
+    needle="$2"
+
+    ret_val=1
+
+    case "$haystack" in
+        *"$needle"* )
+            ret_val=0
+            ;;
+        * )
+            ret_val=1
+            ;;
+    esac
+
+    return $ret_val
+)
+
+
 is_integer()
 {
     [ "$1" != "" ] && [ $1 -eq $1 ] 2>&3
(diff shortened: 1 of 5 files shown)


fedora-toolbox
```diff
@@ -262,6 +264,64 @@ enter()
...
+            echo "$0: failed to remove $3"
+            exit 1
+        fi
+    else
+        if ! rm_output=$($prefix_sudo podman $2 $3 2>&42); then
+            echo "$0: failed to remove $3"
+            exit 1
+        fi
+    fi
+    if ! [[ $rm_output = "" ]]; then
```
debarshiray: Nitpick: would be better to use `[ "$rm_output" != "" ]` instead for POSIX compatibility.

fedora-toolbox
```diff
@@ -262,6 +264,64 @@ enter()
...
+        echo -e "${LRC}$4 $rm_output removed${NC}"
+    fi
+)
+
+
+remove()
+(
+    if [ $4 = "rm" ]; then
+        keyword="container"
```
debarshiray: I wonder if this will be better to split this into separate `remove_images` and `remove_containers` functions. It's nice that you have managed to avoid duplicated code by factoring out the common parts, but it makes the final commands slightly harder to read. Given that we are dealing with shell scripts, it might be better to optimize for readability. I don't know.

At least, I'd suggest copying the function parameters (ie. $1, $2, etc.) into variables with more descriptive names at the top of the function.

fedora-toolbox
```diff
@@ -395,6 +461,36 @@ case $op in
         enter
         exit
         ;;
+    rm | rmi )
+        force=0
+        all=0
+        while [[ "$1" = -* ]]; do
```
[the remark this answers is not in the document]
ozoder: Done

fedora-toolbox
ozoder: Done

fedora-toolbox
ozoder: There are now two separate functions - remove_image and remove_container. Function parameters are also copied to variables for better readability.

ozoder:
I've added the possibility to remove multiple images/containers at once. Toolbox rm and rmi commands can now take more than one name/ID, the same way podman rm and rmi commands do it.

matthiasclasen:
I think long names like 'delete' and 'delete-image' would go better with the existing create/list/enter

debarshiray:
> I think long names like 'delete' and 'delete-image' would go better with the existing create/list/enter

I have been going back and forth on that myself. We can definitely add the longer names as aliases. eg., `ls` as a shorter alias for `list` seems very reasonable.

At the same time `toolbox rm` is such a very thin wrapper around `podman rm` that seems sensible to use the `podman` nomenclature for the short form. It also conveniently matches the `rm(1)` command.

toolbox
```diff
@@ -638,6 +744,42 @@ case $op in
         $ls_containers && list_containers
         exit
         ;;
+    rm | rmi )
```
debarshiray: We need to plug in the `forward_to_host` thing that we added in commit 5b3d234c9e9ef45f.

toolbox
```diff
@@ -638,6 +744,42 @@ case $op in
...
+        force=false
+        all=false
```
debarshiray: Since this is a shell script, and all variables are global, it would be nicer to namespace these variables as `rm_all`, `rm_force`, etc. to prevent them from accidentally leaking into another context and suprising us.

toolbox
```diff
@@ -638,6 +744,42 @@ case $op in
...
+            id=""
+            exit_if_extra_operand $1
+        elif ! $all && [ "$1" = "" ]; then
+            echo "$0: missing argument for '$op'"
+            echo "Try '$0 --help' for more information."
+            exit 1
+        else
+            id=""
+            while [ "$1" != "" ]; do
+                id="$id $1"
```
debarshiray: The variable with the list of IDs will have an extra preceding whitespace. We should  either avoid that in the first place or remove it after the loop to prevent weird behaviour inside the `remove_*` functions. Depending on exactly how you access the variable you either see the whitespace or you don't. This seems to matter when iterating over the IDs.

Shells are weird. It's better to be idiot-proof. :)

toolbox
```diff
@@ -477,6 +478,105 @@ list_containers()
...
+            exit 1
+        fi
+    else
+        if ! rm_output=$($prefix_sudo podman $command_to_run $id_or_name 2>&42); then
+            echo "$0: failed to remove $id_or_name"
+            exit 1
+        fi
+    fi
+    if ! [ "$rm_output" = "" ]; then
+        echo -e "${LRC}$keyword $rm_output removed${NC}"
```
debarshiray: Umm... I think that it's better to stay silent unless there has been an error, or the command explicitly demands some output (eg., `list`) or if it's a long running command that needs some progress indicator.

It seems like we need to find the right balance between the classical UNIX commands like `cp(1)`, `ls(1)`, `rm(1)`, etc. which tend to be more terse and things like `git(1)`, `podman(1)`, etc., which are more chatty. So far, we have been following the former, and we can revisit this if it turns out to be too unfriendly. But for the time being let's stay consistent.

toolbox
```diff
@@ -477,6 +478,105 @@ list_containers()
...
+        IFS=' '
+        export IFS
+        for id_or_name in $id_or_name_list; do
+            if ! inspect_out=$($prefix_sudo podman inspect $id_or_name --type image --format \"{{.Labels}}\" 2>&42); then
+                echo "$0: failed to inspect $id_or_name"
+                exit 1
+            fi
+            if ! [[ $inspect_out = *"com.redhat.component:fedora-toolbox"* ]]; then
+                echo "Not a fedora-toolbox image"
+                exit 1
```
debarshiray: We should probably match the internal behaviour of `podman rm` here. See `man podman-rm` or `man podman-rmi` [1].

In short, when multiple IDs or `--all` are passed, it goes through all of them and tries to remove as many as it can, even if some of them couldn't be removed. However, the exit code is only 0 if all the IDs were successfully removed. To do that, we need to use `return`, not `exit` here.

[1] There's a bug in Podman: https://github.com/containers/libpod/issues/2539

toolbox
```diff
@@ -477,6 +478,105 @@ list_containers()
...
+remove_image()
+(
+    id_or_name_list=$1
+    force_remove=$2
+    remove_all=$3
+
+    if ! $remove_all; then
+        IFS=' '
+        export IFS
+        for id_or_name in $id_or_name_list; do
```
debarshiray: POSIX doesn't support arrays other than the ones for command line parameters (ie., `$1`, `$2`, etc.). We can use a `while` loop with `read` and newline separated elements to work around that.

toolbox
```diff
@@ -477,6 +478,105 @@ list_containers()
...
+                echo "Not a fedora-toolbox container"
+                exit 1
+            fi
+            run_rm "$force_remove" "rm" "$id_or_name" "container"
+        done
+    else
+        if ! output=$($prefix_sudo podman ps --all \
+                --filter "label=com.redhat.component=fedora-toolbox" \
+                --format "{{.ID}}" 2>&42); then
+            echo "$0: failed to list all containers"
```
debarshiray: Using the entire `$0` seems too noisy. We should use just the `basename`, like we do elsewhere.

toolbox

We should probably not touch the current container and it's image when the command is invoked from inside a toolbox. Let's address that in a later PR.

toolbox
debarshiray: Done

toolbox

toolbox

toolbox

toolbox

toolbox

debarshiray:
I took the liberty to address the above nits.

PR #50, opened 2019-02-14
Shorten the prefix for debug and error messages

debarshiray:
Using "$0" leads to printing the entire path to the fedora-toolbox
script, which adds visual noise to the output. Command line tools like
GNU Coreutils and Git remove any leading directory components from
the path, which seems like a reasonable trade-off between aesthetics
and verbosity.

### DIFF
--- a/fedora-toolbox
+++ b/fedora-toolbox
@@ -16,6 +16,10 @@
 #
 
 
+exec 42>/dev/null
+
+base_toolbox_command=$(basename "$0" 2>&42)
+
 source /etc/os-release
 release=$VERSION_ID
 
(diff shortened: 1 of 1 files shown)


debarshiray:
From #silverblue on Freenode:

11:25 <rishi> I wonder if I had gotten netsplit out of the network.
11:25 <rishi> Hey ozoder
11:26 <rishi> I filed a PR to tweak the aesthetics of the toolbox messages a    
      bit: https://github.com/debarshiray/fedora-toolbox/pull/50
11:26 <rishi> Any testing, comments, etc. welcome.
11:26 <rishi> I started to get annoyed by the full path.
11:36 <ozoder> rishi yeah, it looks cleaner this way

PR #55, opened 2019-02-18
Include missing man-pages

ozoder:
Currently, some of the packages inside toolbox images don't have man-pages, e.g. openssl or bash. 
To include them, all those packages need to be reinstalled. This is done by creating a list of packages with missing man-pages and then running dnf reinstall in the Dockerfile for all packages in the list.

I also uploaded the script used to generate lists of missing docs for reference and future use. The script needs to be used like this:
1. `./gen-docs-list collect` executed in Fedora
2. `./gen-docs-list collect --toolbox` executed in the toolbox (--toolbox flag creates lists with different names)
3. `./gen-docs-list generate` executed with the tmp/ directory (containing lists created in 1. and 2.) in the same directory

The script is not polished, as it's just for more convenient generation of the desired list of missing docs.


images/fedora/f28/Dockerfile
```diff
@@ -13,6 +13,10 @@ COPY README.md /
 RUN sed -i '/tsflags=nodocs/d' /etc/dnf/dnf.conf
 RUN dnf -y swap coreutils-single coreutils-full
 
+COPY missing-docs /
+RUN dnf -y --setopt=tsflags='' reinstall $(<missing-docs)
```
debarshiray: It seems like the `--setopt=tsflags=''` isn't necessary. I built an image without it, and found that the packages were reinstalled with their documentation included:
```
host$ buildah bud --tag fedora-toolbox:28 images/fedora/f28
host$ ./toolbox create --release 28
host$ ./toolbox enter --release 28
toolbox$ man gawk
...
<man page shown>
....
```

images/fedora/f29/missing-docs
```diff
@@ -0,0 +1,16 @@
+acl
+bash
+chkconfig
+gnupg2
```
debarshiray: Aren't we missing `curl` and `dbus-daemon` here?

images/fedora/f29/missing-docs
```diff
@@ -0,0 +1,16 @@
...
+httpd
+hunspell
+ibus
+ibus-setup
+info
+openssl
+p11-kit
+pam
+python3
+rpm
```
debarshiray: Aren't we missing `systemd` here?

images/fedora/f28/Dockerfile
ozoder: Yes, the flag is not necessary. Fixed.

ozoder:
Sorry for the chaos in commits, apparently I can't do git rebase.

images/fedora/f28/missing-docs
```diff
@@ -0,0 +1,22 @@
+acl
+bash
+chkconfig
+curl
```
debarshiray: I think we need `dbus` here.

If you do `rpm --list -q dbus | grep man` inside a F28 toolbox, you'll see some `man` pages listed, but those files actually don't exist. If I recall correctly, the `dbus-daemon` package split from the `dbus` package in F29, which probably confused the script.

images/fedora/f28/missing-docs
```diff
@@ -0,0 +1,22 @@
...
+dnf
+gawk
+glusterfs
```
debarshiray: Typo? :)

Did we really have `glusterfs` in the F28 toolbox at some point in time? At least right now, I don't see a `glusterfs` package anymore.

images/fedora/f30/missing-docs
```diff
@@ -0,0 +1,18 @@
+acl
+bash
+chkconfig
+curl
+dbus-daemon
```
debarshiray: Umm... I wonder how I have `/usr/share/man/man8/dnf.8.gz` in my F30 toolbox that was created using the `fedora-toolbox:30` image that's already on `registry.fedoraproject.org` [1]. On the other hand, if I use an image built locally using `buildah bud ...`, I don't have the file.

Looking at a `fedora:30` image [1], I don't see a `/usr/share/man/man8/dnf.8.gz` either.

Any idea?

[1]  `$ toolbox create --release 30`
[2]  `$ podman run -it --rm registry.fedoraproject.org/fedora:30 /bin/bash`

debarshiray: I have a feeling that there's something weird about the `fedora-toolbox:30` image on the registry. There's no `dbus-daemon` package in the `fedora:30` image, but it's there in the toolbox image.

However, we can't push a new image either because currently the F30 and F31 container builds are in a strangely broken limbo. Such is life, I guess. :)

images/fedora/f28/missing-docs
debarshiray: Added.

images/fedora/f28/missing-docs
debarshiray: Removed.

debarshiray:
I also updated the image definition for Fedora 31, and merged. Thanks for working on this!

PR #56, opened 2019-02-18
Remove all bashisms (resolves #53)

### LINKED ISSUE #53: Bashisms with /bin/sh shebang
Currently fedora-toolbox has `/bin/sh` as shebang, implying that only POSIX shell functions are used. For example in Debian/Ubuntu based distributions, `/bin/sh` is not a symlink to bash, but to dash, which does not support bash extensions (see https://wiki.debian.org/DashAsBinSh). However, fedora-toolbox uses a couple of these bash functions. I ran the checkbashisms script to check for this, and it contained a couple of bash-exclusive features as of 4cf58d5b729571152a6255404b9af4cc81dcac58 (see https://gist.github.com/aimileus/e83db1f70344846d98f4e1aff071644b).

aimileus:
I tried to remove all bashisms, and now checkbashisms does not complain anymore:

- POSIX only supports single digit file descriptors. I used 3 instead of 42 as the file descriptor for /dev/null.
- echo -e replaces certain backslash expressions. Since, there were none of them, I just removed the -e flag. The `${variable}` syntax should still work.
- POSIX does not set the $UID environment variable. I replaced it with the custom variable $uid.
- The $'' syntax for backslash expressions does not exists in POSIX.  Since it was only used for the tab character, I just replaced it with the custom $tab variable at the beginning of the script.

In my (limited) testing, create, enter and list still works, using both bash and dash as interpreter. The --verbose option also still works, although I might have forgotten to change the file descriptor somewhere.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -123,7 +123,7 @@ spinner_start()
         return 1
     fi
 
-    echo -n "$message"
+    printf "%s" "$message"
     tput civis 2>&3
 
     exec 4>"$directory/spinner-start"
@@ -139,7 +139,7 @@ spinner_start()
                    break
                 fi
 
-                echo -n "$frame"
+                printf "%s" "$frame"
 
                 frame_len=${#frame}
                 i=0
@@ -737,7 +737,8 @@ list_images()
     fi
 
     if [ "$output" != "" ]; then
-        echo -e "${LBC}Images created by toolbox${NC}"
+        # shellcheck disable=SC2059
+        printf "${LBC}Images created by toolbox${NC}\n"
         echo "$output"
     fi
 
@@ -766,7 +767,8 @@ list_containers()
     fi
 
     if [ "$output" != "" ]; then
-        echo -e "${LBC}Containers created by toolbox${NC}"
+        # shellcheck disable=SC2059
+        printf "${LBC}Containers created by toolbox${NC}\n"
         echo "$output" | head --lines 1 2>&3
 
         echo "$output" | tail --lines +2 2>&3 \
@@ -775,7 +777,8 @@ list_containers()
                       id=$(echo "$container" | cut --delimiter " " --fields 1 2>&3)
                       is_running=$($prefix_sudo podman inspect "$id" --format "{{.State.Running}}" 2>&3)
                       if $is_running; then
-                          echo -e "${LGC}$container${NC}"
+                          # shellcheck disable=SC2059
+                          printf "${LGC}$container${NC}\n"
                       else
                           echo "$container"
                       fi


debarshiray:
Thanks for the excellent patch! Sadly, I ran into a ton of conflicts when trying to rebase it. In the end it seemed like it would be easier to cherry-pick each of the changes manually on top of master.

I have now created pull request #57 for the *single digit file descriptors* change. I hope I got it right.

I can slowly work through the rest of the changes in your patch, but feel free to rebase it yourself, if you prefer.

Once again, thanks for your contribution!

debarshiray:
By the way, you've got a lot of changes here. Feel free to add yourself to the copyright notice. :)

aimileus:
I rebased on master and it seems to work now. And although I changed a lot of lines of code, I do not feel I deserve a place in the copyright notice by replacing &42 by &3. Anyway, thank you very much for the offer!

debarshiray:
Sorry, it took me a fortnight to get back to this. I have now merged the `$UID` replacement via #72

aimileus:
Thanks! I rebased on master again, so you can apply it cleanly, if you want to.

toolbox
```diff
@@ -529,15 +530,15 @@ list_containers()
...
+                            --separator "$tab" \
                             --table \
                             --table-columns "CONTAINER ID,CONTAINER NAME,CREATED,STATUS,IMAGE NAME" 2>&3); then
         echo "$base_toolbox_command: failed to parse list of containers" >&2
         return 1
     fi
 
     if [ "$output" != "" ]; then
-        echo -e "${LBC}Containers created by toolbox${NC}"
+        echo "${LBC}Containers created by toolbox${NC}"
```
debarshiray: I was trying to read up on why the `-e` doesn't matter. The `man` pages for both the bash built-in in `bash(1)` and the external binary in `echo(1)` say that when `-e` is in effect the `\0NNN` sequence is one of the recognized sequences. Wouldn't that cover this case?

Or am I reading it wrong? Another option is to use `printf`.

debarshiray:
I merged the `$'...'` replacement via #77

toolbox
aimileus: First of all, sorry for the delay. Yes, you are correct, I just removed the -e flag and it seemed to work fine, but apparently I was incorrect. I pushed a new commit that uses printf.

debarshiray: No need to apologize. :) Thanks for updating the patch.

toolbox
```diff
@@ -775,7 +775,7 @@ list_containers()
                       id=$(echo "$container" | cut --delimiter " " --fields 1 2>&3)
                       is_running=$($prefix_sudo podman inspect "$id" --format "{{.State.Running}}" 2>&3)
                       if $is_running; then
-                          echo -e "${LGC}$container${NC}"
+                          echo "${LGC}$container${NC}"
```
debarshiray: We need to replace this `echo -e` with `printf` too.

(By the way, when I tested using `echo` without the `-e`, Bash refused to intepret the terminal escape sequences as control characters and printed them as is. So, maybe the behaviour that you saw was specific to Dash.)

debarshiray: I took the liberty to make the above adjustments and merged to `master`. Thanks for your contribution!

aimileus: Thank you for merging!

Issue #58, opened 2019-02-20
[RFE] Add sound support


AsciiWolf:
Fedora Toolbox already supports GUI applications. It would be great if it could also have sound (PulseAudio or ALSA) support. :-)

memnoth:
This might work, ``--volume /dev/snd:/dev/snd``.

Referred to this link, https://github.com/osrf/docker_images/issues/21#issuecomment-239419475

memnoth:
With the mounting of ``/dev/snd``, an application inside container doesn't work well with the sound if it works on the host at the same time.

Errors of mpv
```
ALSA lib pcm_dmix.c:1108:(snd_pcm_dmix_open) unable to open slave
[ao/alsa] Playback open error: Device or resource busy
[ao/oss] Can't open audio device /dev/dsp: No such file or directory
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
[ao/jack] cannot open server
ALSA lib pcm_dmix.c:1108:(snd_pcm_dmix_open) unable to open slave
[ao/sdl] could not open audio: ALSA: Couldn't open audio device: Device or resource busy
[ao] Failed to initialize audio driver 'sdl'
Could not open/initialize audio device -> no sound.
```

tosmi:
could this be implemented via an option e.g.

```
$ toolbox --enable-sound enter
```

my plan is to run vlc, emacs (using emms) or even firefox within the toolbox container and this would require sound support.

jm2c
toni

aimileus:
I tried to use the PulseAudio socket from the host and that seems to be fine: PulseAudio works, even if the host uses audio at the same time.

1. Bind mount `/dev/snd`
2. Install `pulseaudio-utils` in the container with dnf. 
3. Set `$PULSE_SERVER` to the value of `/run/user/$(id -ru)/pulse/native`, so that it uses the session of the host. I did not try to run a seperate PulseAudio session in the container.

tosmi:
i just discovered that sound simply works with `vlc`. `/run/user/<uid>/native` is available within the container. didn't have to install pulseaudio-utils, `$PULSE_SERVER` is not set. i'm using toolbox from `master`.

memnoth:
Thank you all for the information, I will check it out both.

[...]

PR #65, opened 2019-02-27
Work in progress: Support sound

memnoth:
Mount /dev/snd of the host to a container as a volume.

Fix: https://github.com/debarshiray/toolbox/issues/58

It tested with liquidwar game inside containers, the sound works well.


memnoth:
It doesn't work if an application using sound drivers on the host.

memnoth:
Close this issue until finding a solution and keep talking about it in the issue, #58

PR #74, opened 2019-03-06
Give access to Kerberos if KCM credential caches are being used

debarshiray:
There's no easy way to introspect the Kerberos configuration from the
command line. eg., the credential cache type being used, or the value
of the socket_path setting that denotes which socket the KCM service
will listen on. Therefore, it's assumed that the former is KCM if the
socket's path can be parsed from the sssd-kcm.socket unit.

Given the immutable nature of Podman containers, the toolbox container
and its corresponding image will have to be re-created if the host OS
is sufficiently re-configured.

The krb5-libs package was added to the base toolbox images to ensure
the presence of the /etc/krb5.conf.d directory with the correct
permissions. Currently, the package is already pulled in by various
dependencies. Therefore, it doesn't increase the size of the base
image, but serves as a safeguard against any inadvertent changes.

### DIFF
--- a/images/fedora/f28/extra-packages
+++ b/images/fedora/f28/extra-packages
@@ -11,6 +11,7 @@ hostname
 iputils
 jwhois
 keyutils
+krb5-libs
 less
 lsof
 man-db
--- a/images/fedora/f29/extra-packages
+++ b/images/fedora/f29/extra-packages
@@ -12,6 +12,7 @@ hostname
 iputils
 jwhois
 keyutils
+krb5-libs
 less
 lsof
 man-db
--- a/images/fedora/f30/extra-packages
+++ b/images/fedora/f30/extra-packages
@@ -12,6 +12,7 @@ hostname
 iputils
 jwhois
 keyutils
+krb5-libs
 less
 lsof
 man-db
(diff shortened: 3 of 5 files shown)


cgwalters:
I think I know why we want this, but the code could use at least a one-liner rationale comment, and so could the commit message.

debarshiray:
> the code could use at least a one-liner rationale comment,
> and so could the commit message.

This is still work-in-progress. :)

I stuck the branch in a pull request for @hughsie to try it out because I had promised him. As things stand one still has to install the `sssd-kcm` RPM inside the toolbox container to switch the Kerberos credentials cache type to KCM via `/etc/krb5.conf.d/kcm_default_ccache`. However, that package actually has code in it, which we don't need, and pulls in a ton of dependencies. Therefore, we'll either have to split the RPM or hack in the configuration when creating the container.

Having said that, I'd still welcome your feedback. Given the general immutability of Podman containers, I am wondering how best to handle all the variables here:
- host doesn't have `sssd-kcm` installed
- host has `sssd-kcm` installed but has switched the credentials cache to something else
- the UNIX socket doesn't exist on the file system
- etc.

I am expecting to tell the user that under certain circumstances, there's no other option than to re-create the toolbox container.

I have only tested it on a laptop with working KCM. I'd like to at least test it on my other laptop that doesn't use KCM, and see what happens.

debarshiray:
This is a lot more complete now. It works on hosts with KCM set up. I still need to test this on a system without KCM.

PR #75, opened 2019-03-08
added short options for create and enter

tosmi:
having short options for enter and create would be nice as i'm calling these quite often. i didn't add a short option for `create --candidate-registry`, `-r` is already taken and i'm currently not using `--candidate-registry`.

thanks
toni

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -1228,12 +1228,12 @@ usage()
 {
     echo "Usage: toolbox [-v | --verbose]"
     echo "               create [--candidate-registry]"
-    echo "                      [--container <name>]"
-    echo "                      [--image <name>]"
-    echo "                      [--release <release>]"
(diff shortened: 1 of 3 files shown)


tosmi:
what are the chances of getting this merged? makes my life easier :-)

toolbox
```diff
@@ -1085,12 +1085,12 @@ usage()
 {
     echo "Usage: toolbox [-v | --verbose]"
     echo "               create [--candidate-registry]"
-    echo "                      [--container <name>]"
-    echo "                      [--image <name>]"
-    echo "                      [--release <release>]"
+    echo "                      [-c | --container <name>]"
```
debarshiray: We need to update the manuals too.

debarshiray: I took the liberty to update the manuals and updated your branch.

PR #76, opened 2019-03-08
implement run command

tosmi:
this is my shot at implementing a `toolbox run` command. it simply runs a command in the toolbox container given. 

the use case is that we do not have to enter the container before running various commands. i've various aliases defined using the run command:

```
alias emacs='toolbox run -c home emacs'
alias vlc='toolbox run -c home vlc'
```

currently the documentation is missing, i will update the pull request if there is a chance that this gets merged.

i'm not a shell programming expert, so please be patient and excuse any mistakes :-)

thanks for this great tool
toni

**edit:** fixed typo

### DIFF
--- a/doc/toolbox-run.1.md
+++ b/doc/toolbox-run.1.md
@@ -1,0 +1,59 @@
+% toolbox-run(1)
+
+## NAME
+toolbox\-run - Run a command in an existing toolbox container
+
+## SYNOPSIS
+**toolbox run** [*-c | --container NAME*] [*-r | --release RELEASE*] [*COMMAND*]
+
+## DESCRIPTION
+
+Runs a command inside an existing toolbox container. The container
+should have been created using the `toolbox create` command.
+
+A toolbox container is an OCI container. Therefore, `toolbox run` is
+analogous to a `podman start` followed by a `podman exec`.
+
+On Fedora the toolbox containers are tagged with the version of the OS that
+corresponds to the content inside them. Their names are prefixed with the name
+of the base image and suffixed with the current user name.
+
+## OPTIONS ##
+
+The following options are understood:
+
+**-c | --container** NAME
+
+Run command inside a toolbox container with the given NAME. This is
+useful when there are multiple toolbox containers created from the
+same base image, or entirely customized containers created from
+custom-built base images.
+
+**-r | --release** RELEASE
+
+Run command inside a toolbox container for a different operating
+system RELEASE than the host.
+
+## EXAMPLES
+
+### Run ls inside a toolbox container using the default image matching the host OS
+
+```
+$ toolbox run ls -la
+```
+
+### Run emacs inside a toolbox container using the default image for Fedora 30
+
+```
+$ toolbox run --release f30 emacs
+```
+
+### Run uptime inside a custom toolbox container using a custom image
+
+```
+$ toolbox run --container foo cat /etc/fedora-release
+```
+
+## SEE ALSO
+
+`buildah(1)`, `podman(1)`, `podman-exec(1)`, `podman-start(1)`
--- a/toolbox
+++ b/toolbox
@@ -946,6 +946,15 @@ enter()
     create_toolbox_container=false
     prompt_for_create=true
     shell_to_exec=/bin/bash
+    run "PS1=\"$toolbox_prompt\" $shell_to_exec" "$SHELL" "--interactive --tty"
+)
+
+
+run()
+(
+    command="cd $PWD; $1"
+    fallback_command="$2"
+    podman_extra_arguments="$3"
 
     echo "$base_toolbox_command: checking if container $toolbox_container exists" >&3
 
@@ -953,7 +962,7 @@ enter()
                                  --format "{{.ImageName}}" \
                                  --type container \
                                  $toolbox_container 2>&3); then
-        echo "$base_toolbox_command: checking if container $toolbox_container_old exists" >&3
+	echo "$base_toolbox_command: checking if container $toolbox_container_old exists" >&3
 
         if toolbox_image=$($prefix_sudo podman inspect \
                                    --format "{{.ImageName}}" \
@@ -1033,29 +1042,25 @@ enter()
 
     set_environment=$(create_environment_options)
 
-    echo "$base_toolbox_command: looking for $SHELL in container $toolbox_container" >&3
-
-    if $prefix_sudo podman exec "$toolbox_container" test -f "$SHELL" 2>&3; then
-        shell_to_exec="$SHELL"
-    else
-        echo "$base_toolbox_command: $SHELL not found in $toolbox_container; using $shell_to_exec instead" >&3
+    if [ "$fallback_command" != "" ]; then
+	binary=$(echo "$command" | cut -d ' ' -f 1)
+	if $prefix_sudo podman exec $toolbox_container test -x "$binary" 2>&3; then
+	    echo "$base_toolbox_command: $binary not found in $toolbox_container; using $fallback_command instead" >&3
+            command=$fallback_command
+	fi
     fi
 
-    echo "$base_toolbox_command: trying to exec $shell_to_exec in container $toolbox_container" >&3
+    echo "$base_toolbox_command: trying to run '$command' in container $toolbox_container" >&3
 
     # shellcheck disable=SC2016
     # for the command passed to capsh
     # shellcheck disable=SC2086
     $prefix_sudo podman exec \
+            $podman_extra_arguments \
+            $set_dbus_system_bus_address \
             $set_environment \
-            --interactive \
-            --tty \
-            "$toolbox_container" \
-            capsh --caps="" -- -c 'cd "$1"; export PS1="$2"; shift 2; exec "$@"' \
-                    /bin/sh \
-                    "$PWD" \
-                    "$toolbox_prompt" \
-                    "$shell_to_exec" -l 2>&3
+            $toolbox_container \
+            capsh --caps="" -- -c "$command"
 )
 
 
@@ -1473,6 +1478,9 @@ usage()
     echo "   or: toolbox [-y | --assumeyes]"
     echo "               rmi [-a | --all]"
     echo "                   [-f | --force] [<image> ...]"
+    echo "   or: toolbox [-v | --verbose]"
+    echo "               run [-c | --container <name>]"
+    echo "                   [-r | --release <release>]"
     echo "   or: toolbox --help"
 }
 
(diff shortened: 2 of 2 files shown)


debarshiray:
Thanks for the patch @tosmi (and @mildred too for the other implementation in #30) !

A `run` command has been requested by a few people, and I like the idea.

However, I am worried about the toolbox encroaching too much on Flatpak territory. Both of them provider rootless unprivileged containers, and there's significant overlap, but ultimately the toolbox is aimed for hackers to set up a development environment while Flatpak is optimized for non-technical users to run graphical applications. Given that, `toolbox run emacs` becomes a lot like `flatpak run org.gnu.emacs`. :)

This is not to say that we don't want to add a `run`. It's just that I want to get our story straight. :)

debarshiray:
From `#silverblue` on Freenode:
```
15:56 <rishi> otaylor: alexlarsson: Hey! Do you have any thoughts on            
      https://github.com/debarshiray/toolbox/pull/76 ?
15:57 <rishi> In short, people want to be able to do "toolbox run emacs".
15:57 <rishi> And I am worried about encroaching on Flatpak territory.
15:58 <alexlarsson> I don't think its a huge problem.
15:58 <otaylor> rishi: I suspect we need such functionality if we want the      
      toolbox to be a serious tool that people rely on
15:58 <alexlarsson> No 2 tools will be 100% non-overlapping
15:59 <alexlarsson> and i can imagine using this in non-flatpak like way
15:59 <alexlarsson> toolbox run some-service
15:59 <otaylor> rishi: I'd be more worried about adding, say, menu item         
      management so that it looks like 'toolbox run emacs' is a real app
16:00 <alexlarsson> I mean, some people use flatpak for cli stuff
16:00 <rishi> otaylor: alexlarsson: Okay!
16:00 <alexlarsson> which is not quite the point
16:00 <alexlarsson> Still, it works
16:01 <alexlarsson> The main thing is that the design decisions that drive      
      toolbox and flatpak are driven by a particular usecase
16:01 <alexlarsson> Not that they can't be used other ways
16:03 <rishi> Yeah, toolbox is very clearly: "use jhbuild on Silverblue".
16:04 <rishi> walters would say "separate development prefix", but that's       
      about it, I think.
```

cgwalters:
This is one direction for https://discussion.fedoraproject.org/t/developing-applications-using-flatpak-packaged-editors-ides/269 right?  The use case here is "run my editor from the same context with the other developer tools".

toolbox
```diff
@@ -765,6 +807,9 @@ usage()
     echo "               enter [--container <name>]"
     echo "                     [--release <release>]"
     echo "   or: toolbox [-v | --verbose]"
+    echo "               run [-c | --container <name>]"
```
debarshiray: Nitpick: I'd put it after `rm` to keep them sorted.

owtaylor:
> This is one direction for https://discussion.fedoraproject.org/t/developing-applications-using-flatpak-packaged-editors-ides/269 right? The use case here is "run my editor from the same context with the other developer tools".

it's certainly an approach that works *if* the IDE is easy to install in the toolbox. Unfortunately, getting Eclipse (big piles of Java libraries) or VisualStudio Code (electron, lots of nodejs) into RPM form is not easy, and Flatpaks are an attractive target for such applications. So we need to figure out "bridge to the toolbox"

toolbox
tosmi: no problem, will fix that.

tosmi:
thanks for the review. i'll will come up with a manpage asap. 

maybe i should read up on flatpack and how stuff works there. toolbox and running everything in a dev container is just a very easy way to migrate my "old" workflow into a container based one. 

i'm completely aware that toolbox is more aimed at power users than non-technical users. but it's a great tool for me to get my toolchain running and event tinker around with it. wanna try unstable emacs? create a container, install it, use it via `run`, if it doesn't work go back to the old container, super easy.

if toolbox is clearly labeled as a tool for "power" users, it hopefully won't get confused with flatpack.

thanks
toni

toolbox
```diff
@@ -490,6 +490,48 @@ enter()
                     $shell_to_exec -l 2>&3
 )
 
+run()
```
debarshiray: One thing that I forgot to mention before...

Would it be possible to implement `enter` in terms of `run()`? The two functions are almost identical.

It seems to me that entering a toolbox container is synonymous to running `$SHELL` interactively inside the container. There's this fall-back mechanism in `enter` where if the user's `$SHELL` is absent from the toolbox then it falls back to `/bin/bash` which we know is present in the `fedora-toolbox` image. We could handle that by having `run()` accept two arguments - the first mandatory and the second optional. If the first command cannot be found, then it would try the second if it's present. Then `enter` could call `run()` with `$SHELL` and `/bin/bash.

I do not know how practical all this is with POSIX shell, but it seems worth a try at least.

tosmi:
so i've updated my branch:

- added manpage (sorry for any typos, it's already late here :-)
- moved run after rmi
- rebased to current master

thanks
toni

toolbox
tosmi: hm, i not 100% sure if i understand this:

- run slurps up all remaining arguments with "$*", so it is also possible to execute
   `toolbox run ls -la`

but i also felt the tension that run() is mostly enter() with only some lines changed. seems like i don't really understand what is going on in enter(). nevertheless i will try to play around with a common function for run() and enter().

thanks for your comments!

tosmi: so sleeping over the idea helped :-) on my train ride to work i came up with the following PoC:

https://github.com/tosmi/toolbox/blob/tosmi/run/toolbox

so basically run expects now 3 arguments:
- command
- fallback_command
- podman_extra_arguments because run currently runs podman exec without --interactive and --tty.

should i modify my tosmi/exec branch to include these changes? or is a separate pull request the better option? or is this completely stupid implementation?

as always thanks for your time

tosmi: enter() now uses run() for entering the container as requested. i've also rebased to the current master. what are the chances of getting this merged? i'm using run on a daily basis :-)

tosmi:
rebased on current master, cleanup of the history.

thanks
toni

doc/toolbox-run.1.md
```diff
@@ -0,0 +1,59 @@
...
+### Run emacs inside a toolbox container using the default image for Fedora 30
+
+```
+$ toolbox run --release f30 emacs
+```
+
+### Run uptime inside a custom toolbox container using a custom image
+
+```
+$ toolbox run --container foo cat /etc/fedora-release
```
debarshiray: Nitpick: typo alert! s/cat .../uptime/

doc/toolbox-run.1.md
```diff
@@ -0,0 +1,59 @@
+% toolbox-run(1)
+
+## NAME
+toolbox\-run - Run a command in an existing toolbox container
+
+## SYNOPSIS
+**toolbox run** [*-c | --container NAME*] [*-r | --release RELEASE*] [*COMMAND*]
```
debarshiray: Would be nice to put the longer variant first for consistency with the Podman manuals. The `--help` output is the other way round, but that should be replaced with the manuals like `git help`.

toolbox
```diff
@@ -946,14 +946,23 @@ enter()
     create_toolbox_container=false
     prompt_for_create=true
     shell_to_exec=/bin/bash
+    run "PS1=\"$toolbox_prompt\" $shell_to_exec" "$SHELL" "--interactive --tty"
```
debarshiray: The `PS1` is now set by the `profile.d/toolbox.sh` start-up script, and hence this is simpler now. See #148 

`--interactive` and `--tty` are still necessary in `run()` for things like `emacs -nw`.

toolbox
```diff
@@ -1747,6 +1755,35 @@ case $op in
...
+                        exit_if_non_positive_argument --release "$arg"
+                        release=$arg
+                        ;;
+                    * )
+                        exit_if_unrecognized_option "$1"
+                esac
+                shift
+            done
+            update_container_and_image_names
+            run "$*"
```
debarshiray: `"$@"` would be [better](https://unix.stackexchange.com/questions/129072/whats-the-difference-between-and).

doc/toolbox-run.1.md
```diff
@@ -0,0 +1,59 @@
+% toolbox-run(1)
```
debarshiray: This file needs to be added to `doc/meson.build`.

toolbox
```diff
@@ -1473,6 +1478,9 @@ usage()
     echo "   or: toolbox [-y | --assumeyes]"
     echo "               rmi [-a | --all]"
     echo "                   [-f | --force] [<image> ...]"
+    echo "   or: toolbox [-v | --verbose]"
+    echo "               run [-c | --container <name>]"
+    echo "                   [-r | --release <release>]"
```
debarshiray: We should at least update `completion/bash/toolbox` with the command and its options. Would be even nicer to have completion for the command.

toolbox
```diff
@@ -1033,29 +1042,25 @@ enter()
...
     set_environment=$(create_environment_options)
 
-    echo "$base_toolbox_command: looking for $SHELL in container $toolbox_container" >&3
-
-    if $prefix_sudo podman exec "$toolbox_container" test -f "$SHELL" 2>&3; then
-        shell_to_exec="$SHELL"
-    else
-        echo "$base_toolbox_command: $SHELL not found in $toolbox_container; using $shell_to_exec instead" >&3
+    if [ "$fallback_command" != "" ]; then
+	binary=$(echo "$command" | cut -d ' ' -f 1)
```
debarshiray: It would be more robust to let the shell split the command string into an array of arguments, than doing it ourselves with `cut(1)`. Imagine cases where the command has spaces or backslash escapes.

We should also be more pedantic in `run` and disable the smarts introduced for #128.  It's expected that users might use `run` to create ad-hoc *.desktop files, and user interaction wouldn't make sense when used like that. It's also a more advanced command than `enter` and new users are less likely to use it. So it shouldn't affect usability.

toolbox
```diff
@@ -1033,29 +1042,25 @@ enter()
...
 
-    echo "$base_toolbox_command: trying to exec $shell_to_exec in container $toolbox_container" >&3
+    echo "$base_toolbox_command: trying to run '$command' in container $toolbox_container" >&3
 
     # shellcheck disable=SC2016
     # for the command passed to capsh
     # shellcheck disable=SC2086
     $prefix_sudo podman exec \
+            $podman_extra_arguments \
+            $set_dbus_system_bus_address \
```
debarshiray: This is unused.

toolbox
```diff
@@ -1747,6 +1755,35 @@ case $op in
         fi
         exit
         ;;
+    run )
+        if is_integer "$podman_pid"; then
+	    # shellcheck disable=SC2119
+            forward_to_host
```
debarshiray: Some tabs got mixed with spaces.

debarshiray:
I took the liberty to rebase against `master` and made the above changes in PR #149

debarshiray:
Sorry that it took so long to get it merged, and once again thanks for your contribution!

debarshiray:
Merged via #149, closing.

tosmi:
thank you for taking the time merging this request. i'm happy that i can use the upstream version again.

thanks
toni

PR #79, opened 2019-03-13
Give access to /dev/bus for control transfers from USB devices

debarshiray:
This enables things like:
  $ colormgr get-sensor-reading lcd


hughsie:
Works for me here using `colorhug-refresh` -- thanks!

debarshiray:
Thanks, @hughsie

PR #82, opened 2019-03-18
Provide meaningful hint to overcome breakage caused by Podman 1.2.0

debarshiray:
Podman commit 4a02713c57d874c4 [1] broke 'podman exec' for existing
containers.

Note that support for ConmonPidFile in 'podman inspect' was added in
the same version of Podman as the one that has the breakage caused by
commit 4a02713c57d874c4 [1]. Therefore, there's no need to bump the
minimum required version of Podman. Interim Git snapshots between the
two will only cause the extra hint in the error message to be hidden.

[1] https://github.com/containers/libpod/commit/4a02713c57d874c4

https://github.com/containers/libpod/issues/2673

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -698,11 +698,31 @@ enter()
 
     echo "$base_toolbox_command: looking for $SHELL in container $toolbox_container" >&3
 
-    if $prefix_sudo podman exec $toolbox_container test -f "$SHELL" 2>&3; then
-        shell_to_exec=$SHELL
-    else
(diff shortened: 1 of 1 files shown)


owtaylor:
The inability to continue using an existing container seems like it would be a severe Podman regression, and I don't think adding a bunch of code to give the user a "throw away your work" message makes sense. I'd suggest that we just work with the Podman maintainers to encourage them to fix this ASAP upstream and in Fedora.

debarshiray:
The breakage was fixed in Podman itself in https://github.com/containers/libpod/pull/2762

PR #83, opened 2019-03-19
Make the ShellCheck success on toolbox

tpopela:


toolbox
```diff
@@ -737,7 +737,7 @@ list_images()
     fi
 
     if [ "$output" != "" ]; then
-        echo -e "${LBC}Images created by toolbox${NC}"
+        printf "%sImages created by toolbox%s" "${LBC}" "${NC}"
```
debarshiray: Sadly, we can't use `%s` in the format string to escape `LBC`, `NC`, etc.. :(

The problem is that those variables have the terminal escape sequences to change the foreground colour. Unless those variables are expanded inside the format string itself, those sequences won't be interpreted as control characters and will instead be printed as is.

If we keep the variable expansion inside the format string, then we run into [SC2059](https://github.com/koalaman/shellcheck/wiki/SC2059), which needs to be supressed. :/

Anyway, this was addressed in commit 189f366deafed1a225003a2f14f969540b60dbc8 from PR #56.

debarshiray: Rebased on top of master.

toolbox
```diff
@@ -18,7 +18,7 @@
 
 exec 3>/dev/null
 
-arguments="$@"
+arguments="$*"
```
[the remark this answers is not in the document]
debarshiray: I tweaked your commit to use the trick from [Rich's sh tricks](http://www.etalabs.net/sh_tricks.html) in #86. I hope I got it right. I also updated the commit message to explain the change in more detail because of the arcane nature of the shell concepts involved. :)

Comments welcome.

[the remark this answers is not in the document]
debarshiray: I merged #86 after a few hours of testing. Seems to work as intended.

toolbox
```diff
@@ -143,7 +143,7 @@ spinner_start()
 
     (
         while [ -f "$directory/spinner-start" ]; do
-            echo "$spinner_animation" | sed "s/ /\n/g" 2>&3 | while read frame; do
+            echo "$spinner_animation" | sed "s/ /\n/g" 2>&3 | while read -r frame; do
```
debarshiray: ++

I merged this patch to silence SC2162 via #89 after some testing. Thanks Tomas!

toolbox
```diff
@@ -283,7 +283,8 @@ create_environment_options()
         | {
               environment_options=""
               echo "$base_toolbox_command: creating list of environment variables to forward" >&3
-              while read variable; do
+              value=""
```
debarshiray: ++ merged as #90.

toolbox
```diff
@@ -716,9 +716,11 @@ enter()
 
     echo "$base_toolbox_command: trying to exec $shell_to_exec in container $toolbox_container" >&3
 
+    # shellcheck disable=SC2016
+    # for the command passed to capsh
```
debarshiray: ++ merged via #94

juhp:
(Just for the record ShellCheck is packaged in Fedora)

toolbox
```diff
@@ -675,7 +675,7 @@ create()
             --uidmap "$user_id_real":0:1 \
             --uidmap 0:1:"$user_id_real" \
             --uidmap "$uid_plus_one":"$uid_plus_one":"$max_minus_uid" \
-            $kcm_socket_bind \
+            "$kcm_socket_bind" \
```
debarshiray: `kcm_socket_bind` can be a null or empty string too.

toolbox
```diff
@@ -748,8 +748,8 @@ enter()
     # shellcheck disable=SC2016
     # for the command passed to capsh
     $prefix_sudo podman exec \
-            $set_dbus_system_bus_address \
-            $set_environment \
+            ${set_dbus_system_bus_address:+"$set_dbus_system_bus_address"} \
+            "$set_environment" \
```
debarshiray: Sadly, we can't quote these variables like this either. :(

In the case of `set_dbus_system_bus_address`, when the `DBUS_SYSTEM_BUS_ADDRESS` environment variable is present, it evaluates to:
```
podman exec ... "--env DBUS_SYSTEM_BUS_ADDRESS=..." ...
```

The entire `"--env DBUS_SYSTEM_BUS_ADDRESS=..."` string would be considered a single command line argument, instead of separately as `--env` and `DBUS_SYSTEM_BUS_ADDRESS=...`. When that happens, `podman` actually errors out.

In the case of `set_environment`, it evaluates to:
```
podman exec ... "--env=COLORTERM=truecolor --env=DBUS_SESSION_BUS_ADDRESS=..."
```

Again, the entire `"--env=COLORTERM=truecolor --env=DBUS_SESSION_BUS_ADDRESS=..."` string is considered a single command line argument, instead of multiple `--env` flags and their parameters. In this case, `podman` happens to ignore the whole string, but the spawned shell doesn't work properly because the variables aren't set. Missing `SHELL` and `TERM` seems to break Bash quite badly.

For the first case, we could use parameter expansion without the quotes:
```
${set_dbus_system_bus_address:+$set_dbus_system_bus_address}
```
It silences `SC2086` but doesn't prevent the basic problem of [globbing and word splitting](https://github.com/koalaman/shellcheck/wiki/SC2086) that it complains about.

For the second case, I don't know of any other option than to use `$set_environment` (or the identical `${set_environment}`) as before and disable the warning via a directive.

PR #91, opened 2019-03-26
Include images tagged with c.gh.debarshiray.toolbox in 'toolbox list' and the tag to containers and images produced by 'toolbox create'

debarshiray:
See the individual commits for further details.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -262,6 +262,13 @@ configure_working_container()
         return 1
     fi
 
+    if ! $prefix_sudo buildah config \
+                 --label "com.github.debarshiray.toolbox=true" \
+                 "$working_container_name" >/dev/null 2>&3; then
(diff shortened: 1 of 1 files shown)


debarshiray:
@tpopela could you please see if this works for you? It might not solve all your problems, but at least it shouldn't regress anything.

toolbox
```diff
@@ -405,6 +412,26 @@ image_reference_has_domain()
 )
 
 
+images_get_details()
+(
+    images="$1"
+
+    if ! echo "$images" | while read -r image; do
+            if ! $prefix_sudo podman images \
```
debarshiray: We have to be careful about `image` being an empty string. Fixed in commit 4f2eeb091cdcadf9e4a4d255d6d8bfe649ccd698

PR #101, opened 2019-04-04
Include container and images with c.gh.debarshiray.toolbox label in 'rm', 'rmi' and 'list'

tpopela:


toolbox
```diff
@@ -801,14 +801,24 @@ list_images()
...
                               --format "{{.ID}}  {{.Names}}  {{.Created}}  {{.Status}}  {{.Image}}" 2>&3); then
-        echo "$base_toolbox_command: failed to list containers" >&2
+        echo "$base_toolbox_command: failed to list containers with com.redhat.component=fedora-toolbox" >&2
         return 1
     fi
 
+    if ! containers=$($prefix_sudo podman ps \
+                              --all \
+                              --filter "label=com.github.debarshiray.toolbox=true" \
+                              --format "{{.ID}}  {{.Names}}  {{.Created}}  {{.Status}}  {{.Image}}" 2>&3); then
```
debarshiray: I think the `{{.Created}}` values will cause problems when merging the two lists using `sort` and `uniq` because the values can change quite rapidly for freshly created containers. eg., `1 second ago`, `5 seconds ago`, etc.. If such a container has both labels then it won't be merged.

We might have to do something similar as `list_images()`. ie., first get the list of names, merge them, and then get the full details. I am assuming that containers are guaranteed to have a unique name. I am not 100% sure, but it seems true. If that's not true, then we can use the IDs.

tpopela: Thank you Rishi, yes that makes sense. I now remember that you were mentioning that in the office. I went with the name in the end as from what I've found it is really supposed to be unique and for toolbox containers it's always set.

toolbox
```diff
@@ -799,17 +799,51 @@ list_images()
...
+
 list_containers()
 (
-    if ! containers=$($prefix_sudo podman ps \
+    if ! containers_old=$($prefix_sudo podman ps \
                               --all \
                               --filter "label=com.redhat.component=fedora-toolbox" \
-                              --format "{{.ID}}  {{.Names}}  {{.Created}}  {{.Status}}  {{.Image}}" 2>&3); then
-        echo "$base_toolbox_command: failed to list containers" >&2
+                              --format "{{.Names}}" 2>&3); then
```
debarshiray: Nitpick: would be nice to reindent these arguments too.

debarshiray: Rebased and merged as #109

toolbox
```diff
@@ -888,8 +930,13 @@ remove_containers()
...
                                 continue
                             fi
-                            if ! has_substring "$labels" "com.redhat.component:fedora-toolbox"; then
-                                echo "$base_toolbox_command: $id is not a toolbox container" >&2
+
+                            has_substring "$labels" "com.redhat.component:fedora-toolbox"
+                            has_old_id=$?
+                            has_substring "$labels" "com.github.debarshiray.toolbox"
+                            has_new_id=$?
+                            if [ ! $has_new_id ] || [ ! $has_old_id ]; then
```
debarshiray: Umm... doesn't something like this work:
```
if ! has_substring "$labels" "com.github.debarshiray.toolbox" \
   && ! has_substring "$labels" "com.redhat.component:fedora-toolbox"; then
       ...
fi
```

For example, I played around with these:
```
$ if ls /bar >/dev/null 2>/dev/null || ls /foo >/dev/null 2>/dev/null; then echo "one valid path"; fi
$ if ls /etc >/dev/null 2>/dev/null || ls /foo >/dev/null 2>/dev/null; then echo "one valid path"; fi
one valid path
$ if ls /bar >/dev/null 2>/dev/null || ls /tmp >/dev/null 2>/dev/null; then echo "one valid path"; fi
one valid path
$ if ls /etc >/dev/null 2>/dev/null || ls /tmp >/dev/null 2>/dev/null; then echo "one valid path"; fi
one valid path
```

debarshiray: The tweaked version of this patch in #110 seems to work for me.

Issue #59, opened 2019-02-20
onboarding friendliness


matthiasclasen:
I did a fresh install of silverblue. fedora-toolbox was installed. But I didn't quite know how to get started, so I just ran it:

```
$ fedora-toolbox
/usr/bin/fedora-toolbox: missing command
Try '/usr/bin/fedora-toolbox --help' for more information.
```

Thats not terrible, but it could be nicer. 
- Don't show the full path. 'fedora-toolbox' is enough
- Give me some hints how to get started, Something like "Use the 'create' command to create a toolbox", would be helpful
- For a minute, I was confused and thought there was some installation issue where fedora-toolbox was missing some tool it relies on. The error message could be clearer.
- Not so much a newcomer thing, but completion would be great to explore available commands

After this, I tried: man fedora-toolbox - and was greeted by another error message.
Its a commandline tool, it should have a man page.

Eventually, I figured I need to run: fedora-toolbox create. It did something! It did not really give me any for what to do next, but I just tried: fedora-toolbox enter. Sadly, this gives

```
$ fedora-toolbox enter
/usr/bin/fedora-toolbox: failed to start container fedora-toolbox-mclasen:30
```

At this point, I'm stuck. What is wrong? What to do next?

juhp:
I tend to agree I don't really like this trend of `use --help` to get help... just give me the help already. :-)

debarshiray:
> I don't really like this trend of use --help to get help... just give me the
> help already. :-)

Yeah, I'd like to make the command line parsing and associated error messages a bit more fault tolerant but the shell offers very limited facilities. Hopefully things will be better once we rewrite in Go. :)

matthiasclasen:
We changed flatpak to use this style of error message since it can be a bit overwhelming to get confronted with a screenful of --help output for a simple typo.

debarshiray:
> We changed flatpak to use this style of error message since it can be a bit
> overwhelming to get confronted with a screenful of --help output for a simple
> typo.

Ok, good to know.

For the time being, I am trying to make the `toolbox` error messages context-sensitive so that it suggests what the obvious solution might be, with a hint as to how to get more detailed documentation. For example:
```
$ toolbox 
toolbox: missing command

These are some common commands:
create    Create a new toolbox container
enter     Enter an existing toolbox container
list      List all existing toolbox containers and images

Try 'toolbox --help' for more information.
$ toolbox enter
toolbox: container fedora-toolbox-rishi-29 not found
[...]

PR #115, opened 2019-04-15
rm/rmi: give more specific error messages when image/container does not exist/cannot be removed

ozoder:
Currently, we give a generic error message when the image/container specified by the user cannot be removed. This usually happens for one of the following reasons:

1. the user specifies an image or a container that does not exist
2. the image has a child image or is being used by a container
3. the container is paused or running

This PR addresses these three cases and gives more specific message so that the user understands what happened. In order for this to work, the [libpod#2945](https://github.com/containers/libpod/pull/2945) PR needs to be merged.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -973,7 +973,12 @@ remove_containers()
             ret_val=$(echo "$ids" \
                       | (
                             while read -r id; do
-                                if ! $prefix_sudo podman rm $force_option "$id" >/dev/null 2>&3; then
+                                $prefix_sudo podman rm $force_option "$id" >/dev/null 2>&3
+                                ret_code=$?
+                                if [ $ret_code = 2 ]; then
+                                    echo "$base_toolbox_command: cannot remove running or paused container $id" >&2
+                                    ret_val=1
+                                elif [ $ret_code != 0 ]; then
                                     echo "$base_toolbox_command: failed to remove container $id" >&2
                                     ret_val=1
                                 fi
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -964,7 +964,12 @@ remove_containers()
             ret_val=$(echo "$ids" \
                       | (
                             while read -r id; do
-                                if ! $prefix_sudo podman rm $force_option "$id" >/dev/null 2>&3; then
+                                $prefix_sudo podman rm $force_option "$id" >/dev/null 2>&3
+                                ret_code=$?
```
debarshiray: Don't we also need to handle exit code 1?

Even though we have previously seen that the container does exist, it's possible that it got deleted by a simultaneously executing toolbox script or podman command or something else.

toolbox
```diff
@@ -964,7 +964,12 @@ remove_containers()
...
+                                if [ $ret_code = 2 ]; then
+                                    echo "$base_toolbox_command: cannot remove running or paused container $id" >&2
```
debarshiray: As part of #59, we have been trying to make the commands somewhat self-documenting. In this case, it might be nice to suggest the `--force` option and/or `podman stop`. See some of the recent changes done in #59 for ideas.

debarshiray:
Also, somehow your commits are attributed to `ozoder <ozoder@redhat.com>` instead of your full name. It could be a matter of just updating your Git configuration.

debarshiray:
This has been obsoleted by the Go rewrite. The Go implementation already has specific error messages for the scenarios mentioned here.

Issue #97, opened 2019-03-31
Add ttyUSB dialup devices in the toolbox


mildred:
Sometimes, in the toolbox, you need to access external hardware, either to communicate with an external device or flash a chip. Inside the toolbox, `/dev/ttyUSB*` is empty.
It would be nice if those devices were available. Possibly those devices can only appear after the toolbox has been entered. Perhaps sharing the entire `/dev` tree would not be that bad.

debarshiray:
I don't have any `/dev/ttyUSB*` on my host currently. I suppose those are just files under `/dev`, and not sub-directories, right?

I think we should look at bind mounting the entire `/dev`. @rhatdan warned me that PTYs could cause problems, and I haven't looked into that.

mildred:
> I don't have any /dev/ttyUSB* on my host currently. I suppose those are just files under /dev, and not sub-directories, right?

Yes, directly under `/dev`, and they appear dynamically (when you plug an USB serial device). I was convinced for no good reason that it was impossible to share /dev inside the toolbox, but if you do this at the toolbox creation, it works

mildred:
With PR #119, if I manually chown the device to my user/group on the host, I can access it from within the toolbox.

However, if I make myself part of the dialup group (the ttyUSB0 device is owned by root:dialup), the device appears as nobody:nobody in the toolbox and is inaccessible although it's accessible on the host.

thawkins:
You should also support /dev/ttyACMx style devices too, they are often used for arduino, cnc, laser engraver and 3d Printer connections.

debarshiray:
Now that PR #119 has been merged, this issue can be closed.

> You should also support /dev/ttyACMx style devices too, they are often
> used for arduino, cnc, laser engraver and 3d Printer connections.

New toolbox containers will have the entire `/dev` bind mounted from the host. So barring permission problems arising out of the files not being world accessible, things should be working now.

Issue #116, opened 2019-04-16
Nvidia proprietary driver


tfmoraes:
First, great project!

If I'm using Nvidia proprietary driver, OpenGL softwares (like Blender) don't work inside toolbox container. I tried to install the proprietary driver inside the container, it installs but the OpenGL softwares don't work. Is it necessary to install more things? Or set some env variable?

Thanks!

Findarato:
Toolbox is a container, you would have to map your graphics card inside, or do things the way nvidia-docker does.

tfmoraes:
@Findarato You mean add something like `--volume /dev/nvidia0:/dev/nvidia0` and other /dev files?

tpopela:
So to have the NVIDIA stuff working inside the Toolbox I had to do this (inspired by https://github.com/thewtex/docker-opengl-nvidia):

1) You have to patch the Toolbox to bind mount the /dev/nvidia0 and /dev/nvidiactl to the Toolbox and setup the X11 things - see https://github.com/tpopela/toolbox/commit/40231e8591d70065199c0df9b6811c2f9e9d7269

2) Download the NVIDIA proprietary drivers on the host:

```bash
#!/bin/sh

# Get your current host nvidia driver version, e.g. 340.24
nvidia_version=$(cat /proc/driver/nvidia/version | head -n 1 | awk '{ print $8 }')

# We must use the same driver in the image as on the host
if test ! -f nvidia-driver.run; then
  nvidia_driver_uri=http://us.download.nvidia.com/XFree86/Linux-x86_64/${nvidia_version}/NVIDIA-Linux-x86_64-${nvidia_version}.run
  wget -O ~/nvidia-driver.run $nvidia_driver_uri
fi
```

3) Install the drivers while being inside the Toolbox:
```bash
#!/bin/sh

sudo dnf install -y glx-utils kmod libglvnd-devel || exit 1
sudo sh ~/nvidia-driver.run -a -N --ui=none --no-kernel-module || exit 1
glxinfo | grep "OpenGL version"
```

tfmoraes:
@tpopela it worked. Thanks!

tpopela:
I'm glad it worked! But there was a mistake that could lead to malfunctions after the host is restarted - you will need to apply https://github.com/tpopela/toolbox/commit/3db450a8e5762399fd81c848f311da950437dd04 on top of the previous patch.

debarshiray:
[...]

PR #119, opened 2019-04-16
Change volume mounts to include all of /dev.

zerotri:
This is useful for accessing devices such as ttyUSB* which wouldn't
previously appear in the /dev tree when in a toolbox container.

This is a simple solution to #97, passing the entirety of `/dev` into the container, rather than mounting specific portions of it. There may be some security concerns to address when using this particular approach, especially as the container is being run with the `privileged` flag.

EDIT: It should be noted that while the /dev tree is passed into the container, `/dev` belongs to `nobody:nobody` and as such files are unaccessible without changing permissions on the host.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -777,9 +777,7 @@ create()
             --volume "$XDG_RUNTIME_DIR"/.flatpak-helper/monitor:/run/host/monitor \
             --volume "$dbus_system_bus_path":"$dbus_system_bus_path" \
             --volume /etc:/run/host/etc \
-            --volume /dev/bus:/dev/bus \
-            --volume /dev/dri:/dev/dri \
-            --volume /dev/fuse:/dev/fuse \
+            --volume /dev:/dev:rslave \
             --volume /media:/media:rslave \
             --volume /mnt:/mnt:rslave \
             --volume /run/media:/run/media:rslave \


toolbox
```diff
@@ -712,9 +712,7 @@ create()
             --volume "$HOME":"$HOME":rslave \
             --volume "$XDG_RUNTIME_DIR":"$XDG_RUNTIME_DIR" \
             --volume "$dbus_system_bus_path":"$dbus_system_bus_path" \
-            --volume /dev/bus:/dev/bus \
-            --volume /dev/dri:/dev/dri \
-            --volume /dev/fuse:/dev/fuse \
+            --volume /dev:/dev \
```
mildred: It seems podman does not honor this. I'm getting a reduced `/dev` with this version of toolbox.

I suspect that `/dev` is correctly bind mounted, but later on, podman is mounting a reduced `/dev` over it.

mildred: oops, I did not check by creating a new toolbox image :/

mildred: it actually works (you need to create a new toolbox for that), but the file is owned by nobody :(

zerotri: Indeed. This was specified above. You need to change permissions in the host operating system in order to access many of the files.

mildred:
This requires toolbox re-creation.

Also, if I have access to devices outside of the toolbox (because I'm part of the `dialup` group for example), I don't have access inside the toolbox...

Is it possible to imagine something that does not require toolbox re-creation. I would be difficult (creating a new container on enter, or modify podman to allow adding bind mounts after the container has been created, don't know if it's possible.)

or perhaps, a `toolbox recreate` command could take an existing toolbox, commit its filesystem to a new image, and create a new container to replace the previous one with the created image (but the updated bind mounts)

debarshiray:
This also helps with enabling the proprietary NVIDIA driver (#116) because it needs access to `/dev/nvidia0` and `/dev/nvidiactl`.

debarshiray:
I took the liberty to tweak the commit message and rebase on top of current `master`.

Thanks for working on this and the nice patch!

debarshiray:
> Also, if I have access to devices outside of the toolbox (because I'm part of the `dialup` group
> for example), I don't have access inside the toolbox...

Yeah, that's because we can only map one UID:GID pair from the host into the toolbox container. The `dialup` group just doesn't exist. Making the device nodes world readable (and writable?) would work. :/

> Is it possible to imagine something that does not require toolbox re-creation.
>
> [...]
> 
> or perhaps, a `toolbox recreate` command could take an existing toolbox, commit its
> filesystem to a new image, and create a new container to replace the previous one with
> the created image (but the updated bind mounts)

Yes, we need something like that. Want to submit a pull request for that? :)

davidlt:
I was attempting to run QEMU and noticed that `/dev/net/tun` is not in my toolbox container. This PR might help.

Issue #98, opened 2019-04-01
idea: include OS release in hostname


juhp:
It could be nice if the toolbox container hostname included the OS release version.
If one is running multiple toolbox releases it could be confusing which is which.

Maybe could use `toolbox:30` etc.  Well this is just a suggestion.

juhp:
Well, I see there are envvars

```
DISTTAG=f30container
FGC=f30
```

defined.

debarshiray:
Yes, I agree that this is a genuine problem.

mildred:
Around hostname in the toolbox, I miss my machine hostname. Also, I suppose that if I start using silverblue on my servers (which will be the case when upgrading from CoreOS), I might get confused when seeing the prompt with just `toolbox`, not knowing if I'm on localhost or on my servers via ssh.

mildred:
May I suggest that the default hostname should be composed of the machine hostname and the toolbox name ? For example, I run multiple toolboxes (gui, wine, dev, ...) and I have several computers (laptop, server, ...). The toolbox hostname could be a composition of both such as: `laptop-gui`, `laptop-dev`, `server-dev`, ...

In the meantime, I created an init script that executes when the shell starts:

```sh
if [[ "$(hostname)" = toolbox ]] && [[ "$(cat /etc/hostname)" != toolbox ]]; then
  hostname="$(cat /etc/hostname)"
  sh="$SHELL"
  if ! [[ -e "$sh" ]]; then
    sh=/bin/bash
  fi
  echo "Changing hostname from toolbox to $hostname and re-executing $sh..."
  sudo hostname "$hostname" && exec "$sh"
fi
```

Jmennius:
Since there is already a way to name containers on podman level with `toolbox create --container NAME` - we can just use this name for internal hostname (with minor modifications to satisfy hostname limitations).
Essentially:
```diff
@@ -1079,7 +1079,7 @@ create()
             --dns none \
             --env TOOLBOX_PATH="$TOOLBOX_PATH" \
             --group-add "$group_for_sudo" \
-            --hostname toolbox \
+            --hostname ${toolbox_container//_/-} \
             --ipc host \
             --label "com.github.containers.toolbox=true" \
             --label "com.github.debarshiray.toolbox=true" \
```
We can also allow overriding this name (#210), but I don't see why this would be necessary with adequate default.
I can open a pull featuring this change if there is any interest (or should I look into Rust or Go versions?).

Jmennius:
[...]

PR #120, opened 2019-04-16
Modify the prompt to show the currently running container.

### LINKED ISSUE #117: Provide a way to identify container
It would be nice to be able to identify each container from within the containers.

zerotri:
This modifies the prompt to show the currently running container, making it easy to identify which container instance you are working with in the case where you might have many. This would be one way to resolve #117.

This prefixes the diamond character with the container name to
very clearly distinguish the name from the rest of your prompt.

Where previously a bash prompt might read
🔹[zerotri@toolbox toolbox]$

for the container `sysutils` it would now read

sysutils🔹[zerotri@toolbox toolbox]$

EDIT: This would also potentially help resolve the problem defined in #98.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -811,7 +811,7 @@ enter()
             capsh --caps="" -- -c 'cd "$1"; export PS1="$2"; shift 2; exec "$@"' \
                     /bin/sh \
                     "$PWD" \
-                    "$toolbox_prompt" \
+                    "$toolbox_container$toolbox_prompt" \
                     "$shell_to_exec" -l 2>&3
 )
 


Frederik-Baetens:
This is a nice solution, but why not also add the container name to the hostname? That gives the user more flexibility.

bexelbie:
What would do we need to do to revive this PR?  This is something I was about to propose.

Frederik-Baetens:
I read that toolbox will be rewritten in Go, so the current code is probably low priority

HarryMichal:
Hi, we never went ahead and merged this PR. We didn't mean to ignore you (even though we did). Sadly the code here is outdated so I'll close this PR.

But I personally like the idea quite a lot. It certainly looks better than just having 'toolbox' as the hostname. The solution proposed by @Jmennius looks nice, so I'll take a look at it and try to incorporate it into the rewritten Toolbox.

debarshiray:
Yes, this change is a bit too outdated at this point, and not just because of the rewrite in Go.  Soon after this pull request was filed, we [stopped](https://github.com/containers/toolbox/commit/0e38e7d0b31ca866a1b3262fb42c01feae928479) setting the shell's prompt this way.

See all the issues and pull requests linked from https://github.com/containers/toolbox/issues/969 for all the latest discussions around this topic.

Regardless, thanks for your interest in Toolbx!

PR #126, opened 2019-04-23
Make it available inside the toolbox container

debarshiray:
Commit 5b3d234c9e9ef45f had made the toolbox script work inside a
toolbox container, but most people didn't get to use it because it
needed extra effort to get access to the script inside the container.
One either had to grab the Toolbox sources or had to install the RPM.
Both options were inconvenient - the former needed knowing too many
technical details, while the latter would drag in dependencies like
Buildah and Podman that don't work inside the toolbox container.

This makes it easier to use the toolbox script inside a toolbox
container by bind mounting the script from the host inside the
container and keeping track of the path using the TOOLBOX_PATH
environment variable. The environment variable ensures that running
'toolbox create' from inside a toolbox container would continue to
bind mount the same script from the host that was used to create the
current container and is available inside it.

Based on an idea from Colin Walters.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -26,6 +26,7 @@ base_toolbox_image=""
 # https://github.com/containers/libpod/blob/master/libpod/options.go
 container_name_regexp="[a-zA-Z0-9][a-zA-Z0-9_.-]*"
 
+environment=$(set)
 environment_variables="COLORTERM \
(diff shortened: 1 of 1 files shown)


debarshiray:
Merged, but comments still welcome.

Issue #131, opened 2019-04-25
Issue with default image release on non-fedora hosts


andrunko:
While working on integrating toolbox on EndlessOS (eos) I noticed that `toolbox create` (without params) tries to create an image from `fedora-toolbox:3.6.0` by default.

This happens because toolbox gets the default release version from `/etc/os-release` (VERSION_ID) . Given that the default image name is hardcoded to `fedora-toolbox`, toolbox should probably only get the default version from `/etc/os-release` if the `ID` is fedora, otherwise it should default to some fedora release (e.g. 29).

There are a few other options to consider also:
- instead of defaulting to `fedora-toolbox` when no image is provided via cmdline, use the `ID` from `/etc/os-release` as the base name: `$ID-toolbox:$VERSION_ID`
  - this way it at least would not try to download a `fedora-toolbox` image from a completely unrelated version
  - that would probably require changes to the default registry also or maybe we should use a list of registries instead (e.g. using the ones configured in `/etc/containers/registries.conf`)
  - If we go with that, note also that for Debian (`ID=debian`) hosts we should probably get the release version from `/etc/debian_version`, given that `VERSION_ID` is set to a number while the base docker debian images use the release name as version (e.g. `debian:buster` instead of `debian:10`) and I would suspect a potential `debian-toolbox` image would follow the same standard on image naming
- or maybe I should just patch the eos toolbox package directly, but that would mean users from non-fedora hosts trying to use toolbox from git would always (unless `VERSION_ID` matches the ones from fedora) have to pass the image/release from cmdline (maybe not a big deal)

I can provide patches as needed, just want to check first what are your thoughts on the options above (any other ideas?) and whether this is something we want to "fix" - I know it is not a priority for toolbox to work on non-fedora hosts by default, but it would be nice to support it oob when running without extra params.

debarshiray:
Thanks for getting in touch @andrunko !

At the moment the Toolbox implementation totally assumes Fedora everywhere. Not because it's intentional, but simply because it started as a prototype for Fedora Silverblue. I'd love to make adjustments, accept patches, etc. to make it more generic. Although, I have a feeling that some of the things would end up being special-cased for specific distributions. eg., the naming of the OCI images like you mentioned.

To start with, changing the name of the default OCI image seems straightforward.

[...]

PR #132, opened 2019-04-25
Set default release to 29 when running on non-fedora hosts

andrunko:
https://github.com/debarshiray/toolbox/issues/131

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -395,6 +395,14 @@ create_toolbox_image_name()
 )
 
 
+get_host_id()
+(
+    # shellcheck disable=SC1091
+    . /etc/os-release
+    echo "$ID"
+)
+
+
 get_host_version_id()
 (
     # shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -1259,7 +1267,12 @@ usage()
 
 arguments=$(save_positional_parameters "$@")
 
-release_default=$(get_host_version_id)
+host_id=$(get_host_id)
+if [ "$host_id" = "fedora" ]; then
```
debarshiray: Nitpick: would be nice to redirect the standard error to file descriptor 3 for `--verbose` output.

debarshiray: I took the liberty to tweak it as above.

andrunko:
@debarshiray thanks for checking/merging - I guess we can close issue #131 now and open another issue if/when we plan to change the default image based on the host.

debarshiray:
Ok, sure.

PR #133, opened 2019-04-25
completion: Add bash completion

fmuellner:
Make using toolbox a bit more convenient by properly completing its
options. The completions should be complete (that is, there are
completions for all the commands and options shown in --help),
but no attempt is made to filter out conflicting options (for
example "toolbox rm --all my-container").

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -24,6 +24,14 @@ install_data(
   install_dir: get_option('bindir'),
 )
 
+bash_completion = dependency('bash-completion', required: false)
+if bash_completion.found()
+  install_data(
+    'completion/bash/toolbox',
+    install_dir: bash_completion.get_pkgconfig_variable('completionsdir')
+  )
+endif
+
 subdir('data')
 subdir('doc')
 subdir('profile.d')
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -24,6 +24,14 @@ install_data(
   install_dir: get_option('bindir'),
 )
 
+bash_completion = find_program('bash-complation', required: false)
```
debarshiray: Typo alert: s/bash-complation/bash-completion/

meson.build
```diff
@@ -24,6 +24,14 @@ install_data(
...
+if bash_completion.found()
+  install_data(
+    'completion/bash/toolbox',
+    install_dir: bash_complation.get_pkgconfig_variable('completionsdir')
```
debarshiray: Ditto.

meson.build
debarshiray: Fixed and updated.

meson.build
debarshiray: Fixed and updated.

PR #137, opened 2019-04-26
Fix to work with column cmd from bsdmainutils

andrunko:
Fedora ships with 'column' cmd from util-linux which supports more
options than the one shipped by Debian (from bsdmainutils).
This change updates toolbox to also compatible with column from
bsdmainutils.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -506,6 +506,7 @@ images_get_details()
                 echo "$base_toolbox_command: failed to get details for image $image" >&2
                 return 1
             fi
+            echo
          done; then
         return 1
     fi
@@ -1047,6 +1048,8 @@ run()
 
 list_images()
 (
+    output=""
+
     if ! images_old=$($prefix_sudo podman images \
                               --filter "label=com.redhat.component=fedora-toolbox" \
                               --format "{{.Repository}}:{{.Tag}}" 2>&3); then
@@ -1066,11 +1069,12 @@ list_images()
         return 1
     fi
 
-    if ! output=$(echo "$details" \
-                  | sed "s/ \{2,\}/\t/g" 2>&3 \
-                  | column --separator "$tab" --table --table-columns "IMAGE ID,IMAGE NAME,CREATED" 2>&3); then
-        echo "$base_toolbox_command: failed to parse list of images" >&2
-        return 1
+    if [ "$details" != "" ] 2>&3; then
+        table_data=$(printf "%s\t%s\t%s\n" "IMAGE ID" "IMAGE NAME" "CREATED"; echo "$details")
+        if ! output=$(echo "$table_data" | sed "s/ \{2,\}/\t/g" 2>&3 | column -s "$tab" -t 2>&3); then
+            echo "$base_toolbox_command: failed to parse list of images" >&2
+            return 1
+        fi
     fi
 
     if [ "$output" != "" ]; then
@@ -1106,6 +1110,8 @@ containers_get_details()
 
 list_containers()
 (
+    output=""
+
     if ! containers=$(list_container_names); then
         return 1
     fi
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -490,6 +490,7 @@ images_get_details()
                 echo "$base_toolbox_command: failed to get details for image $image" >&2
                 return 1
             fi
+            echo
```
debarshiray: Is this really needed? Or is it a spurious change? It seems to work without it, but I am not sure if I am missing something.

toolbox
```diff
@@ -1040,9 +1041,11 @@ list_images()
...
     fi
 
-    if ! output=$(echo "$details" \
-                  | sed "s/ \{2,\}/\t/g" 2>&3 \
-                  | column --separator "$tab" --table --table-columns "IMAGE ID,IMAGE NAME,CREATED" 2>&3); then
+    table_data=$(printf "%s\t%s\t%s\n" "IMAGE ID" "IMAGE NAME" "CREATED"; \
+                 echo "$details" \
+                 | sed "s/ \{2,\}/\t/g" 2>&3)
+    if ! output=$(echo "$table_data" \
+                  | column -s "$tab" -t 2>&3); then
```
debarshiray: Does `bsdmainutils` not have the longer form of the options? ie., `--separator`, `--table`.

toolbox
andrunko: Yes, this is needed at least on podman 1.0.1, but looks like we can avoid it with 1.2.0 at least. Do we have a minimal podman version we support here?

toolbox
andrunko: Unfortunately it doesnt - see https://manpages.debian.org/buster/bsdmainutils/column.1.en.html.

toolbox
andrunko: Yes, I confirm it works without this and with podman 1.2.0.

debarshiray: Ok, thanks for checking!

It's a good question - what's the minimum supported Podman version. :) I don't know. We either avoided using anything that needed a newer Podman (eg., `podman exec --workdir` from https://github.com/containers/libpod/pull/2097), or we added a runtime check for it (eg., `podman create --dns=none --no-hosts` from https://github.com/containers/libpod/pull/2747).

There is no real reason for it, other than the fact that rootless Podman has been growing and maturing along with the Toolbox. So not having a hard requirement made it easier to revert to an older Podman to work around regressions, `git bisect`, implement something before Podman had support for it, and so on.

But it's possible that we are already hard-depending on features added to a certain version of Podman, and that we can exploit that to drop some of our compatibility code. If you find something like that then please submit a pull request.

toolbox
debarshiray: Ok, fine. Thanks for the link.

toolbox
```diff
@@ -1040,9 +1041,11 @@ list_images()
         return 1
     fi
 
-    if ! output=$(echo "$details" \
-                  | sed "s/ \{2,\}/\t/g" 2>&3 \
-                  | column --separator "$tab" --table --table-columns "IMAGE ID,IMAGE NAME,CREATED" 2>&3); then
+    table_data=$(printf "%s\t%s\t%s\n" "IMAGE ID" "IMAGE NAME" "CREATED"; \
+                 echo "$details" \
+                 | sed "s/ \{2,\}/\t/g" 2>&3)
```
debarshiray: Nitpick: it would be nice if we could also handle errors from the `sed` invocation. I am just paranoid because of the inherent fragility of parsing output from random commands (eg., `details`). One easy way to do that would be to move it into the block below.

toolbox
```diff
@@ -1040,9 +1041,11 @@ list_images()
         return 1
     fi
 
-    if ! output=$(echo "$details" \
-                  | sed "s/ \{2,\}/\t/g" 2>&3 \
-                  | column --separator "$tab" --table --table-columns "IMAGE ID,IMAGE NAME,CREATED" 2>&3); then
+    table_data=$(printf "%s\t%s\t%s\n" "IMAGE ID" "IMAGE NAME" "CREATED"; \
```
debarshiray: We should skip all this when `details` is an empty string. This was currently being taken care of by `column` but not anymore.

debarshiray:
@andrunko I took the liberty to update the branch to address the above nits, and also tweaked the commit message to mention the differences between `util-linux` and `bsdmainutils` with a link to the manual that you showed me on IRC.

Does this look ok to you?

andrunko:
@debarshiray thanks, change lgtm.

PR #148, opened 2019-05-03
Retain the PS1 across su(1) and sudo(8)

debarshiray:
The shell start-up scripts are where the PS1 is meant to be set. So
far, the absence of a toolbox-specific start-up script was being worked
around by setting the PS1 as part of the 'podman exec' invocation. This
came with certain limitations. eg., using su(1) or sudo(8) to get a
root shell can overwrite the PS1 set during 'podman exec' with a value
set by the operating system's existing start-up scripts depending on
which environment variables were being retained.

Now that the toolbox has it's own /etc/profile.d/toolbox.sh start-up
script, it's time to move the PS1 to its rightful home.

Since the start-up script and /run/.toolboxenv are present in older
toolbox containers, this change should be fully backwards compatible
and lead to a more robust PS1 without breaking older containers.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -24,23 +24,26 @@ if [ -f /run/ostree-booted ] \
 fi
 
 if [ -f /run/.containerenv ] \
-   && ! [ -f "$toolbox_welcome_stub" ] \
    && [ -f /run/.toolboxenv ]; then
-    echo ""
-    echo "Welcome to the Toolbox; a container where you can install and run"
-    echo "all your tools."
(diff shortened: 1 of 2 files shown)


debarshiray:
To test, ensure that you have `/etc/profile.d/toolbox.sh`. If not, then 'sudo cp profile.d/toolbox.sh /etc/profile.d`.

Then, use `su -` and `sudo su -` inside a toolbox container and verify that the blue diamond hasn't disappeared from the shell's prompt. It should work with both new and existing containers.

PR #150, opened 2019-05-06
Use a magenta hexagon instead of 🔹 in the PS1

debarshiray:
... because of its likeness to the Toolbox logo. Note that the magenta
foreground colour is requested through a terminal escape sequence with
SGR parameters [1]. The specific colour code for magenta is 35.

[1] https://en.wikipedia.org/wiki/ANSI_escape_code


debarshiray:
Committed as de67ff4bcc09a4d667943bbae1a98d34d3de9f1d

I don't know why it's showing as *closed*.

PR #156, opened 2019-05-13
Fix issue 152

ludwigd:
Missing dashes prevented execution of commands in `buildah unshare` session, which led to the fact that `/run/host/etc/resolv.conf` and `/run/host/etc/hosts` were not correctly linked into the toolbox, leaving the container without network connectivity.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -261,7 +261,7 @@ configure_working_container()
 
     echo "$base_toolbox_command: checking if 'buildah unshare' supports sh -c" >&3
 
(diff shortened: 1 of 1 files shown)


debarshiray:
Is it OK if I attribute the commit to `Damian Ludwig <damian.ludwig@uni-siegen.de>` ?

ludwigd:
> Is it OK if I attribute the commit to `Damian Ludwig <damian.ludwig@uni-siegen.de>` ?

Yes.

PR #165, opened 2019-05-17
Revert "Ensure that the toolbox user runs with 'wheel' as an ..."

debarshiray:
The --group-add option only affects the entry point of the toolbox
container.

This reverts commit 4bda42d4146f3d3baa8003428d1c0fa4b71239bd.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -728,7 +728,6 @@ create()
     $prefix_sudo podman create \
             $dns_none \
             $toolbox_path_set \
-            --group-add wheel \
             --hostname toolbox \
             --label "com.github.debarshiray.toolbox=true" \
             --name $toolbox_container \


debarshiray:
After a bit of testing, I realized that `podman create --group-add` doesn't only affect the entry point of the container. It also affects `podman exec`. So, if we just drop the option then we lose the `wheel` group membership.

With `master`:
```
⬢[rishi@toolbox ~]$ id
uid=1000(rishi) gid=1000(rishi) groups=1000(rishi),10(wheel)
```

With this PR:
```
⬢[rishi@toolbox ~]$ id
uid=1000(rishi) gid=1000(rishi) groups=1000(rishi)
```

So this change is wrong.

The more correct thing to do would be to move the `--group-add` to the relevant `podman exec` calls. Right now, the entry point process, which runs as `root:root`, also belongs to the `wheel` group, which is unnecessary. However, that's currently not possible because `podman exec` doesn't have any such option, even though the underlying `runc exec` command does have a `--additional-gids` option. However, that's something to be done in a different PR.

PR #167, opened 2019-05-20
Determine sudo group on init

andrunko:


toolbox
```diff
@@ -832,12 +831,21 @@ init_container()
             fi
         fi
 
+        groups=
```
debarshiray: Sadly, the idea behind PR #165 wasn't very sound, so we have to do the same thing for `podman create` too.

toolbox
```diff
@@ -832,12 +831,21 @@ init_container()
...
+        if getent group sudo > /dev/null 2>&3; then
+            groups="sudo"
+        elif getent group wheel > /dev/null 2>&3; then
+            groups="wheel"
+        else
+            echo "$base_toolbox_command: failed to setup groups for user $init_container_user, required group 'sudo' or 'wheel' missing" >&2
+        fi
```
debarshiray: We should `return 1`, to avoid attempting an invalid `useradd` later.

toolbox
```diff
@@ -832,12 +831,21 @@ init_container()
...
+
         if ! useradd \
                      --home-dir "$init_container_home" \
                      --no-create-home \
                      --shell "$init_container_shell" \
                      --uid "$init_container_uid" \
-                     --groups wheel \
+                     --groups $groups \
```
debarshiray: We should quote the `$groups` to avoid triggering [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) in future. ShellCheck sometimes misses some things but can start warning later. :/

toolbox

toolbox

debarshiray:
I took the liberty to fix the above nitpicks. The updated code works for me on Fedora. Could you please confirm that it works on Debian and Endless too?

toolbox

andrunko:
Thanks for the updates - I tested here and they work fine for me (tested on both fedora and endless).

PR #168, opened 2019-05-20
Unbreak the initialization of /etc/hosts and /etc/resolv.conf

debarshiray:
The unary logical negation operator (ie. !) was getting associated with
the 'cd /etc' instead of the entire sequence. As a result, neither
/etc/hosts nor /etc/resolv.conf were getting symlinked.

Fallout from 8b84b5e4604921fad818ede5267591de7f0993e6


debarshiray:
Thanks @jamescassell

PR #177, opened 2019-05-27
Implements executing the init system and the toolbox recreate command

mildred:
The recreate command is saving the current container as a new image, removing the container, and recreating a new container with all the updated options.

Implements debarshiray/toolbox#143

Remove PID namespace sharing and run the init system as PID 1 instead of sleep when starting up the toolbox. implements debarshiray/toolbox#176

Shell executed within the toolbox are now running in a separate PID namespace. Waiting on issue containers/libpod#3205


mildred:
Updated the PR to make this dependent upon the `--init` command-line option

ibotty:
Any updates?  Is that a viable approach?  Would that possibly get merged if the conflicts get resolved?

mildred:
Just updated this PR on top of latest master, and removed the recreate command to simplify this PR.

I'm using this feature and it works well.

mildred:
I'm using containers created by this PR daily and it works really great.

mjburling:
Perhaps the use-case for running an RDBMs is too high-minded. From toolbox, I just want to run:

1. `emacs --daemon=foo`
2. `emacsclient --socket-name=foo --create-frame &`

This seems to be able to satisfy that... What's holding up the merge?

mildred:
Just rebased my branch to latest master. If someone can take a look at it, I would be grateful. The change is really light (just replace the sleep by init and change just a few container options : unshare pid namespace, host dbus and mount /run as tmpfs)

mildred:
Sorry, this is no longer working with the newer version. And having journald mounted from within the container does not work.

bam80:
Progress?

PR #189, opened 2019-06-07
Fix check for whether host's /home is a symlink.

### LINKED ISSUE #185: /home is not symlinked in container on Silverblue systems
According to commit 66e982af, /home should be symlinked to /var/home in the toolbox container on Fedora Silverblue hosts. However, this does not seem to be the case with the latest toolbox 0.0.10.1 on SB 30. I suspect the culprit is commit 8b84b5e, which (among other things) which moved the logic introduced in 66e982af to init_container(). As that function is only executed within the container, it cannot detect whether /home on the host is symlinked to /var/home.

cjao:
Check in create() if /home on the host redirects to /var/home,
and modify the container accordingly in init-container(). This
restores the logic introduced in commit 66e982a which was
accidentally undone by commit 8b845e.

Closes #185

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -642,6 +642,7 @@ create()
 
     dbus_system_bus_address="unix:path=/var/run/dbus/system_bus_socket"
     dns_none=""
+    home_link=""
     kcm_socket=""
     kcm_socket_bind=""
     monitor_host=""
@@ -741,6 +742,20 @@ create()
     max_minus_uid=$((max_uid_count - user_id_real))
     uid_plus_one=$((user_id_real + 1))
 
+    if ! home_canonical=$(readlink --canonicalize "$HOME" 2>&3); then
+        echo "$base_toolbox_command: failed to canonicalize $HOME" >&2
+        return 1
+    fi
+
+    echo "$base_toolbox_command: $HOME canonicalized to $home_canonical" >&3
+
+    echo "$base_toolbox_command: checking if /home is a symbolic link to /var/home" >&3
+
+    if [ "$(readlink /home)" = var/home ] 2>&3; then
+	echo "$base_toolbox_command: /home is a symbolic link to /var/home" >&3
+	home_link="--home-link"
+    fi
+
     echo "$base_toolbox_command: calling org.freedesktop.Flatpak.SessionHelper.RequestSession" >&3
 
     if ! gdbus call \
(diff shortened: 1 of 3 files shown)


toolbox
```diff
@@ -751,6 +752,19 @@ create()
...
+	need_home_link=true;
+	echo "$base_toolbox_command: /home is a symlink to /var/home" >&3
+    else
+	need_home_link="";
+    fi
+
+    REALHOME=$HOME
+
+    if [ $need_home_link ] && [ ! "${HOME##/home*}" ]; then
+	REALHOME=/var$REALHOME
```
debarshiray: It would be better to use `readlink --canonicalize "$HOME"` instead. If nothing else, it's a lot more readable. :)

toolbox
```diff
@@ -781,12 +795,13 @@ create()
             --volume /media:/media:rslave \
             --volume /mnt:/mnt:rslave \
             --volume /run/media:/run/media:rslave \
-            --workdir "$HOME" \
+            --workdir "/" \
```
debarshiray: The default is `/`, so we can drop this flag now.

toolbox
```diff
@@ -772,7 +786,7 @@ create()
             $kcm_socket_bind \
             $toolbox_path_bind \
             $toolbox_profile_bind \
-            --volume "$HOME":"$HOME":rslave \
+            --volume "$REALHOME":"$REALHOME":rslave \
```
debarshiray: It would be good to mention something about this in the commit message. Given the subtle difference between Fedora 29 and 30, it took me a while to figure out why this was necessary.

toolbox
```diff
@@ -875,15 +891,14 @@ init_container()
...
 
+
     if ! id -u "$init_container_user" >/dev/null 2>&3; then
-        if [ "$(readlink /home)" = var/home ] 2>&3; then
+        if  $init_container_home_link ; then
             # shellcheck disable=SC2174
-            if ! rmdir /home 2>&3 \
-                 && mkdir --mode 0755 --parents /var/home 2>&3 \
-                 && ln --symbolic var/home /home 2>&3; then
+            if ! (rmdir /home 2>&3 && ln --symbolic /var/home /home 2>&3); then
```
debarshiray: It would be good to mention something about this in the commit message, or split it into a separate commit. Apparently this isn't the first time I have made this typo. :/

By the way, the `mkdir` command got lost.

toolbox
```diff
@@ -51,6 +51,7 @@ environment_variables="COLORTERM \
         XDG_VTNR"
 fgc=""
 
+need_home_link=""
```
debarshiray: Umm... this shouldn't be global.

debarshiray:
I took the liberty to make the above adjustments.

debarshiray:
Thank you very much for your contribution!

PR #195, opened 2019-06-14
Fix PS1 line wrapping

### LINKED ISSUE #190: PS1 line wrapping on long input
In the terminal, long text wraps back onto itself.

wrapping

I believe this line might be the issue.

https://github.com/debarshiray/toolbox/blob/0886fb466e7a73e7382255cf80c9459ffb1d84ac/profile.d/toolbox.sh#L28

The color sequences should probably be enclosed in `\[\]`.

LingCalvin:
Enclose non-printable sequences in the `PS1` with `\[\]` so its length gets calculated properly. Fixes #190.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -25,7 +25,7 @@ fi
 
 if [ -f /run/.containerenv ] \
    && [ -f /run/.toolboxenv ]; then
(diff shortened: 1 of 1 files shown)


debarshiray:
Excellent, thanks. Merged after elaborating the commit message a bit.

PR #200, opened 2019-06-20
Update toolbox help screen

HarryPhoon:
Hi, I recently started to experiment with Fedora Silverblue and naturally with toolbox and during it's usage I found the help screen to be a bit confusing and too verbose. So I took inspiration from other projects and reorganized it a little bit.

Current help screen:
Screenshot from 2019-06-20 17-49-16

Help screen with update:
Screenshot from 2019-06-20 17-48-51

### DIFF
--- a/doc/toolbox-help.1.md
+++ b/doc/toolbox-help.1.md
@@ -1,0 +1,31 @@
+% toolbox-help(1)
+
+## NAME
+toolbox\-help - Display help information about Toolbox
+
+## SYNOPSIS
+**toolbox help** [*COMMAND*]
+
+## DESCRIPTION
+
+When no COMMAND is specified, the `toolbox(1)` manual is shown. If a COMMAND
+is specified, a manual page for that command is brought up.
+
+Note that `toolbox --help ...` is identical to `toolbox help ...` because the
+former is internally converted to the latter.
+
+This page can be displayed with `toolbox help help` or `toolbox help --help`.
+
+## EXAMPLES
+
+### Show the toolbox manual
+
+```
+$ toolbox help
+```
+
+### Show the manual for the create command
+
+```
+$ toolbox help create
+```
--- a/toolbox
+++ b/toolbox
@@ -1189,6 +1189,18 @@ run()
 )
 
 
+help()
+(
+    to_help_command="$1"
+
+    if [ "$to_help_command" = "" ] 2>&3 || [ "$to_help_command" = "$base_toolbox_command" ] 2>&3; then
+        exec man toolbox 2>&1
+    fi
+
+    exec man toolbox-"$to_help_command" 2>&1
+)
+
+
 list_images()
 (
     output=""
@@ -1677,44 +1689,6 @@ update_container_and_image_names()
 }
 
 
-usage()
-{
-    echo "Usage: toolbox [-v | --verbose]"
-    echo "               [-y | --assumeyes]"
-    echo "               create [--candidate-registry]"
-    echo "                      [-c | --container <name>]"
-    echo "                      [-i | --image <name>]"
-    echo "                      [-r | --release <release>]"
-    echo "   or: toolbox [-v | --verbose]"
-    echo "               [-y | --assumeyes]"
-    echo "               enter [-c | --container <name>]"
-    echo "                     [-r | --release <release>]"
-    echo "   or: toolbox [-v | --verbose]"
-    echo "               [-y | --assumeyes]"
-    echo "               init-container --home"
-    echo "                              --home-link"
-    echo "                              --monitor-host"
-    echo "                              --shell"
-    echo "                              --uid"
-    echo "                              --user"
-    echo "   or: toolbox [-v | --verbose]"
-    echo "               [-y | --assumeyes]"
-    echo "               list [-c | --containers]"
-    echo "                    [-i | --images]"
-    echo "   or: toolbox [-y | --assumeyes]"
-    echo "               rm [-a | --all]"
-    echo "                  [-f | --force] [<container> ...]"
-    echo "   or: toolbox [-y | --assumeyes]"
-    echo "               rmi [-a | --all]"
-    echo "                   [-f | --force] [<image> ...]"
-    echo "   or: toolbox [-v | --verbose]"
-    echo "               [-y | --assumeyes]"
-    echo "               run [-c | --container <name>]"
-    echo "                   [-r | --release <release>] <command>"
-    echo "   or: toolbox --help"
-}
-
-
 arguments=$(save_positional_parameters "$@")
 
 host_id=$(get_host_id)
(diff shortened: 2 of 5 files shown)


debarshiray:
Thanks for the nice patch!

Yes, the `--help` output definitely needs some improvement. I had been thinking of just showing the `toolbox(1)` manual for that, and add a `help` command that shows the manuals for each of the other commands. Just like Git.

The idea is to avoid having the documentation spread across too many places, and make it easier to keep them updated.

Would you like to do something like that?

HarryPhoon:
Sounds like the optimal solution to this. I never did something like this so I'll have to do some research first but I'll try to do something like that. I'll push the changes, when I have them, to this PR.

debarshiray:
> I never did something like this so I'll have to do some research
> first but I'll try to do something like that. I'll push the changes, when
> I have them, to this PR.

Eventually this will boil down to `exec man ...`.

You can look at the Git sources for the `git help` command for more details and ideas. They aren't very hard to read. :)

HarryPhoon:
I'll go through the sources and see if I can do something.

toolbox
```diff
@@ -1732,8 +1747,8 @@ while has_prefix "$1" -; do
             assume_yes=true
             ;;
         -h | --help )
-            usage
-            exit
+            help "$2"
```
debarshiray: We should also update the `toolbox --help` output and a manual page for the new `toolbox help` command.

toolbox
```diff
@@ -1679,39 +1692,41 @@ update_container_and_image_names()
...
-    echo "                  [-f | --force] [<container> ...]"
-    echo "   or: toolbox [-y | --assumeyes]"
-    echo "               rmi [-a | --all]"
-    echo "                   [-f | --force] [<image> ...]"
-    echo "   or: toolbox [-v | --verbose]"
-    echo "               [-y | --assumeyes]"
-    echo "               run [-c | --container <name>]"
-    echo "                   [-r | --release <release>] <command>"
-    echo "   or: toolbox --help"
+    echo "Usage: toolbox [arguments] COMMAND [options]"
```
debarshiray: Yes, this format is definitely better than before! :)

Although, I wonder if we should also convert the `--help` output to show the `toolbox(1)` manual page? Piping the output from `man(1)` works fine (eg., `man ls | cat` and such) so it wouldn't hinder people from grepping the output for whatever reason.

It's tempting to not have to maintain the documentation across multiple places, but maybe there are drawbacks I haven't thought about.

What do you think?

toolbox
HarryPhoon: What do you mean specifically by updating? I would just specify that for invoking help you can use both flags `-h` and `--help`. And yeah, I need to update the man page and create one specifically for the command.

toolbox
HarryPhoon: It would simplify the maintenance for sure. And it shouldn't be too hard to implement.

If we were to use the output of man how should we format it? The same way it is now?? Or should we make it more git-like? Just the synopsis, all the commands + quick description and possible arguments?

HarryPhoon:
Just added the new usage functions that scrapes info from man. I think that it looks quite well. If it were to stay like this, I would make some minor changes to the man page:

- different wording for the -v | --verbose flag
- a bit different structure of synopsis

HarryPhoon:
Just noticed a new problem with the new help functionality. When inside of toolbox, there are no man pages of toolbox. So every time a user request help for some function, there is no output. And when the general help output is requested then no info is printed.

How should I fix this?? Copy the manual pages inside of the container when it's created or add an exception for the help function when used from inside of toolbox (and show again some hardcoded text)??

HarryPhoon:
The problem should be solved. I added a new case ('help') when toolbox is checking if it is inside of a container and then forwards the command to host. I had to redirect some outputs of commands to make it work in both environments (host/container). Should be ready to be merged.

toolbox
debarshiray: I meant that we should mention the new `help` command in the `toolbox --help` output.

toolbox
debarshiray: I meant that `toolbox --help` should be identical to `toolbox help`. ie., just run `man toolbox`. I am sorry for not being more clear. I see that you put in a ton of effort to reformat the manual. Sorry about that. :(

I don't quite see the usefulness of the raw textual format for showing usage information. It's not paged and there's no markup, so the format is more rudimentary as compared to the manual; and things like `toolbox help create | wc -l` work, so the fancier manual doesn't really seem to prevent any text processing.

debarshiray:
> I had to redirect some outputs of commands to make it work in both
> environments (host/container). Should be ready to be merged.

Nice! But do you know why the redirects were needed? :)

It turns out that `less(1)`, which is the pager used to render the manual, tries to open `/dev/tty` to get to the keyboard for accepting input. For any process `/dev/tty` represents its [controlling terminal](https://en.wikipedia.org/wiki/POSIX_terminal_interface#Controlling_terminals_and_process_groups), which is the keyboard and display attached to the process.

When we use `flatpak-spawn` from inside the container to find and render the manual on the host across D-Bus, the process on the host doesn't have a controlling terminal. This is because the controlling terminal is inherited by child processes when they are forked from their parents. However, that's not how `flatpak-spawn` operates. It sends a message over the D-Bus socket and the process on the host is forked from the D-Bus daemon, not from the `flatpak-spawn` process.

This means that `less(1)`, when run via `flatpak-spawn`, fails to open `/dev/tty` to get to the keyboard. It handles the failure by trying to use the standard error output to get to the keyboard - bizarre, I know. :)

That's why we need to redirect standard error to the standard output to having a working `less(1)`!

toolbox
```diff
@@ -1778,14 +1812,17 @@ fi
 echo "$base_toolbox_command: TOOLBOX_PATH is $TOOLBOX_PATH" >&3
 
 if [ "$1" = "" ]; then
-    echo "$base_toolbox_command: missing command" >&2
-    echo >&2
-    echo "These are some common commands:" >&2
```
debarshiray: On second thoughts, it would be better to retain this super abbreviated list. We had deliberately [added it](https://github.com/debarshiray/toolbox/commit/3afb2bc4c74db705af2e78bd92ee244597193970) to provider a softer landing ground than the full reference documentation for absolute beginners.

In some sense you can think of it is a self-documenting error message than documentation. Just like some of the other (error) messages that suggest possible solutions or next steps.

doc/toolbox-help.1.md
```diff
@@ -0,0 +1,31 @@
+% toolbox-help(1)
```
debarshiray: Oh, one more thing. We also need to update the Bash completion in `completion/bash`.

debarshiray:
I took the liberty to make the above adjustments, and tweaked the commit message to explain the bizarre `less(1)` quirk and why the redirections are needed. It's the sort of thing that's totally obscure and impossible to understand by just reading the code.

PR #202, opened 2019-06-24
Don't use 'podman cp' to copy toolbox.sh to old containers

debarshiray:
```
Copying files into a running container is considered inherently hacky.
Rootful Podman can pause a container using 'podman cp --pause ...'
during the copy, but that's not possible when used rootless.

Secondly 'podman cp' has suffered from a series of regressions lately.
First there was the problem with how the --pause flag was handled [1],
and then /etc/profile.d/toolbox.sh was getting created as a
directory [2], not regular file, by:
  $ podman cp \
            --pause=false \
            /etc/profile.d/toolbox.sh \
            "$container":/etc/profile.d

Try to side-step all that by using $XDG_RUNTIME_DIR as a conduit to
share the file with the container and using plain cp(1) to place it in
the toolbox container's /etc/profile.d.

[1] Commit e715ff2f9b231271
    https://github.com/debarshiray/toolbox/pull/193

[2] https://github.com/containers/libpod/issues/3384

https://github.com/debarshiray/toolbox/issues/196
```

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -248,7 +248,36 @@ copy_etc_profile_d_toolbox_to_container()
 (
     container="$1"
 
-    pause_false=""
+    profile_d_lock="$toolbox_runtime_directory"/profile.d-toolbox.lock
+
+    # shellcheck disable=SC2174
+    if ! mkdir --mode 700 --parents "$toolbox_runtime_directory" 2>&3; then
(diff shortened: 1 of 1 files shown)


grdryn:
This fixes the issue for me, I'm able to `./toolbox enter`. Thanks @debarshiray! :tada:

PR #205, opened 2019-06-26
update fallback fedora version from 29 to 30

juhp:


toolbox
```diff
@@ -1659,7 +1659,7 @@ host_id=$(get_host_id)
 if [ "$host_id" = "fedora" ] 2>&3; then
     release_default=$(get_host_version_id)
 else
-    release_default="29"
+    release_default="30"
```
debarshiray: This default was added by @andrunko for [running Toolbox on EndlessOS](https://github.com/debarshiray/toolbox/commit/ca9ca9a3ded056408eb4934a6905b2488152b166).

@andrunko would it be fine if we bumped the default Fedora version to 30?

PR #210, opened 2019-06-30
Make the hostname of the container into a command line option

### LINKED ISSUE #98: idea: include OS release in hostname

imciner2:
This creates a new command line option `-h` or `--hostname` for the `toolbox create` command that allows the user to specify the hostname of the container. If none is specified, it defaults to `toolbox`.

This would allow the user to specify the name, so it would fix #98 by allowing them to manually assign the name.

### DIFF
--- a/completion/bash/toolbox
+++ b/completion/bash/toolbox
@@ -17,7 +17,7 @@ __toolbox() {
   local commands="$verbose_commands rm rmi"
 
   declare -A options
-  local options=([create]="--candidate-registry --container --image --release" \
+  local options=([create]="--candidate-registry --container --hostname --image --release" \
                  [enter]="--container --release" \
                  [help]="$commands" \
                  [init-container]="--home --home-link --monitor-host --shell --uid --user" \
--- a/doc/toolbox-create.1.md
+++ b/doc/toolbox-create.1.md
@@ -46,6 +46,12 @@ customized containers from custom-built base images.
 Change the NAME of the base image used to create the toolbox container. This
 is useful for creating containers from custom-built base images.
 
+**--hostname** HOSTNAME
+
+Change the hostname of the toolbox container. This is the name displayed at the
+terminal prompt, so changing this can be useful to differentiate containers
+once they are open.
+
 **--release** RELEASE, **-r** RELEASE
 
 Create a toolbox container for a different operating system RELEASE than the
@@ -75,6 +81,11 @@ $ toolbox create --container foo --image bar
 
 ```
 $ toolbox create --candidate-registry
+```
+
+### Create a toolbox with the default container name but a hostname of devserver
+```
+$ toolbox create --hostname devserver
 ```
 
 ## SEE ALSO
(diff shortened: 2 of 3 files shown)


imciner2:
@debarshiray I have rebased this PR on the current master. Can you take a look at it?

imciner2:
I was just thinking about this some more, and an alternative default could be to default to the container name instead of `toolbox`, and then have the `--hostname` flag override this default. Thoughts?

bhavin192:
Hey @imciner2 thanks for working on this, while using toolbox for working with two different containers I wanted make similar change (using container name as hostname) and I found this PR :smile:

gainhad:
Any chance this will be merged soon?

maxwell-k:
@gainhad I don't have an official answer, but I have read that other merge requests that change functionality are blocked waiting for the golang rewrite in #318 . I imagine having a fixed set of features is very important when re-implementing and testing for compatibility. That rewrite is "finished" and going through a review process right now.

buck-ross:
Is there any update on this? It's been nearly 3 years since this was proposed, and #318 has been resolved since August of 2020. What is still standing in the way of this becoming a feature?

imciner2:
Frankly, I have given up on this and everything Toolbox related. I was initially using Toolbox for development and fixed this for my needs, but considering there was absolutely no communication from anyone in the project during the entire time I had this PR open (and this PR was open before the Go rewrite was proposed in a PR), I lost any hope that it would be merged. That coupled with several severe errors that made Toolbox unusable as a manager of persistent containers on my machine (it would give an error and the only way to fix it was by rebuilding the container), and the lack of support that was given on the BZs for it (there were also others experiencing the error), I gave up on the tool completely. With the switch to Go, I am even less likely to do anything with this, because I have no need to learn another language to fix a tool I don't use any more.

ma3yta:
@imciner2 try https://github.com/89luca89/distrobox
https://fedoramagazine.org/run-distrobox-on-fedora-linux/

debarshiray:
Sorry about the delay.  This is implementing https://github.com/containers/toolbox/issues/563

> I was just thinking about this some more, and an alternative default
> could be to default to the container name instead of `toolbox`, and
> then have the `--hostname` flag override this default. Thoughts?

One potential [problem](https://github.com/containers/toolbox/issues/98#issuecomment-1318592340) with using the name of the container as it's hostname is that it would make the shell's prompt quite long if `\h` was used in it.

There's more discussion in all the issues and pull requests linked from https://github.com/containers/toolbox/issues/969

debarshiray:
>  There's more discussion in all the issues and pull requests linked from https://github.com/containers/toolbox/issues/969

Also https://github.com/containers/toolbox/issues/98

debarshiray (commented):
For what it's worth, https://github.com/containers/toolbox/pull/1007 is a more complete and updated implementation of this idea.  It validates the string passed to the `--hostname` option and has tests.  Also, I suspect that the rules for a valid container name and a valid hostname are probably different.

PR #216, opened 2019-07-04
Fix copying profile.d/toolbox.sh to container

andrunko:
This fixes a couple of issues I am running with latest master (one of which also applies to 0.0.11) and podman 1.2.0 on containers created with toolbox 0.0.10.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -534,10 +534,11 @@ is_etc_profile_d_toolbox_a_bind_mount()
 {
     container="$1"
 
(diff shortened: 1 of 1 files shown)


juhp:
I haven't looked at the patch sorry, but just wondered if this is still needed with podman-1.5?

andrunko:
I can confirm toolbox 0.0.13 works fine with podman 1.5, closing this one as not needed anymore.

PR #229, opened 2019-07-23
Use host ulimits when creating containers

HarryPhoon:
https://github.com/debarshiray/toolbox/issues/213

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -692,6 +692,7 @@ create()
     monitor_host=""
     no_hosts=""
     toolbox_profile_bind=""
+    ulimit_host=""
 
     # shellcheck disable=SC2153
     if [ "$DBUS_SYSTEM_BUS_ADDRESS" != "" ]; then
@@ -748,6 +749,14 @@ create()
         no_hosts="--no-hosts"
 
         monitor_host="--monitor-host"
+    fi
+
+    echo "$base_toolbox_command: checking if 'podman create' supports --ulimit host" >&3
+
+    if man podman-create 2>&3 | grep "You can pass host" >/dev/null 2>&3; then
+        echo "$base_toolbox_command: 'podman create' supports --ulimit host" >&3
+
+        ulimit_host="--ulimit host"
     fi
 
     if ! pull_base_toolbox_image; then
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -711,6 +712,18 @@ create()
...
+    echo "$base_toolbox_command: checking if 'podman create' supports --ulimit host" >&3
+
+    if man podman-create 2>&3 | grep "You can pass 'host'" >/dev/null 2>&3; then
+        echo "$base_toolbox_command: 'podman create' supports --ulimit host" >&3
+
+        ulimit="--ulimit host"
+    else
+        echo "$base_toolbox_command: 'podman create' doesn't support --ulimit host" >&3
+        # shellcheck disable=SC2039
+        ulimit="$--ulimit nofile=$(ulimit -S -n):$(ulimit -H -n)"
```
debarshiray: Is that a stray `$`? :)

Given that `podman create --ulimit host` was added right after our baseline Podman version of 1.4.0, I'd rather not implement a fall back path at all. The fall back code paths are usually painful to test, and as a result have a tendency of starting to bit rot. Specifically, in this case, not preserving the ulimits leads to a failure that's specialized enough that those who really care should really get themselves a newer Podman.

toolbox
```diff
@@ -661,6 +661,7 @@ create()
     monitor_host=""
     no_hosts=""
     toolbox_profile_bind=""
+    ulimit=""
```
debarshiray: Nitpick: might be better to call it `ulimit_host` or such to steer clear of collisions with Bash's `ulimit` built-in.

toolbox
```diff
@@ -711,6 +712,18 @@ create()
         monitor_host="--monitor-host"
     fi
 
+    echo "$base_toolbox_command: checking if 'podman create' supports --ulimit host" >&3
+
+    if man podman-create 2>&3 | grep "You can pass 'host'" >/dev/null 2>&3; then
```
debarshiray: The single quotes around `host` broke the `grep` for me. Are they really needed?

debarshiray: I removed the single quotes for the time being.

toolbox
debarshiray: I dropped the fallback calculation to avoid worrying about the `$`. We can put it back in a subsequent commit if you feel strongly about it.

PR #233, opened 2019-08-01
Fix duplicates showing in list output

### LINKED ISSUE #226: Duplicates in toolbox list
Duplication in list if you use a period in the name.

Container ID's match and both will disappear if you run toolbox rm on it.

```
[dorian@silverblue ~]$ toolbox list
Images created by toolbox
IMAGE ID      IMAGE NAME                                        CREATED
e9beb8775c6c  localhost/fedora-toolbox-dorian:29                27 hours ago
3e43cfbe60b7  registry.fedoraproject.org/f29/fedora-toolbox:29  6 days ago
Containers created by toolbox
CONTAINER ID  CONTAINER NAME            CREATED       STATUS   IMAGE NAME
7f9b4c82621d  fedora-toolbox-dorian-29  27 hours ago  Created  localhost/fedora-toolbox-dorian:29

[dorian@silverblue ~]$ toolbox create --container fed29 --release 29
Created container: fed29
Enter with: toolbox enter --container fed29

[dorian@silverblue ~]$ toolbox list
Images created by toolbox
IMAGE ID      IMAGE NAME                                        CREATED
e9beb8775c6c  localhost/fedora-toolbox-dorian:29                27 hours ago
3e43cfbe60b7  registry.fedoraproject.org/f29/fedora-toolbox:29  6 days ago
Containers created by toolbox
CONTAINER ID  CONTAINER NAME            CREATED        STATUS   IMAGE NAME
bc9838c56733  fed29                     8 seconds ago  Created  registry.fedoraproject.org/f29/fedora-toolbox:29
7f9b4c82621d  fedora-toolbox-dorian-29  27 hours ago   Created  localhost/fedora-toolbox-dorian:29

[dorian@silverblue ~]$ toolbox create --container fed29.1 --release 29
Created container: fed29.1
Enter with: toolbox enter --container fed29.1

[dorian@silverblue ~]$ toolbox list
Images created by toolbox
IMAGE ID      IMAGE NAME                                        CREATED
e9beb8775c6c  localhost/fedora-toolbox-dorian:29                27 hours ago
3e43cfbe60b7  registry.fedoraproject.org/f29/fedora-toolbox:29  6 days ago
Containers created by toolbox
CONTAINER ID  CONTAINER NAME            CREATED         STATUS   IMAGE NAME
48257c617c1b  fed29.1                   4 seconds ago   Created  registry.fedoraproject.org/f29/fedora-toolbox:29
bc9838c56733  fed29                     24 seconds ago  Created  registry.fedoraproject.org/f29/fedora-toolbox:29
48257c617c1b  fed29.1                   4 seconds ago   Created  registry.fedoraproject.org/f29/fedora-toolbox:29
7f9b4c82621d  fedora-toolbox-dorian-29  27 hours ago    Created  localhost/fedora-toolbox-dorian:29
[dorian@silverblue ~]$ 
```

HarryMichal:
This fixes #226. I just removed an additional step in the list function.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -1950,8 +1950,14 @@ case $op in
             fi
         fi
 
-        $ls_images && [ "$images" != "" ] && echo "$images"
-        $ls_containers && [ "$containers" != "" ] && echo "$containers"
+        if $ls_images && [ "$images" != "" ]; then
+            echo "$images"
+        fi
+
+        if $ls_containers && [ "$containers" != "" ]; then
+            echo "$containers"
+        fi
+
         exit
         ;;
     rm | rmi )
--- a/toolbox
+++ b/toolbox
@@ -507,29 +507,6 @@ image_reference_has_domain()
 )
 
 
-images_get_details()
-(
-    images="$1"
-
-    if ! echo "$images" | while read -r image; do
-            [ "$image" = "" ] 2>&3 && continue
-
-            if ! $prefix_sudo podman images \
-                         --format "{{.ID}} {{.Repository}}:{{.Tag}} {{.Created}}" \
-                         --noheading \
-                         "$image" 2>&3; then
-                echo "$base_toolbox_command: failed to get details for image $image" >&2
-                return 1
-            fi
-            echo
-         done; then
-        return 1
-    fi
-
-    return 0
-)
-
-
 is_etc_profile_d_toolbox_a_bind_mount()
 {
     container="$1"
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -1992,8 +1950,14 @@ case $op in
             fi
         fi
 
-        $ls_images && [ "$images" != "" ] && echo "$images"
-        $ls_containers && [ "$containers" != "" ] && echo "$containers"
```
debarshiray: Good catch! I merged this patch via https://github.com/debarshiray/toolbox/pull/258 since it's obviously correct. I tweaked the commit message to make the problem more obvious.

toolbox
```diff
@@ -1236,23 +1196,25 @@ list_images()
 )
 
 
-containers_get_details()
+get_containers()
```
debarshiray: There's a reason why we first get the names, and then the rest of the details. :) Granted, it's a very obscure reason.

Since Podman's filters don't support using a logical `or`, we had to separately query the `com.redhat.component` and `com.github.debarshiray.toolbox` tagged images and merge the results into a single list.

*(I wonder if we can conjure up a logical `or` now that Podman filters support regular expressions.)*

Merging the lists using `uniq(1)` is tricky. We use attributes like `{{.Created}}` for each item, which can potentially change across two different Podman invocations if the item is new enough to be less than a minute old, and hence prevent `uniq(1)` from combining them into one entry.

That's why the names are first queried, merged, and then the attributes are read.

Now that we bumped our minimum Podman requirement to 1.4.0, it would've been nice if the regular expression support was also part of 1.4.0. Sadly, it isn't. It landed only in 1.5.0. Given that we still run into the occasional Podman regression (eg., https://github.com/containers/libpod/issues/3946) I am not comfortable hard bumping the baseline to 1.5.0. At least not yet. It would be nice to be able to keep using 1.4.0 on a best effort basis.

This might mean that we either figure out a way to detect the regular expression support at run-time and use it, or we fix this bug as part of our rewrite in Go because we'd just vendor in the necessary `libpod` version. I am leaning towards the latter option, myself.

PR #236, opened 2019-08-09
profile.d: make silverblue check more specific

mike-nguyen:
The Silverblue welcome message was being displayed incorrectly on
other ostree based OS's (Fedora Atomic Host, Fedora CoreOS).
Silverblue is the only one that runs a desktop so the gdm check
was used to distinguish between the other Fedora variants.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -5,8 +5,13 @@ toolbox_config="$HOME/.config/toolbox"
 host_welcome_stub="$toolbox_config/host-welcome-shown"
 toolbox_welcome_stub="$toolbox_config/toolbox-welcome-shown"
 
+# shellcheck disable=SC1091
+. /usr/lib/os-release
+
 if [ -f /run/ostree-booted ] \
-   && ! [ -f "$host_welcome_stub" ]; then
+   && ! [ -f "$host_welcome_stub" ] \
+   && [ "${ID}" = "fedora" ] \
+   && { [ "${VARIANT_ID}" = "workstation" ] || [ "${VARIANT_ID}" = "silverblue" ]; }; then
     echo ""
     echo "Welcome to Fedora Silverblue. This terminal is running on the"
     echo "host system. You may want to try out the Toolbox for a directly"


profile.d/toolbox.sh
```diff
@@ -6,6 +6,7 @@ host_welcome_stub="$toolbox_config/host-welcome-shown"
 toolbox_welcome_stub="$toolbox_config/toolbox-welcome-shown"
 
 if [ -f /run/ostree-booted ] \
+   && [ -d /run/gdm ] \
```
cgwalters: Since AFAIK Endless is also packaging this project, I'd instead do something like:

```
. /usr/lib/os-release
...
 && [ "${ID}" = "fedora" -a "${VARIANT_ID}" = "workstation" -o "${VARIANT_ID}" = "silverblue" ]
```

or so.

Also I'd just repeat I'm still developing https://github.com/cgwalters/coretoolbox since I don't think bash is appropriate for this.

cgwalters: And probably a better fix is to move this bit into https://pagure.io/workstation-ostree-config/ anyways.

mike-nguyen: > And probably a better fix is to move this bit into https://pagure.io/workstation-ostree-config/ anyways.

@cgwalters Move which bits? the bash profile?

cgwalters: Right.

mike-nguyen: In workstation-ostree-config, `toolbox` is not included in the f25-f29 branches.  It doesn't make sense to move the profile unless we add toolbox to the configs in those branches also.  For now I just update this PR to check os-release.

profile.d/toolbox.sh
```diff
@@ -5,8 +5,13 @@ toolbox_config="$HOME/.config/toolbox"
 host_welcome_stub="$toolbox_config/host-welcome-shown"
 toolbox_welcome_stub="$toolbox_config/toolbox-welcome-shown"
 
-if [ -f /run/ostree-booted ] \
```
debarshiray: Umm... I think that we need to retain the check for `/run/ostree-booted` in addition to the new ones that you added, otherwise the blurb will show up on legacy Workstation too. I haven't actually checked because I don't have a Workstation system at hand, only Silverblue.

Things have been complicated by the fact that none of the stable versions of Silverblue that have shipped so far had *silverblue* as the `VARIANT_ID`.

debarshiray: I took the liberty to tweak the patch myself.

mike-nguyen:
Thanks @debarshiray

PR #238, opened 2019-08-13
Check if /run/media is available

tagoh:
No /run/media is available on Chromebook so it should be optional to get this work on Chromebook.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -691,6 +691,7 @@ create()
     libvirt_system_directory_bind=""
     monitor_host=""
     no_hosts=""
+    run_media_path_bind=""
     toolbox_profile_bind=""
     ulimit_host=""
 
@@ -799,6 +800,10 @@ create()
         toolbox_profile_bind="--volume /etc/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"
     fi
 
+    if [ -d /run/media ] 2>&3; then
+        run_media_path_bind="--volume /run/media:/run/media:rslave"
+    fi
+
     if ! home_canonical=$(readlink --canonicalize "$HOME" 2>&3); then
         echo "$base_toolbox_command: failed to canonicalize $HOME" >&2
         return 1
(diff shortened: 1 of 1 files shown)


debarshiray:
Does ChromeOS not use UDisks2?

tagoh:
Apparently not. it seems that they have own implementation for that, so-called cros-disks: https://chromium.googlesource.com/chromiumos/platform2/+/master/cros-disks/

debarshiray:
> Apparently not. it seems that they have own implementation for that,
> so-called cros-disks:
> https://chromium.googlesource.com/chromiumos/platform2/+/master/cros-disks/

Ok, I see.

As far as I can make out ChromeOS uses Upstart, not systemd, so the [systemd-tmpfiles based mechanism](https://github.com/debarshiray/toolbox/commit/0c848abd1893764f2e0bb62334972d67f95dafd1) for creating `/run/media` will not work. So, I guess we need this after all. :)

toolbox
```diff
@@ -864,7 +869,7 @@ create()
             --volume "$home_canonical":"$home_canonical":rslave \
             --volume /etc:/run/host/etc \
             --volume /dev:/dev:rslave \
-            --volume /media:/media:rslave \
+            $media_path_bind \
```
debarshiray: Is this a typo? :)

Wasn't `/run/media` the missing path? This line is replacing the `/media` entry, not `/run/media`.

toolbox
```diff
@@ -693,6 +693,7 @@ create()
     no_hosts=""
     toolbox_profile_bind=""
     ulimit_host=""
+    media_path_bind=""
```
debarshiray: Maybe call the variable `run_media_path_bind`?

There's another request about making the `/media` bind mount conditional because it doesn't exist on openSUSE. This way the variables won't conflict.

toolbox
tagoh: Doh, you're right.

toolbox
tagoh: Sure.

tagoh:
Okay, updated.

debarshiray:
> Okay, updated.

Thanks. I added the URL to this PR to the commit message and merged.

PR #239, opened 2019-08-18
Optionally bind $HOME/.cache/toolbox/f[release] to /var/cache/dnf

cjao:
Newly created containers must download more than 90mb of repository metadata before they can install anything. However, much of this metadata can be safely shared between containers based on the same image; for instance, the 79mb main Fedora repository is frozen. 

This change allows containers to share their dnf metadata cache, potentially saving substantial time (and data) when creating new containers, while allowing one to opt-out of the sharing via a new `--private-dnf-cache` flag.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -27,6 +27,8 @@ base_toolbox_image=""
 # https://github.com/containers/libpod/blob/master/libpod/options.go
 container_name_regexp="[a-zA-Z0-9][a-zA-Z0-9_.-]*"
 
+dnf_cache_dir=""
+
 environment=$(set)
 environment_variables="COLORTERM \
         DBUS_SESSION_BUS_ADDRESS \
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -769,6 +772,20 @@ create()
 	home_link="--home-link"
     fi
 
+    if [ -z "$prefix_sudo" ]; then
+	dnf_cache_dir="$HOME/.cache/toolbox/$fgc"
```
debarshiray: I must be missing something here, so please bear with me. How does `~/.cache/toolbox/$fgc` get populated on the host?

debarshiray:
This entire pull request needs to be ported to Go.

debarshiray:
`/var/cache/dnf` is usually owned by `root:root`. If we use `~/.cache/toolbox` to back that location via a bind mount, then won't we end up with files with funny permissions inside the user's $HOME? That would be a little unfortunate.

debarshiray:
> This entire pull request needs to be ported to Go.

Closing. Please feel free to submit a new pull request against the Go implementation.

PR #243, opened 2019-08-28
Inherit COLUMNS and LINES at host

tagoh:
to keep same terminal window size on host.

Both COLUMNS and LINES isn't actually environment variables.
So the existing create_environment_options() can't be used for this purpose.

This fixes https://github.com/debarshiray/toolbox/issues/242

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -322,7 +322,28 @@ create_enter_command()
 
 
 create_environment_options()
-{
+(
+    columns=""
+    lines=""
+
+    if terminal_size=$(stty size 2>&3); then
+        columns=$(echo "$terminal_size" | cut --delimiter " " --fields 2 2>&3)
+        if ! is_integer "$columns"; then
+            echo "$base_toolbox_command: failed to parse the number of columns from the terminal size" >&3
+            columns=""
+        fi
+
+        lines=$(echo "$terminal_size" | cut --delimiter " " --fields 1 2>&3)
+        if ! is_integer "$lines"; then
+            echo "$base_toolbox_command: failed to parse the number of lines from the terminal size" >&3
+            lines=""
+        fi
+
+        environment_options="$environment_options --env=COLUMNS=$columns --env=LINES=$lines"
+    else
+        echo "$base_toolbox_command: failed to read terminal size" >&3
+    fi
+
     echo "$environment_variables" \
         | sed "s/ \+/\n/g" 2>&3 \
         | {
@@ -339,12 +360,20 @@ create_environment_options()
                   fi
               done
 
+              if [ "$columns" != "" ] 2>&3; then
+                  environment_options="$environment_options --env=COLUMNS=$columns"
+              fi
+
+              if [ "$lines" != "" ] 2>&3; then
+                  environment_options="$environment_options --env=LINES=$lines"
+              fi
+
               environment_options=${environment_options#" "}
               echo "$base_toolbox_command: created options for environment variables to forward" >&3
               echo "$environment_options" >&3
               echo "$environment_options"
           }
-}
+)
 
 
 create_toolbox_container_name()


juhp:
LGTM and fixes the issue for me too (which seemed to start with podman-1.5?)

toolbox
```diff
@@ -347,6 +347,17 @@ create_environment_options()
 }
 
 
+set_terminal_size()
```
debarshiray: Maybe move this into `create_environment_options` itself?

toolbox
```diff
@@ -347,6 +347,17 @@ create_environment_options()
...
+{
+    stty size|if read -r lines cols; then
```
debarshiray: Umm... why the `if`? I guess you wanted to handle cases where the `stty` invocation might have failed?

If so, then it would be more readable to move it out into a separate `if` branch like we do for other commands; and then use `cut` to extract the size fields because `read` is a bit weird to use when there's no conditional branch or loop to go with it.

debarshiray:
I took the liberty to make the above tweaks, but I don't have enough time to convince myself that I didn't break anything. :)

Does it look ok to you?

toolbox
tagoh: Sure. it could be.

toolbox
tagoh: You're right. okay, let me try as you suggested.

tagoh:
Updated.

debarshiray:
I wonder if there's some race involved here. For some reason this bug stopped reproducing today out of the blue, even with the same software versions that were reliably reproducing it until last Thursday.

toolbox
```diff
@@ -323,6 +323,15 @@ create_enter_command()
 
 create_environment_options()
 {
+    # have to call this outside of pipe to avoid "Inappropriate ioctl for device" error in stty.
+    terminal_size=$(stty size)
+    cols=""
+    lines=""
+    if [ "$terminal_size" != "" ] 2>&3; then
+        cols=$(echo "$terminal_size"|cut -d' ' -f2)
```
debarshiray: Nitpick: could be nice to validate the numbers using `is_integer`.

toolbox
```diff
@@ -323,6 +323,15 @@ create_enter_command()
 
 create_environment_options()
 {
+    # have to call this outside of pipe to avoid "Inappropriate ioctl for device" error in stty.
+    terminal_size=$(stty size)
```
debarshiray: Yes, this is definitely better than using `/dev/tty` because it will continue to work when `toolbox(1)` is used from inside a container and everything is forwarded to the host over D-Bus.

Nitpick: would be better to explain it in more detail in the commit message than a one-line comment. :)

debarshiray:
I made the above tweaks and pushed to master! Thanks for your contribution. :)

toolbox
```diff
@@ -322,7 +322,28 @@ create_enter_command()
...
+            columns=""
+        fi
+
+        lines=$(echo "$terminal_size" | cut --delimiter " " --fields 1 2>&3)
+        if ! is_integer "$lines"; then
+            echo "$base_toolbox_command: failed to parse the number of lines from the terminal size" >&3
+            lines=""
+        fi
+
+        environment_options="$environment_options --env=COLUMNS=$columns --env=LINES=$lines"
```
debarshiray: Eek! This was a copy paste fiasco on my part. Embarassing. :(

toolbox
```diff
@@ -322,7 +322,28 @@ create_enter_command()
 
 
 create_environment_options()
-{
+(
+    columns=""
+    lines=""
+
+    if terminal_size=$(stty size 2>&3); then
```
debarshiray: I missed that just `stty(1)` isn't enough when `toolbox(1)` is invoked from inside a container. In that case `stty(1)` will get run against the inner pseudo-terminal pair created by `podman exec --tty` which doesn't have the right sizes. I think we should forward the `COLUMNS` and `LINES` variables from the outer environment created by `toolbox(1)`.

debarshiray:
I created https://github.com/debarshiray/toolbox/pull/255 for these issues as a follow-up.

One part of me really wishes that https://github.com/containers/libpod/issues/3946 gets fixed as soon as possible and we can revert this workaround. :)

Issue #68, opened 2019-03-01
Add test suite


debarshiray:
The toolbox has grown enough glue on top of Buildah and Podman that it's worth testing them more methodically. The Podman test suite might serve as a decent starting point for researching how such a test suite might look.

PR #250, opened 2019-09-03
Add zuul-ci

HarryMichal:
This adds configuration for zuul-ci provided by softwarefactory, the first batch of system tests that I scrambled together and a fix for bind mounting toolbox.sh inside of containers. This should be ready to be merged. If it will be, also close this #68.


TristanCacqueray:
About the /run/media issue, you can try adding to the First comment of this change:
```
Depends-On: https://github.com/debarshiray/toolbox/pull/251
```

.zuul.yaml
```diff
@@ -0,0 +1,43 @@
...
+    pre-run: playbooks/setup-rpm-env.yaml
+    run: playbooks/test-with-podman-stable.yaml
+
+- job:
+    name: test-with-podman-rawhide
+    description: Test toolbox with latest (rawhide) Podman
+    nodeset:
+      nodes:
+        - name: ci-node-30
+          label: cloud-fedora-30
```
TristanCacqueray: We recently added rawhide with this label: `cloud-fedora-rawhide-small` . It seems like this nodeset definition is not consistent with the job name.

HarryMichal: That's sweet! The Rawhide should indicate that we're testing against the latest Podman. But having a Fedora Rawhide image is surely handy and I'll use it.

TristanCacqueray:
Job's timeout can be increased by setting this attribute: https://zuul-ci.org/docs/zuul/user/config.html#attr-job.timeout

HarryMichal:
> Job's timeout can be increased by setting this attribute: https://zuul-ci.org/docs/zuul/user/config.html#attr-job.timeout

Yeah, that seems like an important setting. The jobs started to hang when I force-pushed while they were running. Thanks

TristanCacqueray:
Force pushing shouldn't affect the performance of the next batch of job. It seems like the job failed with `FAIL: Toolbox should fail with: no such container` and then the `[remove_all_images_but_default]` step seems to got stuck in that situation. Thus increasing the timeout doesn't seem to be the solution here afterall.

HarryMichal:
Still, the timeout can be useful if some other situation arose. I'll take a look at the step.

HarryMichal:
I updated the PR. Let me know if there is anything else you want me to adjust.

I agree that it is a bit of a contradiction that I introduced BATS even though we want to move from shell/bash. At the time I started the PR it looked like the optimal solution because it's quite easy to get started. If in the future we find a better solution then I'm not against replacing.

playbooks/fedora-30/test-with-podman-stable.yaml
```diff
@@ -0,0 +1,21 @@
...
+        chdir: '{{ zuul.project.src_dir }}'
+
+    - name: Install toolbox
+      become: yes
+      command: ninja -C builddir install
+      args:
+        chdir: '{{ zuul.project.src_dir }}'
+
+    - name: Run system tests
+      command: bats ./test/system
```
debarshiray: Nitpick: we should swap the order of commits so that `./test/system` is in place before we start referring to it.

playbooks/shellcheck.yaml
```diff
@@ -0,0 +1,22 @@
...
+  tasks:
+    - name: Install requirements
+      become: yes
+      package:
+        name:
+          - golang-github-cpuguy83-go-md2man
+          - ninja-build
+          - meson
+          - ShellCheck
+          - bash-completion
```
debarshiray: Nitpick: I think `bash-completion` isn't necessary for `ninja test`. We could potentially save a second or two on every CI run.

playbooks/fedora-rawhide/setup-rpm-env.yaml
```diff
@@ -0,0 +1,25 @@
...
+          - golang-github-cpuguy83-md2man
+          - ninja-build
+          - meson
+          - flatpak-session-helper
+          - systemd
+          - bats
+          - ShellCheck
+          - bash-completion
+          - udisks2
+          - wget
```
debarshiray: Typo? Or do we really need `wget`?

HarryMichal: Very possibly a typo.

PR #251, opened 2019-09-03
Do not mount /run/media when missing

TristanCacqueray:
This change prevents this exception:
Error: error checking path "/run/media": stat /run/media: no such file or directory

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -661,6 +661,7 @@ create()
     monitor_host=""
     no_hosts=""
     toolbox_profile_bind=""
+    run_media_bind=""
 
     # shellcheck disable=SC2153
     if [ "$DBUS_SYSTEM_BUS_ADDRESS" != "" ]; then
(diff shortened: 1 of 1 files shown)


debarshiray:
Thanks for stopping by!

Why do you not have `/run/media`? :) Are you running it straight out of the Git tree?

Toolbox itself [takes care](https://github.com/debarshiray/toolbox/commit/0c848abd1893764f2e0bb62334972d67f95dafd1) of creating it, and these days [UDisks2 creates it](https://github.com/storaged-project/udisks/commit/3b7cf561f42c84a45c4dcd38cf8becb71f23451b) as well. Maybe it's a matter of installing Toolbox system-wide, specifically the `systemd-tmpfiles(8)` snippet, or a newer UDisks2?

I'd rather not add a run-time check for this. We should really be able to rely on `/run/media` being present.

debarshiray:
I missed that it's a duplicate of https://github.com/debarshiray/toolbox/pull/238, which has an interesting use case - Chromebooks!

TristanCacqueray:
This was an issue when running toolbox in cloud instance for zuul ci: https://github.com/debarshiray/toolbox/pull/249

PR #252, opened 2019-09-03
take advantage of podman --userns=keep-id

### LINKED ISSUE #244: Simpler uidmap options for podman
In cfcf4eb you did some nice investigatory work in tracking down how to call `--uidmap` properly to map the users id outside the container to the same id inside rootless podman container. I think we can achieve the same goal now with `podman --userns=keep-id`. At least it seems to be the case in my testing if I remove the `--uidmap` options and replace with `--userns=keep-id`: 

```
[vagrant@vanilla-f30 ~]$ toolbox enter
⬢[vagrant@toolbox ~]$ cat /proc/self/uid_map 
      1000          0          1
         0          1       1000
      1001       1001      64536
⬢[vagrant@toolbox ~]$ exit
logout
[vagrant@vanilla-f30 ~]$ sudi vi /usr/bin/toolbox 
-bash: sudi: command not found
[vagrant@vanilla-f30 ~]$ sudo vi /usr/bin/toolbox 
[vagrant@vanilla-f30 ~]$ 
[vagrant@vanilla-f30 ~]$ toolbox enter
⬢[vagrant@toolbox ~]$ 
⬢[vagrant@toolbox ~]$ cat /proc/self/uid_map 
      1000          0          1
         0          1       1000
      1001       1001      64536

```

Do you agree?

dustymabe:
This lets podman do the calculations for mapping our host UID into
the user namespace within the container. See cfcf4eb for original
context.

Fixes #244

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -751,10 +751,6 @@ create()
         toolbox_profile_bind="--volume /etc/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"
     fi
 
-    max_uid_count=65536
-    max_minus_uid=$((max_uid_count - user_id_real))
-    uid_plus_one=$((user_id_real + 1))
-
     if ! home_canonical=$(readlink --canonicalize "$HOME" 2>&3); then
         echo "$base_toolbox_command: failed to canonicalize $HOME" >&2
         return 1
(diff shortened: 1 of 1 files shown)


debarshiray:
Thanks for the nice patch, @dustymabe !

Merged via https://github.com/debarshiray/toolbox/pull/253

I wanted to add the Podman version information to the commit message, but I somehow I didn't have the permission to update this branch.

PR #253, opened 2019-09-04
Simplify code by taking advantage of 'podman create --userns=keep-id'

debarshiray:
This lets podman do the calculations for mapping the host UID into the
user namespace within the container. See cfcf4eb31e14b3a3 for original
context.

The '--userns=keep-id' option was introduced in Podman 1.4.0, which is
old enough to be in even RHEL 7.

https://github.com/debarshiray/toolbox/issues/244


debarshiray:
Tweaks the commit message of #252 to mention the Podman version information.

PR #255, opened 2019-09-09
Try to unbreak the terminal size when working nested

debarshiray:
When running nested, stty(1) is invoked against the inner
pseudo-terminal pair created by 'podman exec --tty' which may not have
a valid size due to: https://github.com/containers/libpod/issues/3946

In such cases, the COLUMNS and LINES variables set by toolbox(1) in the
outer environment should be forwarded.

This should have been part of commit 05544fb2714ec096.

https://github.com/debarshiray/toolbox/issues/242

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -338,8 +338,6 @@ create_environment_options()
             echo "$base_toolbox_command: failed to parse the number of lines from the terminal size" >&3
             lines=""
         fi
-
-        environment_options="$environment_options --env=COLUMNS=$columns --env=LINES=$lines"
     else
         echo "$base_toolbox_command: failed to read terminal size" >&3
     fi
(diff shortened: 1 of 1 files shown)


debarshiray:
@tagoh what do you think about this?

tagoh:
LGTM

As noted, COLUMNS and LINES aren't the exported variables though, it is always synced up when the terminal window size is changed. so the above approach should works as expected on nested.

debarshiray:
> As noted, COLUMNS and LINES aren't the exported variables though,
> it is always synced up when the terminal window size is changed. so the
> above approach should works as expected on nested.

The issue is that `podman exec --tty` creates a second pseudo-terminal pair that's nested inside the original pseudo-terminal created by the terminal emulator (eg., GNOME Terminal). This inner PTY is the one that doesn't have a valid size. So we run `stty(1)` before `podman exec --tty` to get the size, and then forward it via the variables.

When we invoke `toolbox(1)` from inside a container (ie., a `podman exec --tty` session), `stty(1)` gets run against the inner PTY pair, which might not have a valid size if the terminal window wasn't resized. In that case, the `stty(1)` output will be invalid, and our only hope is to forward the variables.

PR #257, opened 2019-09-09
Give access to the system libvirt instance

debarshiray:
This is useful when the session libvirt instance doesn't offer all the
bells and whistles needed for running virtual machines.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -688,6 +688,7 @@ create()
     flatpak_system_directory_bind=""
     kcm_socket=""
     kcm_socket_bind=""
+    libvirt_system_directory_bind=""
     monitor_host=""
     no_hosts=""
     toolbox_profile_bind=""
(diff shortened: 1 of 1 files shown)


cgwalters:
Been doing this in coretoolbox for a while https://github.com/cgwalters/coretoolbox/blob/47d4d1507521f0d7300e1b6f5c93358f0e1db8be/src/coretoolbox.rs#L35

debarshiray:
> Been doing this in coretoolbox for a while
> https://github.com/cgwalters/coretoolbox/blob/47d4d1507521f0d7300e1b6f5c93358f0e1db8be/src/coretoolbox.rs#L35

Yes, I noticed. Is that all that's needed for system libvirt use? If so, great!

I don't use the system libvirt instance at all myself, so my only test was to start virt-manager and boot a Fedora live ISO.

PR #259, opened 2019-09-13
Added build and install instructions

sudo-give-me-coffee:


README.md
```diff
@@ -39,3 +39,37 @@ This will create a container called `fedora-toolbox-<version-id>`.
...
 ```
+
+## Install dependencies, build and install toolbox
+
+### Install dependencies:
+The following dependencies is required to build and install:
+- meson
+- go-md2man
+- shellcheck
+- bash-completion
```
debarshiray: Don't we also need systemd to build? We have this:
```
tmpfilesdir = systemd_dep.get_pkgconfig_variable('tmpfilesdir')
```

README.md
```diff
@@ -39,3 +39,37 @@ This will create a container called `fedora-toolbox-<version-id>`.
 [user@hostname ~]$ toolbox enter
 ⬢[user@toolbox ~]$
 ```
+
+## Install dependencies, build and install toolbox
+
+### Install dependencies:
```
debarshiray: I wonder if we should also list out the run-time dependencies. eg., currently we require Podman >= 1.4.0 but I don't think that's clearly mentioned anywhere other than in the Git commit messages and Fedora package.

README.md
```diff
@@ -39,3 +39,37 @@ This will create a container called `fedora-toolbox-<version-id>`.
...
+```
+mkdir build
+```
+2. Enter in build directory:
+```
+cd build
+```
+3. Configure `toolbox` using `meson`:
+```
+meson ..
```
debarshiray: I wonder if we should combine the first three steps into one as:
```
$ meson builddir
```

No strong opinion.

However, I think that we should at least mention the `-Dprofile_dir` option. Currenty, most people will need `-Dprofile_dir=/etc/profile.d`.

(The `/etc/profile.d` situation is a bit odd.

Modern distributions try to have clear separation between configuration shipped by the OS vendor and changes made by the local system administrator - somewhere inside `/usr` for the former and `/etc` for the latter. eg., /usr/lib/os-release, /usr/lib/sysctl.d, etc.. However, in most cases, this hasn't happened with the shell / environment variable configuration setup. Most distributions continue to dump everything in `/etc/profile.d`.

Flatpak also had to deal with this: https://github.com/flatpak/flatpak/commit/3b64fbc4f51692ca821c1e197b671ac4b22bf726

We currently default to `/usr/share/profile.d` to be future proof and apply some pressure on distributions, but for distributions still using the old pattern (ie., most of them) we have to use the build flag.)

README.md
HarryMichal: That sounds reasonable. We can use this section of README where we track this information.

debarshiray:
Looks the Git branch for this pull request has since vanished.

Nowadays Toolbox is written in Go. The dependencies and installation instructions need to be updated to reflect that.

debarshiray:
I took the liberty to update the text and merged it via https://github.com/containers/toolbox/pull/492

PR #264, opened 2019-09-18
Expose a few more host locations inside the container under /run/host

debarshiray:
This is meant to alleviate some of the pain of not being able to modify
the list of bind mounts once a toolbox container has been created. For
some cases, where read-only access is enough, one can get by with
setting up symbolic links inside the toolbox container.

Based on an idea from Colin Walters.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -872,6 +872,10 @@ create()
             --volume /dev:/dev:rslave \
             --volume /media:/media:rslave \
             --volume /mnt:/mnt:rslave \
+            --volume /run:/run/host/run:rslave \
+            --volume /tmp:/run/host/tmp:rslave \
+            --volume /usr:/run/host/usr:rslave \
+            --volume /var:/run/host/var:rslave \
             "$base_toolbox_image_full" \
             toolbox --verbose init-container \
                     --home "$HOME" \


toolbox
```diff
@@ -872,6 +872,10 @@ create()
             --volume /dev:/dev:rslave \
             --volume /media:/media:rslave \
             --volume /mnt:/mnt:rslave \
+            --volume /run:/run/host/run:rslave \
```
cgwalters: Hm, but I see you chose `/run/host` instead of `/host`.  Standardizing that path as much as we can is useful because other tools may come to rely on it.  For example, things like Systemtap may want to look at `/host/usr/lib/modules` to find the host kernel.

I chose `/host` because `oc debug node` from OpenShift did.

debarshiray: Interesting. I didn't know OpenShift uses `/host`. I did notice that you had went with `/host`, not `/run/host`.

I went with `/run/host` for no other reason than that we already had it [for resolv.conf](https://github.com/debarshiray/toolbox/commit/04732139074b60dabe3087c37afbd99bdbf14a8a), and the original reason was just that it's the location that Flatpak uses.

I am happy to switch to `/host` or add some symbolic links or whatever.

debarshiray: Let's merge this, and think of `/run/host` versus `/host` in a separate PR or issue.

debarshiray:
> For some reason, I cannot enter a toolbox that was created without that line removed
> (on current F31 Silverblue).

You mean this line:
```
        --volume /usr:/run/host/usr:rslave \
```

We are also discussing it in https://github.com/containers/libpod/issues/4061

debarshiray:
> For some reason, I cannot enter a toolbox that was created without that line
> removed (on current F31 Silverblue).

Should be fixed by https://github.com/debarshiray/toolbox/pull/276

We should roll a new release as soon as possible to address this regression.

debarshiray:
This is also affected by https://github.com/containers/crun/pull/120 on Fedora 31 Silverblue.

PR #269, opened 2019-09-23
Rely on 'podman system migrate' always being present

debarshiray:
The 'podman system migrate' command was added in Podman 1.3.0, which is
within the current minimum required Podman version of 1.4.0.


debarshiray:
Merged via commit 2c77778f04e883895d760eb2cd7ced336fbb8674

PR #274, opened 2019-09-24
Reclaim the terminal size

tagoh:
Some applications (e.g. vi) relies on the terminal size from TTY device.
This changes reclaims the terminal size in environment variables to TTY device
to work the sort of issues around.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -1,4 +1,17 @@
 [ "$BASH_VERSION" != "" ] || [ "$ZSH_VERSION" != "" ] || return 0
+if terminal_size=$(stty size); then
+    columns=$(echo "$terminal_size" | cut --delimiter " " --fields 2)
+    if [ "$columns" = "0" ] && [ -n "$COLUMNS" ]; then
+	stty cols $COLUMNS
+    fi
+    lines=$(echo "$terminal_size" | cut --delimiter " " --fields 1)
+    if [ "$lines" = "0" ] && [ -n "$LINES" ]; then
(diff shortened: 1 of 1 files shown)


debarshiray:
Umm... fixed versions of Podman and conmon are already on their way to the repositories. Maybe we don't need to polish the workaround in Toolbox any further?

tagoh:
Not all of distributions have a solution apparently. that may depends on the time though. I'll leave it to you.

debarshiray:
> Not all of distributions have a solution apparently.

Do you have any specific distribution in mind? Conmon 2.0.1 should be in place for all supported Fedoras at least.

tagoh:
As I've contributed a bit about Chromebook recently, they are based on Debian streatch (old-stable) and I'm getting podman packages installed from Ubuntu bionic PPA where are able to install packages without any breakage. there are no 2.0.1 available so far on their PPA for bionic. dunno why. So I'm still working on 1.5.1.

tagoh:
Oh, wait. it seems working fine now even if "stty size" returns "0 0" on the terminal. as you said, we don't need a workaround anymore. we can close this. thanks.

debarshiray:
> As I've contributed a bit about Chromebook recently, they are based on Debian
> streatch (old-stable) and I'm getting podman packages installed from Ubuntu bionic
> PPA where are able to install packages without any breakage. there are no 2.0.1
> available so far on their PPA for bionic. dunno why. So I'm still working on 1.5.1.

I see. You could bring this up in #podman if the packages don't get updated.

PR #284, opened 2019-10-04
Drop the coloured heading from 'list'

debarshiray:
It seems cleaner to limit the use of colour to only marking running
containers. It's redundant to mention that the containers and images
were created by Toolbox because they are being shown by 'toolbox list'
anyway; and there's a second uncoloured heading in a different case,
that differentiates containers from images.


debarshiray:
Thanks for taking a look, @cgwalters

PR #285, opened 2019-10-04
Remove the hidden --sudo option and the /etc/sudoers.d snippet

debarshiray:
A year ago, when rootless Podman was in its infancy, it was often
necessary to run rootful to test and shake out bugs in Podman. Things
are lot more mature now and this hasn't been necessary in the past few
months. Therefore, it's time to sunset this option.

Removing the --sudo option doesn't break backwards compatibility
because it was never documented or advertised to the user in any way.
It was a hidden option only meant to be used by those hacking on
Toolbox itself.

Note that this is different from running 'sudo toolbox ...', which is a
different use-case and uses separate code paths. This is about running
the rest of toolbox(1) as non-root and only invoking the container
tools like Podman as root.

This reverts commit 66ab4da724d066bdeacaf7e03f027d44ecc37f05.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -53,7 +53,6 @@ environment_variables="COLORTERM \
         XDG_VTNR"
 fgc=""
 
-prefix_sudo=""
 registry="registry.fedoraproject.org"
 registry_candidate="candidate-registry.fedoraproject.org"
 release=""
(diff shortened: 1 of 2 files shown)


cgwalters:
Yes, 100% agreed that tools running `sudo` internally is a major anti-pattern.  (Which in fact https://github.com/coreos/toolbox does)

debarshiray:
@cgwalters Thanks for taking a look!

HarryMichal:
@debarshiray You did it before I had time to look at this :)

debarshiray:
> @debarshiray You did it before I had time to look at this :)

Sorry about that. With things settling down for the Fedora 31 final freeze, I was a bit bored and was looking at what would need to change to make `sudo toolbox ...` work, and I couldn't hold back my urge to remove the `--sudo` flag. :)

PR #289, opened 2019-10-08
Add a '--very-verbose | -vv' option

HarryMichal:
Toolbox only offers a --verbose | -v option that shows debug information
only of Toolbox itself. We have no way of getting debug info right from
Podman without altering the code or running the commands themselves.

This just adds a new variable 'podman' that translates to a string
depending on the presence of '--very-verbose | -v' option. If this
option is used, every Podman command in the code is ran with '--log-level
debug' option. Use wisely, it can be 'very verbose'.

### DIFF
--- a/completion/bash/toolbox
+++ b/completion/bash/toolbox
@@ -28,12 +28,12 @@ __toolbox() {
   _init_completion -s || return
 
   if [ "${COMP_CWORD}" -eq 1 ]; then
-    mapfile -t COMPREPLY < <(compgen -W "--assumeyes --help --verbose $commands" -- "$2")
+    mapfile -t COMPREPLY < <(compgen -W "--assumeyes --help --verbose --very-verbose $commands" -- "$2")
     return 0
   fi
 
   case "$prev" in
-    --verbose | -v)
+    --verbose | -v | --very-verbose | -vv)
       mapfile -t COMPREPLY < <(compgen -W "$commands" -- "$2")
       return 0
       ;;
@@ -52,7 +52,7 @@ __toolbox() {
   esac
 
   local command
-  if [ "${COMP_WORDS[1]}" = --verbose ]; then
+  if [ "${COMP_WORDS[1]}" = --verbose ] || [ "${COMP_WORDS[1]}" = --very-verbose ]; then
     command=${COMP_WORDS[2]}
   else
     command=${COMP_WORDS[1]}
--- a/doc/toolbox.1.md
+++ b/doc/toolbox.1.md
@@ -43,6 +43,7 @@ Print a synopsis of this manual and exit.
 **--verbose, -v**
 
 Print debug information including standard error stream of internal commands.
+Use `-vv` for more detail.
 
 ## COMMANDS
 
--- a/toolbox
+++ b/toolbox
@@ -54,6 +54,7 @@ environment_variables="COLORTERM \
         XDG_VTNR"
 fgc=""
 
+podman_command="podman"
 registry="registry.fedoraproject.org"
 registry_candidate="candidate-registry.fedoraproject.org"
 release=""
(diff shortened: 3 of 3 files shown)


toolbox
```diff
@@ -70,6 +70,7 @@ toolbox_image=""
 toolbox_runtime_directory="$XDG_RUNTIME_DIR"/toolbox
 user_id_real=$(id -ru 2>&3)
 verbose=false
+podman="podman"
```
debarshiray: Nice.

Nitpick: There's no need to describe that you added a `podman` variable to the code in the commit's change log, because it's clearly visible in the `diff`. The change log is mainly useful for things that are not immediately obvious from looking at the code changes. Things like why the change was made, what other approaches were considered and what the trade-offs were, any weird nuances of the code, etc..

toolbox
```diff
@@ -1791,6 +1792,11 @@ while has_prefix "$1" -; do
             exec 3>&2
             verbose=true
             ;;
+        -vv | --very-verbose )
```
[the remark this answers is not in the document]
debarshiray: @matthiasclasen pointed out on IRC that Flatpak allows piling on `-v`:s. It's likely difficult to achieve that in our command line parsing implementation in shell, so we could live with `-v -v` or `--verbose --verbose` in the short-term. A better programming language will enable us to do better.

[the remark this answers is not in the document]
HarryMichal: I personally don't really like the idea of folding `podman --log-level debug` into the existing `--verbose` option because of the noise.

toolbox
HarryMichal: Thanks for the comment. I'll rework the message a bit.

toolbox
```diff
@@ -269,7 +270,7 @@ copy_etc_profile_d_toolbox_to_container()
 
     echo "$base_toolbox_command: copying $toolbox_runtime_directory/toolbox.sh to container $container" >&3
 
-    if ! podman exec \
+    if ! $podman exec \
```
debarshiray: We should quote these. eg., `"$podman".

HarryMichal: As we discussed, is it really necessary? In some places it isn't interpreted well (lines 1287, 1294,..).

debarshiray: Yes, you are right.

I was worried about triggering [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086), but sometimes you just have to silence the warning as you can see (look for `SC2086`) from the code.

completion/bash/toolbox
```diff
@@ -52,7 +52,7 @@ __toolbox() {
   esac
 
   local command
-  if [ "${COMP_WORDS[1]}" = --verbose ]; then
+  if [ "${COMP_WORDS[1]}" = --verbose || "${COMP_WORDS[1]}" = --very-verbose ]; then
```
debarshiray: It should be `[ a ] || [ b ]`, not `[ a || b ]`.

We should enable`shellcheck` for this file. It would have triggered [SC2109](https://github.com/koalaman/shellcheck/wiki/SC2109).

Sorry for not noticing this earlier.

toolbox
debarshiray: While trying to rebase this branch before merging, I realized that it would be handy if the name of the variable was slightly different from just `podman`. eg., if the variable is `podman_command`, then one can quickly grep for `podman ` to locate direct invocations left unconverted in the code.

doc/toolbox.1.md
```diff
@@ -42,7 +42,11 @@ Print a synopsis of this manual and exit.
 
 **--verbose, -v**
 
-Print debug information including standard error stream of internal commands.
+Print toolbox debug information including standard error stream of internal commands.
+
+**--very-verbose, -vv**
```
debarshiray: This wording seems a bit awkward and I can't think of any better way to phrase it. eg., `-vv` might affect more than just `podman` in the future.

Moreover, we want to advertise `-vv` as the mechanism to increase the log level. The fact that it's implemented as a separate `--very-verbose` instead of `--verbose --verbose` is an unimportant detail that's a consequence of various technical decisions.

Once we have better command line parsing facilities (in the rewritten Go version), I'd rather not have a separate `--very-verbose` any more.

Given all that, it seems easier to say something like *use -vv for more*.

debarshiray:
I took the liberty to make the above tweaks and merged to master. This seems like something people will find useful in the short-term, so let's not let it bit rot any longer.

debarshiray:
Thanks for working on this, @HarryMichal !

PR #292, opened 2019-10-10
Enhance bash complete

HarryMichal:


completion/bash/toolbox
```diff
@@ -10,11 +10,10 @@ __toolbox_images() {
 }
 
 __toolbox() {
-  local MIN_VERSION=29
-  local RAWHIDE_VERSION=31
+  local MIN_VERSION=28
```
debarshiray: I think `MIN_VERSION` should continue to be 29 because Fedora 28 is now end-of-life. Thanks for updating `RAWHIDE_VERSION`, though.

HarryMichal: That's true. I didn't realize that.

PR #295, opened 2019-10-11
Add a reset command

debarshiray:
The 'reset' command is meant to factory reset a local Podman and
Toolbox installation. Every now and then early adopters and testers of
Toolbox have to do this when their local Podman state has gotten
irrecoverably broken due to some Podman bug.

It's useful to have a command that encapsulates all the steps to do a
factory reset, as opposed to having to spell them out separately. It's
easier to document, helps with user support, and can enable less opaque
error messages that suggest a way forward when nothing is working.

Note that when running as root, some care has been taken to avoid
removing directories that might be owned by the operating system. eg.,
on Fedora /var/lib/containers/sigstore is owned by the
containers-common RPM.

### DIFF
--- a/completion/bash/toolbox
+++ b/completion/bash/toolbox
@@ -13,7 +13,7 @@ __toolbox() {
   local MIN_VERSION=29
   local RAWHIDE_VERSION=32
 
-  local commands="create enter help init-container list rm rmi run"
+  local commands="create enter help init-container list reset rm rmi run"
 
   declare -A options
   local options=([create]="--candidate-registry --container --image --release" \
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -11,6 +11,7 @@ manuals = [
   'toolbox-init-container.1',
   'toolbox-help.1',
   'toolbox-list.1',
+  'toolbox-reset.1',
   'toolbox-rm.1',
   'toolbox-rmi.1',
   'toolbox-run.1',
--- a/doc/toolbox-reset.1.md
+++ b/doc/toolbox-reset.1.md
@@ -1,0 +1,26 @@
+% toolbox-reset(1)
+
+## NAME
+toolbox\-reset - Remove all local podman (and toolbox) state
+
+## SYNOPSIS
+**toolbox reset**
+
+## DESCRIPTION
+
+Removes all existing podman (and toolbox) containers, images and configuration.
+This can be used to factory reset your local Podman and Toolbox installations
+when something has gone irrecoverably wrong with the `podman(1)` and
+`toolbox(1)` commands.
+
+This command can only be used on the host, and not from within a toolbox
+container, and is only expected to be used right after a fresh boot before any
+other `podman(1)` or `toolbox(1)` commands have been invoked.
+
+## EXAMPLES
+
+### Reset a broken Podman and Toolbox installation
+
+```
+$ toolbox reset
+```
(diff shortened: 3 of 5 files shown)


debarshiray:
@rhatdan @mheon @giuseppe could one of you please sanity check that I have got it right or haven't missed anything?

rhatdan:
I think this would be better done within podman, since the configuation files could have been changed for the location of storage.

`podman system reset`

But until that command exists, I think you should also clean out the XDG_RUNTIME_ container directories.

debarshiray:
> I think this would be better done within podman, since the configuation files could have
> been changed for the location of storage.
> 
> `podman system reset`

Ok, makes sense.

> But until that command exists, I think you should also clean out the XDG_RUNTIME_ container
> directories.

I saw `/run/user/1000` in `podman info` but I don't see a `/run/user/1000/containers` on my filesystem. So I am confused. Which path is it exactly?

rhatdan:
This is a case of us not doing a good job os separating.

Do you see /run/user/1000/overaly*

debarshiray:
> This is a case of us not doing a good job os separating.
> 
> Do you see /run/user/1000/overaly*

Yes, I see those! Thanks @rhatdan 

What's the best way to remove ~/.local/share/containers other than using `sudo`? `podman unshare ...` sometimes works, but at other times errors out as:
```
rm: cannot remove '/home/rishi/.local/share/containers/storage/overlay/7fdb6e40e7e92a74f245405325dae00ae53335dbc6de644dac3bddbc2e86f452/merged': Device or resource busy
```

Also, what are the chances of `podman unshare` working if things in in `~/.local/share/containers` and `~/.config/containers` are badly borked?

debarshiray:
Filed https://github.com/containers/libpod/issues/4240 for implementing this in Podman itself.

rhatdan:
Most likely there are running containers, which would block it.  

The reset should force stop all containers and run a cleanup on eachone to make sure everything is in as good of shape as possible.  Then it should attempt to rm -rf .

debarshiray:
> The reset should force stop all containers and run a cleanup on each one to make
> sure everything is in as good of shape as possible. Then it should attempt to rm -rf .

I thought I had added `podman stop --all` to `reset` but I see that it's missing. That might actually be a good thing because can we expect `podman stop` to work when things are completely borked? Maybe we should suggest the user to reboot to stop all containers and terminate all `exec` sessions?

Also, it seems like `podman unshare` relies on having access to a working OCI runtime, which can't be guaranteed under the circumstances where people might be using `reset`. Maybe I should look at using `unshare` and `newuidmap` instead?

giuseppe:
> Also, it seems like `podman unshare` relies on having access to a working OCI runtime, which can't be guaranteed under the circumstances where people might be using `reset`. Maybe I should look at using `unshare` and `newuidmap` instead?

`podman unshare` shouldn't need one.  What error have you seen without a working OCI runtime?

mheon:
Unshare needs to make a libpod.Runtime (not entirely sure why) and that
won't work without access to conmon, an OCI runtime, etc.

debarshiray:
For rootless `reset`, I think I have managed to replace `podman share` with a combination of `unshare`, `newgidmap`, `newuidmap` and `flock`; and I am side-stepping the use of `podman stop` by asking the user to reboot if `$XDG_RUNTIME_DIR/overlay*` is present.

The rootful case needs more testing.

PR #298, opened 2019-10-13
First attempt at an Ubuntu toolbox container image.

kronenpj:
This isn't as clean as the fedora/centos variants. There's also a problem on first run where toolbox
emits 'toolbox: failed to create /run/.toolboxenv in container' but runs normally from there.


images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM docker.io/library/ubuntu:19.04
+
+ENV NAME=ubuntu-toolbox VERSION=19.04
+LABEL com.github.debarshiray.toolbox="true" \
```
debarshiray: We should add `com.github.containers.toolbox="true"` to the list. See commit de5e5df9b793edc746cde7125cd437b69f15ccd7.

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+      com.redhat.component="$NAME" \
```
debarshiray: `com.redhat.component` shouldn't be necessary anymore. I'd suggest removing it and filing an issue or PR if something breaks.

Moreover, it refers to the Fedora bugzilla and I doubt we will be filing bugs against the Ubuntu image on the Fedora bugzilla. :)

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+      name="$FGC/$NAME" \
```
debarshiray: I suspect that this should be `name="$NAME"`.

`FGC` is a Fedora thing. It used to be recommended by a previous version of the [Fedora container guidelines](https://docs.fedoraproject.org/en-US/containers/guidelines/creation/), apparently not anymore.

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+      version="$VERSION" \
+      usage="This image is meant to be used with the toolbox command" \
+      summary="Base image for creating Ubuntu 19.04 toolbox containers" \
+      maintainer="Debarshi Ray <rishi@fedoraproject.org>"
```
debarshiray: May I convince you to be the maintainer of these Ubuntu images? :)

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+RUN apt-get update
+
+RUN apt-get -y upgrade
+
+COPY extra-packages /
+RUN apt-get -y install $(cat extra-packages | xargs)
+RUN rm /extra-packages
+
+RUN sed -i -e 's/ ALL$/ NOPASSWD:ALL/' /etc/sudoers
```
debarshiray: Is this really necessary? Toolbox itself calls `passwd --delete root` inside the container to delete the `root` password.

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+      maintainer="Debarshi Ray <rishi@fedoraproject.org>"
+
+COPY README.md /
+
+RUN apt-get update
+
+RUN apt-get -y upgrade
+
+COPY extra-packages /
+RUN apt-get -y install $(cat extra-packages | xargs)
```
debarshiray: Does this not work:
```
RUN apt-get -y install $(<extra-packages)
```

Would be nice to avoid needless divergences in the Dockerfiles.

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN rm /extra-packages
+
+RUN sed -i -e 's/ ALL$/ NOPASSWD:ALL/' /etc/sudoers
+
+# Restore documentation
+RUN yes | /usr/local/sbin/unminimize
+RUN apt-get clean
+
+RUN echo VARIANT_ID=container >> /etc/os-release
```
debarshiray: Interesting. Do the Ubuntu base images have any `VARIANT_ID`? If so, what's it?

images/ubuntu/19.04/extra-packages
```diff
@@ -0,0 +1,18 @@
+bash-completion
+git
+keyutils
+libcap2-bin
+lsof
+man-db
+mlocate
+mtr
+packagekit-command-not-found
```
debarshiray: `packagekit-command-not-found` shouldn't be included. See commit 0dee71b13f71fca9a37c1595f73a71727cafb67f

images/ubuntu/19.04/Dockerfile
kronenpj: Unfortunately yes. For some reason I haven't figured out - it's required for sudo to work.

images/ubuntu/19.04/Dockerfile
kronenpj: Probably not, I only use Centos and Fedora. Debian and its derivatives aren't anything I work with normally. I really just did this one as a demonstration.

images/ubuntu/19.04/Dockerfile
kronenpj: It's not part of the base image, and toolbox barfs if it's not there.

images/ubuntu/19.04/extra-packages
kronenpj: Agreed, it doesn't work anyway.

images/ubuntu/19.04/Dockerfile
kronenpj: Nope, it skips over the line because $(<extra-packages) ends up being empty. My guess is that there's a difference in the bash versions / behaviors.

images/ubuntu/19.04/Dockerfile
kronenpj: Agreed. See upcoming update to PR.

images/ubuntu/19.04/Dockerfile
kronenpj: :) That would be a bad idea..

images/ubuntu/19.04/Dockerfile
nanonyme: I think toolbox is wrong here, VARIANT_ID is an optional field.

evan-a-a:
I just spent about an hour putting together a Ubuntu toolbox container - then I found this! Looks like we came to the same conclusions on what's needed to make Ubuntu happy inside of podman/toolbox.

I do have an issue with sudo complaining about not being able to resolve the `toolbox` hostname. I'm guessing this is due to it not being listed in `/etc/hosts`, but it's the same for Fedora and no complaints there. Ubuntu must be more strict about hostnames in /etc/hosts.

nanonyme:
About /etc/hosts https://github.com/containers/toolbox/blob/master/toolbox#L1079 it's supposed to get indirected. Maybe check if it's already a symlink in the image?

images/ubuntu/19.04/Dockerfile
debarshiray: Ok.

Before we announce support for Ubuntu, we would need somebody who cares enough about Ubuntu to ensure that Toolbox keeps working on a day to day basis with Ubuntu hosts and images. I am a little apprehensive of taking up that responsibility myself. I could, but it seems wise to spread the load around a bit.

debarshiray: Anyway, in the meantime, we can work towards removing the rough edges before thinking of making any announcements. :)

images/ubuntu/19.04/Dockerfile
debarshiray: Yes, the `VARIANT_ID` shouldn't be critical - the Toolbox logic doesn't need it to be.

Feel free to file a separate PR to fix the code to not barf.

images/ubuntu/19.04/extra-packages
```diff
@@ -0,0 +1,17 @@
+bash-completion
```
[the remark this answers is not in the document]
kronenpj: That's part of the reason I don't want to maintain the Ubuntu toolbox image. I don't actually use the distribution, so I can't say what a "usual" user experience would be for it.

images/ubuntu/19.04/README.md
```diff
@@ -0,0 +1,41 @@
+<img src="data/logo/toolbox-logo-landscape.svg" alt="Toolbox logo landscape" width="800"/>
```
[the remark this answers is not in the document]
kronenpj: It does hit the buildah bug on my box. One difficulty with living on the edge eh? :)
However, I just downgraded buildah to 1.7-18 and it gave the same secure join error.
Also, using docker to create the image, I get a more descriptive / interesting error:
COPY failed: Forbidden path outside the build context: README.md ().

images/ubuntu/19.04/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM docker.io/library/ubuntu:19.04
+
+ENV NAME=ubuntu-toolbox VERSION=19.04
+LABEL com.github.containers.toolbox="true" \
```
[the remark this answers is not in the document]
kronenpj: I've restored the label in the PR.

images/ubuntu/19.04/Dockerfile
kronenpj: Sounds good to me.

images/ubuntu/19.04/Dockerfile
kronenpj: I'll experiment a bit with omitting that one, but it'll be a PR of its own.

evan-a-a:
> About /etc/hosts https://github.com/containers/toolbox/blob/master/toolbox#L1079 it's supposed to get indirected. Maybe check if it's already a symlink in the image?

It's being properly indirected inside the container, but `/etc/hosts` contains the hostname "dockerhost" and not "toolbox".

markstos:
Having a Ubuntu container could help driving conversions from Ubuntu users. It looks like this has been pending a review since early November. Supporting a Ubuntu LTS container would be great too.

markstos:
I tested this image as follows:

 1. I downloaded the files in images/ubuntu/19.04 to my home directory on Silverblue 31.
 2. I ran `podman build images/ubuntu/19.04`. (I tried running `podman build images/19.04/Dockerfile`, but that segfault'ed)
 3. I ran `podman images` to confirm the image was built and what the image ID was.
 4. I ran `toolbox create --container ubuntu-disco --image 703d46da649e` to use the local image
 5. I ran `toolbox enter --container ubuntu-disco` to enter the local container.
        
Unfortunately, the container exits immediately with an return value of 127. According to `man podman-exec`, this return status means `contained command cannot be found`.

Here's the final output of running the `toolbox enter` command with `sh -x`:

```
+ podman exec --interactive --tty --user mark --workdir /var/home/mark --env=COLORTERM=truecolor --env=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus --env=DESKTOP_SESSION=gnome --env=DISPLAY=:0 --env=LANG=en_US.UTF-8 --env=SHELL=/bin/bash --env=SSH_AUTH_SOCK=/run/user/1000/keyring/ssh --env=TERM=xterm-256color --env=VTE_VERSION=5803 --env=WAYLAND_DISPLAY=wayland-0 --env=XDG_CURRENT_DESKTOP=GNOME --env=XDG_DATA_DIRS=/var/home/mark/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop --env=XDG_MENU_PREFIX=gnome- --env=XDG_RUNTIME_DIR=/run/user/1000 --env=XDG_SESSION_DESKTOP=gnome --env=XDG_SESSION_TYPE=wayland --env=COLUMNS=158 --env=LINES=37 ubuntu-disco capsh --caps= -- -c 'exec "$@"' /bin/sh /bin/bash -l
+ ret_val=127
+ true
+ printf '\033]777;container;pop;;\033\'
+ exit 127
+ exit
```

kronenpj:
@markstos: Looking into your concern. I can't get toolbox to create a container from the image number:
```
$ toolbox create --container ubuntu-disco --image e29401ae3372
toolbox: failed to get RepoTag for base image e29401ae3372
```
Once I tag that image, I can create the toolbox container:
```
$ podman tag e29401ae3372 ubuntu-tool
$ toolbox create --container ubuntu-disco --image ubuntu-tool
Created container: ubuntu-disco
Enter with: toolbox enter --container ubuntu-disco
```
The container then "works for me:"
```
$ toolbox enter --container ubuntu-disco
⬢[kronenpj@toolbox toolbox]$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="19.04 (Disco Dingo)"
...

I'm not sure what's going on here. @debarshiray thoughts?

**Comment by markstos** [comment]
@kronenpj I got the image ID from the output of `podman images`. Is that where you got the value that failed as well?

**Comment by kronenpj** [comment]
@markstos Yes, I did leave that step out. I think it was actually `podman image ls` but it's the same output.

**Comment by martinpitt** [comment]
For anyone who wants to use ubuntu containers with toolbox now: I [have a script](https://piware.de/gitweb/?p=bin.git;a=blob;f=build-debian-toolbox) that can build a debian or ubuntu container for any release from the stock dockerhub images.

**Comment by markstos** [comment]
@martinpitt Thanks. I see a comment in there about systemd. Does systemd work inside these containers?

I also see that `libwww-perl` is being installed as a dev tool. What's it used for these days? I used to install it to use the `GET` and `HEAD` binaries, but these days `httpie` seems like better HTTP CLI tool.

**Comment by martinpitt** [comment]
@markstos : systemd is not really being used in the containers. It's just commonly getting installed as a build dependency, and /var/log/journal/ being a bind mount to the real system confuses the package postinst script, so the package fails to install.

I forgot about the particular reason for libwww-perl, I think a tool like dget or dput needs it. I reduced dependencies with `--no-install-recommends`, but that was one which at least I need. You can of course try with dropping deps that you don't need, and just install them later on demand.

**Comment by mjlbach** [comment]
@martinpitt When I use your script to try to make an ubuntu container, I get the following error:
```
❯ bash make_debian_container.sh 18.04 ubuntu
+ RELEASE=18.04
+ DISTRO=ubuntu
+ toolbox rm -f 18.04
+ toolbox -y create -c 18.04 --image docker.io/ubuntu:18.04
Created container: 18.04
Enter with: toolbox enter --container 18.04
+ podman start 18.04
Error: unable to start container "18.04": error looking up supplemental groups for container 86982a6dffded63fb77cfe6c74a80e08f349012672d0de7ee9466817a93993fb: Unable to find group wheel
```

martinpitt:
@mjlbach : Indeed I get the same now, on any Debian/Ubuntu release. That smells like a recent podman regression, I'll look into it this evening.

heyakyra:
What's the current status of this?

j1mc:
I'm grateful everyone's work on this thus far. I know that Ubuntu might not be Redhat / Fedora folk's primary target here, but any update would be appreciated.

reesericci:
> What's the current status of this?

looking for the same

acmel:
I'm also interested in running Ubuntu in a container on fedora silverblue, any update on this?

Jmennius:
Please try images from #483 PR - they should be more complete and few recent LTS releases are available.

debarshiray:
Hey @kronenpj did you receive my email to your `kronenpj@gmail.com` inbox earlier this year to work out the specifics of offering out-of-the-box Ubuntu support? It was titled *Toolbox: Ubuntu suport*, and was from `rishi.is@lostca.se`.

images/ubuntu/19.04/Dockerfile
debarshiray: I see. The other option is to add the `nullok` option to the `PAM(8)` configuration.

debarshiray:
> Hey @kronenpj did you receive my email to your `kronenpj@gmail.com`
> inbox earlier this year to work out the specifics of offering out-of-the-box
> Ubuntu support? It was titled _Toolbox: Ubuntu suport_, and was from
> `rishi.is@lostca.se`.

My apologies. I just remembered that you had already mentioned on this pull request that you don't want to maintain the Ubuntu support because you don't actually use it and you only did it as a demo.

I have marked those conversation threads as *unresolved* so that they don't get hidden in the noise.

debarshiray:
I am closing this pull request since @kronenpj doesn't want to maintain the Ubuntu support long-term, and we can't merge stuff that has nobody maintaining it.

debarshiray:
Let's focus on https://github.com/containers/toolbox/pull/483 which is more active. :)

Issue #301, opened 2019-10-14
OpenShift Prow onboarding: Create OWNERS file to use


cgwalters:
The CoreOS group is [increasingly using OpenShift Prow](https://github.com/coreos/fedora-coreos-tracker/issues/263).

If you wish to use the OpenShift Prow instance, please [create an `OWNERS` file](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md) in this repository to indicate agreement.  

[Here's an example](https://github.com/containers/libpod/blob/master/OWNERS) from this Github organization.

If you do not wish to have this repository onboarded to OpenShift Prow, simply close this issue.

cgwalters:
For background on this, I think having a "merge bot" is really useful, and Prow is a fine instance of it.  The OpenShift CI cluster also can indirectly sponsor projects with CI resources via using its Prow instance.

debarshiray:
Yes, sure. I added an `OWNERS` file in https://github.com/containers/toolbox/pull/302

Two things came to my mind.

@HarryMichal did a bunch of work adding continuous integration tests using [Zuul](https://zuul-ci.org/) to test Toolbox changes against Podman and vice versa in https://github.com/containers/toolbox/pull/250 They have fallen off our radar due to the ongoing chaos caused by Cgroups v2, but we should finalize them soon. How are they affected by OpenShift Prow?

Should we synchronize the contents of `OWNERS` with the current roles assigned in the project settings? The project settings have evolved in a somewhat ad hoc and organic way. eg., some of the admins are only there to allow moving issues to other components in the *containers* organization. Maybe we should clean that up by having one of us be an admin in the *containers* organization?

HarryMichal:
> For background on this, I think having a "merge bot" is really useful, and Prow is a fine instance of it. The OpenShift CI cluster also can indirectly sponsor projects with CI resources via using its Prow instance.

I'm a bit against having a `merge bot`. Using merge bot could lead to some unnecessary commit junk. I think our current strategy of handling PRs is quite alright (as in speed) even though it's done manually. And if we split the responsiblity in the future then it'll be even more fine.

cgwalters:
The argument for a merge bot is multiple.  In Project Atomic CI we used Homu, see
https://github.com/servo/homu#why-is-it-needed

It's not quite about speed so much as automating rules around merging.

debarshiray:
Ping.

Where do we stand on this? Shall we merge https://github.com/containers/toolbox/pull/302 ?

[...]

PR #302, opened 2019-10-14
Add OWNERS for OpenShift Prow

debarshiray:
https://github.com/containers/toolbox/issues/301

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,5 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+  - cgwalters
+  - debarshiray


cgwalters:
Let's check if anyone else in the coreos/ and containers/ organizations wants to pitch in?  I think at least some of them should be approvers.

cgwalters:
For example several people from the coreos group contributed to https://github.com/coreos/toolbox/commits/master

cgwalters:
Filed https://github.com/coreos/toolbox/issues/60 to ping there

OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+  - cgwalters
+  - debarshiray
+reviewers:
```
cgwalters: Contrary to what the Prow docs imply, it's not necessary to duplicate approvers into reviewers.  If you don't want a separate set at all (which I don't think we do right now, since they're the same, we can just drop `reviewers:`).
Example:
https://github.com/openshift/machine-config-operator/blob/master/OWNERS

debarshiray: Ok, sure. I had copied [libpod's example](https://github.com/containers/libpod/blob/master/OWNERS), so probably the documentation isn't at fault. :)

debarshiray:
> Let's check if anyone else in the coreos/ and containers/ organizations
> wants to pitch in? I think at least some of them should be approvers.

Ok, sure.

I just put something together to avoid blocking anything. Feel free to handle it whichever way you prefer.

HarryMichal:
I'm closing this as we never got to merge this before. I kinda understand the motivation behind this but I don't think it to be very beneficial.

PR #303, opened 2019-10-16
Import coretoolbox

cgwalters:
We are aiming to combine efforts.  I'd like to merge coretoolbox
into this repository so that it's convenient to use as a reference
when we work on really merging them.

Also, this way there's only one place to review pull requests
and the future is clear.


cgwalters:
Hm, I was trying this https://nuclearsquid.com/writings/subtree-merging-and-you/ but it apparently didn't work.  Retrying...

cgwalters:
Ah, I think I see.  Github is telling me this PR can't be fast-forwarded.  But, a repo administrator should be able to merge?

That said, I think it may be best to verify this all works with some disposable test repositories first. I'm going to play with that.

Alternatively, we could toss the history, but...there's some useful stuff there.

dustymabe:
> That said, I think it may be best to verify this all works with some disposable test repositories first. I'm going to play with that.

how did that go?

debarshiray:
Thanks for the pull request, @cgwalters 

I am fine with whichever trade-off works for you. In the worst case, as we absorb more and more changes, we can explain everything in the new commit messages.

PR #305, opened 2019-10-17
Make 'podman exec' wait for container init

HarryMichal:
In the 'run' script are two main Podman commands. 'podman start' and
'podman exec'. 'podman start' is used for container initalization using
the 'init_container' function. But because this function is executed in
a subshell the next call doesn't wait for it to finish. Such an issue
isn't very common but it may happen.

This solves the potential issue by using a dirty hack. A simple while
loop that'll block the function until a 'block file' is deleted by the
preceeding process.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -753,7 +753,7 @@ unshare_userns_rm()
     unshare --user sh -c "flock $unshare_directory/map rm --force --recursive $path" 2>&3 &
     unshare_pid="$!"
 
-    echo "$base_toolbox_command: setting UID map of user namespace" >&3
+    echo "$base_toolbox_command: setting GID and UID map of user namespace" >&3
 
     if ! newgidmap "$unshare_pid" 0 "$user_id_real" 1 1 "$userns_gid_start" "$userns_gid_len" 2>&3; then
         echo "$base_toolbox_command: failed to set GID mapping of user namespace" >&2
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -1085,6 +1085,12 @@ EOF
         fi
     fi
 
+    echo "$base_toolbox_command: removing blocker file" >&3
+
+    cont_toolbox_runtime_dir="/run/host/run/user/1000/toolbox"
+
+    rm --force "$cont_toolbox_runtime_dir"/toolbox_init_lock 2>&3
```
debarshiray: Umm... since `XDG_RUNTIME_DIR` is shared between the host and all containers, this means that different `podman start` invocations by different instances of `toolbox enter` can step on each others toes by deleting the same file. That would be bad. :)

I think it's a practical solution to use the same file for all containers. Otherwise, if it was an unique path for each container, then we'd have to somehow pass the path to the entry-point during `podman create` and then dig it back up in `run()`. That's a bit painful to implement.

However, as long as we have a global path for locking, we have to somehow prevent different instances from colliding against each other. Maybe we do need `flock` here after all.

debarshiray: Backwords compatibility is another hairy issue. Older Toolbox containers didn't have `toolbox init-container` as their entry-point. They won't have anything to cancel the waiting, and cause `toolbox enter` to get stuck. We need to handle this case.

debarshiray:
I am merging this for the time being, so that we can put this out as a zero-day update for the upcoming Fedora 31 release. However, comments, improvements, etc. welcome.

PR #308, opened 2019-10-23
README: add "Goals and Use Cases" section

dustymabe:


dustymabe:
cc @debarshiray @cgwalters @tpopela @matthiasclasen @miabbott

README.md
```diff
@@ -24,6 +24,65 @@ image. On Fedora this is the `fedora-toolbox` image. This image is used to
 create a toolbox container that seamlessly integrates with the rest of the
 operating system.
 
+## Goals and Use Cases
```
debarshiray: How about putting the `Goals and Use Cases` section below `Usage`? Just so that first-time readers and newcomers get a quick practical introduction before getting into the more nitty-gritty details.

Not a very strong opinion, though. :)

dustymabe: yeah I thought about that, but it seemed like we are jumping right to how to use it without explaining what it is for first. I'm open to moving it.

@cgwalters @matthiasclasen - any opinion here?

matthiasclasen: the suggested change is fine with me

dustymabe: ok I moved it to after `Usage`!

miabbott:
LGTM!

I left the one comment around the types of containers users would run via `toolbox`, but don't think it is enough to block merging.

README.md
```diff
@@ -41,6 +41,65 @@ This will create a container called `fedora-toolbox-<version-id>`.
 ⬢[user@toolbox ~]$
 ```
 
+## Goals and Use Cases
+
+### High Level Goals
+
+- Provide a CLI convenience interface to run containers (via `podman`) easily
```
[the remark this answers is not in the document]
dustymabe: yeah we can probably be more specific here.. probably saying "pet containers" or "snowflake containers" or something.

PR #318, opened 2019-10-31
Rewrite in Go

HarryMichal:
This is where the Go rewrite is happening. I welcome any kind of comment/contribution.

The shell script has been renamed to toolbox.sh so that toolbox.go can
be compiled.

The new codebase is based on [Cobra](https://github.com/spf13/cobra) which takes care of the CLI heavy lifting.


cmd/create.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Here you will define your flags and configuration settings.
+
+	// Cobra supports Persistent Flags which will work for this command
+	// and all subcommands, e.g.:
+	// createCmd.PersistentFlags().String("foo", "", "A help for foo")
+
+	// Cobra supports local flags which will only run when this command
+	// is called directly, e.g.:
+	createCmd.Flags().StringVarP(&candidateRegistry, "canidate-registry", "", false, "Pull the base image from candidate-regostry.fedoraproject.org")
```
cgwalters: Let's not copy this one from shell-toolbox; I think we should just support pulling any base image and not special case "the" toolbox image too much.

cmd/create.go
```diff
@@ -0,0 +1,59 @@
...
+func init() {
+	rootCmd.AddCommand(createCmd)
+
+	// Here you will define your flags and configuration settings.
+
+	// Cobra supports Persistent Flags which will work for this command
+	// and all subcommands, e.g.:
+	// createCmd.PersistentFlags().String("foo", "", "A help for foo")
+
+	// Cobra supports local flags which will only run when this command
```
cgwalters: These comments aren't very useful.

cgwalters:
So first:  Let's try to get a vote among contributors here - do you really *really* want Go?  I think the code in https://github.com/containers/toolbox/pull/319 is pretty good, Rust's "structopt" is a lot nicer than anything Go can do, etc.  If you don't know Rust (and don't have the time to learn), that's a totally valid argument.  It's also valid that podman is in Go.

But I'd hope people at least take a careful look at the toolbox-rs code and think about it!

But I write Go a lot, and I just really dislike the language.  The "Go strength" in web services doesn't apply here at all.

On the other hand, hopefully this project doesn't exceed 2000 lines, so it doesn't matter too much.

cmd/create.go
dustymabe: would a config file with a default be useful?

dustymabe:
> So first: Let's try to get a vote among contributors here - do you really _really_ want Go? I think the code in #319 is pretty good, Rust's "structopt" is a lot nicer than anything Go can do, etc. 

> But, if you don't know Rust (and don't have the time to learn), that's a totally valid argument against. 

I really want to learn more rust and I like the concept of the language.

> Go _is_ very easy to learn. It's also valid that podman is in Go.

I think these points are the biggest one in favor of Go and why we have centered on go in previous discussions.

> 
> But I'd hope people at least take a careful look at the toolbox-rs code and think about it!

+1 - I looked at it earlier today. I don't quite understand it, because I don't know rust, but it at least looks nice.

> 
> But I write Go a lot, and I just really dislike the language. The "Go strength" in web services doesn't apply here at all.

Understood.

> 
> On the other hand, hopefully this project doesn't exceed 2000 lines, so it doesn't matter too much.

Agree, Either one should be fine implmentations for this. We should hoist advanced features back into podman.

cmd/create.go
HarryMichal: Indeed. Those were created by the generator and I didn't remove them, yet.

cmd/create.go
HarryMichal: Ok, I was just mirroring the options of containers/toolbox.

evelineraine:
Switching to either opens opportunities for me to contribute to the code (I can't write bash).
Regarding the choice I feel the same as @dustymabe. In a vacuum, I would prefer Rust over Go, because I really like it's design principles (unlike Go's) and toolbox would be a perfect opportunity for me to practice Rust.

However, toolbox not in a vacuum and entire container domain is almost Go-exclusive, which is not exactly good, but means moving toolbox to Go would attract more contributions from people that already work on container projects and also new people (Go is really easy to get started with, after all). Podman being in Go is really important point.

Still, it seems to me an even better option is a scripting/interpreted language. As a wrapper around a complete container solution (podman), I'm not sure what benefit there will be from features of a compiled language like Rust or Go (performance, advanced memory management, complex data representations, direct system calls, etc.).
On the other hand, a scripting language allows easier inspection of the code and easier changing of it, lowering entry-barrier for new people to contribute. And really, until toolbox supports tons of runtime customization, being able to easily mess with the code easily is very beneficial.

I would make many points in favor of Python, but as I understood because of CoreOS it's a no-go. *Is it the case?*

dustymabe:
> I would make many points in favor of Python, but as I understood because of CoreOS it's a no-go. _Is it the case?_

It is true that the CoreOS team has a goal of keeping Python out of the base. See https://github.com/coreos/fedora-coreos-tracker/blob/master/Design.md#approach-towards-shipping-python

cmd/create.go
cgwalters: To rephrase here:

Right - part of my point here is I don't think we should think of this as "rewriting original shell toolbox in Go" - it's more like "merge our toolbox efforts".  Using original shell toolbox for inspiration of course makes total sense, but I'd like to *also* use as a reference the code in https://github.com/containers/toolbox/pull/319

HarryMichal: That's my intention. I'm still getting to know Go and Cobra so I took the thing I knew the most and started to experiment.

evelineraine:
> It is true that the CoreOS team has a goal of keeping Python out of the base. See https://github.com/coreos/fedora-coreos-tracker/blob/master/Design.md#approach-towards-shipping-python

I see, thanks for the reference. Python is not an option then *(unless you feel like freezing binaries)* :(
So to me it seems Go has more advantages if starting from scratch. At the same time there is [cgwalters/coretoolbox](https://github.com/cgwalters/coretoolbox) which is an already functional implementation. Between the two, I'm not sure what I would choose. Very curious to see how this decision goes.

giuseppe:
If you pick Go, I'd also suggest to use podman directly from the CLI, or alternatively through varlink.

The `libpod` version vendored by `toolbox` may be different than what Podman uses.  I don't think we are ready for supporting two versions of `libpod` that run and manage the same storage.

dustymabe:
> The `libpod` version vendored by `toolbox` may be different than what Podman uses. I don't think we are ready for supporting two versions of `libpod` that run and manage the same storage.

Yes. Agree. This is something we had previously discussed as well and we agreed to not vendor libpod.

morucci:
recheck

stbischof:
Is there a Timeline for this pr? It looks complete. Would be nice to have it in fedora silverblue silverblue 32.

HarryMichal:
> Is there a Timeline for this pr? It looks complete. Would be nice to have it in fedora silverblue silverblue 32.

It's awaiting review.

src/go.mod
```diff
@@ -0,0 +1,29 @@
...
+	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
+	github.com/briandowns/spinner v1.9.0
+	github.com/cpuguy83/go-md2man/v2 v2.0.0
+	github.com/fatih/color v1.9.0 // indirect
+	github.com/go-ole/go-ole v1.2.4 // indirect
+	github.com/godbus/dbus/v5 v5.0.3
+	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
+	github.com/mattn/go-colorable v0.1.6 // indirect
+	github.com/mcuadros/go-version v0.0.0-20190830083331-035f6764e8d2
+	github.com/mitchellh/go-homedir v1.1.0
```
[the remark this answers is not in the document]
HarryMichal: You are right. We don't need it.

debarshiray:
Some of the error messages and diagnostics might have to be adjusted.

For example, currently we have:
```
$ toolbox 
toolbox: missing command

These are some common commands:
create    Create a new toolbox container
enter     Enter an existing toolbox container
list      List all existing toolbox containers and images

Try 'toolbox --help' for more information.
```

But now it is:
```
$ toolbox 
Toolbox is a tool that offers a familiar RPM based environment for
developing and debugging software that runs fully unprivileged using Podman.

Usage:
  toolbox [command]

Available Commands:
  create         Create a new toolbox container
  enter          Enter a toolbox container for interactive use
  help           Help about any command
  list           List existing toolbox containers and images
  rm             Remove one or more toolbox containers
  rmi            Remove one or more toolbox images
  run            Run a command in an existing toolbox container

Flags:
  -y, --assumeyes          Automatically answer yes for all questions
  -h, --help               help for toolbox
      --log-level string   Log messages above specified level: trace, debug, info, warn, error, fatal or panic (default "warn")
      --log-podman         Show the log output of Podman. The log level is handled by the log-level option
      --version            version for toolbox

Use "toolbox [command] --help" for more information about a command.
```

This is a somewhat cosmetic issue, but we already spent some effort in the past polishing the diagnostics, so it would be good to not lose them.

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+	Args: cobra.MaximumNArgs(1),
+}
+
+func init() {
+	rootCmd.AddCommand(createCmd)
+
+	flags := createCmd.Flags()
+	flags.StringVarP(&createFlags.image, "image", "i", "", "Change the name of the base image used to create the toolbox container")
+	flags.StringVarP(&createFlags.release, "release", "r", "", "Create a toolbox container for a different operating system release than the host")
+	flags.BoolVar(&createFlags.unsafe, "unsafe", false, "Allow the use of a non-Toolbox-compliant image")
```
debarshiray, line 98: I am curious about the `--unsafe`` flag? Was there a specific reason for it?

If really necessary it will be better to add it through a separate PR to keep the Go rewrite as conservative in scope as possible.

src/cmd/run.go
```diff
@@ -0,0 +1,334 @@
...
+						if response == "y" || response == "yes" {
+							createContainer = true
+						}
+
+						break
+					}
+				}
+
+				if !createContainer {
+					logrus.Fatal("A container can be created later with the 'create' command.")
```
debarshiray, line 142: This is showing up as `FATAL` when `toolbox enter` is invoked without any existing containers. It should be a "normal" message.

src/cmd/root.go
```diff
@@ -0,0 +1,227 @@
...
+	rootCmd.PersistentFlags().StringVar(&rootFlags.loglevel, "log-level", "warn", "Log messages above specified level: trace, debug, info, warn, error, fatal or panic")
+	rootCmd.PersistentFlags().BoolVar(&rootFlags.logPodman, "log-podman", false, "Show the log output of Podman. The log level is handled by the log-level option")
+	viper.BindPFlag("log-podman", rootCmd.PersistentFlags().Lookup("log-podman"))
+	// This flag is kept for compatibility reasons. In the future it would be better removed.
+	rootCmd.PersistentFlags().BoolVar(&rootFlags.verbose, "verbose", false, "Set log-level to 'debug'")
+	rootCmd.PersistentFlags().MarkDeprecated("verbose", "use 'log-level' instead.")
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
```
debarshiray, line 123: Do we need this? We don't have a config file at the moment.

HarryMichal:
> Some of the error messages and diagnostics might have to be adjusted.
> 
> For example, currently we have:
> 
> ```
> $ toolbox 
> toolbox: missing command
> 
> These are some common commands:
> create    Create a new toolbox container
> enter     Enter an existing toolbox container
> list      List all existing toolbox containers and images
> 
> Try 'toolbox --help' for more information.
> ```
> 
> But now it is:
> 
> ```
> $ toolbox 
> Toolbox is a tool that offers a familiar RPM based environment for
> developing and debugging software that runs fully unprivileged using Podman.
> 
> Usage:
>   toolbox [command]
> 
> Available Commands:
>   create         Create a new toolbox container
>   enter          Enter a toolbox container for interactive use
>   help           Help about any command
>   list           List existing toolbox containers and images
>   rm             Remove one or more toolbox containers
>   rmi            Remove one or more toolbox images
>   run            Run a command in an existing toolbox container
> 
> Flags:
>   -y, --assumeyes          Automatically answer yes for all questions
>   -h, --help               help for toolbox
>       --log-level string   Log messages above specified level: trace, debug, info, warn, error, fatal or panic (default "warn")
>       --log-podman         Show the log output of Podman. The log level is handled by the log-level option
>       --version            version for toolbox
> 
> Use "toolbox [command] --help" for more information about a command.
> ```
> 
> This is a somewhat cosmetic issue, but we already spent some effort [in the past](https://github.com/containers/toolbox/issues/59) polishing the diagnostics, so it would be good to not lose them.

The new "help" screen is currently the default format set by Cobra. It is possible to create our own format that could be more structured so that it provides better onboarding experience.

We could make two help outputs for the base command:
1. User just types `toolbox`: custom structured with categories/pointers/anything to help the onboarding
2. User types `toolbox --help`: the current full list of commands/flags + description

Maybe this is not a good approach because it introduces two different screens which could be confusing.

Overall I agree that Toolbox should help more with the onboarding process but still, it is a command-line tool that has a considerable set of options that the user should have easy access to them/info about them.

src/cmd/root.go
HarryMichal: No, we don't really need it now but it could be useful in the future. Until that time comes it can be omitted.

src/cmd/run.go
HarryMichal: Nice catch! That should be printed with fmt.Println()

src/cmd/create.go
HarryMichal: People try to use non-toolbox images to create containers. It's not critical but something that can help with testing new images.

But surely we could separate it.

src/cmd/list.go
```diff
@@ -0,0 +1,178 @@
...
+		logrus.Fatal(err)
+		logrus.Fatalf("%+v", err)
+	}
+
+	return err
+}
+
+func GetContainers() ([]map[string]interface{}, error) {
+	logrus.Info("Fetching containers with label=com.github.debarshiray.toolbox=true")
+	args := []string{"-a", "--filter", "label=com.github.debarshiray.toolbox=true"}
```
debarshiray, line 88: Typo alert! Should be `label=com.redhat.component=fedora-toolbox"`

src/cmd/list.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	containers := utils.JoinJSON("ID", Dcontainers, Ccontainers)
+	containers = utils.SortJSON(containers, "Names", false)
+	return containers, err
+}
+
+func GetImages() ([]map[string]interface{}, error) {
+	logrus.Info("Fetching images with label=com.github.debarshiray.toolbox=true")
+	args := []string{"--filter", "label=com.github.debarshiray.toolbox=true"}
```
debarshiray, line 110: Ditto.

debarshiray:
> > Some of the error messages and diagnostics might have to be adjusted.
> >
> > [...]
> >
> > This is a somewhat cosmetic issue, but we already spent some effort
> > [in the past](https://github.com/containers/toolbox/issues/59) polishing the diagnostics,
> > so it would be good to not lose them.
> 
> The new "help" screen is currently the default format set by Cobra. It is possible to create
> our own format that could be more structured so that it provides a better onboarding experience.
> 
> We could make two help outputs for the base command:
> 
>     1. User just types `toolbox`: custom structured with categories/pointers/anything to help
> the onboarding
> 
>     2. User types `toolbox --help`: the current full list of commands/flags + description

I was able to accomplish this with Cobra's [SetHelpFunc](https://gowalker.org/github.com/spf13/cobra#Command_SetHelpFunc) and [SetUsageFunc](https://gowalker.org/github.com/spf13/cobra#Command_SetUsageFunc).

src/cmd/list.go
```diff
@@ -0,0 +1,178 @@
...
+
+func init() {
+	rootCmd.AddCommand(listCmd)
+
+	flags := listCmd.Flags()
+	flags.BoolVarP(&listFlags.listContainers, "containers", "c", false, "List only toolbox containers, not images")
+	flags.BoolVarP(&listFlags.listImages, "images", "i", false, "List only toolbox images, not containers")
+}
+
+func list(cmd *cobra.Command, args []string) error {
```
debarshiray, line 51: The `error` thrown by this function doesn't seem to be handled by anything. Maybe we should drop the `error`?

src/pkg/podman/podman.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// CheckVersion compares provided version with the version of Podman.
+//
+// Takes in one string parameter that should be in the format that is used for versioning (eg. 1.0.0, 2.5.1-dev).
+//
+// Returns 0 if required version equals Podman version
+// Returns 1 if required version is higher than Podman version
+// Returns -1 if required version is lower than Podman version
+func CheckVersion(requiredVersion string) int {
```
debarshiray, line 57: While it's nice that the underlying `version` package follows the same convention as C's `strcmp`, I think `CheckVersion` should be a simpler boolean check like [GLIB_CHECK_VERSION](https://developer.gnome.org/glib/stable/glib-Version-Information.html#GLIB-CHECK-VERSION:CAPS).

The vast majority of times (possibly always?) we are only interested in checking if Podman is new enough or not, and we don't really care whether the versions exactly match or not. This means we no longer care about a tristate reply, but just a boolean is enough.

If nothing else, the code becomes easier to read. :)

src/cmd/root.go
```diff
@@ -0,0 +1,224 @@
...
+	return nil
+}
+
+func migrate() error {
+	configDir, err := os.UserConfigDir()
+	if err != nil {
+		return fmt.Errorf("Could not get the user config directory: %w", err)
+	}
+	toolboxConfigDir := fmt.Sprintf("%s/toolbox", configDir)
+	migrateStampPath := fmt.Sprintf("%s/podman-system-migrate", configDir)
```
debarshiray, line 176: The file locking seems to have gone missing. We probably don't want multiple `toolbox` processes to step on each other's toes trying to migrate.

src/cmd/initContainer.go
```diff
@@ -0,0 +1,363 @@
...
+				logrus.Error(err)
+			}
+		}
+
+		// TODO: flatpak-session-helper cannot be used over dbus as root; some kind of workaround has to be implemented
+		if utils.PathExists("/run/host/monitor") {
+			logrus.Info("Path /run/host/monitor exists. Mount binding to that path will happen now.")
+			localtimeTarget, err := os.Readlink("/etc/localtime")
+			if err != nil || localtimeTarget != "/run/host/monitor/localtime" {
+				err = redirectPath("/run/host/monitor/localtime", "/etc/localtime", false)
```
debarshiray, line 144: The semantics of source and target seems to have gotten switched. We redirect `/etc/localtime` to `run/host/monitor/localtime`.

src/cmd/root.go
```diff
@@ -0,0 +1,224 @@
...
+		return fmt.Errorf("Could not get the version of Podman: %w", err)
+	}
+	logrus.Debugf("Current Podman version is %s", podmanVersion)
+
+	err = os.MkdirAll(toolboxConfigDir, 0664)
+	if err != nil {
+		return fmt.Errorf("Configuration directory not created: %w", err)
+	}
+
+	output, err := ioutil.ReadFile(migrateStampPath)
```
debarshiray, line 190: I wonder if we should run the contents of the file through [strings.TrimSpace](https://golang.org/pkg/strings/#TrimSpace)

src/cmd/root.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
```
debarshiray, line 192: Are you sure it isn't `os.PathError`? I don't get a `os.ErrNotExist` when the file doesn't exist.

src/cmd/root.go
```diff
@@ -0,0 +1,224 @@
...
+		logPodman bool
+		assumeyes bool
+		verbose   bool
+	}
+	rootCmd = &cobra.Command{
+		Use:   "toolbox",
+		Short: "Unprivileged development environment",
+		Long: `Toolbox is a tool that offers a familiar RPM based environment for
+developing and debugging software that runs fully unprivileged using Podman.`,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
debarshiray, line 47: Would be better to use a separate top-level function for this. If nothing else, the code wouldn't be so heavily indented.

src/cmd/list.go
HarryMichal: True. But shouldn't all three labels be used?

src/cmd/list.go
HarryMichal: Yes, we should drop it.

src/cmd/list.go
```diff
@@ -0,0 +1,178 @@
...
+
+var listFlags struct {
+	listContainers bool
+	listImages     bool
+}
+
+var listCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List existing toolbox containers and images",
+	Run: func(cmd *cobra.Command, args []string) {
```
debarshiray, line 38: We should switch to the `RunE` hook, and keep throwing the errors up the call stack instead of logging them immediately using `logrus`. That way, Cobra will present the error message in a consistent way to the user and we can use the right exit code.

Currently, what's happening is that some errors get logged twice, or errors from the command line parsing are presented differently than other kinds of errors and so on.

src/cmd/rmi.go
```diff
@@ -0,0 +1,147 @@
...
+
+		var idKey string
+		if podman.CheckVersion("1.8.2") < 0 {
+			idKey = "ID"
+		} else {
+			idKey = "id"
+		}
+		images := utils.JoinJSON(idKey, Dimages, Cimages)
+		for _, image := range images {
+			logrus.Infof("Deleting image %s", image[idKey].(string))
```
debarshiray, line 75: Isn't `image[idKey]` already a `string`? Is the cast necessary?

src/cmd/rmi.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		for _, imageID := range args {
+			// Check if the container exists
+			logrus.Infof("Inspecting image %s", imageID)
+			args := []string{"inspect", "--format", "json", "--type", "image", imageID}
+			output, err := podman.CmdOutput(args...)
+			if err != nil {
+				if errors.As(err, &podman.ErrInternal) {
+					logrus.Fatalf("Image '%s' does not exist", imageID)
```
debarshiray, line 99: Even though this is going to be correct way many more times than not, we shouldn't be guessing like this. `podman inspect` doesn't use a specific exit code to denote missing images. We should just throw a generic error, unless we can find a way to exactly catch the *image not found* case.

src/cmd/initContainer.go
```diff
@@ -0,0 +1,363 @@
...
+		logrus.Fatal("Failed to create /run/.toolboxenv")
+	}
+
+	if initContFlags.monitorHost {
+		logrus.Info("Monitoring host")
+
+		if utils.PathExists("/run/host/etc") {
+			logrus.Info("Path /run/host/etc exists. Mount binding to that location will happen now.")
+
+			err = mountBind("/run/host/etc/machine-id", "/etc/machine-id", "ro")
```
debarshiray, line 101: With a *real* programming language we ought to be using things like arrays and `for` loops. :)

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+
+		if podman.ImageExists(imageName) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func checkIfToolboxImage(imageName string) (bool, error) {
```
debarshiray, line 417: This looks like something that could be used to implement `rmi`. Maybe we should split it out into `pkg/utils`?

src/cmd/root.go
HarryMichal: It's possible. I tried the first error that seemed to match and it worked.

src/cmd/root.go
HarryMichal: Right!

src/cmd/list.go
HarryMichal: Yes, you are right. `logrus` should only be used for logging debug information (with different levels of severity) but it shouldn't be used to control the flow of the program (methods `Fatal` and `Panic`).

src/cmd/rmi.go
HarryMichal: Children of `image` are of type `interface{}` which is a vague type that can store anything -> no generics. So the cast is necessary (though I did not cast it in some cases). In some cases casting is not needed (for `logrus` or `fmt`) but I would call it safer.

src/pkg/utils/utils.go
```diff
@@ -0,0 +1,283 @@
...
+	"os"
+	"os/exec"
+	"os/user"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/containers/toolbox/pkg/podman"
+	"github.com/shirou/gopsutil/host"
```
debarshiray, line 15: Does `github.com/shirou/gopsutil/host` use `os-release`? Maybe we should use `github.com/acobaugh/osrelease` instead?

src/pkg/utils/utils.go
```diff
@@ -0,0 +1,283 @@
...
+			if release == "rawhide" {
+				release = "32"
+			}
+		} else {
+			release = viper.GetString("RELEASE_DEFAULT")
+		}
+	}
+
+	if imageName == "" {
+		if hostPlatform == "fedora" {
```
debarshiray, line 173: Isn't `imageName` the same for both *fedora* and *!fedora* ? Or am I misreading the code? :)

src/pkg/utils/utils.go
HarryMichal: Fine by me. `github.com/shirou/gopsutil/host` was the first thing that did what I needed but it may be unnecessarily big for our use case.

src/pkg/utils/utils.go
HarryMichal: I'm not sure if I follow here. Could you, please, elaborate?

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+	logrus.Info("Checking if toolbox.sh profile exists")
+	if utils.PathExists("/etc/profile.d/toolbox.sh") {
+		logrus.Info("Found /etc/profile.d/toolbox.sh")
+
+		toolboxProfileBind = []string{"--volume", "/etc/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"}
+		createArgs = append(createArgs, toolboxProfileBind...)
+	} else if utils.PathExists("/usr/share/profile.d/toolbox.sh") {
+		logrus.Info("Found /usr/share/profile.d/toolbox.sh")
+
+		toolboxProfileBind = []string{"--volume", "/usr/share/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"}
```
debarshiray, line 218: `toolboxProfileBind` doesn't get appended in this code path.

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+
+var (
+	createFlags struct {
+		image   string
+		release string
+		unsafe  bool
+	}
+	ulimitHost               = []string{}
+	homeCanonical            = ""
+	homeLink                 = ""
```
debarshiray, line 46: Do these variables really need to be global?

I understand the desire to make `createFlags` and `preservedEnvVars` global - the flags apply to the entire command and there is value in clearly advertising the list of environment variables. However, the rest are just implementation details as to how we cobble together the `podman create ...` call.

It's already bad enough that all top-level symbols inside a Go package are global within the package, so maybe we shouldn't be making it worse. :)

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+	mediaLink                = ""
+	mntLink                  = ""
+	toolboxProfileBind       = []string{}
+	sudoGroup                = ""
+	kcmSocket                = ""
+	kcmSocketBind            = []string{}
+	usrMountPoint            = ""
+	usrMountSourceFlags      = ""
+	usrMountDestinationFlags = "ro"
+	dbusSystemBusAddress     = ""
```
debarshiray, line 56: Maybe this should be initialized with the default system bus address, which is `unix:path=/var/run/dbus/system_bus_socket` ?

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+	if err != nil {
+		logrus.Error(err)
+	}
+
+	dbusSystemBusAddressBind := []string{"--volume", fmt.Sprintf("%s:%s", dbusSystemBusAddress, dbusSystemBusAddress)}
+	createArgs = append(createArgs, dbusSystemBusAddressBind...)
+
+	logrus.Info("Preparing sssd-kcm socket")
+	args = []string{"show", "--value", "--property", "Listen", "sssd-kcm.socket"}
+	output, err := utils.SystemctlOutput(args...)
```
debarshiray, line 291: Now that we are using a *proper* programming language, we should use systemd's D-Bus API to find out the location of the KCM socket, instead of parsing `systemctl`. D-Bus APIs are meant to be consumed programmatically, as opposed to command line utilities which are often targetted towards humans and their outputs can subtly change.

Podman uses `github.com/coreos/go-systemd`, so we could try that.

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+
+	logrus.Info("Checking if 'podman create' supports option '--ulimit host'")
+	if podman.CheckVersion("1.5.0") {
+		logrus.Info("Option '--ulimit host' is supported")
+		ulimitHost = []string{"--ulimit", "host"}
+	} else {
+		logrus.Info("Option '--ulimit host' is not supported")
+	}
+
+	homeEnv := strings.Split(viper.GetString("HOME"), "=")[0]
```
debarshiray, line 327: Is the `strings.Split` really necessary? A few lines above, it's not split.

Either way, we could use `user.Current().HomeDir`, can't we? We already query `user.Current()` for the UID.

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+	conn, err := dbus.SessionBus()
+	if err != nil {
+		logrus.Error("Failed to connect to Session Bus")
+	}
+	defer conn.Close()
+
+	logrus.Info("Calling org.freedesktop.Flatpak.SessionHelper.RequestSession")
+	SessionHelper := conn.Object("org.freedesktop.Flatpak", "/org/freedesktop/Flatpak/SessionHelper")
+	call := SessionHelper.Call("org.freedesktop.Flatpak.SessionHelper.RequestSession", 0)
+	if call.Err != nil {
```
debarshiray, line 353: Can we immediately check the `Err` immediately after creating the `Call` object. [It might be](https://godoc.org/github.com/godbus/dbus#Call) that we need to wait for the call to complete by using the `Done` channel.

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+		logrus.Fatal("Failed to call org.freedesktop.Flatpak.SessionHelper.RequestSession")
+	}
+
+	// Add the environment variables that hold a value
+	for _, env := range preservedEnvVars {
+		value := viper.GetString(env)
+		if len(value) != 0 {
+			createArgs = append(createArgs, fmt.Sprintf("--env=%s=%s", env, value))
```
debarshiray, line 361: This is meant to be done only in `run`, not `create`. Or am I missing something?

HarryMichal: I think the duplication is also in Shell Toolbox. I kept it in. We can drop it.

src/cmd/create.go
debarshiray: After actually trying out the D-Bus bindings, it turned out to be the `Store` method on the `Call` object.

playbooks/fedora-31/setup-rpm-env.yaml
```diff
@@ -5,20 +5,15 @@
...
+          - golang
 
     - name: Setup environment (create missing /run/media)
       command: sudo systemd-tmpfiles --create
 
     - name: Check versions of crucial packages
-      command: rpm -q podman crun conmon fuse-overlayfs flatpak-session-helper
-
-    - include_tasks: ./pre-common.yaml
+      command: rpm -q podman crun conmon fuse-overlayfs flatpak-session-helper golang
```
elimisteve, line 19: Might you consider adding CRIU so we can create snapshots/container state that can survive a reboot and/get moved between hosts?

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+
+	// Inside of a toolbox we want to be able to access dbus for using flatpak-spawn and for users, who work with dbus.
+	logrus.Info("Preparing dbus system bus address")
+	dbusSystemBusAddress = strings.Split(viper.GetString("DBUS_SYSTEM_BUS_ADDRESS"), "=")[1]
+	dbusSystemBusAddress, err = filepath.EvalSymlinks(dbusSystemBusAddress)
+	if err != nil {
+		logrus.Error(err)
+	}
+
+	dbusSystemBusAddressBind := []string{"--volume", fmt.Sprintf("%s:%s", dbusSystemBusAddress, dbusSystemBusAddress)}
```
elimisteve, line 286: `dbusSystemBusAddress` could equal empty string here, which may not be what you want :-)

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+		logrus.Error("Failed to evaluate if /home is a symbolic link")
+	}
+	if homeSymPath == "/var/home" {
+		logrus.Info("/home is a symbolic link to /var/home")
+		command = append(command, "--home-link")
+	}
+
+	conn, err := dbus.SessionBus()
+	if err != nil {
+		logrus.Error("Failed to connect to Session Bus")
```
elimisteve, line 346: Considering how much `conn` is used below, perhaps this error should be fatal?

playbooks/fedora-31/setup-rpm-env.yaml
HarryMichal: This is not related to this PR. For a feature request open a new issue, please :).

src/cmd/run.go
```diff
@@ -0,0 +1,335 @@
...
+			} else {
+				logrus.Fatal("Create a new container with 'toolbox create' or 'toolbox enter' an already existing container")
+			}
+		} else {
+			logrus.Fatalf("Container '%s' not found", containerName)
+		}
+	}
+
+	// Prepare Flatpak session-helper
+	conn, err := dbus.SessionBus()
```
debarshiray, line 159: This could be factored out into a function, because the code is shared with `create.go`.

src/cmd/run.go
```diff
@@ -0,0 +1,335 @@
...
+		logrus.Fatal(err)
+	}
+
+	containerConfig := containerInfo["Config"].(map[string]interface{})
+	containerState := containerInfo["State"].(map[string]interface{})
+	containerEntryPoint := containerConfig["Cmd"].([]interface{})[0]
+
+	logrus.Infof("Entry point of container '%s' is '%s'", containerName, containerEntryPoint)
+
+	if containerEntryPoint == "toolbox" {
```
debarshiray, line 190: We should simply reject containers that don't have `toolbox` itself as their entry point. Such containers have been marked as deprecated since the 0.0.17 release in November 2019.

src/cmd/run.go
```diff
@@ -0,0 +1,335 @@
...
+			}
+		}
+		logrus.Infof("Container '%s' is properly initialized", containerName)
+	} else {
+		logrus.Warnf("Container '%s' uses deprecated features", containerName)
+		logrus.Warn("Consider recreating it with Toolbox version 0.0.17 or newer")
+	}
+
+	logrus.Info("Checking if the current working directory is in the container")
+	args = []string{"exec",
```
debarshiray, line 225: Nitpick: it would be nice to not repurpose the function's `args` parameter for some internal use. It changes the semantics of the variable, which would be fine if it was some loop counter or such, but an incoming argument is a bit more important. :)

src/cmd/run.go
```diff
@@ -0,0 +1,335 @@
...
+		"--user", viper.GetString("USER"),
+		containerName,
+		"sh", "-c", `command -v "$1"`, "sh", commands[0]}
+	err = podman.CmdRun(args...)
+	if err != nil {
+		if runFlags.fallbackToBash {
+			logrus.Infof("%s not found in '%s'; using /bin/bash instead", commands[0], containerName)
+			commands = nil
+			commands = []string{"/bin/bash"}
```
debarshiray, line 245: Do we really need to set `commands` to `nil` before this?

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+	preservedEnvVars         = []string{
+		"COLORTERM",
+		"COLUMNS",
+		"DBUS_SESSION_BUS_ADDRESS",
+		"DBUS_SYSTEM_BUS_ADDRESS",
+		"DESKTOP_SESSION",
+		"DISPLAY",
+		"LANG",
+		"LINES",
+		"SSH_AUTH_SOCK",
```
debarshiray, line 66: Aren't we missing `SHELL`?

src/cmd/create.go
```diff
@@ -0,0 +1,490 @@
...
+		"TERM",
+		"USER",
```
debarshiray, line 68: Aren't we missing `TOOLBOX_PATH`?

Is there a reason why we need `SHELL`? I am curious because we weren't propagating it into the container thus far.

src/cmd/run.go
```diff
@@ -0,0 +1,335 @@
...
+		"--user", viper.GetString("USER"),
+		"--workdir", viper.GetString("PWD")}
+
+	if !runFlags.notty {
+		args = append(args, "--interactive", "--tty")
+	}
+
+	// Add the environment variables that hold a value
+	for _, env := range preservedEnvVars {
+		value := viper.GetString(env)
```
debarshiray, line 263: We should differentiate between environment variables that are unset, and those that are set to the empty string.

debarshiray:
Are we missing the `flatpak-spawn`-based *forward-to-host* code paths?

HarryMichal:
> Are we missing the `flatpak-spawn`-based _forward-to-host_ code paths?

Yes, we currently do. I was not sure how to handle it...

src/cmd/run.go
HarryMichal: Ah! That is true. I didn't realize that when I wrote it.

src/cmd/create.go
HarryMichal: They don't need to be. When I started with the rewrite, I did not know how should I handle those variables. So I left them to be global. I guess that's what 'viper' could be used for.

src/cmd/create.go
HarryMichal: I think we're not. `SHELL` is passed to the 'init-container' command and there it is set during user creation.

src/cmd/create.go
HarryMichal: `TOOLBOX_PATH` is being added manually (line 187). It could be passed together with other env vars.

`USER` came from [coretoolbox](https://github.com/cgwalters/coretoolbox/blob/master/src/coretoolbox.rs#L45).

src/cmd/create.go
HarryMichal: Good catch!

abitrolly:
Is it possible to apply incremental approach to this PR? I asked on the Go forum how to do that. Still quiet there.

https://forum.golangbridge.org/t/incremental-transition-from-bash-to-go-cmdline-api/18693

This will make the development process more distributed and allow people like to get the fixes while other can have their fun porting the code to Go, Rust or whatever language they like.

HarryMichal:
Hi @abitrolly! That was the original intention of how the rewrite was supposed to go. But I made the call to do the rewrite without including calls to the shell version (which I now see as a mistake).

At this stage, I don't think it's worth it because the rewrite is on my part already done. @debarshiray  is currently doing the review and making the PR prettier. Hopefully, the new version will come out soon.

src/cmd/run.go
```diff
@@ -0,0 +1,335 @@
...
+
+				if !createContainer {
+					fmt.Println("A container can be created later with the 'create' command")
+					return nil
+				}
+
+				create([]string{containerName})
+			} else if containersCount == 1 {
+				containerName = fmt.Sprint(containers[0]["Names"])
+				logrus.Infof("Entering container %s instead", containerName)
```
debarshiray, line 149: We should fallback like this only when the container that we were trying to run was the default one.

The idea behind this was, if somebody only created a specific non-default toolbox container, then we can let her be slightly lazier by letting `toolbox enter` automatically fallback to her custom container. Of course, there's always a chance that this wasn't the desired outcome, and hence the diagnostic.

However, if someone really went out of their way to ask for something specific, then it's safe to assume that they had good reasons to do so, and this sort of fallback will be more annoying than helpful.

src/pkg/utils/utils.go
```diff
@@ -0,0 +1,283 @@
...
+
+// GetCgroupsVersion returns the cgroups version of the host
+//
+// The information is taken from the output of `podman info` command
+func GetCgroupsVersion() (string, error) {
+	podmanInfo, err := podman.PodmanInfo()
+	if err != nil {
+		return "", err
+	}
+	cgroupVersion := fmt.Sprint(podmanInfo["host"].(map[string]interface{})["CgroupVersion"])
```
debarshiray, line 32: Podman 1.9.0 has changed the JSON key to `cgroupVersion`. :/

I think we will be better off using code similar to what Podman itself uses to check the Cgroups version. @mheon pointed me to `IsCgroup2UnifiedMode` in `libpod/pkg/cgroups`.

debarshiray:
I think that we should still retain the `reset` command to avoid getting tangled up in tracking down all the documentation across the Internets and updating them. The Go implementation could just be a wrapper around `podman system reset` with a deprecation notice on top.

src/go.mod
```diff
@@ -0,0 +1,28 @@
...
+	github.com/mattn/go-colorable v0.1.6 // indirect
+	github.com/pelletier/go-toml v1.6.0 // indirect
+	github.com/shirou/gopsutil v2.20.2+incompatible
+	github.com/sirupsen/logrus v1.4.2
+	github.com/spf13/afero v1.2.2 // indirect
+	github.com/spf13/cast v1.3.1 // indirect
+	github.com/spf13/cobra v0.0.6
+	github.com/spf13/jwalterweatherman v1.1.0 // indirect
+	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/spf13/viper v1.6.2
```
debarshiray, line 23: I think that we should avoid the dependency on `github.com/spf13/viper`, because that reduces the size of the `toolbox` binary from 14 MB to 7.6 MB - almost by half.

In many cases, we can replace it with `os.Getenv` and `os.Setenv`.Then we can have a `podman.SetLogLevel` function that updates a global variable to implement the Podman logging.

Ultimately, we don't need a fully featured configuration management framework, because we almost don't have any settings at the moment. It can, of course, change in the future.

src/cmd/rm.go
```diff
@@ -0,0 +1,142 @@
...
+
+	flags := rmCmd.Flags()
+	flags.BoolVarP(&rmFlags.deleteAll, "all", "a", false, "Remove all toolbox containers")
+	flags.BoolVarP(&rmFlags.forceDelete, "force", "f", false, "Force the removal of running and paused toolbox containers")
+}
+
+func rm(args []string) error {
+	if rmFlags.deleteAll {
+		logrus.Info("Fetching containers with label=com.github.debarshiray.toolbox=true")
+		args := []string{"--filter", "label=com.github.debarshiray.toolbox=true"}
```
debarshiray, line 54: Typo alert! The `--all` went missing. :)

src/cmd/root.go
```diff
@@ -0,0 +1,230 @@
...
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.PersistentFlags().BoolVarP(&rootFlags.assumeyes, "assumeyes", "y", false, "Automatically answer yes for all questions")
+	rootCmd.PersistentFlags().StringVar(&rootFlags.loglevel, "log-level", "warn", "Log messages above specified level: trace, debug, info, warn, error, fatal or panic")
+	rootCmd.PersistentFlags().BoolVar(&rootFlags.logPodman, "log-podman", false, "Show the log output of Podman. The log level is handled by the log-level option")
+	viper.BindPFlag("log-podman", rootCmd.PersistentFlags().Lookup("log-podman"))
+	// This flag is kept for compatibility reasons. In the future it would be better removed.
+	rootCmd.PersistentFlags().BoolVar(&rootFlags.verbose, "verbose", false, "Set log-level to 'debug'")
```
debarshiray, line 122: We could turn `--verbose` into a *count* flag to trivially retain compatibility with `-vv`, abbreviated form of `--very-verbose`.

src/pkg/podman/podman.go
```diff
@@ -0,0 +1,317 @@
...
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	if logLevel == "debug" || logLevel == "trace" {
+		fmt.Fprint(os.Stderr, stderr.String())
+	}
+
+	if err != nil {
+		if strings.Contains(stderr.String(), "invalid status code from registry 503 (Service Unavailable)") || strings.Contains(stderr.String(), "received unexpected HTTP status: 503 Service Temporarily Unavailable") {
+			return ErrServiceUnavailable
```
debarshiray, line 224: We shouldn't rely on looking for sub-strings in the error messages, unless these strings have been blessed as API. If these are just *informal* user-facing strings then they might get tweaked in a future release and break our error handling.

Instead, we should [add a specific exit code](https://github.com/containers/libpod/issues/6190) to `podman pull` to denote network failures, which can then be considered as API.

It's good to be paranoid about these error handling code paths. They don't get exercised under normal use, and sometimes it's not trivial to trigger these code paths even just for testing. So, if things silently break, then we are less likely to notice, and users who face the error in the wild would have even less of a clue.

HarryMichal: Yes, the sub-string as an error is not the most optimal way of handling this. I filed some time ago an issue that pointed out this behaviour but it was closed: https://github.com/containers/libpod/issues/5288.

abitrolly:
No way! :D

PR #319, opened 2019-10-31
Import cgwalters/coretoolbox, add `TOOLBOX_VARIANT option

cgwalters:
This is per the decision to merge the projects.  Add a mechanism
to "dispatch" transiently or persistently to `/usr/libexec/toolbox-rs`
so people can easily try it out.

Basically: `mkdir -p ~/.config/toolbox && echo rs > ~/.config/toolbox/variant`

We'll ship both, and then a 3rd new version later.


meson.build
```diff
@@ -35,3 +35,11 @@ endif
 subdir('data')
 subdir('doc')
 subdir('profile.d')
+
+#mytarget = custom_target('toolbox-rs',
```
cgwalters: I don't know meson, tried only briefly to figure out how to glue this together but we can just stick it in the spec file or wrap it in a `Makefile`.

debarshiray:
Does the Rust code actually work?

I had tried `coretoolbox` in the past with older Podman versions (eg., `podman-1.6.2-2.fc30` and a few builds older than that) and `toolbox run` kept erroring out for me on Fedora. Today I tried again with `podman-1.7.0-0.4.rc2.fc31.x86_64` and got:
```
$ ~/.cargo/bin/toolbox-rs run
mount: /home/rishi: special device /host/home/rishi does not exist.
error: Initializing container (static): Child ["mount" "--rbind" "/host/home/rishi" "/home/rishi"] exited: exit code: 32
Error: non zero exit code: 1: OCI runtime error
```

I don't recall if this is the exact same error as the ones with older Podman versions, but I can try again, if you want.

(Sorry, I thought I had already left a comment some weeks back, but apparently I didn't.)

debarshiray:
It's the same error that @HarryMichal reported in https://github.com/cgwalters/coretoolbox/issues/3#issuecomment-548917810

njam:
Ha! I didn't know there's a version written in Rust. I like `run_pid1`().

I implemented my own experimental version in Rust: https://gitlab.com/njam/toolbox-njam
It's not 1:1 feature compatible, rather I'm using it to experiment with some ideas (like mounting working directories other than $HOME).

Anyways, looking forward to the Go-rewrite, hopefully once done new features and bug fixes can be implemented in toolbox :)

debarshiray:
The Go version of Toolbox is alive and thriving. So I don't think this is relevant anymore.

Please feel free to re-open if you have reason to think otherwise.

PR #328, opened 2019-11-07
Arch Toolbox container image

erazemkokot:
I'm creating an Arch base for toolbox.

Currently I get the error `toolbox: failed to get the tag of image ./Dockerfile` and I'm new to Podman/Buildah, so I don't know how to tag the file or if I should even use either of those to make the image.

* I've taken most of the packages from the Fedora images, but have replaced them with Arch equivalents.
* The VERSION variable doesn't really apply to Arch, so for now I've set it to "rolling", but any input on this would be very helpful.

Todo:

* [x] Check if other packages are needed due to Arch changing [base](https://www.archlinux.org/news/base-group-replaced-by-mandatory-base-package-manual-intervention-required/)
* [ ] Test image


images/arch/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+ENV NAME=arch-toolbox VERSION=rolling
+LABEL com.github.containers.toolbox="true" \
+      com.github.debarshiray.toolbox="true" \
+      name="$FCG/$NAME" \
+      version="$VERSION" \
+      usage="This image is meant to be used with the toolbox command" \
+      summary="Base image for creating Arch toolbox containers" \
+      maintainer="Erazem Kokot <contact at erazem dot eu>"
+
+RUN pacman -Syu
```
HarryMichal: This also needs `--noconfirm` flag.

images/arch/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Install packages from the extra-packages file
+COPY extra-packages /
+RUN pacman -Sy --noconfirm $(<extra-packages)
```
HarryMichal: I would most likely include the 'base' group you mentioned in your comment.

images/arch/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN rm /extra-packages
+
+# Clean up all local caches
+RUN pacman -Scc
+
+CMD /bin/sh
```
HarryMichal: I tried to build the image and when I got to this point, it errored out with `no such command`. That'll probably need some investigation??

images/arch/extra-packages
```diff
@@ -0,0 +1,27 @@
+bash-completion
```
HarryMichal: If you were to include the 'base' group, then most likely some packages can be removed from this list.

erazemkokot:
@HarryMichal Thanks for the reply. Can you please tell me what I need to do to properly run the container, since I get the error in the first comment and can't even start it.

images/arch/Dockerfile
erazemkokot: I've tested pacman -Scc and it works fine, did you mean no such command for /bin/sh or pacman?

HarryMichal: For `CMD /bin/sh`.

HarryMichal:
Disclaimer: I'm no expert at building container images. I'm still discovering this.

First, you need to build the image before you can use it with Toolbox. My process looked like this:
```
$ cd <toolbox_dir>/images/arch
$ podman build .
```

The podman command can be replaced by `buildah build-using-dockerfile`. But this command doesn't work for me even with the already existing fedora Dockerfiles, so there is most likely more to it.

Maybe we should expand the README (or the GitHub Wiki or even the Fedora Wiki) with some instructions for image building. What do you think @debarshiray, @juhp??

images/arch/extra-packages
erazemkokot: Yup, I've removed the packages that belong to base and base-devel.

erazemkokot:
> Maybe we should expand the README (or the GitHub Wiki or even the Fedora Wiki) with some instructions for image building.

I haven't read much documentation on podman yet, but I, and I'm sure a lot of other people, would appreciate a section in the readme dedicated to manually building images.

images/arch/Dockerfile
erazemkokot: Added to extra-packages

images/arch/Dockerfile
erazemkokot: Ok, I'll check it out (I'm pretty sure it was because `bash` is a part of the `base` metapackage now and I haven't added it yet when you were testing the image).

kronenpj:
@HarryMichal 
> The podman command can be replaced by `buildah build-using-dockerfile`. But this command doesn't work for me even with the already existing fedora Dockerfiles, so there is most likely more to it.

If you're using an up-to-date `buildah` there's a [bug report](https://github.com/containers/buildah/issues/1952) open which, when fixed, will address a message like:
`error building at STEP "COPY README.md /": secure join: too many levels of symbolic links`
A temporary workaround is to `rm README.md` and `cp ../../../README.md .`

Otherwise using `buildah bud -t f32-toolbox .` worked for me just now.

kronenpj:
@erazemkokot 
I'm not that familiar with Arch, but if the default "FROM" container has removed the documentation or man pages, you may want to re-install them. This way the person using the toolbox has the documentation available from the toolbox distribution.

In Fedora the approach was to take a selection of the packages and re-install them. Ubuntu container had a script `/usr/local/sbin/unminimize` that identified and re-installed packages where documentation was missing. Arch may have a similar approach.

erazemkokot:
> I'm not that familiar with Arch, but if the default "FROM" container has removed the documentation or man pages, you may want to re-install them. This way the person using the toolbox has the documentation available from the toolbox distribution.

I'll check, thanks for the warning.

images/arch/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN pacman -Syu --noconfirm
+
+# Install packages from the extra-packages file
+COPY extra-packages /
+RUN pacman -Sy --noconfirm $(<extra-packages)
+RUN rm /extra-packages
+
+# Clean up all local caches
+RUN pacman -Scc --noconfirm
+
```
njam, line 22: sudo should be enabled.

```suggestion

# Allow sudo for users in group "sudo"
RUN groupadd -f sudo
RUN echo '%sudo ALL=(ALL) ALL' >/etc/sudoers.d/group-sudo
```

images/arch/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM docker.io/archlinux/base:latest
```
njam, line 1: I think "archlinux" is more official. It's referenced from https://www.archlinux.org/download/

```suggestion
FROM docker.io/archlinux:latest
```

erazemkokot:
I'm going to close this to give it a new - clean - shot, keeping in mind all the tips I got so far.

PR #332, opened 2019-11-15
Fix shift count out of range error

tagoh:
`run` should take one argument for command at least

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -2533,6 +2533,7 @@ case $op in
             esac
             shift
         done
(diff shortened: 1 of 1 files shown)


toolbox
```diff
@@ -2530,6 +2530,7 @@ case $op in
         if ! update_container_and_image_names; then
             exit 1
         fi
+	exit_if_missing_argument "$op" "$1"
```
debarshiray: Maybe do it even earlier before `update_container_and_image_names`?

PR #362, opened 2020-01-13
Check /usr/share/profile.d when bind mounting toolbox.sh

debarshiray:
/usr/share/profile.d is the default location where toolbox.sh is
installed, even though, in practice, most (all?) distributions use
/etc/profile.d. It's reasonable to at least make the code work with the
default build values.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -989,6 +989,10 @@ create()
 
     if [ -f /etc/profile.d/toolbox.sh ] 2>&3; then
         toolbox_profile_bind="--volume /etc/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"
+    elif [ -f /usr/share/profile.d/toolbox.sh ] 2>&3; then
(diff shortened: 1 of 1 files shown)


debarshiray:
I split this out from @HarryMichal 's branch for hhttps://github.com/containers/toolbox/pull/250

toolbox
```diff
@@ -989,6 +989,10 @@ create()
 
     if [ -f /etc/profile.d/toolbox.sh ] 2>&3; then
         toolbox_profile_bind="--volume /etc/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"
+    elif [ -f /usr/share/profile.d/toolbox.sh ] 2>&3; then
+        toolbox_profile_bind="--volume /usr/share/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"
+    else
+        echo "$base_toolbox_command: failed to bind mount toolbox.sh"
```
debarshiray: This should go to the standard error output or be a debug log entry.

debarshiray: I took the liberty to tweak this before merging.

Issue #369, opened 2020-02-01
toolbox enter fails silently when pwd doesn't exist in container


bureado:
`toolbox` already checks if a program exists inside the container, but it doesn't check if `$PWD` exists in the container. In such cases, `toolbox enter` fails silently.

Ideally, `toolbox` should exit gracefully with a helpful error message.

**Steps to reproduce:**

1. `cd` into a directory that is unlikely to exist in the container: `mkdir -p /tmp/1337 && cd /tmp/1337`
2. Run `toolbox enter` in that directory:
```
foo@bar:/tmp/1337$ toolbox enter --container ubuntu-toolbox-latest
toolbox: /etc/profile.d/toolbox.sh not found
toolbox: /run/user/1000/toolbox/toolbox.sh not found
foo@bar:/tmp/1337$ echo $?
127
```
3. `cd` to your home profile and try `toolbox enter` again:
```
foo@bar:/tmp/1337$ cd 
foo@bar:~$ toolbox enter --container ubuntu-toolbox-latest
toolbox: /etc/profile.d/toolbox.sh not found
toolbox: /run/user/1000/toolbox/toolbox.sh not found
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.
...
```

I'll submit a suggestion for a patch shortly.

Issue #348, opened 2019-12-11
Adjust home path


ckaspro:
I like to use several toolboxes, with different content. To properly distinguish these, I'd like to mount them in subdirectories (e.g. container1:/home/user -> host:/var/home/user/container1). It's a similar thing to #183 , but not as security- centric.
Is there a way to achieve this at the moment / via consistent workaround (I don't want to change a line of code and after the next update I'm saving files somewhere else ;)  )?

Update: I wouldn't actually mind (maybe even appreciate) having these under different user-contexts - but in the same general directory. That way there is a clear distinction and separation.
For presenting this to the end- user in a "nice way", one could add an extra link opening the file-browser as sudo... (just thinking loud here)

Thanks in advance,
Chris

yilkalargaw:
I usually just do ```HOME=/home/user1/container1``` before creating the containers and just alias use an alias to open them. (```alias toolbox1='HOME=/home/user1/container1/toolbox "$@"'```)

debarshiray:
Yeah, overriding `$HOME` is one way to do it.

JaneSmith:
I'd also like to see something like this. My main use case for Toolbox is to set up throwaway development environments, for either working on projects, or just quickly compiling some software from source. I don't want to pollute my system with random files from dependencies, etc. just to compile something once. Toolbox seemed perfect for this, as it claims to provide an isolated container to keep the host clean.

This didn't turn out to work quite how I expected though. While Toolbox *does* keep the host OS clean, it doesn't keep the home directory clean at all. That still gets totally polluted. For example, using a throwaway container to build something that uses `rust`, my home directory ended up getting altered in several ways:

* A `~/.cargo` directory was silently created, containing many Rust-related binaries, source packages, etc. for a total of 123 MB.
* A `~/.rustup` directory was silently created, containing many more Rust-related binaries, for a total of 683 MB.
* My `~/.bash_profile` file was silently altered to add the `~/.cargo/bin` directory to my $PATH ahead of everything else.
* My `~/.profile` file was silently altered to add the `~/.cargo/bin` directory to my $PATH ahead of everything else.
* Who knows what else.

Yikes. I realise now from the comments above that I can override the $HOME environment variable manually to try to work around this, but this was not intuitive or expected to me. Since Toolbox is supposed to be (or at least as I understand it) a simplified and user-friendly way to get into containers, I would appreciate if this could be handled in some better way.

[...]

PR #370, opened 2020-02-01
Adding a test for $PWD in the container

bureado:
This is a suggestion to address #369, input and changes welcome.

I didn't want to count on `coreutils` so ran `test` inside `sh`, but perhaps that's fine.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -1603,6 +1603,16 @@ run()
 
     $emit_escape_sequence && printf "\033]777;container;pop;;\033\\"
 
+    if [ "$ret_val" -eq 127 ] 2>&3; then
+        # shellcheck disable=SC2016
+        if ! $podman_command exec \
+                     --user "$USER" \
+                     "$toolbox_container" \
+                     sh -c 'test -d "$1"' sh "$PWD" >/dev/null 2>&3; then
+            echo "$base_toolbox_command: directory '$PWD' not found in container $toolbox_container" >&2
+        fi
+    fi
+
     exit "$ret_val"
 )
 
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -314,7 +314,11 @@ func runCommand(container string,
 	case 126:
 		err = fmt.Errorf("failed to invoke command %s in container %s", command[0], container)
 	case 127:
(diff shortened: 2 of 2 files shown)


bureado:
Also FYI, I'm not sure if cgwalters/coretoolbox is affected because I couldn't get past its attempt to bind itself, but a cursory look at the code didn't raise any flags. Just mentioning because of #319

HarryMichal:
Hi @bureado! I'm currently looking at your PR and I'm wondering that instead of just showing the error message Toolbox could try to mount the missing path. The reason why I think this could be viable is that most paths on the host are inside of the container (they are bind-mounted in /run/host).

Also now that I'm thinking about it. This seems to be kinda similar to #100 and #348. Both are a bit different but in both cases, one of the steps to solve them is to add user-specifiable bind mounts.

But for now, the easiest solution is to use your approach. I'll add it to the Go version of Toolbox.

abitrolly:
@HarryMichal why not to commit that to `master` until Go rewrite is ready?

toolbox
```diff
@@ -1580,6 +1580,14 @@ run()
         fi
     fi
 
+    if ! $podman_command exec \
+                 --user "$USER" \
+                 "$toolbox_container" \
+                 sh -c "test -d $PWD" >/dev/null 2>&3; then
+        echo "Directory $PWD does not exist in container $toolbox_container, try cd && !!" >&2
```
debarshiray, line 1587: I wish we could avoid another potentially slow `podman exec` (do the `exec`, invoke `sh`, etc.) before the *main* `podman exec` that offers the interactive shell. Most people running `toolbox enter` or `toolbox run` are hopefully not going to be hitting this error, so it makes sense to optimize for the non-failure case.

How about doing it after the *main* `podman exec` fails with an exit code of 127? That way the extra `podman exec` will only be run when things have already failed.

(I wish we could also somehow avoid the other `podman exec` that's already there to do the fallback-to-bash, but that one seems more difficult to get rid off because it involves the binary that gets invoked.)

debarshiray: I took the liberty to adjust it as above, and I see that @HarryMichal already ported your change to his Go version.

debarshiray:
Thank you for your contribution, and apologies for the delay in getting to it!

PR #372, opened 2020-02-07
Load fedora-toolbox image in the pre-phase to prevent ci failure

TristanCacqueray:
This change adds a pre task to load the fedora-toolbox image from
the registry to prevent false positive failure when:
Trying to pull registry.fedoraproject.org/f31/fedora-toolbox:31...
  invalid status code from registry 503 (Service Unavailable)

### DIFF
--- a/playbooks/fedora-30/setup-rpm-env.yaml
+++ b/playbooks/fedora-30/setup-rpm-env.yaml
@@ -20,3 +20,5 @@
 
     - name: Check versions of crucial packages
       command: rpm -q podman runc conmon fuse-overlayfs flatpak-session-helper
+
+    - include_tasks: ../pre-common.yaml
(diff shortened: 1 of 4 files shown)


TristanCacqueray:
See the failures in the last periodic run: https://softwarefactory-project.io/zuul/t/local/buildset/b532a118c7044bcaa522f085052f6261

TristanCacqueray:
That doesn't seems enough since the test seems to rm the image...

HarryMichal:
> That doesn't seems enough since the test seems to rm the image...

That is true. I'm not very happy with the state the tests are in. I need to adjust them a bit. We'll use this certainly.

HarryMichal:
@TristanCacqueray I took the liberty of using your PR commit in #375 where I updated the tests to prevent false-positive results. If you don't mind, I'll close this PR.

TristanCacqueray:
@HarryMichal thanks!

PR #374, opened 2020-02-08
Add Code of Conduct

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

As the title says.

PR #375, opened 2020-02-10
Rework the system tests

HarryMichal:
The tests introduced by #250 have proven to be rather unstable due to
mistakes in their design. The tests very quite chaotically structured.
Because of that images were deleted and pulled too often which caused
several false positives (#374, #372).

This changes the strucutre of the tests in a major way. The tests (resp.
commands) are now ran in a manner to kinda simulate the way Toolbox is
used. From clean state, through creating containers, using them and in
the end deleting them. This should reduce the strain on the bandwidth
and possibly even speed up the tests themselves.

More information in the README.md in the directory with the tests.


test/system/README.md
```diff
@@ -47,6 +28,9 @@ suggestions or even their implementation.
 
 Make sure you have `bats` and `podman` with `toolbox` installed on your system.
 
+**Important**
+Before you start the tests, you need to have present two images: the default fedora-toolbox image (depending on your version of Fedora) and fedora-toolbox-29
```
debarshiray: Rookie mistake: `fedora-toolbox:29`, not `fedora-toolbox-29`. :)

Image names have colon, container names don't.

test/system/README.md
```diff
@@ -1,43 +1,24 @@
 # System tests
 
 These tests are built with BATS (Bash Automated Testing System). They are
-strongly influenced by libpod project.
+strongly influenced by the [libpod](https://github.com/containers/libpod) project.
```
debarshiray: I took the liberty to split these cosmetics out into https://github.com/containers/toolbox/pull/377

test/system/002-help.bats
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "Show usage screen when no command is given" {
+  run_toolbox 1
+  is "${lines[0]}" "toolbox: missing command" "Usage line 1"
+}
```
debarshiray: Nitpick: some of these new files don't have a trailing newline, which Git complains about with `\ No newline at end of file`.

playbooks/fedora-30/pre-common.yaml
```diff
@@ -0,0 +1,13 @@
+- name: Load registry image f30/fedora-toolbox:30 in pre-phase and auto restart when registry is flaky
```
debarshiray: Nitpick: just `Pull registry.fedoraproject.org/f30/fedora-toolbox:30` might be better when looking at the test logs. The logs already list this under `pre`, so the details about the phase might not be necessary.

test/system/README.md
```diff
@@ -1,43 +1,24 @@
...
-  - [ ] create several containers and hop between them (series of enter/exit)
-  - [ ] create a container, enter it, run a series of basic commands (id,
-        whoami, dnf, top, systemctl,..)
-  - [ ] enter a container and test basic set of networking tools (ping,
-        traceroute,..)
-
-The list of tests is stil rather basic. We **welcome** PRs with test
-suggestions or even their implementation.
+- **0xx (Info)**
+  - In this section are ran commands that are not dependent on the presence/number of containers/images (eg. getting version, help,..).
```
debarshiray: Nightpick: would be nice to keep the documentation files limited to 80 characters per line.

When it comes to code, depending on the programming language, the ability to use more than 80 characters can be hugely convenient in practical terms. Human languages are usually easier.

test/system/README.md
debarshiray: Fixed.

playbooks/fedora-30/pre-common.yaml
debarshiray: Tweaked.

test/system/README.md
debarshiray: Tweaked.

test/system/002-help.bats
debarshiray: Added the newlines.

HarryMichal:
Thank you for fixing the mentioned issues!!

PR #380, opened 2020-02-26
Support /etc/resolv.conf symlinks on the host

martinpitt:
If /etc/resolv.conf is already a symlink on the host [1], fix the link
target in toolbox to the corresponding /run/host/<original target>.

https://bugzilla.redhat.com/show_bug.cgi?id=1785244

[1] It's common to point it to /run/NetworkManager/resolv.conf or
    /run/systemd/resolve/resolv.conf

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -1209,12 +1209,17 @@ init_container()
             fi
 
             if ! readlink /etc/resolv.conf >/dev/null 2>&3; then
-                echo "$base_toolbox_command: redirecting /etc/resolv.conf to /run/host/etc/resolv.conf" >&3
+                # /etc/resolv.conf on the host may itself be a symlink
+                resolv_conf_target="/run/host/etc/resolv.conf"
+                if T=$(readlink "$resolv_conf_target"); then
+                    resolv_conf_target="/run/host/$T"
+                fi
+                echo "$base_toolbox_command: redirecting /etc/resolv.conf to $resolv_conf_target" >&3
 
                 if ! (cd /etc 2>&3 \
                       && rm --force resolv.conf 2>&3 \
-                      && ln --symbolic /run/host/etc/resolv.conf resolv.conf 2>&3); then
-                    echo "$base_toolbox_command: failed to redirect /etc/resolv.conf to /run/host/etc/resolv.conf" \
+                      && ln --symbolic "$resolv_conf_target" resolv.conf 2>&3); then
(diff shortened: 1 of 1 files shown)


martinpitt:
I tested this with /etc/resolv.conf being a plain file and a symlink to /run/NetworkManager/resolv.conf (my usual configuration, to avoid writing to /etc), and it now behaves correctly in both cases.

Before that, networking was always broken by default, and I had to [apply a hack](https://piware.de/gitweb/?p=bin.git;a=blob;f=build-devtoolbox#l17). I didn't go with letting it point to /run/host/monitor/resolv.conf though, as that's just a bind-mount from the host's flatpak helper and I'd rather not introduce this assumption/dependency to resolv.conf handling.

martinpitt:
Argh, the `ls -ld` was a debugging leftover. I force-pushed to drop it. @kronenpj , do you mind to re-approve? Thanks for the review!

martinpitt:
I can't look into that podman-rawhide failure (the logs are private or require logging in, etc?), but I don't think it's something that I caused. The previous version (with the leftover debug `ls -ld`) was passing, and the f30/f31 ones work.

kronenpj:
I looked into that a bit as well. I'm guessing only a 'member' of the project can kick the job off again. You might ask @debarshiray for access.

martinpitt:
I force-pushed to trigger the job again, but I suppose it's rather some change/regression with podman itself in rawhide.

HarryMichal:
Hi, @martinpitt! Seems like there's a problem with Zuul itself because the playbook won't even start. The other tests seem to be ok.

morucci:
recheck

kronenpj:
recheck

martinpitt:
FTR, the exact same issue applies to the Go rewrite, but now of course this PR does not apply any more.

debarshiray:
@HarryMichal ported this to the Go implementation in https://github.com/containers/toolbox/pull/460

Thanks for pursuing this @martinpitt ! Much appreciated, and my apologies for the delay in getting to it.

debarshiray:
> I didn't go with letting it point to /run/host/monitor/resolv.conf though,
> as that's just a bind-mount from the host's flatpak helper and I'd rather
> not introduce this assumption/dependency to resolv.conf handling.

Yes, I agree.

We do currently use `flatpak-session-helper` for some of the configuration files, but we need to move away from it because it doesn't work with `sudo toolbox ...`.

martinpitt:
Thanks @debarshiray ! Glad to see that this is fixed at last.

PR #383, opened 2020-02-29
Use container name as container hostname

Jmennius:
..to distinguish different toolboxes more easily inside a toolbox,
this includes version information too (by default).
Replace underscores with dashes since underscores are not allowed in hostnames.

Fixes: #98, #120

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -419,7 +419,7 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 	createArgs = append(createArgs, xdgRuntimeDirEnv...)
 
 	createArgs = append(createArgs, []string{
-		"--hostname", "toolbox",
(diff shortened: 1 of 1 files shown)


elimisteve:
Good idea

debarshiray:
> I know that your immediate need has
> [been](https://github.com/containers/podman/issues/6192)
> [addressed](https://github.com/romkatv/powerlevel10k/issues/1560)
> already. Would you still be interested in updating this pull
> request to implement #969 ?
> 
> Even if we don't use the hostname of the container directly in
> the shell prompt, #969 still brings up some good points about
> the default hostname.

https://github.com/containers/toolbox/pull/1086 is very close to implementing https://github.com/containers/toolbox/issues/969  Let's focus on that one.

Regardless, thanks for all your contributions to the Toolbx project, @Jmennius   It's a pleasure to have you around.  :)

Issue #218, opened 2019-07-07
Improve the terminal integration to preserve the current container for new terminals


rugk:
I basically want to _not_ do https://github.com/debarshiray/toolbox/issues/135 /https://github.com/debarshiray/toolbox/pull/199.

When I <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd> (open a new tab) when I am in a `toolbox` in Fedora Silverblue, the new terminal also is in that toolbox?

So:
1. Can I disable this "feature"? How?
2. (If not) Can I at least somehow exit the toolbox in a new tab then without closing that tab?
  I.e. if I am in that toollbox in a new tab and enter `exit`, it actually closes this tab. Can I somehow get out of this without closing this tab?

---

original question from here: https://ask.fedoraproject.org/t/can-i-disable-automatic-toolbox-entering-when-new-terminal-tab-is-opened-in-gnome-terminal/2412?u=rugk
Also basically requested [here](https://discussion.fedoraproject.org/t/gnome-terminal-launch-command-toolbox-enter-on-launch/1954?u=rugk) and @basvdlei also seems to ask for a way to disable that behaviour in https://github.com/debarshiray/toolbox/issues/135#issuecomment-508447430 if I understand correctly.

shaman007:
+1, very annoying by default.

cgwalters:
I also think this is trying to solve an important problem in the wrong way.  Personally I use custom gnome-terminal profiles, including one for my toolbox that gives me the tab behavior, without affecting the main Terminal app.

mildred:
I quite like opening the toolbox automatically on a new tab, especially when not running the default toolbox. There is one thing though, when all tabs in a terminal are running a toolbox, we can only open a new window on the host.

You can probably disable the feature yourself in your dotfiles if you run inside the toolbox :

    printf "\033]777;container;pop;;\033\\"

from https://github.com/debarshiray/toolbox/pull/199/files

Did not test that though

TheFrozenFire:
:+1:, but to the second question of how to exit a toolbox. When you want to run system-level commands (e.g. rpm-ostree), there's no clear way to get out of toolbox.

shaman007:
@TheFrozenFire Ctrl+D?

TheFrozenFire:
@shaman007 That's too obvious and simple for me. Are you sure you don't have a dbus command for that? (/s)

Thanks. :)

bkhl:
[...]

Issue #135, opened 2019-04-26
Running `toolbox enter` for every terminal prompt will get annoying quickly


allanday:
On Silverblue, if we expect users to generally be using a toolbox as their terminal environment, then having to run `toolbox enter` every time you open a new terminal will get annoying really quickly.

Potential solutions to this:

 1. Show separate terminal launchers - one for the host, one for each toolbox
 2. Have the terminal open in the last used environment (could be a toolbox or the host); one issue here: how you then get from the toolbox container to the host

From a pure UX perspective I think I prefer the second option, because it puts less stress on the user: with the first option, you have to think about where you want to open the terminal and pick the right launcher (potentially from a list), which is harder work than just launching the terminal it generally doing the right thing.

debarshiray:
>     2. Have the terminal open in the last used toolbox (one issue here: how you then
> get from the toolbox container to the host)

We could offer a command to open a shell on the host. Could be a `toolbox` command or something separate.

allanday:
> We could offer a command to open a shell on the host. Could be a `toolbox` command or something separate.

That could be cool.

allanday:
Another issue with option 1: someone has to learn to use the toolbox terminal. Their first experience is likely be to try what they know, which will be to run the standard terminal app.

There is another option of course:

 3. Have one terminal, and make it default to a toolbox environment

The obvious issue with this is that it will be painful for anyone who needs to use the host environment. There's a bit of an unknown here: we're expecting most people to use the toolbox most of the time, but there might be cases we don't know about where people need the host.

debarshiray:
https://gitlab.freedesktop.org/terminal-wg/specifications/issues/17 is meant to reduce the need to keep typing `toolbox enter` over and over again. The idea is that you'd run `toolbox enter` once, and the terminal will take care of spawning the right environment from there onwards.

Related to that is this GNOME Terminal issue about retaining the current working directory for commands other than the host shell: https://gitlab.gnome.org/GNOME/gnome-terminal/issues/126

debarshiray:
Part of the solution was implemented in pull request #199 When used with [matching](https://bodhi.fedoraproject.org/updates/FEDORA-2019-5aca07c69c) GNOME Terminal and VTE versions, the toolbox container will be preserved when opening a new terminal.

There's still no `toolbox` command to request a host shell from inside the container, though.

kamalmarhubi:
[...]

PR #384, opened 2020-02-29
Allow to opt out of terminal container tracking

Jmennius:
By providing a --once (-o) switch.
Issuing a `toolbox enter --once` means that you do not want
your terminal to track the container in use.
Tracking is enabled by default.

Fixes: #218

### DIFF
--- a/doc/toolbox-enter.1.md
+++ b/doc/toolbox-enter.1.md
@@ -6,6 +6,7 @@ toolbox\-enter - Enter a toolbox container for interactive use
 ## SYNOPSIS
 **toolbox enter** [*--distro DISTRO* | *-d DISTRO*]
               [*--release RELEASE* | *-r RELEASE*]
+              [*--once* | *-o*]
               [*CONTAINER*]
 
 ## DESCRIPTION
@@ -39,6 +40,10 @@ host.
 
 Enter a toolbox container for a different operating system RELEASE than the
 host.
+
+**--once**, **-o**
+
+Do not track container in use, only enter the container once.
 
 ## EXAMPLES
 
--- a/src/cmd/enter.go
+++ b/src/cmd/enter.go
@@ -32,6 +32,7 @@ var (
 		container string
 		distro    string
 		release   string
+		once      bool
 	}
 )
 
@@ -62,6 +63,12 @@ func init() {
 		"r",
 		"",
 		"Enter a toolbox container for a different operating system release than the host")
+
+	flags.BoolVarP(&enterFlags.once,
+		"once",
+		"o",
+		false,
+		"Do not track container in use, only enter the container once.")
 
 	if err := enterCmd.RegisterFlagCompletionFunc("container", completionContainerNames); err != nil {
 		logrus.Panicf("failed to register flag completion function: %v", err)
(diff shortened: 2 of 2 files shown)


Jmennius:
recheck

Jmennius:
recheck

bam80:
What is the status of this?

Jmennius:
I really hope that maintainers will get back to actively working on issues and PRs really soon.

But this probably needs to be reimplemented in Go from what I see...?

CI failures I believe are just flakiness (recheck?).

HarryMichal:
@bam80, we merged today the [rewritten version of Toolbox](#318). When we get an initial release out, we'll turn attention to this.

Jmennius:
@HarryMichal should I look into implementing this for Go-based verion or is it better left for you guys?

HarryMichal:
Hey @Jmennius! I'm not entirely sure how this behaviour should be handled properly. But I'd rather this not to be an option but to be solved on the default level... But I'm definitely open to a discussion.

Jmennius:
Updated this to a Go-based implementation (dunno why I did not push it earlier - was using this for over a year).

doc/toolbox-enter.1.md
```diff
@@ -40,6 +41,10 @@ host.
 Enter a toolbox container for a different operating system RELEASE than the
 host.
 
+**--once**, **-o**
+
+Do not track container in use, only enter the container once.
```
cben, line 46: This formulation assumes user already understands the default behavior #135 but it's not yet documented elsewhere.  Consider something more educational like
```suggestion
By default, an escape sequence is printed that some terminal emulators honor, so that new tabs/windows also automatically launch in the toolbox.
After you exit the shell opened directly by this command, Toolbx undoes the escape sequence and you get back to the shell or process that launched `toolbox enter`.

(Effectively, opening a new tab can "copy" the toolbox status from current tab, similar to working directory getting copied.)

This flag disables printing such escape sequences.  It will be up to you to run `toolbox enter` in each terminal where you want it independently.
```

Jmennius: I agree. I'll fix it as soon as there is interest in meeting this.

kepstin:
Just to confirm, will this work with configuration? So I could add e.g.
```
[general]
once = true
```
to my ~/.config/containers/toolbox.conf to get the "once" behaviour without having to manually type it every time?

The command-line flag isn't a great solution for me, since it's very easy to forget the "--once" option and accidentally get your terminal stuck in the container.

debarshiray (commented):
Oops, I missed that there was this pull request for https://github.com/containers/toolbox/issues/218  Sorry about that.

As I [mentioned](https://github.com/containers/toolbox/issues/218#issuecomment-1863095554) a year ago in https://github.com/containers/toolbox/issues/218, the long-term solution is to switch to a terminal emulator like [Ptyxis](https://gitlab.gnome.org/chergert/ptyxis), which was earlier known as Prompt.

It is [now the default](https://pagure.io/fedora-workstation/issue/417) terminal emulator on Fedora 42 Silverblue & Workstation.  Others can get it from [Flathub](https://flathub.org/apps/app.devsuite.Ptyxis).

debarshiray:
Closing.  My apologies for not commenting earlier.

PR #387, opened 2020-03-06
Centos 8.1

kronenpj:
The usual first attempt at a CentOS 8.1 toolbox container. Nothing Earth shattering here.
Unfortunately the CentOS container project isn't using an '8.1' tag so it'll be difficult to follow the latest image without changing the Dockerfile.


kronenpj:
Never mind. I didn't catch that the centos:8 tag is the same as centos:8.1.1911, as of today anyhow.

PR #396, opened 2020-03-27
Disable detach-keys

### LINKED ISSUE #394: Being in a toolbox breaks Ctrl-p for readline
I use the readline commands for navigating my command history in bash. That is: C-p for previous, C-n for next. C-p should do the same thing as up-arrow, and C-n should do the same as down-arrow.

When I am in a toolbox and I press C-p to get my previous command, nothing shows up. If I press return, the command shows up, and executes, but it was not visible until I press return. If I press C-p twice, I see the two-commands-ago command line, and then pressing C-n gives me the previous command.

This only happens when I am inside a toolbox on Fedora Silverblue. I have not tried it on other platforms. It does not happen on the same machine, when I am _not_ in a toolbox.

up-arrow and down-arrow work just fine.

EnigmaCurry:
This fixes the problem of not being able to use the Ctrl-p shortcut in readline and Emacs. This disables the podman detach-keys (default Ctrl-p, Ctrl-q). Without this, things like bash readline, and Emacs suffer from not being able to use Ctrl-p key.

I don't know if there is any good reason to actually enable detach keys when using toolbox. If so, then this should be made configurable somehow, but I thought I would share my fix regardless. Thanks!

This fixes #394

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,18 @@
+Overview of changes in 0.0.93
+=============================
+
+* Ensure reproducible builds by using the -trimpath build flag
+* Fix the test suite to work with the Go implementation
+* Make listing of containers and images more robust against changes in the JSON
+  returned by Podman
+* List out dependencies and installation instructions
+* Re-enable highlighting of running containers
+* Show the spinner only when connected to a terminal
+* Speed things up by caching the Podman version at runtime
+* Update hint after creating a container to use the new syntax
+* Use the correct verb format for string
+
+
 Overview of changes in 0.0.92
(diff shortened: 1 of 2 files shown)


HarryMichal:
Hi @EnigmaCurry! I apologize for prematurely closing your PR. I wanted to update it to make it work with the latest version of Toolbox (it has been rewritten in Go) but I messed up. I'll open a new PR from my branch where I have your original commit that I've updated. You're still it's original author, so git statistics magic should give the credit to you :).

Thank you for your work!

PR #401, opened 2020-04-02
Drop unnecessary and wrong group detection on host

martinpitt:
Don't call get_group_for_sudo() on the host during create(). That runs
on the host, and thus will check which sudo group exists on the host.
But that is entirely irrelevant for sudo inside the container, and it
breaks when trying to create a Debian or Ubuntu based toolbox on a
Fedora host (or vice versa).

Also, there is no point in running the `podman create` command with an
extra sudo group, normal user privileges are just fine.

init_container() will call get_group_for_sudo() inside the container and
initialize the groups correctly there.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -191,11 +191,6 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 	toolboxPathEnvArg := "TOOLBOX_PATH=" + toolboxPath
 	toolboxPathMountArg := toolboxPath + ":/usr/bin/toolbox:ro"
 
-	sudoGroup, err := utils.GetGroupForSudo()
-	if err != nil {
-		return err
-	}
-
 	logrus.Debug("Checking if 'podman create' supports '--ulimit host'")
 
 	var ulimitHost []string
@@ -339,7 +334,6 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 		"create",
 		"--dns", "none",
 		"--env", toolboxPathEnvArg,
-		"--group-add", sudoGroup,
 		"--hostname", "toolbox",
 		"--ipc", "host",
 		"--label", "com.github.containers.toolbox=true",
--- a/toolbox
+++ b/toolbox
@@ -982,11 +982,6 @@ create()
         return 1
     fi
 
-    if ! group_for_sudo=$(get_group_for_sudo); then
-        echo "$base_toolbox_command: failed to create container $toolbox_container: group for sudo not found" >&2
-        return 1
-    fi
-
     if [ -f /etc/profile.d/toolbox.sh ] 2>&3; then
         toolbox_profile_bind="--volume /etc/profile.d/toolbox.sh:/etc/profile.d/toolbox.sh:ro"
     elif [ -f /usr/share/profile.d/toolbox.sh ] 2>&3; then
@@ -1078,7 +1073,6 @@ create()
     $podman_command create \
             --dns none \
             --env TOOLBOX_PATH="$TOOLBOX_PATH" \
-            --group-add "$group_for_sudo" \
             --hostname toolbox \
             --ipc host \
             --label "com.github.containers.toolbox=true" \


martinpitt:
With toolbox, the reproducer is this:

    toolbox -y create -c sid --image docker.io/debian:sid && podman start sid

martinpitt:
This fails the tests as it breaks the `init-container` stage -- at that point `get_group_for_sudo()` is called *inside* the container, where indeed there is no podman command and no image:
```
++ get_group_for_sudo
+++ podman run --rm '' getent group sudo wheel
/usr/bin/toolbox: line 515: podman: command not found
```
So the fix is even easier -- just drop the redundant "outer" group detection.

Jmennius:
This is revealed/caused by [a fix in libpod](https://github.com/containers/libpod/commit/b41c864d569357a102ee2335a4947e59e5e2b08a) introduced in 1.8.1.
Thanks for the fix!

masterzen:
Can this be merged in ?
It isn't possible at all to run toolbox with a fedora image on Fedora CoreOS (which has `sudo` supplemental group) for the moment, and drastically reduces the ability to troubleshoot issues.

martinpitt:
My hunch is that the maintainers now ignore this project as they are working full steam on the Go rewrite. I haven't tested that at all yet (it's not yet in Fedora 32 or even Rawhide). It would still be nice to land this, as this project is really easy to work on and run straight out of git.

fslds:
I would also like to see this PR being approved and merged in master and consequently released as a package...
It is currently breaking cross-compatibility between Debian and RedHat derivatives, if one is the host and the other is a container.

crahan:
I'd like to add my vote for merging this fix as well.

evan-a-a:
The associated code in the Go variant is here:

https://github.com/containers/toolbox/blob/master/src/cmd/create.go#L193
https://github.com/containers/toolbox/blob/master/src/cmd/create.go#L341

The change to remove this setting in Go can probably be added to this pull request as well to ensure this is remains fixed when the Go version becomes official.

HarryMichal:
I took the liberty of updating your PR to reflect your original changes in the new Go code.

I tried this change on a CoreOS host myself and it really makes rootless Toolbox work. @debarshiray, this should be good to be merged.

Thank you very much for working on this!

martinpitt:
The tests succeeded in April when I sent this, but the run yesterday (triggered by @HarryMichal ?) now fails with

    meson.build:8:0: ERROR: Program(s) ['go'] not found or not executable

That looks like it needs to be fixed in your CI?

HarryMichal:
The CI is being fixed in a separate PR.

Ramblurr:
Any traction on this?

Toolbox is a daily important tool we use. I appreciate the go rewrite, and look forward to it. But the backlog of simple and good bug fixes backing up here is taking its toll.

src/cmd/initContainer.go
```diff
@@ -231,7 +231,7 @@ func initContainer(cmd *cobra.Command, args []string) error {
 
 		sudoGroup, err := utils.GetGroupForSudo()
 		if err != nil {
-			return fmt.Errorf("failed to add user %s: %s", initContainerFlags.user, err)
+			return fmt.Errorf("failed to get group for sudo: %w", err)
```
debarshiray, line 234: I took the liberty to split this typo fix (thanks!) into a [separate commit](https://github.com/containers/toolbox/pull/545)

debarshiray:
Looks like this fix for https://github.com/containers/toolbox/issues/423 had an unintended side-effect.

Earlier, inside the container, we'd get:
```
⬢[rishi@toolbox ~]$ id
uid=1000(rishi) gid=1000(rishi) groups=1000(rishi),10(wheel)
```

... but now:
```
⬢[rishi@toolbox ~]$ id
uid=1000(rishi) gid=1000(rishi) groups=1000(rishi)
```

Notice how the `wheel` group doesn't get listed. It does show up in `id rishi`, and doesn't seem to affect `sudo` capabilities. However, this can still trip up users and programs alike.

debarshiray:
> Looks like this fix for https://github.com/containers/toolbox/issues/423 had an unintended side-effect.

Owen chased it down the stack and fixed it in crun:
* https://github.com/containers/podman/issues/9986
* https://github.com/containers/crun/issues/644
* https://github.com/containers/crun/pull/645

So, all good!

PR #404, opened 2020-04-05
When bind-mounting a file, create the target path if it doesn't exist

njam:
If the target path doesn't exist, 'mount' fails with this error:

> mount: <TARGET>: mount point does not exist.

I got this error when in a toolbox the path `/etc/machine-id` does not exist.  

This PR just `touch`es the target path if it doesn't exist.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -683,12 +683,22 @@ mount_bind()
 
     ! [ -d "$source" ] 2>&3 && ! [ -f "$source" ] 2>&3 && return 0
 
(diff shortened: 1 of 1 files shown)


njam:
Closing because this PR is against the old shell-script version.

debarshiray:
The underlying issue also affects the Go version, because it's ultimately a [bug in Podman](https://github.com/containers/libpod/issues/6735).

PR #405, opened 2020-04-05
Don't try to remove /media if it doesn't exist

njam:
Otherwise `rmdir` will result in an error, and "init_container" will abort.


njam:
I see this code was changed in https://github.com/containers/toolbox/pull/340, with the goal of handling the case when `/media` doesn't exist, but I think it wasn't solved fully?

njam:
Closing because this PR is against the old shell-script version.

PR #412, opened 2020-04-18
Add dnf_cache_path_bind overlay volume

cryobry:
Apply @rhatdan's tip to add a dnf cache overlay to speed up dnf in new toolboxes.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -904,6 +904,7 @@ create()
     mnt_link=""
     mnt_path_bind=""
     run_media_path_bind=""
+    dnf_cache_path_bind=""
     toolbox_profile_bind=""
     ulimit_host=""
     usr_mount_destination_flags="ro"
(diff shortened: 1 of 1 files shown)


HarryMichal:
Hey @cryobry! Did you test the patch on your machine? According to the tests, there is a problem while creating containers.

toolbox
```diff
@@ -1019,6 +1020,10 @@ create()
         run_media_path_bind="--volume /run/media:/run/media:rslave"
     fi
 
+    if [ -d /var/cache/dnf ] 2>&3; then
+        dnf_cache_path_bind="--volume /var/cache/dnf:/var/cache/dnf:O"
```
HarryMichal, line 1024: Should the option really be `O`? I don't think that is valid. What should the behaviour be like?

HarryMichal:
Also a heads-up: This will, most likely, not be accepted because Shell Toolbox will be replaced by a [new version of Toolbox](https://github.com/containers/toolbox/pull/318) written in Go. When it is replaced, would you mind updating the PR to the new version?

toolbox
cryobry: See "Overlay Volume Mounts" section: https://github.com/containers/buildah/blob/master/docs/buildah-bud.md

HarryMichal: Toolbox operates on top of Podman. Not Buildah (but it used to). Also `buildah` is strictly for work with images while `podman` is capable of both (by including `buildah`) The exact command used in the case of the create() function is `podman create`. Online manage here: https://www.mankier.com/1/podman-create

debarshiray:
Yes, this needs to be ported to the Go implementation. Feel free to submit a new pull request.

PR #435, opened 2020-05-13
Prepare 0.0.90

HarryMichal:


NEWS
```diff
@@ -1,3 +1,21 @@
+Overview of changes in 0.0.90
+============================
+
+* Rewrite Toolbox in Go
+* Add option --log-podman to regulate output of Podman
+* Add option --version to show current Toolbox version
+* Fix duplication in the output of the list command
+* Make option --container in commands create and enter an argument
+* Make bash completion a bit more reliable
```
debarshiray, line 9: Did we touch the Bash completion in this release? I couldn't find anything.

NEWS
```diff
@@ -1,3 +1,21 @@
...
+* Make the create command abortable
+* Fix leaking of spinners when aborted
+* Replace option --verbose and --very-verbose with --log-level while keeping
+  verbose for backwards compatibility
+* Mark the reset command as deprecated (replaced by 'podman system reset')
+* Drop some no longer needed code
```
debarshiray, line 15: We should be more explicit by saying something like: *we dropped compatibility with ...*.

NEWS
HarryMichal: I made some changes [here](https://github.com/HarryMichal/toolbox/commit/9822d593d5f8946be52a6217820ae34ffb2f17e6). The changes did not get in.

NEWS
HarryMichal: Yes, true. I did not understand some of those code paths, so I was a bit too general.  The line could be dropped completely.

NEWS
debarshiray: Ok, I will drop that line from this release.

NEWS
debarshiray: We should still mention it though because it might still affect some users. Who knows? We just need to be more explicit about it.

NEWS
```diff
@@ -1,3 +1,21 @@
+Overview of changes in 0.0.90
+============================
+
+* Rewrite Toolbox in Go
+* Add option --log-podman to regulate output of Podman
+* Add option --version to show current Toolbox version
+* Fix duplication in the output of the list command
+* Make option --container in commands create and enter an argument
+* Make bash completion a bit more reliable
+* Make the create command abortable
```
debarshiray, line 10: Is this related to the leaking spinners? Or something else?

debarshiray: I assumed that it's about the leaking spinners. :)

PR #436, opened 2020-05-13
Update README

HarryMichal:
The README was in need of upgrading. I changed the strucutre a bit and
moved the 'Goals and Use Cases' section to the bottom.

Mainly the introduction of Toolbox received a major facelift. The build
instructions were also upgraded and some badges (resp. shields) were
added for quick access to the package info and tests dashboard (Zuul[0]
recently added support for dynamic badges. When the instance in
softwarefactory[1] is updated, we could wire it up later on).

Other sections were also updated. Mainly to clarify meaning of some
messages there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,152 +8,225 @@
 
 [Toolbox](https://github.com/containers/toolbox) is a tool that offers a
 familiar package based environment for developing and debugging software that
-runs fully unprivileged using [Podman](https://podman.io/).
+runs fully rootless using [Podman](https://podman.io/) and integrates with
+your system.
 
-The toolbox container is a fully *mutable* container; when you see
-`yum install ansible` for example, that's something you can do inside your
-toolbox container, without affecting the base operating system.
-
-This is particularly useful on
-[OSTree](https://ostree.readthedocs.io/en/latest/) based operating systems like
-[Fedora CoreOS](https://coreos.fedoraproject.org/) and
-[Silverblue](https://silverblue.fedoraproject.org/).  The intention of these
-systems is to discourage installation of software on the host, and instead
-install software as (or in) containers.
-
-However, this tool doesn't *require* using an OSTree based system — it
-works equally well if you're running e.g. existing Fedora Workstation or
-Server, and that's a useful way to incrementally adopt containerization.
-
-The toolbox environment is based on an [OCI](https://www.opencontainers.org/)
-image. On Fedora this is the `fedora-toolbox` image. This image is used to
-create a toolbox container that seamlessly integrates with the rest of the
-operating system.
-
-## Usage
+## Basic usage
 
 ### Create your toolbox container:
-```
+
+```shell
 [user@hostname ~]$ toolbox create
-Created container: fedora-toolbox-30
+Created container: fedora-toolbox-32
 Enter with: toolbox enter
 [user@hostname ~]$
 ```
-This will create a container called `fedora-toolbox-<version-id>`.
+
+This will create a container called `fedora-toolbox-<version-id>` based on the
+version of Fedora you are using.
 
 ### Enter the toolbox:
-```
+
+```shell
 [user@hostname ~]$ toolbox enter
 ⬢[user@toolbox ~]$
 ```
 
-## Dependencies and Installation
+> Notice: The '⬢' symbol may not be present if you are using a custom shell
+> theme.
 
-Toolbox requires at least Podman 1.4.0 to work, and uses the Meson build
-system.
+### Remove a toolbox container:
 
-The following dependencies are required to build it:
-- meson
-- go-md2man
+```shell
+[user@hostname ~]$ toolbox rm <container-name>
+[user@hostname ~]$
+```
+
+## Tell me more about Toolbox!
+
+Operating systems that are based on [OSTree](https://ostree.readthedocs.io/en/latest/)
+like [CoreOS](https://coreos.fedoraproject.org/) and
+[Fedora Silverblue](https://silverblue.fedoraproject.org/) discourage
+installation of software on the host (they mostly don't even have a *classic*
+package manager), and instead install software as/in containers.
+
+**Hmm, why is that so?**
+
+There are several advantages to this approach including higher reliability,
+easier upgrading and easier debugging. Fedora Magazine has nice articles about
+both [CoreOS](https://fedoramagazine.org/introducing-fedora-coreos/) and
+[Silverblue](https://fedoramagazine.org/what-is-silverblue/).
+
+**OK, what is then Toolbox good for?**
+
+Toolbox is a tool built on top of [Podman](https://podman.io) (which is a tool
+for managing [OCI](https://www.opencontainers.org/)-based containers) providing
+sane default values when creating/running containers (so-called *toolboxes*).
+These containers then *seamlessly integrate* with the rest of the operating
+system (e.g., environmental variables, home folder, shared memory, important
+locations in the root filesystem).
+
+Bear in might that Toolbox is not meant to be "safe" in terms of isolating your
+actions in the container from the host. It strives to extend your workflow
+rather than isolate it using the container technology.
+
+**Uh.., what?**
+
+Let's say you want to try to use `ansible` but you don't want to bloat your
+system (e.g., Fedora Workstation) or you can't really install it comfortably
+(e.g., Silverblue).
+
+Just create a container with `toolbox create`, enter it with `toolbox enter`
+(you should see the '⬢' symbol at the beginning of the command prompt) and
+type `sudo dnf install ansible` (the container is Fedora-based).
+
+When the installation is done you should have `ansible` prepared for use,
+without affecting the base operating system (if you leave the container,
+`ansible` will no longer be available until you return into it).
+
+And because your home folder is automatically mounted in the container, you
+can easily execute your already existing ansible playbooks.
+
+When you're done, you can simply exit the container by typing `exit` or
+pressing `Ctrl+D` and delete the container with `toolbox rm` and the system
+should be in the same state when you started (apart from files you may have
+created/edited/removed in your home folder or other locations shared with the
+host's filesystem).
+
+Also notice that you did not need to run Toolbox with `sudo` or as a
+priviledged user!
+
+**Sounds good! So, to use Toolbox I need one of those mentioned OSs?**
+
+No, you don't! This tool **doesn't require** using an OSTree-based system — it
+works equally well if you're running e.g. existing Fedora Workstation or
+Server. But if you haven't given Silverblue a try then you definitely should!
+
+## Documentation
+
+https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/
+
+## Installation
+
+### Fedora Silverblue
+
+Toolbox is installed by default on Fedora Silverblue.
+
+### Fedora
+
+`$ dnf install toolbox`
+
+[Package definition](https://src.fedoraproject.org/rpms/toolbox)
+
+### Arch Linux
+
+`$ pacman -S toolbox`
+
+[Package definition](https://www.archlinux.org/packages/community/x86_64/toolbox/)
+
+## How to build from source
+
+Toolbox uses [Meson](https://mesonbuild.com) buildsystem for building,
+installing and generating manual pages + bash completion.
+
+> Because `Meson` does [not](https://github.com/mesonbuild/meson/issues/123) have native support for Go, Toolbox uses a wrapper
+> script ([`src/go-build-wrapper`](https://github.com/containers/toolbox/blob/master/src/go-build-wrapper))
+> to execute the build.
+
+**The following dependencies are required to build Toolbox:**
+
+- podman (>=v1.4.0)
+- flatpak-session-helper (monitoring of key system files)
+- go (>=v1.13)
+- meson (>=v0.40.0) + ninja
+- go-md2man (generating manual pages)
 - systemd
 
 The following dependencies enable various optional features:
+
 - bash-completion
+- shellcheck
 
-It can be built and installed as any other typical Meson-based project:
-```
-[user@hostname toolbox]$ meson -Dprofile_dir=/etc/profile.d builddir
-[user@hostname toolbox]$ ninja -C builddir
-[user@hostname toolbox]$ sudo ninja -C builddir install
-```
+Consult the [src/go.mod](https://github.com/containers/toolbox/blob/master/src/go.mod)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,153 +1,232 @@
...
+[![Fedora package](https://img.shields.io/fedora/v/toolbox/rawhide)](https://src.fedoraproject.org/rpms/toolbox/)
+[![Arch Linux package](https://img.shields.io/archlinux/v/community/x86_64/toolbox)](https://www.archlinux.org/packages/community/x86_64/toolbox/)
+
 [Toolbox](https://github.com/containers/toolbox) is a tool that offers a
 familiar package based environment for developing and debugging software that
-runs fully unprivileged using [Podman](https://podman.io/).
+runs fully unprivileged using [Podman](https://podman.io/) and integrates with
+your system .
+
+**Ok, this sounds a bit criptic..**
```
debarshiray, line 14: Typo alert: cryptic

debarshiray (changes requested):
Thanks for working on this, @HarryMichal 

I think that it would be good to have the *Usage* section higher up above *Documentation* and the build instructions for the benefit of casual passers by. Most people who stumble across the Toolbox project won't want to read through the documentation. Since the package is shipped by default on Fedora, there's really no need to build it from source, and even on other distributions people are more likely to use the pre-built packages.  Having a few example commands is more likely to catch their attention.

Or in other words, I think that we should keep the `README.md` optimized for casual and curious readers, who don't necessarily know all the technical jargon, but at the same time avoid being misleading and not give any false information.

So, for example, you could stick the example commands right after you introduce Toolbox with one or two sentences. That will immediately set up a frame of reference in the reader's mind. Then you can dig into it a bit, and explain why someone would want to do this, how it works, etc..

I also wonder if we can avoid highlighting the word *unprivileged* so much. Different people interpret it wildly differently and end up being disappointed. We intend it to mean *rootless*, while others read it as *isolated and sandboxed*.

README.md
HarryMichal: Fixed.

HarryMichal:
> Thanks for working on this, @HarryMichal
> 
> I think that it would be good to have the _Usage_ section higher up above _Documentation_ and the build instructions for the benefit of casual passers by. Most people who stumble across the Toolbox project won't want to read through the documentation. Since the package is shipped by default on Fedora, there's really no need to build it from source, and even on other distributions people are more likely to use the pre-built packages. Having a few example commands is more likely to catch their attention.
> 
> Or in other words, I think that we should keep the `README.md` optimized for casual and curious readers, who don't necessarily know all the technical jargon, but at the same time avoid being misleading and not give any false information.
> 
> So, for example, you could stick the example commands right after you introduce Toolbox with one or two sentences. That will immediately set up a frame of reference in the reader's mind. Then you can dig into it a bit, and explain why someone would want to do this, how it works, etc..
> 
> I also wonder if we can avoid highlighting the word _unprivileged_ so much. Different people interpret it wildly differently and end up being disappointed. We intend it to mean _rootless_, while others read it as _isolated and sandboxed_.

Thank you for the review! I tried to address the points you raised. More info in the commit message :).

README.md
```diff
@@ -1,153 +1,232 @@
...
-systems is to discourage installation of software on the host, and instead
-install software as (or in) containers.
+[![Zuul](https://zuul-ci.org/gated.svg)](https://softwarefactory-project.io/zuul/t/local/builds?project=containers/toolbox)
+[![Daily Pipeline](https://img.shields.io/badge/zuul-periodic-informational)](https://softwarefactory-project.io/zuul/t/local/builds?project=containers%2Ftoolbox&pipeline=periodic)
 
-However, this tool doesn't *require* using an OSTree based system — it
-works equally well if you're running e.g. existing Fedora Workstation or
-Server, and that's a useful way to incrementally adopt containerization.
+[![Fedora package](https://img.shields.io/fedora/v/toolbox/rawhide)](https://src.fedoraproject.org/rpms/toolbox/)
+[![Arch Linux package](https://img.shields.io/archlinux/v/community/x86_64/toolbox)](https://www.archlinux.org/packages/community/x86_64/toolbox/)
```
debarshiray, line 7: I took the liberty to split out the badges into a [separate pull request](https://github.com/containers/toolbox/pull/589), while I try to rebase this PR on top of `master`.

README.md
```diff
@@ -8,152 +8,225 @@
...
-create a toolbox container that seamlessly integrates with the rest of the
-operating system.
-
-## Usage
+## Basic usage
 
 ### Create your toolbox container:
-```
+
+```shell
```
debarshiray, line 18: Nice. I didn't know that it's possible to specify the *language* like that.

However, it seems that it would be [better](https://stackoverflow.com/questions/20303826/highlight-bash-shell-code-in-markdown-files) [to use](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml) `console` instead of `shell`.

README.md
```diff
@@ -8,152 +8,225 @@
...
+[Package definition](https://www.archlinux.org/packages/community/x86_64/toolbox/)
+
+## How to build from source
+
+Toolbox uses [Meson](https://mesonbuild.com) buildsystem for building,
+installing and generating manual pages + bash completion.
+
+> Because `Meson` does [not](https://github.com/mesonbuild/meson/issues/123) have native support for Go, Toolbox uses a wrapper
+> script ([`src/go-build-wrapper`](https://github.com/containers/toolbox/blob/master/src/go-build-wrapper))
+> to execute the build.
```
debarshiray, line 134: It might be wise to avoid going into too many details. The more details we have here, the higher the chances that something will go stale and require updating.

For example, it's enough to say that it uses Meson. If someone is curious about the details of what happens during the build, they can go and read our `meson.build` files. Otherwise, it's not so important whether we generate the manual pages during the build or ship them verbatim or something else.

Same thing about how we shoehorn Go into Meson.

Or, in other words, the `README.md` should give accurate information, but also be brief enough for someone to quickly skim through and get a grasp of the most important aspects of the project. It's not primarily targeted at potential contributors. It's meant for a wide variety of people -- potential users, technical leads of different operating system teams, potential contributors, etc..

README.md
```diff
@@ -8,152 +8,225 @@
...
+
+**The following dependencies are required to build Toolbox:**
+
+- podman (>=v1.4.0)
+- flatpak-session-helper (monitoring of key system files)
```
debarshiray, line 139: Strictly speaking, these two aren't necessary during the build. Also, we are about to stop using `flatpak-session-helper` all together.

I'd also be careful about mentioning too many version numbers. They can go stale really quickly, and end up misleading people. I think the Podman version is the most important. It needs to be present at run-time on the host OS, and can have far-reaching consequences on the rest of the OS (think of OSes with long-term support).

README.md
```diff
@@ -8,152 +8,225 @@
...
+
+**Uh.., what?**
+
+Let's say you want to try to use `ansible` but you don't want to bloat your
+system (e.g., Fedora Workstation) or you can't really install it comfortably
+(e.g., Silverblue).
+
+Just create a container with `toolbox create`, enter it with `toolbox enter`
+(you should see the '⬢' symbol at the beginning of the command prompt) and
+type `sudo dnf install ansible` (the container is Fedora-based).
```
debarshiray, line 81: One good thing about having the example commands at the top of the `README.md` is that we don't need to describe the steps in detail. :)

README.md
```diff
@@ -8,152 +8,225 @@
...
+Server. But if you haven't given Silverblue a try then you definitely should!
 
-Toolbox requires at least Podman 1.4.0 to work, and uses the Meson build
-system.
+## Documentation
 
-The following dependencies are required to build it:
-- meson
-- go-md2man
+https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/
```
debarshiray, line 107: It might be better to consolidate all the documentation in the upstream manuals. eg., we can link to the top-level `toolbox(1)` manual here, and maybe those of the other commands too.

README.md
```diff
@@ -8,152 +8,225 @@
...
+> theme.
+
+### Remove a toolbox container:
+
+```shell
+[user@hostname ~]$ toolbox rm <container-name>
+[user@hostname ~]$
+```
+
+## Tell me more about Toolbox!
```
debarshiray, line 45: I wonder if we can compress this section a bit.

Currently, on this ThinkPad x220, I have to press *page down* ten times to scroll through the entire `README.md`. That's a lot of text! It might be better to aim to have each section fit on one page, with the exception of tables and lists which tend to consume more space.

Since the introductory sentence says *Toolbox ... is a ... environment*, maybe we can start this section more directly? eg., *The Toolbox environment is a Podman container ..." or something on those lines. We don't need to explain too much about what's OSTree or CoreOS or Silverblue. It's enough to link to some nice references and let the reader figure out all the good things like fearless and robust upgrades.

README.md
HarryMichal: And I didn't know there's `console` :). Let's use it!

allanday:
Hey! I've had a quick look through the current version and have some feedback - I hope it's useful.

On the opening line:

 - It says "runs fully rootless using Podman". It's not really clear why this is good; it feels like an implementation detail - not something I'd lead on.
 - It also says "and integrates with your system". This is vague; how does it integrate?
 - As an opening line, what about something like: "Toolbox is a tool for Linux operating systems, which allows the use of containerised command line environments. It is built on top of Podman and other standard container technologies."

General feedback

 - It feels like the basic usage section comes too early in the document- people need to understand what toolbox is and how to get it, before they start using it.
 - The document feels too long and some of the information doesn't seem in scope for a readme; I'd split out the sections on image requirements and goals and use cases
 - Personally, I'm not a huge fan of the conversational style in the "tell me more" section. It feels overly verbose and the headings don't help you navigate the content. I would prioritise efficiency and clarity of structure - help to people get the information they need as quickly as possible.

tpopela:
> * It says "runs fully rootless using Podman". It's not really clear why this is good; it feels like an implementation detail - not something I'd lead on.

Actually this is one of the main selling points if we are comparing Podman/Toolbox to Docker.

debarshiray:
Thanks for the feedback, @allanday !

> On the opening line:
> 
>     * It says "runs fully rootless using Podman". It's not really clear why this is good;
> it feels like an implementation detail - not something I'd lead on.

Yeah, I agree. Also, support for `sudo toolbox ...` makes this even more confusing.

The history here is that the word *unprivileged* was added in [this](https://github.com/containers/toolbox/commit/c047659c1d85ca982374da5c58ee7a24ba3847bd) commit, and it seems that it's a source of confusion for users because *unprivileged* means different things to different people.

I believe @HarryMichal was trying to spell out what *unprivileged* actually means, but I am fine with removing it altogether.

>     * It also says "and integrates with your system". This is vague; how does it integrate?

True. I am not sure what to do about it, though. On one hand it's confusing, on the other it's what Toolbox actually gives you on top of something like Podman.

Do we take it out? Do we add a separate section or page on how it integrates with the host?
 
>     * As an opening line, what about something like: "Toolbox is a tool for Linux operating systems,
> which allows the use of containerised command line environments. It is built on top of Podman and
> other standard container technologies."

Sounds good to me.

> General feedback
> 
>     * It feels like the basic usage section comes too early in the document- people need to understand
> what toolbox is and how to get it, before they start using it.

Two vaguely related thoughts on that:

* *How to get it* might not be that big of an issue because we expect Toolbox to be part of the default OS installation on fully supported OSes like Fedora Silverblue and Workstation, and Endless OS, and on others it's just a matter of installing the `toolbox` package.

* I felt that the introductory section of the current README.md had become a bit too technical after [this](https://github.com/containers/toolbox/commit/c047659c1d85ca982374da5c58ee7a24ba3847bd) change, and might not be so understandable to the uninitiated. I thought some usage examples could help make the text more approachable. Somewhat like screenshots for graphical applications.

debarshiray:
> As an opening line, what about something like: "Toolbox is a tool for Linux
> operating systems, which allows the use of containerised command line
> environments. It is built on top of Podman and other standard container
> technologies."

I submitted a pull request for this bit:
https://github.com/containers/toolbox/pull/597

README.md
debarshiray: Submitted https://github.com/containers/toolbox/pull/598 for this part.

allanday:
> > * It says "runs fully rootless using Podman". It's not really clear why this is good; it feels like an implementation detail - not something I'd lead on.
> 
> Actually this is one of the main selling points if we are comparing Podman/Toolbox to Docker.

Sure. The issue isn't that it's not a valid point - just that it needs framing and putting in the right place. Somewhere we can certainly say "Toolbox is built on top of containers and Podman, and it makes it really easy to use containers as part of your daily workflow. For example, unlike normal container tools, Toolbox can be run as a normal user rather than having to be root."

allanday:
> How to get it might not be that big of an issue because we expect Toolbox to be part of the default OS installation on fully supported OSes like Fedora Silverblue and Workstation, and Endless OS, and on others it's just a matter of installing the toolbox package.

A single sentence to that effect seems fine to me - it doesn't have to be a long section. But I think we do need that information there, because it's somewhat critical to understanding Toolbox and its position/role.

debarshiray:
> > How to get it might not be that big of an issue because we expect
> > Toolbox to be part of the default OS installation on fully supported
> > OSes like Fedora Silverblue and Workstation, and Endless OS, and
> > on others it's just a matter of installing the toolbox package.
> 
> A single sentence to that effect seems fine to me - it doesn't have to
> be a long section. But I think we do need that information there, because
> it's somewhat critical to understanding Toolbox and its position/role.

Ok, I have now done that in https://github.com/containers/toolbox/pull/683

debarshiray:
> The document feels too long and some of the information doesn't
> seem in scope for a readme; I'd split out the sections on image
> requirements and goals and use cases

I've split out the goals and use cases in https://github.com/containers/toolbox/pull/684

debarshiray:
> Personally, I'm not a huge fan of the conversational style in the
> "tell me more" section. It feels overly verbose and the headings
> don't help you navigate the content. I would prioritise efficiency
> and clarity of structure - help to people get the information they
> need as quickly as possible.

I reworked the meat of the `README.md` in https://github.com/containers/toolbox/pull/685

I think it's a little bit more approachable now.

debarshiray:
I am closing this pull request.

We have merged most (all?) of the changes through smaller pull requests after addressing all the comments in the discussion here, and this pull request is beginning to become too long and unwieldy. Feel free to open a new issue or pull request for further changes.

Thanks for all the work here, @HarryMichal, and thanks to everybody else for their feedback!

Issue #410, opened 2020-04-16
Periodic CI randomly fails on fedora-31 for `Echo 'Hello World' inside of the 'running' container`


TristanCacqueray:
From https://softwarefactory-project.io/zuul/t/local/builds?job_name=test-with-podman-stable-f31&pipeline=periodic

The job seems to fail on
````
not ok 13 Echo 'Hello World' inside of the 'running' container
# (from function `die' in file test/system/helpers.bash, line 218,
#  from function `run_toolbox' in file test/system/helpers.bash, line 107,
#  in test file test/system/201-run.bats, line 11)
#   `run_toolbox run -c running echo "Hello World"' failed with status 129
# $ ./toolbox run -c running echo Hello World
# [ rc=129 (** EXPECTED 0 **) ]
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #| FAIL: exit code is 129; expected 0
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Notice that this is only affecting f31, f30 seems to be in the clear: https://softwarefactory-project.io/zuul/t/local/builds?job_name=test-with-podman-stable-f30&pipeline=periodic

HarryMichal:
Thank you @TristanCacqueray for reporting this. I, too, noticed that some time ago. I'm not sure what may be causing the issue. I'll try to investigate when I can.

HarryMichal:
I finally got to look at this and after a few hours of searching, I still don't know the reason why this is happening. The issue has also evolved and now looks like this:

```
not ok 12 Echo 'Hello World' inside of the 'running' container
# (from function `is' in file test/system/helpers.bash, line 261,
#  in test file test/system/201-run.bats, line 12)
#   `is "$output" "Hello World" "Should say 'Hello World'"' failed
# $ /usr/local/bin/toolbox run -c running echo Hello World
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: Should say 'Hello World'
# #| expected: 'Hello World'
# #|   actual: ''
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

[...]

PR #438, opened 2020-05-14
test/system: Don't check the output of 'toolbox run'

HarryMichal:
Now that the rewritten version is the [default](#437), the tests need to be adjusted.

Also, we did not enable tests on Fedora 32 -> enabling with this.

### DIFF
--- a/test/system/201-run.bats
+++ b/test/system/201-run.bats
@@ -4,12 +4,12 @@ load helpers
 
 @test "Echo 'Hello World' inside of the default container" {
   run_toolbox run echo "Hello World"
-  is "$output" "Hello World" "Should say 'Hello World'"
+  # is "$output" "Hello World" "Should say 'Hello World'"
 }
 
 @test "Echo 'Hello World' inside of the 'running' container" {
   run_toolbox run -c running echo "Hello World"
-  is "$output" "Hello World" "Should say 'Hello World'"
+  # is "$output" "Hello World" "Should say 'Hello World'"
 }
 
 @test "Stop the 'running' container using 'podman stop'" {
@@ -19,5 +19,5 @@ load helpers
 
 @test "Echo 'hello World' again in the 'running' container after being stopped and exit" {
   run_toolbox run -c running echo "Hello World"
-  is "$output" "Hello World" "Should say 'Hello World'"
+  # is "$output" "Hello World" "Should say 'Hello World'"
 }


HarryMichal:
Looks like I'll need to adjust the parts where the tests check what was written in stdout/stderr. The spinner breaks those test cases.

debarshiray:
> Looks like I'll need to adjust the parts where the tests check what was written
> in stdout/stderr. The spinner breaks those test cases.

Oh, really? Is it because the spinner just looks different now? I think that the spinner is written to the standard output in both the POSIX shell and Go versions.

playbooks/shellcheck.yaml
```diff
@@ -5,7 +5,8 @@
       become: yes
       package:
         name:
-          - golang-github-cpuguy83-go-md2man
```
debarshiray, line 8: It would be nicer to split the changes related to switching `playbooks/shellcheck.yaml` to Fedora 32 into a separate commit. For a moment, I was thinking that this change from `golang-github-cpuguy83-go-md2man` to `golang-github-cpuguy83-md2man` is related to some sort of typo.

test/system/102-create.bats
```diff
@@ -7,14 +7,14 @@ load helpers
 }
 
 @test "Create a container with a valid custom name ('not-running')" {
-  run_toolbox -y create -c "not-running"
+  run_toolbox -y create "not-running"
```
debarshiray, line 10: We still support the `--container` flag, so we should also continue to test it. :)

HarryMichal: Ah, right.

playbooks/shellcheck.yaml
HarryMichal: Ok, will do.

HarryMichal:
> > Looks like I'll need to adjust the parts where the tests check what was written
> > in stdout/stderr. The spinner breaks those test cases.
> 
> Oh, really? Is it because the spinner just looks different now? I think that the spinner is written to the standard output in both the POSIX shell and Go versions.

I think the problem is in the `output` variable. I think it holds both `stdout` and `stderr`. I'll check the documentation.

playbooks/shellcheck.yaml
debarshiray: I took the liberty to split them out and merge them via https://github.com/containers/toolbox/pull/441 and https://github.com/containers/toolbox/pull/442

HarryMichal:
@debarshiray This should be ready to be merged. The tests are not passing but they are working which is the main point of this PR. I'm just wondering about the additional dependency introduced in https://github.com/containers/toolbox/pull/438/commits/11133ac8bdd091eb4394ea6ab2bf3bd7065d42da. What do you think??

test/system/102-create.bats
HarryMichal: I updated the test cases to use both ways of creating a named container.

HarryMichal:
Tests should be ok now. https://github.com/containers/toolbox/issues/410 still needs to be sorted out but that is a different issue (I suspect this is something introduced by Podman but I'm not sure. We'll see if V2 makes a difference regarding this matter). Rawhide tests are failing but that is due to an old version of Podman being available. Newer versions should have all the necessarry fixes.

HarryMichal:
recheck

debarshiray:
Thanks for updating the branch, @HarryMichal !

It seems that you switched from `golang.org/x/crypto/ssh/terminal` to `github.com/mattn/go-isatty`. I don't really have an opinion, but I am curious about what made you change your mind?

debarshiray:
I split out and [merged](https://github.com/containers/toolbox/pull/495) the patch to drop the tests for Fedora 30 to keep things moving.

src/cmd/create.go
```diff
@@ -181,6 +182,15 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 		return nil
 	}
 
+	s := spinner.New(spinner.CharSets[9], 500*time.Millisecond)
+
+	if logLevel := logrus.GetLevel(); logLevel < logrus.DebugLevel && isatty.IsTerminal(1) {
```
debarshiray, line 187: I am curious. Why are we moving the entire chunk of code that creates the spinner, instead of just adding the extra condition to the existing chunk in-place?

debarshiray:
> It seems that you switched from `golang.org/x/crypto/ssh/terminal` to
> `github.com/mattn/go-isatty`. I don't really have an opinion, but I am
> curious about what made you change your mind?

At first glance, anything that's under `golang.org/x` seems like a better choice because of its semi-official nature.

Looking at the API documentation of [golang.org/x/crypto/ssh/terminal](https://pkg.go.dev/golang.org/x/crypto/ssh/terminal) and [github.com/mattn/go-isatty](https://pkg.go.dev/github.com/mattn/go-isatty), it seems likely that the former will have a bigger footprint in terms of the final binary size. Preventing the binary size from growing is a worthwhile goal.

Podman uses `golang.org/x/crypto/ssh/terminal`, but one of its [dependencies](https://pkg.go.dev/github.com/vbauerster/mpb) uses `github.com/mattn/go-isatty`.

I see that `github.com/mattn/go-isatty` v0.0.12 is packaged in Fedora. However, strangely, it insists on a newer `golang.org/x/sys`, v0.0.0-20200116001909-b77594299b42, while Fedora only has `20200103gitc96a22e` which seems like a 13 days older snapshot.

debarshiray:
> Looking at the API documentation of [golang.org/x/crypto/ssh/terminal](https://pkg.go.dev/golang.org/x/crypto/ssh/terminal)
> and [github.com/mattn/go-isatty](https://pkg.go.dev/github.com/mattn/go-isatty), it seems likely that the former will have a
> bigger footprint in terms of the final binary size. Preventing the
> binary size from growing is a worthwhile goal.

After actually trying out both packages and building the resulting binaries, `golang.org/x/crypto/ssh/terminal` gave the smaller size. 7778323 bytes versus 7782284.

It's not what I expected because `github.com/mattn/go-isatty` seemed like a much smaller package in comparison. It's essentially just:
```
import "golang.org/x/sys/unix"

// IsTerminal return true if the file descriptor is terminal.
func IsTerminal(fd uintptr) bool {
        _, err := unix.IoctlGetTermios(int(fd), unix.TCGETS)
        return err == nil
}
```

But I cannot argue with actual data. :) Unless I am missing something.

debarshiray:
I discussed the spinner changes with @HarryMichal on Telegram, and he confirmed that there's no reason not to use the `golang.org/x/crypto/ssh/terminal` package. So, I took the liberty to tweak his commit accordingly, and merged it via https://github.com/containers/toolbox/pull/496

src/cmd/create.go
```diff
@@ -634,10 +635,9 @@ func pullImage(image, release string) (bool, error) {
 
 	logrus.Debugf("Pulling image %s", imageFull)
 
-	if logLevel := logrus.GetLevel(); logLevel < logrus.DebugLevel {
+	if logLevel := logrus.GetLevel(); logLevel < logrus.DebugLevel && isatty.IsTerminal(1) {
 		s := spinner.New(spinner.CharSets[9], 500*time.Millisecond)
 		s.Prefix = fmt.Sprintf("Pulling %s: ", imageFull)
-		s.Writer = os.Stdout
```
debarshiray, line 640: Is there a reason to remove this line?

debarshiray: After a few rounds of testing it seems accidental to me, so I filed https://github.com/containers/toolbox/pull/498 to put it back.

But let me know if I am missing something. :)

src/cmd/rm.go
```diff
@@ -70,6 +70,8 @@ func rm(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	var encounteredError bool = false
```
debarshiray, line 73: Yes, I had noticed this regression while reviewing the Go rewrite in https://github.com/containers/toolbox/pull/318 I chose to ignore it because I couldn't think of a nice way to do it back then and didn't realize that it would have a negative effect on the tests.

It is a bit ugly to have each command directly call `os.Exit` this way. It's generally nicer to have a single return or exit point. In this case, we'd somehow propagate the error (and/or the exit code) up the function call stack, and have `Execute` in `cmd/root.go` handle it accordingly. However, I don't know off-hand how to achieve that with Cobra, but it should be possible.

In the worst case, given that, as far as we know, nobody is using `/usr/bin/toolbox` as an API from another program, I'd be happier to tweak the test suite to not expect a non-zero exit code, as a short-term temporary measure. Doesn't the test suite already look for the expected error message strings? If so, then I can live with it, while we figure out something nicer.

HarryMichal: Fair enough. I'll drop those two commits and adjust the test cases.

HarryMichal:
recheck

playbooks/fedora-31/setup-env.yaml
```diff
@@ -5,6 +5,7 @@
       become: yes
       package:
         name:
+          - golang
```
debarshiray, line 8: It would be nicer to split this part out into a separate and smaller (eg., *add Go to the list of build requirements*) commit, instead of having it as part of a larger *update tests* commit.

test/system/301-rm.bats
```diff
@@ -3,13 +3,13 @@
 load helpers
 
 @test "Try to remove a nonexistent container" {
-  run_toolbox 1 rm nonexistentcontainer
-  is "$output" "toolbox: failed to inspect $todelete" "Toolbox should fail with: no such container"
+  run_toolbox rm nonexistentcontainer
```
debarshiray, line 6: I'd probably merge the *Don't expect non-zero code on container removal* commit with the *update tests* commit, because that's what this is about - but still mention the reason in the commit message. :)

test/system/201-run.bats
```diff
@@ -4,12 +4,12 @@ load helpers
 
 @test "Echo 'Hello World' inside of the default container" {
   run_toolbox run echo "Hello World"
-  is "$output" "Hello World" "Should say 'Hello World'"
+  # is "$output" "Hello World" "Should say 'Hello World'"
```
debarshiray, line 7: Should we talk to the Zuul folks about this?

I am happy to turn off the assertion as a temporary measure. I just hope that we don't forget about it.

debarshiray:
I [merged](https://github.com/containers/toolbox/pull/508) the renaming of the files and test names.

The only thing left is to figure out if we should notify Zuul about the *toolbox run* test failures.

test/system/201-run.bats
HarryMichal: A similar issue was brought up by the Zuul folks some time ago already (https://github.com/containers/toolbox/issues/410). I think we can track the issue there (the issue title should be updated, though).

debarshiray: Ok, that's great.

Nitpick: we should link to https://github.com/containers/toolbox/issues/410 from this commit to complete the paper trail.

debarshiray:
> The only thing left is to figure out if we should notify Zuul about the _toolbox run_ test failures.

Let's continue discussing that on https://github.com/containers/toolbox/pull/438

juanje:
Hi, I was rewriting the bats tests (for practicing purposes) and I found this issue with the `toolbox run`.

Here the test (using [bats-assert](https://github.com/ztombol/bats-assert)):
```
@test "Echo 'Hello World' inside of the default container" {
  create_default_container
  run toolbox run echo "Hello World"

  assert_success
  assert_output "Hello World"
}
```

And the fail:
```
 ✗ Echo 'Hello World' inside of the default container
   (from function `assert_output' in file test/system/libs/bats-assert/src/assert.bash, line 239,
    in test file test/system/1201-run.bats, line 28)
     `assert_output "Hello World"' failed
   
   -- output differs --
   expected : Hello World
   actual   : Hello World
   --
```

As you can see, the outputs seem to be the same, but after some checking, I found that the `toolbox run echo` added a `CRLF` at the end of the line.

After much testing I tried this:
```
$ toolbox run echo foo > toolbox_output.txt
$ echo foo >  output.txt
$ file *output.txt
output.txt:         ASCII text
toolbox_output.txt: ASCII text, with CRLF line terminators
```

I couldn't make it work at the test comparison, but if I add the flag `-n` to the command `echo` the test work:
```
$ toolbox run echo -n foo > toolbox_output.txt
$ file toolbox_output.txt
toolbox_output.txt: ASCII text, with no line terminators
```

Well, I could make the comparison works with the assert's flag `--partial`, but it looks to me as a hack for the test to pass rather than a solution:
```
assert_output --partial "Hello World"
```

The same with the `echo -n` trick. The problem remains because if I try other commands it doesn't work. For example:
```
$ toolbox run ls > toolbox_output.txt
$ file toolbox_output.txt
toolbox_output.txt: ASCII text, with CRLF line terminators
```

I guess it's something about the toolbox output. I hope this give you some useful info to debug the problem.

PR #221, opened 2019-07-13
Drop the "immutable" term

cgwalters:
I'm still waging a fight against this.  Please stop saying "immutable"
for this, it's more misleading than useful.

Previously e.g. https://github.com/debarshiray/toolbox/pull/43

PR #440, opened 2020-05-14
profile: Reword to say that containers are recommended

cgwalters:
First actually, drop the hardcoded "Welcome to Fedora Silverblue"
which has been wrong from the start since we support installing
toolbox on traditional systems too.

(And it's an anti-pattern for software to hardcode the name of
 the OS, we should have been using the data from `/usr/lib/os-release`
 there too)

Now, the other change here is a followup to https://github.com/containers/toolbox/pull/221

My goal from the very start with rpm-ostree has been to
encourage using it in "image mode" by default.  But package layering
is 100% supported and certainly *allowed*.

Remove the use of the term "allow" in "allow package installation"
which implies it's not allowed.

We recommend containers.  But ultimately, it's *your* computer, not
ours.  So let's not use terms like "allow" or "immutable".

See also
https://blog.verbum.org/2019/12/23/starting-from-open-and-foss/


HarryMichal:
The reasoning sounds very ok to me. I'm just wondering if the profile message is really "hardcoded" considering the message is printed when the host OS is Fedora Workstation or Silverblue.

But I agree that the wording is not right and this is a step in the right direction.

Also, I think this kind of a welcome message should be handled by the OSs themselves (at least in the case of Silverblue, because the workflow there is very different). It's not the job Toolbox to tell the users how to use the OS (or specifially Silverblue).

cgwalters:
Yep agreed!  Maybe it's time to just move this bit into the https://pagure.io/workstation-ostree-config and drop it from here.

cgwalters:
OK, updated to remove the welcome prompt entirely.

cgwalters:
BTW there was an interesting thread here https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/DDKMBPVGGRBRUJCLTTJNI7T6TGEMYNNG/

HarryMichal:
I'm not entirely sure if removing the welcome prompt entirely is right. As I said previously, the welcome message related to the OS (showed when first opening the terminal) should definitely be handled by the OS but the second message is related to Toolbox and is shown when a toolbox is entered for the first time. I think the second message should be kept (but I'd probably upgrade it a bit).

I'll ask Alan to say what he thinks.

allanday:
I don't see any arguments for removing the welcome message here - just for not stating the OS name and for changing some of the terminology. Why remove it completely?

We haven't got very far into thinking about the onboarding experience for a traditional, package-based OS, but I'd imagine that showing some information on first run would be useful. One option might be to keep the welcome message, and rewrite it to be toolbox specific (and OS agnostic).

debarshiray:
> First actually, drop the hardcoded "Welcome to Fedora Silverblue"
> which has been wrong from the start since we support installing
> toolbox on traditional systems too.

The welcome message is only shown on Silverblue systems, not on legacy Workstation, because it's conditional on `/run/ostree-booted`. So, you won't see a *Welcome to Fedora Silverblue* on a legacy Workstation system.

> (And it's an anti-pattern for software to hardcode the name of
> the OS, we should have been using the data from /usr/lib/os-release
> there too)

`/etc/profile.d/toolbox.sh` already uses `/usr/lib/os-release` for various things. It always has.

The reason the *Fedora Silverblue* string isn't parametrized is because Silverblue has the same `os-release` as Workstation. So things like `VARIANT` and `VARIANT_ID` are `Workstation Edition` and `workstation` respectively.

I have to check if this changed in Fedora 32 Silverblue, but probably not. Either way, the point is that it's a typical work-in-progress scenario. Things will get better as the various pieces evolve.

> My goal from the very start with rpm-ostree has been to
> encourage using it in "image mode" by default. But package layering
> is 100% supported and certainly allowed.
>
> Remove the use of the term "allow" in "allow package installation"
> which implies it's not allowed.
>
> We recommend containers. But ultimately, it's your computer, not
> ours. So let's not use terms like "allow" or "immutable".

Sure, I am happy to use some other text that doesn't use *allow*.

However, note that it's about installing packages in the usual way with DNF. The actual text reads:
```
You may want to try out the Toolbox for a directly mutable environment that allows package installation with DNF.
```

A big reason for Toolbox is that layering packages with `rpm-ostree` can quickly become tedious for a developer who is working through a project's dependencies. Hence, the importance of DNF.

debarshiray:
> As I said previously, the welcome message related to the OS (showed when
> first opening the terminal) should definitely be handled by the OS but the
> second message is related to Toolbox and is shown when a toolbox is entered
> for the first time. I think the second message should be kept (but I'd probably
> upgrade it a bit).

One could argue that the terminal emulator, (at least the default) shell and Toolbox are part of the OS. :)

Where the welcome banner actually resides, whether it's shipped by Bash or GNOME Terminal or VTE or something else, is a bit of an implementation detail. There's also the question whether it's very useful for the banner to mention Toolbox if `/usr/bin/toolbox` is missing.

HarryMichal:
My main argument for moving the welcome machine to Silverblue's config is that its appearance is not triggered by interacting with Toolbox but with a different part of the OS (GNOME Terminal or a different terminal).

I guess this discussion should continue in a different issue so that we don't block the PR.

HarryMichal:
Hmm... we never really got to merge this. I don't want to poke a wasp nest but I'm wondering about what we should do about this PR.

A quick summary. The profile script currently has 3 jobs:

1) printing welcome message when opening a terminal for the first time on the host (important for Silverblue)
2) printing welcome message after entering a toolbox for the first time
3) Adding the ⬢ symbol to PS1

This PR does not concern 3.

Based on what Allan said, right now we should not remove either 1. or 2. because the final desired onboarding experience has not been defined, yet.

My personal opinion is that 2. should remain in control of Toolbox and 1. should either:

- be handled by the OS itself (can be OS-specific and in the case of Silverblue can safely talk about Toolbox without having to be concerned of it being installed)
- continue being handled by Toolbox (must be OS-agnostic and needs to be opinionated about on which distros the message will be shown)

All in all, the wording of those messages definitely needs to be updated based on @cgwalters's original PR and could possibly be extended a bit (mainly 2.) to be more welcoming.

Does this sound right to you, @cgwalters @allanday @debarshiray?

allanday:
> Does this sound right to you, @cgwalters @allanday @debarshiray?

No strong opinion here on where the welcome messages should live.

juhp:
This could be closed now I think?

PR #445, opened 2020-05-15
Allow specifying user shell at toolbox creation

### LINKED ISSUE #413: Feature request: specify shell at toolbox creation time
Hello,

Currently, the shell launched when entering a toolbox is the one specified in the `SHELL` environment variable. I'd like to be able to specify which shell will be used by a toolbox at creation time. This would allow having a different shell on the host (bash) and in the toolbox (zsh). It would also allow having several toolboxes, each with its specific shell (maybe a python toolbox using an interactive python interpreter as shell?).

I can try to work on a PR if this idea is accepted.

yann-soubeyrand:
When creating a toolbox the user shell can be specified and this shell will be
used when entering the toolbox.

Signed-off-by: Yann Soubeyrand <yann.soubeyrand@camptocamp.com>

Closes #413

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -46,6 +46,7 @@ var (
 		distro    string
 		image     string
 		release   string
+		shell     string
 	}
 
 	createToolboxShMounts = []struct {
@@ -89,6 +90,12 @@ func init() {
 		"r",
 		"",
 		"Create a toolbox container for a different operating system release than the host")
+
+	flags.StringVarP(&createFlags.shell,
+		"shell",
+		"s",
+		"",
+		"Specify the shell to use in the toolbox.")
 
 	createCmd.SetHelpFunc(createHelp)
 	rootCmd.AddCommand(createCmd)
@@ -156,14 +163,23 @@ func create(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if err := createContainer(container, image, release, true); err != nil {
+	var userShell string
+	if createFlags.shell != "" {
+		userShell = createFlags.shell
+		if !filepath.IsAbs(userShell) {
+			return fmt.Errorf("Invalid shell '%s': must be an absolute path", userShell)
+		}
+		userShell = filepath.Clean(userShell)
+	}
+
+	if err := createContainer(container, image, release, userShell, true); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func createContainer(container, image, release string, showCommandToEnter bool) error {
+func createContainer(container, image, release string, userShell string, showCommandToEnter bool) error {
 	if container == "" {
 		panic("container not specified")
 	}
@@ -371,9 +387,11 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 
 	logLevelString := podman.LogLevel.String()
 
-	userShell := os.Getenv("SHELL")
 	if userShell == "" {
-		return errors.New("failed to get the current user's default shell")
+		userShell = os.Getenv("SHELL")
+		if userShell == "" {
+			return errors.New("failed to get the current user's default shell")
+		}
 	}
 
 	entryPoint := []string{
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -127,25 +127,20 @@ func run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if err := runCommand(container,
+	return runCommand(container,
 		!nonDefaultContainer,
 		image,
 		release,
 		command,
 		false,
-		false,
-		true); err != nil {
-		return err
-	}
-
-	return nil
+		true)
 }
 
 func runCommand(container string,
 	defaultContainer bool,
 	image, release string,
 	command []string,
-	emitEscapeSequence, fallbackToBash, pedantic bool) error {
+	emitEscapeSequence, pedantic bool) error {
 	if !pedantic {
 		if image == "" {
 			panic("image not specified")
@@ -195,7 +190,7 @@ func runCommand(container string,
 				return nil
 			}
 
-			if err := createContainer(container, image, release, false); err != nil {
+			if err := createContainer(container, image, release, "", false); err != nil {
 				return err
 			}
 		} else if containersCount == 1 && defaultContainer {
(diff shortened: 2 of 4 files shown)


src/cmd/run.go
```diff
@@ -120,25 +120,20 @@ func run(cmd *cobra.Command, args []string) error {
...
-
-	return nil
+		true)
 }
 
 func runCommand(container string,
 	defaultContainer bool,
 	image, release string,
 	command []string,
-	emitEscapeSequence, fallbackToBash, pedantic bool) error {
```
yann-soubeyrand, line 141: I removed fallbackToBash as it was only used in the enter command since the shell can now be specified. I'm not sure if I should have kept it or not though.

src/cmd/run.go
```diff
@@ -283,10 +277,14 @@ func runCommand(container string,
...
 		container,
-		"capsh", "--caps=", "--", "-c", "exec \"$@\"", "/bin/sh",
+		"capsh", "--caps=", "--", "-c", commandLauncher, "toolbox",
 	}...)
 
-	execArgs = append(execArgs, command...)
+	if command != nil {
+		execArgs = append(execArgs, command...)
+	} else {
+		command = []string{"\"$SHELL\""}
```
yann-soubeyrand, line 286: command variable cannot be null as it's accessed in the error messages after Podman returns.

yann-soubeyrand: I'm not certain these error handling should be kept as is though, since it can be misleading. For example, if you enter a toolbox, try to launch a command which doesn't exist, and immediately exit the toolbox (with ctrl+d), you'll get "Error: command /bin/bash not found in container fedora-toolbox-32" (assuming your shell is /bin/bash and your toolbox container is named fedora-toolbox-32).

yann-soubeyrand:
> Build failed.
> 
>     * [shellcheck ](https://softwarefactory-project.io/zuul/t/local/build/15af452961464632940b3d9affe9a94d) : SUCCESS in 1m 13s
> 
>     * [test-with-podman-stable-f30 ](https://softwarefactory-project.io/zuul/t/local/build/fbb58d02449848c7abb0b2a3f2af221d) : FAILURE in 2m 49s
> 
>     * [test-with-podman-stable-f31 ](https://softwarefactory-project.io/zuul/t/local/build/c00aefea732d42d8a8a5f0ee78a8fada) : FAILURE in 2m 28s
> 
>     * [test-with-podman-rawhide ](https://softwarefactory-project.io/zuul/t/local/build/d3f373a87e984eac93021e0be87472b7) : FAILURE in 2m 42s

Is it some kind of normal error or did I break something? It seems that meson cannot find the go compiler.

yann-soubeyrand:
Hi,
@HarryMichal, @debarshiray, could you have a look at this PR and tell me if I should rework it please?

HarryMichal:
@yann-soubeyrand, thank you for the patch! I'm currently focusing on v0.1.0 that does not add any new features. When we get that out, I'll review this in-depth.

debarshiray:
What's the use-case for this?

Toolbox containers are so-called *pet* containers, and meant to be somewhat long-lived interactive command-line environments. Hence, I'd expect users to configure their default shell in the usual way.

yann-soubeyrand:
Hello @debarshiray,

Indeed, I configure my default shell (zsh) in my toolbox container. However, I keep the system default shell on the host (bash on Fedora Silverblue). This is for several reasons: I try to avoid installing extra packages on my Silverblue host, and having a different shell on the host and inside my toolbox allows me to quickly differentiate a host shell and a toolbox shell.

As I understood, you validated the idea (https://github.com/containers/toolbox/issues/413#issuecomment-628011839), that's why I've worked on it ;-)

debarshiray:
> However, I keep the system default shell on the host (bash on
> Fedora Silverblue). This is for several reasons: I try to avoid
> installing extra packages on my Silverblue host, and having a
> different shell on the host and inside my toolbox allows me to
> quickly differentiate a host shell and a toolbox shell.

Indeed. Not having to install extra packages on a Silverblue host is a very good reason.

> As I understood, you validated the idea (#413 (comment)), that's why I've worked on it ;-)

Oops, my mistake. I was quickly going through the open PRs and forgot about the earlier issue. Sorry about that.

yann-soubeyrand:
> > As I understood, you validated the idea (#413 (comment)), that's why I've worked on it ;-)
> 
> Oops, my mistake. I was quickly going through the open PRs and forgot about the earlier issue. Sorry about that.

No problem ;-)

I've just rebased my PR so that it applies cleanly.

yann-soubeyrand:
Hello @HarryMichal @debarshiray, is there something blocking this PR which I can act on?

debarshiray:
I suppose this needs to rebased after https://github.com/containers/toolbox/pull/813

src/cmd/create.go
```diff
@@ -90,6 +91,12 @@ func init() {
 		"",
 		"Create a toolbox container for a different operating system release than the host")
 
+	flags.StringVarP(&createFlags.shell,
+		"shell",
```
debarshiray, line 95: It will be better not to add a new flag to `toolbox create`. Instead, I would let users configure it separately through `usermod --shell` or `chsh` inside the container. Those are easy enough to do and well understood, and we can avoid exposing an extra user-facing interface that we would have to maintain forever.

The main part is to somehow get `toolbox enter` to actually use the default shell specified for the user inside the container, instead of always using what's in the `SHELL` environment variable on the host.

src/cmd/run.go
```diff
@@ -265,18 +260,15 @@ func runCommand(container string,
...
 	logrus.Debugf("Container %s is initialized", container)
 
-	if _, err := isCommandPresent(container, command[0]); err != nil {
-		if fallbackToBash {
-			fmt.Fprintf(os.Stderr,
-				"Error: command %s not found in container %s\n",
-				command[0],
-				container)
-			fmt.Fprintf(os.Stderr, "Using /bin/bash instead.\n")
+	commandLauncher := "export SHELL=\"$(getent passwd \"$(whoami)\" | cut -d : -f 7)\"; "
```
debarshiray, line 263: The `toolbox shell` idea described in https://github.com/containers/toolbox/issues/908 would be a better way to do this.

It would avoid all the gymnastics with shell snippets, and might make the fallback code faster.

debarshiray (changes requested):
Sorry for the delay!

debarshiray:
I am going to close this pull request because it would need a complete re-write. See my comments above.

However, we still want the feature, so please feel free to submit a new pull request.

src/cmd/create.go
debarshiray: @yann-soubeyrand has a work in progress at https://github.com/containers/toolbox/pull/1037

src/cmd/run.go
debarshiray: @yann-soubeyrand has a work in progress at https://github.com/containers/toolbox/pull/1037

debarshiray:
> I am going to close this pull request because it would need a complete re-write. See my comments above.
> 
> However, we still want the feature, so please feel free to submit a new pull request.

@yann-soubeyrand has a work in progress at https://github.com/containers/toolbox/pull/1037

PR #446, opened 2020-05-16
Document alternatives (closes #227)

### LINKED ISSUE #227: Are there any alternatives to toolbox? (i.e. other implementations)
To understand some software it is also good to know about the alternatives.

What other implementations are there of software doing similar things as this GitHub project?

abitrolly:


HarryMichal:
Hi @abitrolly! Thanks for the PR! The process of fixing the tests is in process and is tracked in #438. So you can drop commit d3d9076f1fd8328369b3bafe060463f526bb4a1d from the PR.

Also, I think showing a list of alternatives is fine but I wouldn't advertise their pros/cons over this Toolbox.

abitrolly:
@HarryMichal removed the test code. 

As for "advertising" the alternatives, the whole point of mentioning them is to let people select the tool that will suit them without spending few days trying to get `toolbox` do something it is not capable of. As such, the list of alternatives here is not about pros/cons, but about missing features in `toolbox`. If those features are implemented, then I agree that listing them in description would sound like an ad.

PR #448, opened 2020-05-16
go-build-wrapper: Ensure reproducible builds

Foxboron:
The go compiler embeds full paths to modules and dependencies into the
binary. This prevents people from reconstructing a bit-for-bit identical
toolbox binary without going through several hoops.

Without patch:

    $ strings build/src/toolbox | grep "$HOME" | wc -l
    105

With patch:

    $ strings build/src/toolbox | grep "$HOME" | wc -l
    0

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -27,5 +27,5 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -ldflags "-X github.com/containers/toolbox/pkg/version.currentVersion=$3" -o "$2"
+go build -trimpath -ldflags "-X github.com/containers/toolbox/pkg/version.currentVersion=$3" -o "$2"
 exit "$?"


src/go-build-wrapper
```diff
@@ -27,5 +27,5 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -o "$2"
+go build -trimpath -o "$2"
```
debarshiray, line 30: I really don't know what the trade-offs of using `-trimpath` are.

It seems like the Fedora's Go build flags don't include it, but maybe that's because Fedora doesn't build using Go modules?

For the sake of reference Fedora uses:
```
go build -buildmode pie -compiler gc -tags="rpm_crashtraceback ${BUILDTAGS:-}" -ldflags "${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n') -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '" -a -v -x
```

Foxboron: Fedora build guidelines are for fedora packages. These changes are for the binaries you publish on the github release page. What fedora does downstream doesn't really matter in this context.

debarshiray: Distributors like Fedora do care a lot about the reproducibility of their builds. Since this is about reproducible builds, the question arises why the `-trimpath` flag isn't seemingly being used by Fedora. Is it a mistake? Does it not matter for non-module builds?

By the way, (as a workaround?) `GOFLAGS` should be able to introduce the flag. Or is not working for some reason?

Foxboron: > Since this is about reproducible builds, the question arises why the -trimpath flag isn't seemingly being used by Fedora. Is it a mistake? Does it not matter for non-module builds?

Any reproduction in the context of distribution would be using the same build paths. It doesn't really matter for us, but it's included regardless. Fedora doesn't include it simply because they do not care about reproducible builds and hasn't worked towards it the past 4 or so years.

>By the way, (as a workaround?) GOFLAGS should be able to introduce the flag. Or is not working for some reason?

We do this in Arch. But this change is for *your* published builds. Not ours.

Conan-Kudo: > Fedora doesn't include it simply because they do not care about reproducible builds and hasn't worked towards it the past 4 or so years.

This is massively flawed and not true. We haven't used it because we didn't know about it.

Conan-Kudo: Also, the Koji build infrastructure uses fixed build paths for everything, so this likely has not come up before, since replaying a Koji build would already have that input be exactly the same.

debarshiray: It took me a while to get to the bottom of this.

I [filed](https://pagure.io/go-rpm-macros/issue/25) a feature request to include `-trimpath` in Fedora.

Even though Fedora's Go build flags, as defined by RPM macros like `%gobuild` and such, don't include `-trimpath`, searching through the build logs of a typical Go package in Fedora does reveal the use of `-trimpath`. eg., here is a snippet from the Toolbox build logs:
```
...
github.com/spf13/cobra
mkdir -p $WORK/b079/
...
cd /usr/share/gocode/src/github.com/spf13/cobra
/usr/lib/golang/pkg/tool/linux_amd64/compile -o $WORK/b079/_pkg_.a -trimpath "$WORK/b079=>" -shared -p github.com/spf13/cobra -complete -installsuffix shared -buildid MKzTveashal7vQXDb6Ke/MKzTveashal7vQXDb6Ke -goversion go1.14.3 -D "" -importcfg $WORK/b079/importcfg -pack ./args.go ./bash_completions.go ./cobra.go ./command.go ./command_notwin.go ./powershell_completions.go ./shell_completions.go ./zsh_completions.go
/usr/lib/golang/pkg/tool/linux_amd64/buildid -w $WORK/b059/_pkg_.a # internal
...
```

What's actually happening here is that `go build` always invokes the underlying Go compiler, in this case `/usr/lib/golang/pkg/tool/linux_amd64/compile`, to remove the paths of the intermediate object files. However, the paths to the sources are only removed if `go build -trimpath` is used. Also, note that the `-trimpath` flag in `go build` is slightly different from that of the compiler. The former is a boolean, while the latter requires a string argument.

Grep for `trimpath` in the Go [sources](https://github.com/golang/go) for further details.

debarshiray:
Thanks for your contribution, @Foxboron !

Issue #423, opened 2020-04-29
Fedora CoreOS: Unable to find group sudo


Beholder101:
Hello,

I am trying to run toolbox on a Fedora CoreOS instance without any luck. The Toolbox is created as expected and without any errors, but entering the toolbox results in:

```toolbox: running as real user ID 1000
toolbox: resolved absolute path for /usr/bin/toolbox to /usr/bin/toolbox
toolbox: checking if /etc/subgid and /etc/subuid have entries for user core
toolbox: TOOLBOX_PATH is /usr/bin/toolbox
toolbox: running on a cgroups v1 host
toolbox: current Podman version is 1.8.2
toolbox: migration not needed: Podman version 1.8.2 is unchanged
toolbox: Fedora generational core is f31
toolbox: base image is fedora-toolbox:31
toolbox: container is fedora-toolbox-31
toolbox: checking if container fedora-toolbox-31 exists
toolbox: calling org.freedesktop.Flatpak.SessionHelper.RequestSession
toolbox: starting container fedora-toolbox-31
toolbox: /etc/profile.d/toolbox.sh already mounted in container fedora-toolbox-31
Error: unable to start container "fedora-toolbox-31": error looking up supplemental groups for container 673c0263ae66255178c9cd92cc47b61c5ebc19b9d27971408a3eacced61230c4: Unable to find group sudo
toolbox: failed to start container fedora-toolbox-31
```
When i cat /etc/group i can see sudo does exist and that my user core is a member
```root:x:0:
adm:x:4:core
wheel:x:10:core
sudo:x:16:core
systemd-journal:x:190:core
docker:x:982:core
core:x:1000:
fedora-coreos-pinger:x:981:
zincati:x:980:
```

Running the commands as sudo (sudo toolbox create) results in a whole other error i won't bring up here unless requested.

I am running the most recent release
```State: idle
AutomaticUpdates: disabled
Deployments:
* ostree://fedora:fedora/x86_64/coreos/stable
                   Version: 31.20200407.3.0 (2020-04-21T19:37:39Z)
                    Commit: 89e17cc21b6aa3bea8959d1e6957fda157168d57ba6805d8a36142184edc2901
              GPGSignature: Valid signature by 7D22D5867F2A4236474BF7B850CB390B3C3359C4
```

What am i missing here? What am i doing wrong?

Regards, Peter

egeturgay:
I experience the same issue with the testing release Fedora CoreOS 31.20200420.2.0

[...]

PR #456, opened 2020-05-24
cmd/initContainer: write a password-less sudo config for distros that don't have nullok by default

### LINKED ISSUE #452: Toolbox assumes that the container image allows empty passwords
For passwordless `sudo` to work, toolbox currently deletes `root`'s password in the container when running `toolbox init-container`.

However, some images, such as the openSUSE toolbox image (`registry.opensuse.org/opensuse/toolbox`), don't have the `nullok` option in PAM by default, rendering `sudo` totally unusable.

There are two ways of solving this:

1. Require toolbox images to allow null passwords. For openSUSE, this means running `pam-config -a --nullok` while building the image.
2. To keep images as "stock" as possible, have toolbox insert a file in `/etc/sudoers.d` with the content

    ```
    %wheel ALL=(ALL) NOPASSWD: ALL
    ```

    for passwordless `sudo`.

tudurom:
Fixes #452.


tudurom:
This wouldn't work with Ubuntu, as it uses the `sudo` group instead of `wheel`. Maybe require images to supply the group as a container image label if it's not `wheel`?

abderrahim:
> This wouldn't work with Ubuntu, as it uses the `sudo` group instead of `wheel`. Maybe require images to supply the group as a container image label if it's not `wheel`?

toolbox already has a way to figure out the group for `sudo`: you have a variable `sudoGroup` defined in this scope

HarryMichal:
Hi @tudurom! As @abderrahim said, a bit earlier in `init-container` is a call to a function that returns which `sudo` group is available.

tudurom:
Updated!

abderrahim:
I found out this wouldn't work if the sudo group is different between the host and toolbox, see #423. This would have to wait until that issue is fixed.

PR #459, opened 2020-05-27
Update issue templates

HarryMichal:
Bug reporters most of the time don't know what and how much info they should share with us.


HarryMichal:
> Yes, it would be good to have some issue templates.
> 
> One nitpick, though. Would it be possible to use hyphens, instead of underscores, in the filenames? eg., `bug-report.md`, etc. Barring `meson_options.txt`, which is mandated by Meson itself, we use hyphens everywhere.

Yep, should be doable.

PR #460, opened 2020-05-27
cmd/initContainer: Reinforce symlink redirection

### LINKED ISSUE #187: /etc/resolv.conf is broken when it's an absolute symbolic link on the host
systemd-resolved has you symlink `/etc/resolv.conf` to some other location (under `/run/systemd/resolve`). When the host volume /etc is mounted, this location of course does not exist, so all hostname access within the container fails.

HarryMichal:
There has been a long-standing issue[0] when symlinking /etc/resolv.conf
to /run/host/etc/resolv.conf (host's resolv.conf). Many solutions were
proposed. Hardcode other paths, where the file could be, completely rely
on tools like flatpak-session-helper that will track such files in a
single directory, or update the symlinking function to follow symlinks
and update the target path.

The first solution is the easiest short-term but not very good
long-term. The second solution is possibly the best long-term but there
is a problem with using flatpak-session-helper. It cannot be used as
root. That leaves the third option, follow symlinks and update the
target path if the target is an invalid symlink.

This commit takes the third approach to solve the issue. Now the target
of the symlinking is first tested if it is a symlink. If it's not then
the symlinking is done right away. If it's a symlink then it is resolved.
When the target is valid, symlinking proceeds normally. If it it's not
then symlinking still proceeds but in two different ways depending on
target being an absolute or a relative symlink:

  - absolute - target is prepended with /run/host (the target may not be
    invalid)
  - relative - target is not changed (the target will be invalid)

This commit tries to rely on well-made relative symlinks. Those behave
correctly even when they are placed in a different prefix (in Toolbox's
case under /run/host).

Thanks Tudor Roman for raising concern about relative links.

Closes #187

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path/filepath"
 	"strings"
 	"syscall"
 
@@ -407,8 +408,47 @@ func mountBind(containerPath, source, flags string) error {
 	return nil
 }
 
+// redirectPath serves for creating symbolic links for crucial system
+// configuration files to their counterparts on the host's filesystem.
+//
+// containerPath and target must be absolute paths
+//
+// If the target itself is a symbolic link, redirectPath will evaluate the
+// link. If it's valid then redirectPath will link containerPath to target.
+// If it's not, then redirectPath will still proceed with the linking in two
+// different ways depending whether target is an absolute or a relative link:
+//
+//   * absolute - target's destination will be prepended with /run/host, and
+//                containerPath will be linked to the resulting path. This is
+//                an attempt to unbreak things, but if it doesn't work then
+//                it's the user's responsibility to fix it up.
+//
+//                This is meant to address the common case where
+//                /etc/resolv.conf on the host (ie., /run/host/etc/resolv.conf
+//                inside the container) is an absolute symbolic link to
+//                something like /run/systemd/resolve/stub-resolv.conf. The
+//                container's /etc/resolv.conf will then get linked to
+//                /run/host/run/systemd/resolved/resolv-stub.conf.
+//
+//                This is why properly configured hosts should use relative
+//                symbolic links, because they don't need to be adjusted in
+//                such scenarios.
+//
+//   * relative - containerPath will be linked to the invalid target, and it's
+//                the user's responsibility to fix it up.
+//
+// folder signifies if the target is a folder
 func redirectPath(containerPath, target string, folder bool) error {
-	logrus.Debugf("Redirecting %s to %s", containerPath, target)
+	if !filepath.IsAbs(containerPath) {
+		panic("containerPath must be an absolute path")
+	}
+
+	if !filepath.IsAbs(target) {
+		panic("target must be an absolute path")
+	}
+
+	logrus.Debugf("Preparing to redirect %s to %s", containerPath, target)
+	targetSanitized := sanitizeRedirectionTarget(target)
 
 	err := os.Remove(containerPath)
 	if folder {
@@ -421,9 +461,59 @@ func redirectPath(containerPath, target string, folder bool) error {
 		}
 	}
 
-	if err := os.Symlink(target, containerPath); err != nil {
+	logrus.Debugf("Redirecting %s to %s", containerPath, targetSanitized)
+
+	if err := os.Symlink(targetSanitized, containerPath); err != nil {
 		return fmt.Errorf("failed to redirect %s to %s: %w", containerPath, target, err)
 	}
 
 	return nil
+}
+
+func sanitizeRedirectionTarget(target string) string {
+	if !filepath.IsAbs(target) {
+		panic("target must be an absolute path")
+	}
+
+	fileInfo, err := os.Lstat(target)
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Warnf("%s not found", target)
+		} else {
+			logrus.Warnf("Failed to lstat %s: %v", target, err)
+		}
+
+		return target
+	}
+
+	fileMode := fileInfo.Mode()
+	if fileMode&os.ModeSymlink == 0 {
+		logrus.Debugf("%s isn't a symbolic link", target)
+		return target
+	}
+
+	logrus.Debugf("%s is a symbolic link", target)
+
+	_, err = filepath.EvalSymlinks(target)
+	if err == nil {
+		return target
+	}
+
+	logrus.Warnf("Failed to resolve %s: %v", target, err)
+
+	targetDestination, err := os.Readlink(target)
+	if err != nil {
+		logrus.Warnf("Failed to get the destination of %s: %v", target, err)
+		return target
+	}
+
+	logrus.Debugf("%s points to %s", target, targetDestination)
+
+	if filepath.IsAbs(targetDestination) {
+		logrus.Debugf("Prepending /run/host to %s", targetDestination)
+		targetGuess := filepath.Join("/run/host", targetDestination)
+		return targetGuess
+	}
+
+	return target
 }


src/cmd/initContainer.go
```diff
@@ -400,7 +400,41 @@ func mountBind(containerPath, source, flags string) error {
...
+	// If the target is a link, follow it. If needed, repeat the process.
+	for isSymlink {
+		logrus.Debugf("Path %s is a symlink. Resolving.", target)
+		newTarget, err := os.Readlink(target)
+		if err != nil {
+			return fmt.Errorf("Could not read link %s", target)
+		}
+
+		// Try to find the file inside of the container
+		target = fmt.Sprintf("/run/host%s", newTarget)
```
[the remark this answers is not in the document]
HarryMichal: Yes, I am aware of this shortcoming. I don't deny that I put this patch together rather quickly and it is not *the perfect* solution. I currently don't have a solution for this ready but I might at least add to the comment a `TODO` to highlight the need to make this mechaninsm more flexible.

HarryMichal:
I extended the comment in the function to clarify what it does, added a description of the `redirectPath` function and wrapped returned errors around the original errors.

src/cmd/initContainer.go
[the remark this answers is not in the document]
tudurom: One solution might be to check if the symlink is absolute with `filepath.IsAbs()`. If not, join it to the directory it resides in (`filepath.Dir()`). Then join the resulting path to `"/run/host"`. You should join paths with `filepath.Join()` instead of `fmt.Sprintf()`.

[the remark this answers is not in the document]
HarryMichal: I incorporated your suggestion in the latest iteration of the PR. What do you think?

HarryMichal:
@debarshiray when you have time, could you look at this?? https://github.com/containers/toolbox/pull/460/commits/2c4405dbdb690bb17c12915ce36250c045b67603 and https://github.com/containers/toolbox/pull/460/commits/88a7ae5bd614bc9e9f006f3e5a932afde7a3b468 made this a bit larger change but I think it's necessary to include it to keep toolboxes work across system updates (or rebases on Silverblue/CoreOS).

src/cmd/initContainer.go
```diff
@@ -397,22 +398,81 @@ func mountBind(containerPath, source, flags string) error {
...
+			return fmt.Errorf("target %s was not found in the container", target)
+		}
+		return fmt.Errorf("failed to lstat %s: %w", target, err)
+	}
+	isSymlink := (targetInfo.Mode() & os.ModeSymlink) != 0
+
+	// If the target is a link, follow the chain. If needed, repeat the process.
+	for isSymlink {
+		logrus.Debugf("Path %s is a symlink. Resolving.", target)
+		newTarget, err := os.Readlink(target)
```
debarshiray: Am I wrong in thinking that a lot of this can be simplified by using [filepath.EvalSymlinks](https://golang.org/pkg/path/filepath/#EvalSymlinks) ?

For example, I have:
```
$ ls -l /etc/os-release 
lrwxrwxrwx. 1 root root 21 Apr 28 14:50 /etc/os-release -> ../usr/lib/os-release
```

Invoking `filepath.EvalSymlinks("/etc/os-release")` gives me `/usr/lib/os-release`.

For the sake of sanity, I don't mind requiring that the `redirectPath` function accepts only absolute paths, because it's an internal function whose callers are under our control. If we go down this path, then I would check the inputs early on in the function and `panic` out (because it would be a programmer error) if the conditions aren't met.

src/cmd/initContainer.go
```diff
@@ -397,22 +398,81 @@ func mountBind(containerPath, source, flags string) error {
...
+		if !isSymlink {
+			logrus.Debugf("Resolved and redirecting %s to %s", containerPath, newTarget)
+		}
+	}
+
+	err = os.Remove(containerPath)
 	if folder {
 		if err != nil {
-			return fmt.Errorf("failed to redirect %s to %s", containerPath, target)
+			return fmt.Errorf("failed to redirect %s to %s: %w", containerPath, target, err)
```
debarshiray: This part is unrelated and should split into a separate commit.

debarshiray: I took the liberty to split it out into https://github.com/containers/toolbox/pull/527

src/cmd/initContainer.go
debarshiray: Ignoring the exact Go APIs that are used, I think that the `redirectPath` function should roughly do the following:

* If `target` is not a symbolic link, then don't bother checking if it exists or not. Just link `containerPath` to `target`.

* If `target` is a symbolic link, then check if it's resolving. If it's resolving, then link `containerPath` to `target`.

* If `target` is a symbolic link and if it's not resolving, then either the link is absolute or relative. If it's absolute, then prepend `/run/host` to the destination of `target` and link `containerPath` to it, which may or not lead to a valid file. If it's relative, then link `containerPath` to `target` even though we know that it doesn't resolve. In both these cases, we are not trying to guarantee that `containerPath` points to a valid file. We are merely doing the best we can to dig up the right destination, and giving up if we run out of guesses.

* It should not fail the entry-point process even if `target` doesn't exist or resolve. We can log a debug message or such, but we shouldn't stop the container from starting up just because there's something wrong with a configuration file. In this case, a slightly broken container that starts is vastly better than one doesn't start at all. If nothing else, it makes it easier to debug.

At the end of the day, all sane operating systems should be using relative symbolic links on the host. Fedora does that, and this is exactly the reason why relative symlinks are nice. They work even if you change the root from `/` to `/run/host`.

What we are trying to fix, are those few cases where a human created an absolute symbolic link. We should do our best to handle the obvious and simple cases, but if a human has made a mistake, then it's their responsibility to fix that up.

debarshiray (changes requested):

debarshiray:
Given that operating systems are expected to use relative symbolic links, it seems to me that updates aren't such a pressing problem.

When someone updates from Fedora 32 to 33, then their `/etc/resolv.conf` will turn into a relative symbolic link. Existing Toolbox containers had their `/etc/resolv.conf` pointing at `/run/host/etc/resolv.conf`, which will now be a relative symlink and therefore resolve.

Things can break if a user changes their `/etc/resolv.conf` into a symbolic link that either:

* Points somewhere that's not bind mounted into the container. I won't worry about this unless someone actually comes up with a valid example.

* Is absolute, and the change happened after there were already some Toolbox containers with their `/etc/resolv.conf` configured in a way that didn't expect that. In such a case, it's not that hard for the user to switch to a relative symlink. After all they are messing around with the OS' configuration, and this is a pretty minor tweak.

src/cmd/initContainer.go
```diff
@@ -407,10 +408,60 @@ func mountBind(containerPath, source, flags string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to lstat %s: %w", target, err)
+	}
+
+	if (fi.Mode() & os.ModeSymlink) != 0 {
+		logrus.Debugf("Target %s is a symlink. Resolving", target)
+		resolvedTarget, err = filepath.EvalSymlinks(target)
+		if err != nil {
+			logrus.Warnf("Failed to resolve symlink %s: %v", target, err)
+			if filepath.IsAbs(target) {
```
debarshiray, line 451: Umm... this doesn't look correct. Here need to determine if the target of `target` is absolute or not, not target itself.

src/cmd/initContainer.go
```diff
@@ -407,10 +408,60 @@ func mountBind(containerPath, source, flags string) error {
...
+
+	logrus.Debugf("Preparing to redirect %s to %s", containerPath, target)
+
+	// Skip any filepath resolution if target is not a symlink
+	fi, err := os.Lstat(target)
+	if err != nil {
+		return fmt.Errorf("failed to lstat %s: %w", target, err)
+	}
+
+	if (fi.Mode() & os.ModeSymlink) != 0 {
```
debarshiray, line 446: Looking at these deeply nested `if` branches, I am tempted to think that it would be better to split things out into a separate helper function. Something like `sanitizeTarget` or whatever.

Deeply nested structures are hard to read and follow; and we have quite a few subtle conditions based on whether we have an error or not, and from which operation.

debarshiray:
I took the liberty to address the above issues.

Thanks for your work on this, @HarryMichal !

PR #469, opened 2020-06-12
Simplify container & image fetching using labels

HarryMichal:
I did not like the way listing of containers was written. There was a lot of duplication and was very prone to error.

Now the labels are tracked in a variable, function for checking if a container/image is more lightweight (requires data to be passed to it) and functions for getting a list of toolboxes and toolbox images are much easier to understand.

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -34,6 +34,7 @@ type toolboxImage struct {
 	ID      string
 	Names   []string
 	Created string
+	Labels  map[string]string
 }
 
 type toolboxContainer struct {
@@ -42,6 +43,7 @@ type toolboxContainer struct {
 	Status  string
 	Created string
 	Image   string
+	Labels  map[string]string
 }
 
 var (
@@ -323,6 +325,7 @@ func (i *toolboxImage) UnmarshalJSON(data []byte) error {
 		ID      string
 		Names   []string
 		Created interface{}
+		Labels  map[string]string
 	}
 
 	if err := json.Unmarshal(data, &raw); err != nil {
@@ -341,6 +344,8 @@ func (i *toolboxImage) UnmarshalJSON(data []byte) error {
 		i.Created = utils.HumanDuration(int64(value))
 	}
 
+	i.Labels = raw.Labels
+
 	return nil
 }
 
@@ -352,6 +357,7 @@ func (c *toolboxContainer) UnmarshalJSON(data []byte) error {
 		State   interface{}
 		Created interface{}
 		Image   string
+		Labels  map[string]string
 	}
 
 	if err := json.Unmarshal(data, &raw); err != nil {
@@ -394,6 +400,7 @@ func (c *toolboxContainer) UnmarshalJSON(data []byte) error {
 		c.Created = utils.HumanDuration(int64(value))
 	}
 	c.Image = raw.Image
+	c.Labels = raw.Labels
 
 	return nil
 }
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -51,6 +51,12 @@ var (
 		onlyContainers bool
 		onlyImages     bool
 	}
+
+	// toolboxLabels holds labels used by containers/images that mark them as compatible with Toolbox
+	toolboxLabels = map[string]string{
+		"com.github.debarshiray.toolbox": "true",
+		"com.github.containers.toolbox":  "true",
+	}
 )
 
 var listCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


HarryMichal:
Hmm... This does not work for version of Podman before V2 where the json output of `podman images` does not hold info about the image labels. I'd like to get this in before 0.1.0, so I'll fix this ASAP.

HarryMichal:
I'm confident this works with Podman 1.9.3 and 2.0.0-rc7 (with fixes in #470 and #471).

src/pkg/podman/podman.go
```diff
@@ -27,8 +27,18 @@ import (
...
+	Value string
+}
+
 var (
-	LogLevel = logrus.ErrorLevel
+	LogLevel      = logrus.ErrorLevel
+	ToolboxLabels = []label{
+		{"com.github.debarshiray.toolbox", "true"},
+		{"com.github.containers.toolbox", "true"},
+		{"com.redhat.component", "fedora-toolbox"},
```
debarshiray, line 40: I wonder if we are at a point where we can drop the `com.redhat.component` label. We need to check if there are still any supported `fedora-toolbox` images, and the possibility of creating containers, with only this label and none of the others. Digging into the Git history of the labels can help answer this question.

HarryMichal: I guess instead of just dropping the label we could have two lists: one for creating and the other for using in containers??

HarryMichal:
> I like this idea of collecting the container and image labels in one centralized location in the code-base.
> 
> However, this commit feels like it's a bit incomplete. It introduces `ToolboxLabels`, but doesn't actually use it for anything and leaves the labels duplicated.
> 
> It would be better if we could also replace all the other direct uses of the labels with this `ToolboxLabels` in one single self-contained commit.

Yes, this makes sense.

src/pkg/podman/podman.go
HarryMichal: Based on your comment in de5e5df9b793edc746cde7125cd437b69f15ccd7 it should be ok to drop the `com.redhat.component` label.

src/cmd/list.go
```diff
@@ -121,35 +129,22 @@ func list(cmd *cobra.Command, args []string) error {
...
 
 func getContainers() ([]toolboxContainer, error) {
-	logrus.Debug("Fetching containers with label=com.github.containers.toolbox=true")
-	args := []string{"--all", "--filter", "label=com.github.containers.toolbox=true"}
-	containers_old, err := podman.GetContainers(args...)
-	if err != nil {
-		return nil, fmt.Errorf("failed to list containers with label=com.github.containers.toolbox=true: %w", err)
-	}
+	var err error
+	var args []string
```
debarshiray, line 133: Not all of these variables need to be defined separately.

src/cmd/list.go
```diff
@@ -121,35 +129,22 @@ func list(cmd *cobra.Command, args []string) error {
...
 
-	logrus.Debug("Fetching containers with label=com.github.debarshiray.toolbox=true")
-	args = []string{"--all", "--filter", "label=com.github.debarshiray.toolbox=true"}
-	containers_new, err := podman.GetContainers(args...)
+	logrus.Debug("Fetching all containers")
+	args = []string{"--all", "--sort", "names"}
+	containers, err = podman.GetContainers(args...)
 	if err != nil {
-		return nil, fmt.Errorf("failed to list containers with label=com.github.debarshiray.toolbox=true: %w", err)
+		return nil, fmt.Errorf("failed to get containers: %w", err)
```
debarshiray, line 141: Bubbling the errors upwards is fine, but we should be careful not show some weird gibberish to the user under normal operation. In such cases, it's better to either dump the error in the debug log; or detect specific errors, at least the most common ones, and have nicer equivalents for those.

debarshiray (commented):
Doubling the speed of `toolbox list` is a really nice improvement!

Does this work with older versions of Podman, because we are sorting the JSON differently?

HarryMichal:
The compatibility should remain the same as the parsing of the output was only expanded (tracking of labels) and not altered in any other away.

debarshiray:
> The compatibility should remain the same as the parsing of the
> output was only expanded (tracking of labels) and not altered in
> any other away.

I saw that we no longer differentiate between `Id`, `ID` and `id`, when handling the JSON. That's what I was referring to.

I see that we are no longer joining and sorting two separate JSON blurbs. Is that the part that makes us immune to the changes in JSON across Podman versions?

Sorry for being annoying. I am just a bit paranoid about backwards compatibility at this point.

src/cmd/list.go
HarryMichal: You've got a point here. Should I address that immediately? I think there are several places like this across the code base.

src/cmd/list.go
HarryMichal: You're right, I only did it to make it more readable. I guess that's the C programmer in me talking :).

HarryMichal:
We unmarshal raw JSON into a structure for [some time](https://github.com/containers/toolbox/pull/503/files) now. Owen helped with that a lot. That approach sees `Id`, `ID` and `id` as the same thing. This work only builds on top of that.

src/cmd/list.go
debarshiray: See https://github.com/containers/toolbox/pull/799

src/cmd/list.go
debarshiray: Never mind, I think I meant this comment for some other change. I see that we were bubbling the errors up even before, and I just did a quick test and it all looks good.

Sorry about the noise.

src/cmd/list.go
```diff
@@ -121,35 +129,22 @@ func list(cmd *cobra.Command, args []string) error {
...
-		containers = utils.SortJSON(containers, "Names", false)
-	}
-
-	// This section is a temporary solution that is here to prevent a major
-	// redesign of the way how toolbox containers are fetched.
-	// Remove this in Toolbox v0.2.0
-	var toolboxContainers []toolboxContainer
 	for _, container := range containers {
 		var c toolboxContainer
+		var isToolboxContainer bool = false
```
debarshiray, line 146: Nitpick: the ` = false` is redundant.

src/cmd/list.go
```diff
@@ -189,38 +194,22 @@ func listHelp(cmd *cobra.Command, args []string) {
...
+		return nil, fmt.Errorf("failed to get images: %w", err)
 	}
 
-	// This section is a temporary solution that is here to prevent a major
-	// redesign of the way how toolbox images are fetched.
-	// Remove this in Toolbox v0.2.0
-	var toolboxImages []toolboxImage
 	for _, image := range images {
 		var i toolboxImage
+		var isToolboxImage bool = false
```
debarshiray, line 211: Ditto.

PR #470, opened 2020-06-12
Make version checking work for development version

HarryMichal:
I made a mistake when creating the function for checking Podman's version. It did not account for development releases. This should address the issue.


HarryMichal:
Scenario that does not work:

- Installed version of Podman is `2.0.0-rc7`
- `podman.CheckVersion` is called with argument `2.0.0`
- the function should return `true` but returns `false`

Basically the version checking function should work even for in-development releases.

PR #471, opened 2020-06-12
Podman V2 compatibility fixes

HarryMichal:
Podman is nearing a new major version (V2). We need to preserve compatibility.

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -161,7 +161,10 @@ func listImages() ([]map[string]interface{}, error) {
 	}
 
 	var images []map[string]interface{}
-	if podman.CheckVersion("1.8.3") {
+	if podman.CheckVersion("2.0.0") {
+		images = utils.JoinJSON("Id", images_old, images_new)
+		images = utils.SortJSON(images, "Names", true)
+	} else if podman.CheckVersion("1.8.3") {
 		images = utils.JoinJSON("ID", images_old, images_new)
 		images = utils.SortJSON(images, "Names", true)
 	} else {
@@ -178,7 +181,11 @@ func listOutput(images, containers []map[string]interface{}) {
 		fmt.Fprintf(writer, "%s\t%s\t%s\n", "IMAGE ID", "IMAGE NAME", "CREATED")
 
 		var idKey, nameKey, createdKey string
-		if podman.CheckVersion("1.8.3") {
+		if podman.CheckVersion("2.0.0") {
+			idKey = "Id"
+			nameKey = "Names"
+			createdKey = "Created"
+		} else if podman.CheckVersion("1.8.3") {
 			idKey = "ID"
 			nameKey = "Names"
 			createdKey = "Created"
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -119,8 +119,15 @@ func listContainers() ([]map[string]interface{}, error) {
 		return nil, errors.New("failed to list containers with label=com.github.debarshiray.toolbox=true")
 	}
 
-	containers := utils.JoinJSON("ID", containers_old, containers_new)
-	containers = utils.SortJSON(containers, "Names", false)
+	var containers []map[string]interface{}
+	if podman.CheckVersion("2.0.0") {
+		containers = utils.JoinJSON("Id", containers_old, containers_new)
+		containers = utils.SortJSON(containers, "Names", true)
+	} else {
+		containers = utils.JoinJSON("ID", containers_old, containers_new)
+		containers = utils.SortJSON(containers, "Names", false)
+	}
+
 	return containers, nil
 }
 
@@ -194,6 +201,7 @@ func listOutput(images, containers []map[string]interface{}) {
 			nameKey = "names"
 			createdKey = "created"
 		}
+
 		for _, image := range images {
 			id := utils.ShortID(image[idKey].(string))
 			name := image[nameKey].([]interface{})[0].(string)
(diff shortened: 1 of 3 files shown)


src/cmd/create.go
```diff
@@ -338,7 +338,6 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 		"create",
 		"--dns", "none",
 		"--env", toolboxPathEnvArg,
-		"--group-add", sudoGroup,
```
debarshiray, line 341: Do we really not need the group for `sudo`?

src/cmd/create.go
```diff
@@ -347,14 +346,13 @@ func createContainer(container, image, release string, showCommandToEnter bool)
...
 		"--pid", "host",
-		"--privileged",
 		"--security-opt", "label=disable",
 	}
 
 	createArgs = append(createArgs, ulimitHost...)
 
 	createArgs = append(createArgs, []string{
-		"--userns=keep-id",
+		// "--userns=keep-id",
```
debarshiray, line 355: This cannot be right. :)

Wasn't there a bug in Podman that broke `--userns=keep-id` for rootless containers that later got fixed?

src/cmd/list.go
```diff
@@ -211,12 +216,17 @@ func listOutput(images, containers []map[string]interface{}) {
...
+		var id, name, created, status, imageName string
 		for _, container := range containers {
-			id := container["ID"].(string)[:12]
-			name := container["Names"].(string)
-			created := container["Created"].(string)
-			status := container["Status"].(string)
-			imageName := container["Image"].(string)
+			id = container["Id"].(string)[:12]
+			if podman.CheckVersion("2.0.0") {
+				name = container["Names"].([]interface{})[0].(string)
```
debarshiray, line 223: Ugh. Another JSON change. :/

I wonder if this is needed in other places.

src/cmd/create.go
HarryMichal: Yes, exactly. I was playing with this and pushed it with it being commented. It should be uncommented, of course.

src/cmd/create.go
HarryMichal: Well, in the case of Fedora-based images it works but other distros may use `wheel` as the sudoers group and this can't be tested on the host but only inside of the container.

src/cmd/list.go
HarryMichal: I have some new code in https://github.com/containers/toolbox/pull/469 that makes this a bit less painful. But yeah, we need to be careful about this.

HarryMichal:
This relies on #469 so that the number of version checks against Podman is kept to minimum.

src/cmd/create.go
debarshiray: For those following along at home, this was because of a [regression](https://github.com/containers/libpod/pull/6715) in Podman where the `--group-add` and `--privileged` flags won't work together.

src/cmd/create.go
debarshiray: For those following along at home, I was referring to https://github.com/containers/libpod/issues/6652

src/cmd/list.go
```diff
@@ -211,12 +216,29 @@ func listOutput(images, containers []map[string]interface{}) {
...
+		}
+
+		var id, name, created, status, imageName interface{}
 		for _, container := range containers {
-			id := container["ID"].(string)[:12]
-			name := container["Names"].(string)
-			created := container["Created"].(string)
-			status := container["Status"].(string)
-			imageName := container["Image"].(string)
+			id = utils.ShortID(container[idKey].(string))
```
debarshiray, line 233: Nitpick: this bit should be split out into a separate commit since it's unrelated to Podman 2.0.

See: https://github.com/containers/toolbox/pull/479

HarryMichal: Ah, right. Thanks for the catch.

src/cmd/list.go
```diff
@@ -187,10 +187,15 @@ func listOutput(images, containers []map[string]interface{}) {
...
+
+		if podman.CheckVersion("2.0.0") {
+			idKey = "Id"
+		}
+
 		for _, image := range images {
 			id := utils.ShortID(image[idKey].(string))
-			name := image[nameKey].([]interface{})[0].(string)
-			created := image[createdKey].(string)
+			name := image[nameKey].([]interface{})[0]
```
debarshiray, line 197: After pondering over this a bit, I don't think we should drop the type cast just like that.

I understand that the strict cast makes the code more sensitive to (future) changes in the JSON, and currently it doesn't matter because everything is a `string` anyway. However, we should also be in control of what we are printing, and we shouldn't just throw some unknown type at the `%s` format specifier in the hope that it will somehow work out.

Go does offer some nice features like [type assertions](https://tour.golang.org/methods/15) and [type switches](https://tour.golang.org/methods/16). These could be used to come up with a more elegant solution that's more tolerant towards JSON changes, and at the same time lets us retain control over our types.

At the very least, if something is not of the expected type, we should log some debug message with details about the actual type. Parsing the JSON and dealing with all the `interface` types can be tricky at times, so logging the actual failure can be useful.

As for the actual output, we could either print an empty string for all values that aren't of the expected type, or we can skip the entire column altogether. Or maybe something else. It somewhat depends on how practical each of these options are.

Either way, we should use a separate pull request for this topic because it's not directly connected to the issue of *JSON changes in Podman 2.0*. :)

HarryMichal: Sure! This just looked to me like the easiest solution if something were to change and we weren't fast enough to change it. Using type assertions/switches definitely gives us more power over dealing with what to show/hide in case of an error.

I'll look into that in the future.

PR #472, opened 2020-06-13
utils: Use Lstat over Stat in PathExists

HarryMichal:
`utils.PathExists` uses `os.Stat` under the hood but `stat` be definition follows symlinks. I think better is to use `os.Lstat` because `lstat` returns info about symlinks themselves.

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -636,9 +636,9 @@ func parseReleaseRHEL(str string) (string, error) {
 	return str, nil
 }
 
-// PathExists wraps around os.Stat providing a nice interface for checking an existence of a path.
+// PathExists wraps around os.Lstat providing a nice interface for checking an existence of a path.
 func PathExists(path string) bool {
-	if _, err := os.Stat(path); !os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


src/pkg/utils/utils.go
```diff
@@ -430,9 +430,9 @@ func ParseRelease(str string) (string, error) {
 	return release, nil
 }
 
-// PathExists wraps around os.Stat providing a nice interface for checking an existence of a path.
+// PathExists wraps around os.Lstat providing a nice interface for checking an existence of a path.
 func PathExists(path string) bool {
-	if _, err := os.Stat(path); !os.IsNotExist(err) {
+	if _, err := os.Lstat(path); !os.IsNotExist(err) {
```
debarshiray, line 435: This is why symbolic links are messy. :)

Did you check all our uses of `PathExists` to be sure that it won't introduce some subtle bug?

HarryMichal: I perused the code and I think the change should be alright. In some cases, the function can be used to check a possible symlink but I think this modification will not affect even those cases.

PR #478, opened 2020-06-23
Rpm Go build fixes

HarryMichal:
This is a small thing we need to fix for the Toolbox rpm.

### DIFF
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -27,5 +27,5 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -o "$2"
+go build -ldflags "-B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n')" -o "$2/toolbox"
 exit "$?"


src/go-build-wrapper
```diff
@@ -27,5 +27,5 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -o "$2"
+go build -ldflags "-B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n')" -o "$2/toolbox"
```
debarshiray, line 30: Nitpick: is the `$2/toolbox` really necessary? Isn't `$2` enough?

HarryMichal: Yes, it is necessary. The installation `meson install` would fail to find the binary because it was created as `src/src`.

HarryMichal:
> Yes, I saw this on https://fedoraproject.org/wiki/PackagingDrafts/Go
> 
> While this might be the bare minimum that's necessary to get things to build on Fedora, ideally we should let Fedora dictate all the build flags. In this case that would be the `%{gobuild}` RPM macro. Use `rpm --eval "%{gobuild}"` to look at the expansion of the macro.
> 
> Other distributions would want to use other flags.
> 
> I wonder what could be the best way to achieve this. For the immediate short-term we can somehow hack this up via Fedora-specific patches in the RPM, but going forward we'd need something more civilized. :)

Yes, much more ideal would be if we could simply call whatever is in the `%{gobuild}` macro. I couldn't figure out a way to do it though so instead, I adjusted the wrapper. I guess it would be better if we maintained the change as a patch in dist-git.

HarryMichal:
I'm closing this in favor of making this a downstream patch.

src/go-build-wrapper
debarshiray: Interesting. I see what you are saying. It does fail when built in Koji, but if I build it locally using `rpmbuild -ba toolbox.spec` then it works because it creates both `src/src` and `src/toolbox`.

What kind of dark voodoo is this?! :)

PR #480, opened 2020-06-26
[cmd/root] Use string verb format for string path instead of slice

masch:
Minor fix on root command, it was using slice verb format(`%p`) on an string instead of the corresponding verb (`%s`).


debarshiray:
Thanks!

I took the liberty to fix the attribution to `Mario Sebastian Chacon <the.masch@gmail.com>`, instead of `masch <the.masch@gmail.com>`

PR #481, opened 2020-06-26
pkg/podman: Remember Podman's version during runtime

HarryMichal:
We call `podman version` every time we want to check Podman's version. This slows Toolbox down considerably. It's more efficient if we call `podman version` just once and remember the output during runtime and use the cached value.

On Intel i5-6300U calling `time podman version --format json` takes roughly a bit over 100ms.

### DIFF
--- a/src/pkg/podman/podman.go
+++ b/src/pkg/podman/podman.go
@@ -35,14 +35,14 @@ var (
 //
 // Takes in one string parameter that should be in the format that is used for versioning (eg. 1.0.0, 2.5.1-dev).
 //
-// Returns true if the Podman version is equal to or higher than the required version.
+// Returns true if the current version is equal to or higher than the required version.
 func CheckVersion(requiredVersion string) bool {
-	podmanVersion, _ := GetVersion()
+	currentVersion, _ := GetVersion()
 
-	podmanVersion = version.Normalize(podmanVersion)
+	currentVersion = version.Normalize(currentVersion)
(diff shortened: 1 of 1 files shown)


src/pkg/podman/podman.go
```diff
@@ -29,6 +29,7 @@ import (
 
 var (
 	LogLevel = logrus.ErrorLevel
+	Version  = ""
```
debarshiray, line 32: We shouldn't export this variable from the `podman` package. This pull request itself is a good example of why we shouldn't do this. :)

If clients of this package use the `GetVersion` function to get the version as a string, we have a lot more flexibility and elbow room to change the underlying implementation in future. eg., this pull request changes the implementation from always querying the version from Podman, to caching it.

Another example is https://github.com/containers/toolbox/pull/487 that changed the implementation of the `GetVersion` function in the `version` package without affecting any of the client code.

However, I see that we have a tricky naming problem - `version` is already taken by `github.com/HarryMichal/go-version`, and there's a `podmanVersion` that's local to the `CheckVersion` function. Maybe we can rename the variable in `CheckVersion` to `currentVersion` or something, it seems to match its `requiredVersion` counterpart, to free up a name?

HarryMichal: Nice catch! I did not realize the variable should not be exported.

Concerning the variable name, make whatever edits you think are right 👍.

PR #482, opened 2020-06-26
Add errors to pkg/podman & wrap errors across codebase

HarryMichal:
In a lot of places in the code, we use Go errors in this fashion: `fmt.Errorf("container %s...", container)`. This is alright if we use the error messages immediately and don't care if a function may return different errors. At the moment we need to check for different errors coming from functions, this style becomes bothersome.

This is the most noticeable in the case of removing containers/images. Commands `podman rm/rmi` may fail with different errors. Either a container/image does not exist or it is running/used. Without dedicated error variables, this is rather hard to catch. We've been going about it by using the approach briefly shown above.

The solution is to add error variables to the package (`podman`) that can be checked for using the `errors` (`errors.Is` and `errors.As`) library of Go.

The second part of this PR is that in a lot of places in the code, we return errors but don't wrap them around the errors that hold the information about the cause. I encountered this recently when Podman V2 introduced a regression about the option `--userns keep-id` not working properly. It caused Toolbox to crash (specifically the container entry-point `init-container`) due to being unable to find info about current user but there was no helpful error message because we dropped it when returning Toolbox's error message.

A good side-effect of this is that commands `toolbox rm/rmi` now detect if a container is running/image has dependent containers and print a helpful message with a hint on how to delete them.

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -137,7 +137,7 @@ func initContainer(cmd *cobra.Command, args []string) error {
 
 	toolboxEnvFile, err := os.Create("/run/.toolboxenv")
 	if err != nil {
-		return errors.New("failed to create /run/.toolboxenv")
+		return fmt.Errorf("failed to create /run/.toolboxenv: %w", err)
 	}
 
 	defer toolboxEnvFile.Close()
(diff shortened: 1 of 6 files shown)


HarryMichal:
I created a separate PR (#519) for moving the remove functions. I'll mark this PR as a draft until #519 gets merged.

Issue #188, opened 2019-06-07
NAME and VERSION leak into toolbox container env


juhp:
I noticed that envvar for `ENV NAME=fedora-toolbox VERSION=30` leak into the toolbox environment.
I am not sure if this is desirable: eg NAME affects changelog entry generation in Emacs rpm-spec-mode. And probably having VERSION in the container is not that useful either.

debarshiray:
Yeah, those variables are exported by Podman. It always felt odd that they aren't namespaced. I wonder if Docker does the same, and so it's done for compatibility.

debarshiray:
I wish I could transfer this to `libpod`. Doesn't look like I have the necessary permissions for that. :(

@juhp could you please (re)file this against `libpod` ? We could do something to suppress these variables, but they come directly from Podman, so might be better to at least discuss it a bit first. I have no idea why such generically named and non-namespaced variables were chosen in the first place.

debarshiray:
Oh wait, they actually come from the Dockerfile for the `fedora-toolbox` images.

debarshiray:
They are mandated by the [Fedora guidelines](https://docs.fedoraproject.org/en-US/containers/guidelines/creation/).

debarshiray:
On closer reading, it doesn't look like the guidelines actually require them, but they are part of the example Dockerfile snippets.

cverna:
So from a quick look at OSBS it seems that these labels are used to populate some data in koji post build. I don't think we can remove them.

I ll look into more details next week if there is a work around.

debarshiray:
Thanks for clarifying that, @cverna !

debarshiray:
@cverna confirmed in `#fedora-containers` on Freenode that `NAME` and `VERSION` really do need to be present in the Dockerfile.

HarryMichal:
@debarshiray, I guess we could unset the env vars when the container is being initialized. That way we keep them in the Dockerfiles but they will not cause problems during runtime.

Jmennius:
@debarshiray Can you clarify why exactly are they required?
Should we have them in non-Fedora images?

debarshiray:
> I guess we could unset the env vars when the container is
> being initialized. That way we keep them in the Dockerfiles
> but they will not cause problems during runtime.

The `NAME` and `VERSION` environment variables make their way into the toolbox containers from their corresponding images, and are part of the containers' metadata. See:
```
$ podman inspect --type container <container>
```

[...]

Issue #176, opened 2019-05-27
systemctl does not work inside the toolbox


mildred:
I'm setting up a toolbox for a development environment for a web application that requires postgresql running. I'm installing postgresql-server on the toolbox and then trying to start it using the usual method:

```
INS mildred@toolbox$ sudo systemctl start postgresql.service
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

This is understandable as I'm running within a container, and systemd usually is not started in containers. Still, it means I have to manually start up postgresql by reading the systemd service file.

It would be great if systemd was started as entrypoint in the toolbox image, instead of `sleep +Inf`. Of course, udev and low level services would have to be prevented from starting so that a blank toolbox would not start any service.

mildred:
Trying out running systemd with the toolbox, I find a problem with it. systemd is not running as PID 1 and tries to start in user mode and not in system mode.

This is because there is the `--pid=host` option passed to `podman create`.

May I suggest that `podman create` keeps the PID namespace (so systemd can run) but `podman exec` gets a new option to keep the PID namespace of the host. it is possible with nsenter, why not podman exec ?

found a nice post here: https://developers.redhat.com/blog/2019/04/24/how-to-run-systemd-in-a-container/

mildred:
Trying to use nsenter to have systemd running in a PID namespace and toolbox shells running in the host PID namespace does not work so well. the `/proc` filesystem mounted within the container represents the container PID namespace and not the host. As such, commands are complaining with errors like:

    ls: cannot read symbolic link '/proc/self': No such file or directory

Perhaps we can have different features enabled for the toolboxes, and we can choose for each toolbox if we want to

- run an init system but have a separate PID namespace
- do not run an init system and keep the same PID namespace

It corresponds to two different use cases.

mildred:
There are a few other issues running systemd in the container, systemd is mounting tmpfs to `/run`, overriding bind mounts that were there.

mildred:
It also means we cannot share the system dbus socket inside the toolbox because the system bus needs to points to the toolbox system instead. I don't know what are the consequence of that.

debarshiray:
> There are a few other issues running systemd in the container, systemd is mounting
> tmpfs to `/run`, overriding bind mounts that were there.

I thought `podman create --systemd ...` is meant to take care of that. Maybe it doesn't work rootless because `/run` inside the container appears to be `fuse-overlayfs`?

[...]

PR #483, opened 2020-06-27
Introduce Ubuntu LTS and latest non-LTS images

Jmennius:
This adds Ubuntu 16.04, 18.04, 20.04, 21.10 and 22.04 images that play well with `toolbox`:
password-less `sudo`, able to resolve its own hostname (so that `sudo` does not complain), SELinux is masked off, etc.
Since 18.04 Ubuntu images on the Docker Hub are 'minimized', so we have to undo that.
Pushed images to https://hub.docker.com/r/jmennius/ubuntu-toolbox for the time being.

Users may want to generate their locale after toolbox creation with `sudo sed -i "/${LANG}/s/^# //g" /etc/locale.gen && sudo locale-gen` (assuming that you have `LANG` set) to avoid `perl` and others complain about it.

I've been using an 18.04 image for ~several months~ 1.5yrs now with great success.

Some things I'd like to resolve or discuss at least (need help/guidance/decisionmaking/PRs):
- [ ] Use actual quay toolbx robot credentials.
- [x] READMEs are not symlinked. Didn't get this to work as symlink target is out of build context. #723
- [x] Moreover, I'd also like to share `extra-packages` between images. Maybe we can store all Containerfiles in the `ubuntu` and a single readme and extra-packages alongside?
- [x] align READMEs with the rest of the repo before merging
- [x] Needs #401 to function
- [x] I can mention myself as the image maintainer if that's the way to go.
- [x] Any CI/CD integration? Yes, see #1019 
- [x] ~I've only tested this with a shell-based toolbox.~ Works perfectly with golang implementation.
- [x] Did not use `NAME` and `VERSION` variables (discussed in #188)
 
I'd be grateful for any testing and feedback on their use-case and environment.

Alternative-to: #298

### DIFF
--- a/.github/workflows/ubuntu-images.yaml
+++ b/.github/workflows/ubuntu-images.yaml
@@ -1,0 +1,61 @@
+name: "Images: Build and push Ubuntu toolbx images"
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - images/ubuntu/**
+      - .github/workflows/ubuntu-images.yaml
+  schedule:
+    - cron: '0 0 * * MON'
+
+# Prevent multiple workflow runs from racing
+concurrency: ${{ github.workflow }}
+
+env:
+  latest_release: '22.04'
+
+jobs:
+  build-and-push-images:
+    strategy:
+      matrix:
+        release: ['16.04', '18.04', '20.04', '22.04', '22.10', '23.04']
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: 'toolbx+github'
+          password: ${{ secrets.QUAY_ROBOT_TOKEN }}
+
+      - name: Build and push Ubuntu ${{ matrix.release }} toolbox image
+        uses: docker/build-push-action@v3
+        with:
+          context: images/ubuntu/${{ matrix.release }}
+          file: images/ubuntu/${{ matrix.release }}/Containerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          no-cache: true
+          tags: quay.io/toolbx/ubuntu-toolbox:${{ matrix.release }}
+
+      - name: Push latest tag
+        if: env.latest_release == matrix.release
+        uses: docker/build-push-action@v3
+        with:
+          context: images/ubuntu/${{ matrix.release }}
+          file: images/ubuntu/${{ matrix.release }}/Containerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: quay.io/toolbx/ubuntu-toolbox:latest


Jmennius:
I'd really appreciate some feedback ;) @debarshiray ?

HarryMichal:
> I'd really appreciate some feedback ;) @debarshiray ?

I'd love to give this a spin but I'm currently working on something else.. :(

Jmennius:
@HarryMichal no problem :)

What do you think about sharing extra-packages and READMEs? Any nice way to do it?
Maybe we can drop READMEs from images themselves?

HarryMichal:
In my opinion, sharing of those files mainly depends on the way the images will be built. `fedora-toolbox` images are built in Fedora's build system and their definitions are maintained there (meaning the definitions in this repository could be called a mirror - or is it the other way @debarshiray?). If the container definitions are tracked in a got repository where every version is in a separate branch than it's not really possible to share. If they are in a single branch but separated in their own directories, than a single file shared by all definitions could be placed in the root directory. But that approach may be dangerous a bit if the package is renamed in some of those version (though in some cases very unlikely).

I myself never worked with Docker Hub and other tools so I can't really help much with this.

Dropping READMEs seems okay-ish to me. But on the other hand having some kind of file that holds some info about the image/tool is nice. Maybe some specialized file that will be defined once, will be shorter than the project's README that will not require to be updated with every tweak to the project's README?

mfocko:
used the Dockerfile for 20.10, seems to be running OK :+1:

mtalexan:
Probably out-of-scope for the intended changes here, but support for `ubuntu` as a distribution under the `(-d|--distribution)` and the related releases under the `(-r|--release)` options would require updating the src/pkg/utils/utils.go `supportedDistros` map to add a `Distro` structure  (and the related functions to fill it) for an `ubuntu` key.  Otherwise the results of these builds are limited to use of the `--image` option.

Jmennius:
cc @debarshiray

Jmennius:
I now get `WARNING: libcap needs an update (cap=40 should have a name).` when entering a 20.04 toolbox.
This is likely because I am using Linux 5.10 (Fedora 33) but Ubuntu 20.04 userspace ships with `libcap2` 2.32 which is not aware of `CAP_CHECKPOINT_RESTORE` (40).
Everyhting works fine though.

runfalk:
I too would really like to see Ubuntu support being included one way or another. My use case is doing firmware development on Fedora when the rest of the team uses Ubuntu. They have set up install scripts and debs for installing the all the required cross compilation targets.

I tried the image in this project using:

```bash
$ toolbox create --image registry.hub.docker.com/jmennius/ubuntu-toolbox:20.04 firmaredev
```

The `/README.md` talks about using systemctl for managing services, but it doesn't seem like systemd is actually running in the container:

```
$ systemctl
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

This seems to be documented in another issue as well: https://github.com/containers/toolbox/issues/176.

Other than some dbus errors (probably because systemd isn't running) and the libpcap error I don't see any other problems with these containers. I tried both 20.04 and 18.04.

I shouldn't hijack this MR, but is there a way of running services like SSH in the container?

Jmennius:
Guys, any updates or feedback?

Jmennius:
> Probably out-of-scope for the intended changes here, but support for `ubuntu` as a distribution under the `(-d|--distribution)` and the related releases under the `(-r|--release)` options would require updating the src/pkg/utils/utils.go `supportedDistros` map to add a `Distro` structure (and the related functions to fill it) for an `ubuntu` key. Otherwise the results of these builds are limited to use of the `--image` option.

I've implemented this when I ported it to Go (back in April).

Jmennius:
> The `/README.md` talks about using systemctl for managing services, but it doesn't seem like systemd is actually running in the container:
> 
> ```
> $ systemctl
> System has not been booted with systemd as init system (PID 1). Can't operate.
> Failed to connect to bus: Host is down
> ```
> 
> This seems to be documented in another issue as well: #176.
> 
> Other than some dbus errors (probably because systemd isn't running) and the libpcap error I don't see any other problems with these containers. I tried both 20.04 and 18.04.
> 
> I shouldn't hijack this MR, but is there a way of running services like SSH in the container?

Well, I wouldn't expect any of this to really work, the README should probably be corrected.
I don't think that running sshd is feasible, but you can probably google that with regards to containers in general.

Jmennius:
> mounting a tmpfs on /tmp made it work !
> 
> mount -o mode=1777,nosuid,nodev -t tmpfs tmpfs /tmp

Interesting.. so you didn't have a tmpfs on /tmp on your host (Debian)?
If so - that probably deserves a separate issue to be fixed properly.

dabrain34:
For now i havent investigated into toolbox code to see how I could propose a patch but the other problem I faced was that on the tip, the binary was not exectutable.

Jmennius:
Few updates:
* decided to also provide the image for the latest non-LTS release - 21.10 image
* I've noticed a delay when using `sudo` in new images (and that has probably been there forever... no idea how I lived with it for over a year!?) - fixed by moving `myhostname` `nsswitch` entry to the proper place manually instead of trusting package postinst with it.
* few updates to align with fedora image changes
* rebuilt and pushed all images

yuval-k:
i'm using the 18.04 pushed image; so far so good... thank you for adding this

djrenren:
I'm using 20.04 and getting issues with locales: 

image

Doesn't seem to break anything but `perl` isn't too happy

Jmennius:
> I'm using 20.04 and getting issues with locales:
> 
> image
> 
> Doesn't seem to break anything but `perl` isn't too happy

This is kind of a known issue with [slim] container images that do not have any locale generated/installed.
I believe this was discussed somewhere, but I couldn't find it here or in my notes (in `toolbox` context)...
Anyway, there is nothing we can do at container image build time because user may need/want a different locale.
IMO, a proper solution would be to implement a `toolbox create` time feature/hook that would let us automatically generate your current locale.

For now, just run `sudo sed -i "/${LANG}/s/^# //g" /etc/locale.gen && sudo locale-gen` (given that you have `LANG` set).

QuarkTheAwesome:
These images are great! I do have a small issue with running GUI/X11 apps inside them. Sometimes (seemingly related to the creation of a new window or menu panel) input on my host system freezes up for a moment.

Repro:
- Create a new Ubuntu toolbox (e.g. jmennius/ubuntu-toolbox:20.04)
- Install a Qt app (mgba-qt is in the default repos, but I see it with other apps too. xterm doesn't seem to have this issue)
- Open the app (/usr/games/mgba-qt)
- Open the File menu and then scroll the mouse left and right to open all the different menus. Sometimes the system will freeze up for about a second at a time.

I suspect the toolbox container is involved here since I 1. never saw this on a Fedora container and 2. can fix it by adding `toolbox` to `/etc/hosts` as a localhost alias. Since hosts is shared between the host and the container I am unsure exactly who's at fault though.

Seen on Fedora KDE 35, using Wayland+XWayland

Jmennius:
> These images are great! I do have a small issue with running GUI/X11 apps inside them. Sometimes (seemingly related to the creation of a new window or menu panel) input on my host system freezes up for a moment.
> 
> Repro:
>     * Create a new Ubuntu toolbox (e.g. jmennius/ubuntu-toolbox:20.04)
>     * Install a Qt app (mgba-qt is in the default repos, but I see it with other apps too. xterm doesn't seem to have this issue)
>     * Open the app (/usr/games/mgba-qt)
>     * Open the File menu and then scroll the mouse left and right to open all the different menus. Sometimes the system will freeze up for about a second at a time.
> 
> 
> I suspect the toolbox container is involved here since I 1. never saw this on a Fedora container and 2. can fix it by adding `toolbox` to `/etc/hosts` as a localhost alias. Since hosts is shared between the host and the container I am unsure exactly who's at fault though.
> 
> Seen on Fedora KDE 35, using Wayland+XWayland

I've tried reproducing, but it seems fine (scrolling through the menus)...
I agree with you though that it does look like an issue (actually very similar to what we've had before - see this https://github.com/containers/toolbox/pull/483#issuecomment-968385425 it was easy to reproduce the delay with just `sudo`).
Do you have `myhostname` on `hosts:` line in `/etc/nsswitch.conf` in your image?

QuarkTheAwesome:
> Do you have `myhostname` on `hosts:` line in `/etc/nsswitch.conf` in your image?

Yup:
```
hosts:          files myhostname mdns4_minimal [NOTFOUND=return] dns
```

Since the mouse cursor locks up too, I wonder if the Wayland compositor on the host is trying to resolve the name `toolbox`?

Jmennius:
> Since the mouse cursor locks up too, I wonder if the Wayland compositor on the host is trying to resolve the name `toolbox`?

I guess we can try making separate hosts files for host and container and experiment with that to understand where does it come from.

HarryMichal:
This seems to be similar to #1059. @QuarkTheAwesome, could you, please, try to reproduce your bug with VSCodium? If it is the same bug, we could move the discussion there.

But that is just a hunch and you've already written you don't experience this particular problem in a Fedora container.

Jmennius:
Few updates:
* rebased, squashed
* refreshed all images (on Docker Hub)
* added `22.04` image
* moved all common files into common directory (still would be nice to have README.md symlinked)
* added `BUILD.md`

Jmennius:
Added some unit tests as well.

Jmennius:
Added GitHub actions workflow for images CD as discussed in #1019.
Robot account credentials pending.

tjanez:
I've just tested this with Fedora 35's Toolbox and Ubuntu 20.04 images.

Everything appears to work, I just saw a warning related to libcap needing an update:

```
WARNING: libcap needs an update (cap=40 should have a name).
```

Full reproducer:
```
[tadej@toronto temp]$ toolbox create -i docker.io/jmennius/ubuntu-toolbox:20.04
Created container: ubuntu-toolbox-20.04
Enter with: toolbox enter ubuntu-toolbox-20.04
[tadej@toronto temp]$ toolbox enter ubuntu-toolbox-20.04
WARNING: libcap needs an update (cap=40 should have a name).
[tadej@toolbox temp]$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.4 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.4 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

Jmennius:
> Everything appears to work, I just saw a warning related to libcap needing an update:
> 
> ```
> WARNING: libcap needs an update (cap=40 should have a name).
> ```

That's expected and fine, see [my earlier comment](https://github.com/containers/toolbox/pull/483#issuecomment-803642248).

tjanez:
@debarshiray, I think this PR has matured quite well. What are the next steps for getting this merged and released?

kriptolix:
I've tested this with Fedora SB 36 Toolbox and Ubuntu 18.04 image and get "Error: invalid entry point PID of container ubuntu-toolbox". 20.04 image works fine.

heyakyra:
> I've tested this with Fedora SB 36 Toolbox and Ubuntu 18.04 image and get "Error: invalid entry point PID of container ubuntu-toolbox".

No problems for me, also on fedora silverblue 36. What exact commands did you use? Anything helpful in verbose mode?

heyakyra:
Inside the U18.04 toolbox on a F36 SB host, I'm unable to run `pbuilder create` due to an error `mknod: /var/cache/pbuilder/build//171608/test-dev-null: Operation not permitted` which causes `Cannot install to target '/var/cache/pbuilder/build/39472' mounted with noexec or nodev`:

```
$ sudo pbuilder create --debug --configfile /etc/custom/pbuilder/pbuilderrc
…
I: Applying user distribution dist build script /etc/custom/pbuilder/hooks/mybuilder
+ which /var/cache/pbuilder/debootstrap/usr/sbin/debootstrap
/var/cache/pbuilder/debootstrap/usr/sbin/debootstrap
+ :
+ /var/cache/pbuilder/debootstrap/usr/sbin/debootstrap --include=apt --variant=buildd --force-check-gpg --keyring=/etc/custom/pbuilder/trusted.gpg mybuild /var/cache/pbuilder/build//171608 http://apt.mycustombuild.com/main /etc/custom/pbuilder/hooks/mybuild
mknod: /var/cache/pbuilder/build//171608/test-dev-null: Operation not permitted
E: Cannot install into target '/var/cache/pbuilder/build//171608' mounted with noexec or nodev
+ log.e '/var/cache/pbuilder/debootstrap/usr/sbin/debootstrap failed'
+ case "${LOGLEVEL-I}" in
+ log 'E: /var/cache/pbuilder/debootstrap/usr/sbin/debootstrap failed'
+ set +x
E: /var/cache/pbuilder/debootstrap/usr/sbin/debootstrap failed
+ log.e 'Tail of debootstrap.log:'
+ case "${LOGLEVEL-I}" in
+ log 'E: Tail of debootstrap.log:'
+ set +x
E: Tail of debootstrap.log:
+ tail -n 50 /var/cache/pbuilder/build//171608/debootstrap/debootstrap.log
tail: cannot open '/var/cache/pbuilder/build//171608/debootstrap/debootstrap.log' for reading: No such file or directory
```

I don't see any SELinux issues reported by the alert browser (setroubleshootd/sealert). Any ideas? 

> I've tested this with Fedora SB 36 Toolbox and Ubuntu 18.04 image and get "Error: invalid entry point PID of container ubuntu-toolbox".

No problems entering the toolbox for me, also on fedora silverblue 36. What exact commands did you use? Anything helpful in verbose mode?

travier:

debarshiray:
I finally have some time to focus on pushing Toolbx forward again, as opposed to keeping the lights on.  So, I want to get Ubuntu supported out-of-the-box as soon as possible.

I have rebased this branch against current `main` and synchronized the image definitions with https://github.com/toolbx-images/images/tree/main/ubuntu  So far, I have found an issue with the `ubuntu-toolbox:18.04` image due to `capsh(1)` being missing (https://github.com/toolbx-images/images/issues/73).

debarshiray (commented):
/cc @Jmennius and @travier 

I would like to play with the images a bit more.  Until then I have some random questions.  None of them are blockers or anything.  Just questions that occurred to me.

Do you prefer to delete the image definitions for Ubuntu releases that have reached End of Life?  eg., https://github.com/toolbx-images/images/commit/c2e2f0d3a526d55ee0ec9b54af338a493a29a2c2  We have kept all the image definitions for Fedora, just so that someone can locally build an image for an unsupported release, in case they need it, but for the Ubuntu images I want to leave it up to you.

Do you know of a way in which we can run our test suite on Ubuntu hosts?  Like we do currently on Fedora and CentOS Stream 9 hosts.  We would need [at least](https://github.com/containers/toolbox/issues/1247) Ubuntu 22.10, if we want to avoid building our own [Shadow](https://github.com/shadow-maint/shadow/) that's new enough to offer `libsubid.so`.  Note that we would need Ubuntu hosts, not containers.

travier:
> Do you prefer to delete the image definitions for Ubuntu releases that have reached End of Life? eg., [toolbx-images/images@c2e2f0d](https://github.com/toolbx-images/images/commit/c2e2f0d3a526d55ee0ec9b54af338a493a29a2c2) We have kept all the image definitions for Fedora, just so that someone can locally build an image for an unsupported release, in case they need it, but for the Ubuntu images I want to leave it up to you.

I think we'll keep all images available in https://github.com/toolbx-images/images until they no longer build.

> Do you know of a way in which we can run our test suite on Ubuntu hosts? Like we do currently on Fedora and CentOS Stream 9 hosts. We would need [at least](https://github.com/containers/toolbox/issues/1247) Ubuntu 22.10, if we want to avoid building our own [Shadow](https://github.com/shadow-maint/shadow/) that's new enough to offer `libsubid.so`. Note that we would need Ubuntu hosts, not containers.

No idea sorry.

Jmennius:
@debarshiray Please let me double check image definitions (will do on the weekend).
I just want to make sure nothing else was mangled along the way of all the imports and rebases (https://github.com/toolbx-images/images/issues/72#issuecomment-1474184253)...

Jmennius:
@debarshiray 
* need to pull changes to fix 18.04 image (toolbx-images/images#75)
* We can drop 21.10 as it is not supported anymore (it also has `ENV NAME=ubuntu-toolbox VERSION=21.10` which we don't want)
* I'd like to move `flatpak-xdg-utils` from extra-packages to containerfile itself (we can do it later)

Otherwise looks good!

Jmennius:
> Do you prefer to delete the image definitions for Ubuntu releases that have reached End of Life? eg., [toolbx-images/images@c2e2f0d](https://github.com/toolbx-images/images/commit/c2e2f0d3a526d55ee0ec9b54af338a493a29a2c2) We have kept all the image definitions for Fedora, just so that someone can locally build an image for an unsupported release, in case they need it, but for the Ubuntu images I want to leave it up to you.

Yeah, sounds reasonable to leave image definition but remove it from CI (after [EOL](https://wiki.ubuntu.com/Releases)).
(then we can leave 21.10 and I can resurrect 20.10 which was here in the past :))

debarshiray:
/cc @Jmennius and @travier

Thanks for clearing those up!  I have rebased this pull request against `main`, synchronized `images/ubuntu` with https://github.com/toolbx-images/images/tree/main/ubuntu, and added a bunch of new tests.

This means that the image definitions for Ubuntu 21.10 and 20.10 are currently not present because they are also absent from https://github.com/toolbx-images/images/tree/main/ubuntu.  I will leave it to you to decide if you want to bring them back or not.

debarshiray (commented):
I merged the first two patches in this pull request through https://github.com/containers/toolbox/pull/1284

I am not really qualified to understand the details of the third patch that adds the GitHub action to publish the images.  It seems to me that it needs to be updated to publish them to `quay.io/toolbx-images`.  Is that right?  If so, could you please update it?

Also, since @Jmennius is listed as the maintainer in the `Containerfiles`, I think we should update `.github/CODEOWNERS` to reflect that.

travier:
> I am not really qualified to understand the details of the third patch that adds the GitHub action to publish the images. It seems to me that it needs to be updated to publish them to `quay.io/toolbx-images`. Is that right? If so, could you please update it?

What's the long term plan? Do you want to move images back into this repo? Should we move the toolbx-images repo into the containers org?

debarshiray:
> Also, since @Jmennius is listed as the maintainer in the
> `Containerfiles`, I think we should update `.github/CODEOWNERS`
> to reflect that.

@Jmennius I sent you an invitation with `write` permissions and updated `.github/CODEOWNERS`: https://github.com/containers/toolbox/pull/1287  Does that look good to you?

@travier do you want to be listed as well?

travier:
I think we should have discussion about the future of https://github.com/toolbx-images/images (which could be anything: moving it to the containers org, or keeping it but just for some images) before we start including other images here.

Do we want to produce images from this repo and push them to quay? Shouldn't we keep all images in a separated repo like we did in https://github.com/toolbx-images/images to make this de-coupled from toolbox maintenance itself?

Right now the Fedora images are built on Fedora infra. Where / who will build the other images? How can we claim that they are official?

travier:
My preference would be that we move the https://github.com/toolbx-images/images repo to the containers org and keep all images definitions there, pushing to Quay. This will keep them clearly marked as "community" images.

debarshiray:
> > I am not really qualified to understand the details of the third patch that adds the GitHub action to publish the images. It seems to me that it needs to be updated to publish them to `quay.io/toolbx-images`. Is that right? If so, could you please update it?
> 
> What's the long term plan? Do you want to move images back into this repo?

Long-term, when we land built-in support for a distribution through the `--distro` option, the upstream image definitions should also live in `toolbox.git/images`.

Some distributions might want to have downstream copies of the image sources.  It's on those distribution maintainers to keep those copies synchronized.  eg., the RHEL images will probably always be copied downstream.

If someone who isn't a Red Hat employee wants to submit a change to Toolbx that requires changes to the images, then they shouldn't have to worry about the downstream copy.  However, since the RHEL image sources are part of `toolbox.git`, the contributor can't leave them broken as a result of their changes and will have to ensure that they continue to work.  I think that this strikes a decent balance between a contributor and the RHEL maintainers.

A contributor to `toolbox.git` should only have to ensure that the sources in the repository work well together.  They shouldn't have to track down various other Git repositories.  Similarly, those landing the changes shouldn't have to worry about coordinating the merges.

I also covered this in greater detail in https://github.com/containers/toolbox/issues/1019

> Should we move the toolbx-images repo into the containers org?

I don't know.  :)

I answered this in https://github.com/containers/toolbox/issues/1019#issuecomment-1489510386

debarshiray:
> Do we want to produce images from this repo
> and push them to quay?

Yes, sure.  :)

> Shouldn't we keep all images in a separated
> repo like we did in
> https://github.com/toolbx-images/images to
> make this de-coupled from toolbox maintenance
> itself?

It's difficult to avoid the fact that Toolbx is composed of two parts that need to work well together - the `toolbox(1)` command and the OCI images.  Sometimes one requires a change in the other, and they need to be tested together, etc..

> Right now the Fedora images are built on Fedora
> infra and the Containerfiles here are just informative.

Umm... I consider it the other way around.  :)

When working on Toolbx, it's a lot easier to run a `podman build` against the images in `toolbox.git/images` and test the outcome against different versions of the `toolbox(1)` command.  It doesn't work very well the other way around.  eg., I [don't know](https://src.fedoraproject.org/container/fedora-toolbox/pull-request/4#comment-132284) how to do a scratch build for OCI images on the Fedora infra'.  It's even more cumbersome for the RHEL images.

So, I primarily copy the changes downstream and build them once I am sure that the upstream image definition works.

I want to get to a point where we `podman build` the images as part of our test suite, and run the tests both against the images that are currently on the registries and the ones that were built from source.  That will make the upstream copies in `toolbox.git/images` a lot more authoritative.

> Where / who will build the other images?

It depends on the image maintainers.  They can build and publish directly from here to some registry, or they can keep separate copies and use a different infra' like Fedora and RHEL, or something else.

Ultimately, the upstream Toolbx project can only support what's in `toolbox.git`.  We have been a bit hand-wavy about it because we only supported Fedora until recently, and we are all Fedora contributors, so it was all the same.

As we add more and more distributions, this is quickly going to get very complicated, because the test matrix will explode.  I am already feeling it with the RHEL images.  Hence, I want all the sources for the built-in images to be in `toolbox.git/images` so that we can test, commit, merge, etc. changes in single atomic and coherent units, without worrying about chasing and coordinating across multiple repositories.  Otherwise they will never be first-class citizens.

> How can we claim that they are official?

They are *official* from the perspective of the upstream Toolbx project because they are wired to `--distro`.  They are *official* from the perspective of the distribution because some people packaged up `toolbox(1)` for that distro and some decided to make an image.  Hopefully there would be an overlap between those two groups.

In practice, free software is chaotic, so sometimes there might be competing images for the same distribution, but for the most part people will band around the image that's best-in-breed, and that's the one we will bless upstream.

I mean, are the Ubuntu [images](https://hub.docker.com/_/ubuntu) on Docker Hub official?  Are they official because it says so in the title of the page, or because it's marked as maintained by Canonical and a Debian Developer?  Those clearly don't come from a `ubuntu.com` sub-domain.  :)

Are the `fedora-toolbox` images official?  Is it because they come from a `fedoraproject.org` sub-domain, because [registry.fedoraproject.org](https://registry.fedoraproject.org/repo/fedora-toolbox/tags/) has very little information?  Or are they offiicial because they are known to be first-class citizens?

I also covered this in greater detail in https://github.com/containers/toolbox/issues/1019#issuecomment-1489468834

travier:
> > How can we claim that they are official?
> 
> I mean, are the Ubuntu [images](https://hub.docker.com/_/ubuntu) on Docker Hub official? Are they official because it says so in the title of the page, or because it's marked as maintained by Canonical and a Debian Developer? Those clearly don't come from a `ubuntu.com` sub-domain. :)

They are official because they are pushed by Ubuntu maintainers to the Dockerhub repo.

> Are the `fedora-toolbox` images official? Is it because they come from a `fedoraproject.org` sub-domain, because [registry.fedoraproject.org](https://registry.fedoraproject.org/repo/fedora-toolbox/tags/) has very little information? Or are they offiicial because they are known to be first-class citizens?

Similarly they are official because they are pushed by Fedora maintainers and built on Fedora infra.

>> Right now the Fedora images are built on Fedora infra and the Containerfiles here are just informative.
>
> Umm... I consider it the other way around. :)

OK, but it's written nowhere and does not reflect the actual setup (as they are not built from the Containerfiles here) so it's confusing.

I'm fine with building and hosting Ubuntu images in the toolbx org in Quay.io. Let me know so that I can remove them from the toolbx-image repo and redirect folks to them here.

Jmennius:
> I merged the first two patches in this pull request through #1284

Awesome!
 
> I am not really qualified to understand the details of the third patch that adds the GitHub action to publish the images. It seems to me that it needs to be updated to publish them to `quay.io/toolbx-images`. Is that right? If so, could you please update it?

Will do.

> > Also, since @Jmennius is listed as the maintainer in the
> > `Containerfiles`, I think we should update `.github/CODEOWNERS`
> > to reflect that.
> 
> @Jmennius I sent you an invitation with `write` permissions and updated `.github/CODEOWNERS`: #1287 Does that look good to you?

Yep.

Jmennius:
> I am not really qualified to understand the details of the third patch that adds the GitHub action to publish the images. It seems to me that it needs to be updated to publish them to `quay.io/toolbx-images`. Is that right? If so, could you please update it?

Regarding that - `quay.io/toolbx-images` is the registry/organization of `toolbx-images.git`.
If we intend to push there - we have to coordinate workflows between our repositories (disable the one in `toolbx-images.git`, remove image definitions?). Or should we use the `toolbx` organization we've already established some time ago?
Anyway, let's continue discussing this.

Jmennius:
I've synchronized the workflow mostly, excluding some templating, @travier - what do you think, should we leave it?
TODO:
- [ ] decide which registry to use (and credentials for the robot account)

travier:
I don't have admin access to this repo or to the `toolbx` org on Quay.io so I can't do anything here. I can grant access to the `toolbx-images` repos on GitHub & Quay.io to folks if needed.

Jmennius:
@travier @debarshiray [toolbx](https://quay.io/organization/toolbx?tab=teams) quay organization has two owners (I guess @Foxboron is one of the owners.), and several contributors - Timothée an me are on the list.
Lets use that one so it doesn't conflict with `toolbox-images`?
I figure if we add somebody from containers org to the list of owners (i.e. @debarshiray ) - we could use that one for 'official' images.

Jmennius:
BTW, I've opened #1291 and #1292.

Foxboron:
> (I guess @Foxboron is one of the owners.),

I don't think so. I can only see contributors in that org.

debarshiray:
> > > How can we claim that they are official?
> > 
> > 
> > I mean, are the Ubuntu [images](https://hub.docker.com/_/ubuntu) on Docker Hub official? Are they official because it says so in the title of the page, or because it's marked as maintained by Canonical and a Debian Developer? Those clearly don't come from a `ubuntu.com` sub-domain. :)
> 
> They are official because they are pushed by Ubuntu maintainers to the Dockerhub repo.

*nod*

I think it's reasonable to expect that the *official* Toolbx image for a distribution will be maintained by folks working on that distribution.  So, people with a `@archlinux.org` email for the Arch Linux image, `@fedoraproject.org` for the Fedora image, etc..

That's already the case for the folks who maintain the packages shipping the `toolbox(1)` binary, and true for many of the images.  eg., @Foxboron is [@archlinux.org](https://archlinux.org/people/trusted-users/#Foxboron).  I think @Jmennius is a prominent exception here.  :)

Since the `toolbox(1)` binary and the image works together, it's reasonable to expect that there will be an overlap between the two groups.  So, I think sooner or later @Jmennius will acquire a `@ubuntu.com` identity.  Until then, I am wondering if we should invite the folks maintaining the Ubuntu [package](https://code.launchpad.net/ubuntu/+source/golang-github-containers-toolbox) to participate in taking care of the image.  I see [Andrej Shadura](https://nm.debian.org/person/andrewsh/) and [Ileana Dumitrescu](https://nm.debian.org/person/ildumi/) committing to the package.

Anyway, @Jmennius is a good example of why I am a bit wary of making it a hard requirement for the image maintainer to have an official connection to the distribution.  After all, if someone from outside volunteers to create an image and maintain it over time, and nobody else from the distribution has a competing image, then do we really want to block it?  :)

> > > Right now the Fedora images are built on Fedora infra and the Containerfiles here are just informative.
> > 
> > 
> > Umm... I consider it the other way around. :)
> 
> OK, but it's written nowhere and does not reflect the actual setup (as they are not built from the Containerfiles here) so it's confusing.

Yes, I agree.  Things grew organically and we never formalized them.

> I'm fine with building and hosting Ubuntu images in the toolbx org in Quay.io. Let me know so that I can remove them from the toolbx-image repo and redirect folks to them here.

Yes, please, that will be good.  Could you co-ordinate the merging of the GitHub Action with @Jmennius ?

debarshiray (commented):
I am not very qualified to review this.  I think you and @travier understand this way better than I do.

By the way, don't you want to add yourself to `.github/CODEOWNERS` as the owner of this GitHub Action for Ubuntu?  What about @travier ?

Jmennius:
> I am not very qualified to review this. I think you and @travier understand this way better than I do.
> 
> By the way, don't you want to add yourself to `.github/CODEOWNERS` as the owner of this GitHub Action for Ubuntu? What about @travier ?

I've implemented the action initially but @travier has used and evolved it a lot in `toolbx-images/images`, so it would be great to have an Ack from him.

Regarding [toolbx](https://quay.io/organization/toolbx) quay org - I've been invited by you @debarshiray :) We also have @travier in members there. I don't really see a way to discover who those two admins are :) 
What we need to proceed with this quay repo - create a robot account in the organization and add its token as [Github secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets). Update references to robot name and secret name and that's all.

Jmennius:
> Since the `toolbox(1)` binary and the image works together, it's reasonable to expect that there will be an overlap between the two groups. So, I think sooner or later @Jmennius will acquire a `@ubuntu.com` identity. Until then, I am wondering if we should invite the folks maintaining the Ubuntu [package](https://code.launchpad.net/ubuntu/+source/golang-github-containers-toolbox) to participate in taking care of the image. I see [Andrej Shadura](https://nm.debian.org/person/andrewsh/) and [Ileana Dumitrescu](https://nm.debian.org/person/ildumi/) committing to the package.

Yeah, I agree, it is only natural that you need the Ubuntu image for `toolbox(1)` on Ubuntu. I'd be happy to hand over those images!
BTW, @andrewshadura is already a maintainer in `toolbx-images` and he has shown interest in toolbx-images/images#17. I'll ping him there.

.github/workflows/ubuntu-images.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: 'jmennius+toolbox_test'  # TODO: credentials
+          password: ${{ secrets.QUAY_TEST_ROBOT_TOKEN }}
```
debarshiray, line 41: Shouldn't these be `toolbx-images+github` and `secrets.QUAY_ROBOT_TOKEN` ?

Of course, we might have to tweak the settings of this project for them to actually work, but it's time to get the code ready, no?

.github/workflows/ubuntu-images.yaml
```diff
@@ -0,0 +1,61 @@
...
+concurrency: ${{ github.workflow }}
+
+env:
+  latest_release: '22.04'
+
+jobs:
+  build-and-push-images:
+    strategy:
+      matrix:
+        release: ['16.04', '18.04', '20.04', '22.04', '22.10']
```
debarshiray, line 23: We need to add `23.04` to this list.

debarshiray (changes requested):
Some fly-by comments ...

I saw that the corresponding GitHub Action in [github.com/toolbx-images/images](https://github.com/toolbx-images/images/blob/main/.github/workflows/ubuntu.yaml) has `permissions: read-all` which is missing here.  Is that important?

debarshiray:
> I don't have admin access to this repo

@travier I added you as an admin to this repository.

> or to the `toolbx` org
> on Quay.io so I can't do anything here.

Let's see what's the situation there.

> I can grant access to
> the `toolbx-images` repos on GitHub & Quay.io to folks if needed.

I think you had once set me an invitation to `toolbx-images` but it got buried in my inbox and expired before I could act on it.  Feel free to send me another invitation if you think I can help with something there.

debarshiray:
> > Since the `toolbox(1)` binary and the image works together, it's reasonable to expect that there will be an overlap between the two groups. So, I think sooner or later @Jmennius will acquire a `@ubuntu.com` identity. Until then, I am wondering if we should invite the folks maintaining the Ubuntu [package](https://code.launchpad.net/ubuntu/+source/golang-github-containers-toolbox) to participate in taking care of the image. I see [Andrej Shadura](https://nm.debian.org/person/andrewsh/) and [Ileana Dumitrescu](https://nm.debian.org/person/ildumi/) committing to the package.
> 
> Yeah, I agree, it is only natural that you need the Ubuntu image for `toolbox(1)` on Ubuntu. I'd be happy to hand over those images! BTW, @andrewshadura is already a maintainer in `toolbx-images` and he has shown interest in [toolbx-images/images#17](https://github.com/toolbx-images/images/issues/17). I'll ping him there.

@Jmennius instead of handing over Ubuntu images, you could also get yourself a @ubuntu.com identity, and co-maintain them with @andrewshadura It never hurts to have a second pair of eyes. :)

debarshiray:
> > I don't have admin access to this repo
> 
> @travier I added you as an admin to this repository.
> 
> > or to the `toolbx` org
> > on Quay.io so I can't do anything here.
> 
> Let's see what's the situation there.

I keep getting lost with Quay.io.  I clearly remember creating the [toolbx](https://quay.io/organization/toolbx) a few months ago, but I just can't log in as the owner of it.

These days Quay.io keeps insisting to *sign in with Red Hat*, which looks like some sort of single sign-on shared with the Red Hat Developer [portal](https://developers.redhat.com/), and I see that I have accumulated three (!) accounts.  Two corresponding with different `@redhat.com` aliases and one with my personal identity.

The problem is that it's not easy to switch through these accounts to see which one owns the `toolbx` organization, and potentially clean up the accounts.  Once I have signed in, I simply can't sign out and log in with something else.  That *sign in with Red Hat* thing insists on remembering my last log-in.

travier:
OK, I have access to the toolbx org on Quay.io now.

.github/workflows/ubuntu-images.yaml
```diff
@@ -0,0 +1,61 @@
...
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to Quay.io
+        uses: docker/login-action@v2
+        with:
+          registry: quay.io
+          username: 'jmennius+toolbox_test'  # TODO: credentials
```
travier, line 40:
```suggestion
          username: 'toolbx+github'
```

.github/workflows/ubuntu-images.yaml
travier, line 41:
```suggestion
          password: ${{ secrets.QUAY_ROBOT_TOKEN }}
```

travier:
I've set up the robot and permissions on Quay and added the token in the repo here.

travier:
Note that to clearly distinguish PR & commit workflows, I've used two distinct GitHub Action definition in the toolbx-images repo. This PR is missing the one for PRs (https://github.com/toolbx-images/images/blob/main/.github/workflows/ubuntu-pr.yaml) and thus it is not triggered here.

debarshiray:
> I've set up the robot and permissions on Quay and added the token in
> the repo here.

Thanks for doing that, @travier !

What are the next steps?  Are we waiting for @Jmennius to update the GitHub Action in this pull request?

> Note that to clearly distinguish PR & commit workflows, I've used
> two distinct GitHub Action definition in the toolbx-images repo. This
> PR is missing the one for PRs
> (https://github.com/toolbx-images/images/blob/main/.github/workflows/ubuntu-pr.yaml)
> and thus it is not triggered here.

I will let @Jmennius respond to this one.

debarshiray:
Ping, @travier @Jmennius :  what's the next step here?

The most pressing issue right now for Ubuntu support is to get the `ubuntu-toolbox` images for 23.04 published.

.github/workflows/ubuntu-images.yaml
debarshiray: Done.

.github/workflows/ubuntu-images.yaml
debarshiray: Done.

.github/workflows/ubuntu-images.yaml
debarshiray: Done, as @travier suggested.  It's actually `toolbx+github` now, not `toolbx-images+github`, because we want to publish these images at `quay.io/toolbx/ubuntu-toolbox`.

debarshiray:
> The missing bits are in my review [#483 (review)](https://github.com/containers/toolbox/pull/483#pullrequestreview-1436078151). Then we should be able to merge this one.

Okay!  I took the liberty to make those changes and merged.  I hope I got it right.

debarshiray:
I filed a pull request to switch the `toolbox(1)` code and the system tests to use the new location: https://github.com/containers/toolbox/pull/1306

PR #484, opened 2020-06-29
Don't use Go's semantic import versioning

debarshiray:
Fedora doesn't support Go modules when building Go programs. This
means that source code using semantic import versioning can't be built.


HarryMichal:
I'd rather not merge this as Go modules are enabled by default and could actually be called a standard way to consume Go (respectively Go packages) nowadays. The inability to work with Go modules is a problem of Fedora (which should be fixed by Fedora 34 as a part of a bigger change in the Fedora Go RPM stack) and patches should be made downstream rather than upstream where it would in my opinion cause more harm then good.

debarshiray:
> I'd rather not merge this

Ok, fair enough. Let's not merge this. Closing.

PR #491, opened 2020-07-07
cmd/initContainer: Give access to the udev database

### LINKED ISSUE #468: libudev in toolbox fails to retreive data set by udev rules on the host
I am using a toolbox to develop pipewire, which uses libudev to enumerate alsa and v4l2 devices. While device enumration works fine in the toolbox, the devices seem to be missing some properties that are normally present on the host...

After some digging I realized that these properies are set by udev rules on the host and stored in the udev database. When enumerating, libudev uses `/run/udev/data` to access them.

In the container, this directory is available in `/run/host/run/udev/data` instead. To work this around, I added a symlink in the container: `ln -s /run/host/run/udev /run/udev`

HarryMichal:
Enumeration of audio devices using libudev seems to work without this
but some properties are missing. This complicates the development of
software like pipewire.

Closes #468

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -53,6 +53,7 @@ var (
 		{"/etc/machine-id", "/run/host/etc/machine-id", "ro"},
 		{"/run/libvirt", "/run/host/run/libvirt", ""},
 		{"/run/systemd/journal", "/run/host/run/systemd/journal", ""},
+		{"/run/udev/data", "/run/host/run/udev/data", ""},
 		{"/var/lib/flatpak", "/run/host/var/lib/flatpak", "ro"},
 		{"/var/log/journal", "/run/host/var/log/journal", "ro"},
 		{"/var/mnt", "/run/host/var/mnt", "rslave"},


src/cmd/initContainer.go
```diff
@@ -172,6 +172,14 @@ func initContainer(cmd *cobra.Command, args []string) error {
 				}
 			}
 
+			if _, err := os.Readlink("/run/udev"); err != nil {
```
debarshiray, line 175: It would be better to do it as a bind mount instead.

For individual regular files we prefer symbolic links instead of bind mounts, because symbolic links can track updates to the original file. Whereas, with bound mounts, if the original file was updated [atomically](http://www.microhowto.info/howto/atomically_rewrite_the_content_of_a_file.html), which is what most programs do, then the atomic rename changes the inode, and the two diverge.

However, symbolic links are always weird and introduce various oddities (eg., `stat` versus `lstat`), so bind mount is better when possible.

debarshiray (changes requested):

src/cmd/initContainer.go
debarshiray: I took the liberty to switch it to a bind mount.

PR #493, opened 2020-07-07
cmd/create: Update hint after creating a container

### LINKED ISSUE #489: toolbox 0.0.92 - unify toolbox create command with blog post
Hello! I saw HarryMichal's blog post about new version of podman rewritten in go - thank you for improving toolbox!
I compiled binary myself on Silverblue and one thing occurred to me. 
In blog post ( http://harrymichal.undo.it/posts/2020/wheres-toolbox-0.0.9x-update/ ) I got an impression that toolbox commands got unified with podman and new (preferred) way to enter container is 

`toolbox enter <container-name>` 

but when I create new container message returned from toolbox displays old way of entering with `--container` parameter:

```
~ ➜ toolbox create my-container
Created container: my-container
Enter with: toolbox enter --container my-container
```
Not sure if that's intended or not so I decided to create this simple to fix issue, can create PR if someone decides that it's preferred way.

Thanks

HarryMichal:
Rewritten version of Toolbox supports newer syntax for entering a
container: `toolbox enter <name-of-container>`. This new workflow should
be shown in the hint after creating a container with `toolbox create`.

Before this change:

```
$ toolbox create 123
Created container: 123
Enter with: toolbox enter --container 123
```

After this change:

```
$ toolbox create 123
Created container: 123
Enter with: toolbox enter 123 | toolbox enter --container 123
```

Closes #489

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -466,7 +466,7 @@ func getEnterCommand(container, release string) string {
 	case containerNamePrefixDefaultWithRelease:
 		enterCommand = fmt.Sprintf("%s enter --release %s", executableBase, release)
 	default:
-		enterCommand = fmt.Sprintf("%s enter --container %s", executableBase, container)
+		enterCommand = fmt.Sprintf("%s enter %s", executableBase, container)
 	}
 
 	return enterCommand


src/cmd/create.go
```diff
@@ -463,7 +463,11 @@ func getEnterCommand(container, release string) string {
 	case containerNamePrefixDefaultWithRelease:
 		enterCommand = fmt.Sprintf("%s enter --release %s", executableBase, release)
 	default:
-		enterCommand = fmt.Sprintf("%s enter --container %s", executableBase, container)
+		enterCommand = fmt.Sprintf("%s enter %s | %s enter --container %s",
```
debarshiray, line 466: I think it's better to just mention the new syntax, and completely leave out the old one. :)

We still support the old syntax in the code for the sake of compatibility with old documentation and old habits. We should switch our user interface, documentation and such to advertise the new syntax, so that finally after some time down the road we can remove the old syntax.

debarshiray:
I noticed that changing the hint didn't require changes to the tests. Maybe we should check the hint in our test suite?

PR #494, opened 2020-07-08
cmd/list: Re-enable highlighting of running containers

HarryMichal:
The rewritten Toolbox did not highlight running containers because the
code uses a special table writter from 'text/tabwriter' that does not
support ANSI color codes[0]. The solution used here is to use color
codes on _every_ lines of the output (including headers) and make the
color codes use the same number of characters.

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -217,14 +217,20 @@ func listOutput(images, containers []map[string]interface{}) {
 	}
 
 	if len(containers) != 0 {
+		const boldGreenColor = "\033[1;32m"
+		const defaultColor = "\033[0;00m" // identical to resetColor, but same length as boldGreenColor
+		const resetColor = "\033[0m"
+
 		writer := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 		fmt.Fprintf(writer,
-			"%s\t%s\t%s\t%s\t%s\n",
+			"%s%s\t%s\t%s\t%s\t%s%s\n",
+			defaultColor,
 			"CONTAINER ID",
 			"CONTAINER NAME",
 			"CREATED",
 			"STATUS",
-			"IMAGE NAME")
+			"IMAGE NAME",
+			resetColor)
 
 		var idKey, createdKey, statusKey string
 		if podman.CheckVersion("2.0.0") {
@@ -252,7 +258,27 @@ func listOutput(images, containers []map[string]interface{}) {
 			status := container[statusKey].(string)
 			imageName := container["Image"].(string)
 
-			fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\n", id, nameString, created, status, imageName)
+			isRunning := false
+			if podman.CheckVersion("2.0.0") {
+				isRunning = status == "running"
+			}
+
+			var color string
+			if isRunning {
+				color = boldGreenColor
+			} else {
+				color = defaultColor
+			}
+
+			fmt.Fprintf(writer,
+				"%s%s\t%s\t%s\t%s\t%s%s\n",
+				color,
+				id,
+				nameString,
+				created,
+				status,
+				imageName,
+				resetColor)
 		}
 
 		writer.Flush()


debarshiray:
Nice. Well done!

Both GNOME Terminal and Tilix ultimately use [VTE](https://wiki.gnome.org/Apps/Terminal/VTE) for terminal emulation; so I tried xterm to have something different, and it seems to work well there too.

I will look at the actual code now.

src/cmd/list.go
```diff
@@ -252,7 +259,23 @@ func listOutput(images, containers []map[string]interface{}) {
...
 			imageName := container["Image"].(string)
 
-			fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\n", id, nameString, created, status, imageName)
+			// Prepare color for a line based on toolbox's status
+			// By default don't colorize a line
+			isRunning := false
+			if podman.CheckVersion("2.0.0") {
+				isRunning = status == "running"
+			} else {
+				isRunning, _ = regexp.MatchString("^Up [0-9]* [a-z]* ago$", status)
```
debarshiray, line 268: I salute your perseverance!

If I were to nitpick, I'd suggest first compiling the regular expression separately outside the loop and then matching it. See [regexp.Compile](https://golang.org/pkg/regexp/#Compile) and [regexp.Regexp.MatchString](https://golang.org/pkg/regexp/#Regexp.MatchString).

Note that the `regexp` APIs can fail if there's a mistake in the regular expression, which would be a programmer error. That makes me think that we should use [regexp.MustCompile](https://golang.org/pkg/regexp/#MustCompile) in some place like `init`. That way, if we ever typo the regular expression, `toolbox` will panic on start-up and it will be a lot harder for us to miss it.

All that said, why not use just [strings.HasPrefix](https://golang.org/pkg/strings/#HasPrefix) and [strings.HasSuffix](https://golang.org/pkg/strings/#HasSuffix)? No need to handle any errors, and likely faster than regular expressions. :)

To be honest, I would still be very happy if we coloured the output only with Podman >= 2.0.0.

src/cmd/list.go
```diff
@@ -217,14 +218,20 @@ func listOutput(images, containers []map[string]interface{}) {
 	}
 
 	if len(containers) != 0 {
+		// Because text/tabwriter does not support escape sequences for colors, we have to do some
+		// shenanigans (string of the same length; adding color to every line) to make this work
+		noColor := "\033[0m" // No color
+		color := "\033[0;00m"
```
debarshiray, line 224: Terminal escape sequences are like voodoo, so I wonder if we can add some comment or reference explaining the SGR sequences.

If I understood correctly, your clever hack was to use `\033[0;00m`, instead of the more idiomatic `\033[0m`, to match the number of characters in `\033[1;32m`. Right? I think it would be good to mention this in the commit message, because most documentation on SGR sequences will mention `\033[0m` and readers will be left confused what the other thing means.

src/cmd/list.go
```diff
@@ -252,7 +259,23 @@ func listOutput(images, containers []map[string]interface{}) {
...
+			// Prepare color for a line based on toolbox's status
+			// By default don't colorize a line
+			isRunning := false
+			if podman.CheckVersion("2.0.0") {
+				isRunning = status == "running"
+			} else {
+				isRunning, _ = regexp.MatchString("^Up [0-9]* [a-z]* ago$", status)
+			}
+			if isRunning {
+				color = "\033[1;32m" // Light Green Color
```
debarshiray, line 271: Nitpick: I know the POSIX shell implementation calls this *light green color*, but I think it's more important to mention that it's supposed to be *bold*. The first SGR parameter is 1, and it stands for bold. Whether 32 is green or light green is not such a big thing because we don't have any other shade of green, or even any other color.

src/cmd/list.go
debarshiray: I took the liberty to move most of the explanation into the Git commit message, and added some extra text about the matching lengths.

(I generally prefer the commit message over code comments. The former is set in stone and closely associated with code changes, while the latter can slowly drift out of sync as the surrounding code changes and people forget to update the comment.)

src/cmd/list.go
debarshiray: I am going to leave this for you to decide. :)

debarshiray: We can use a separate pull request for it.

debarshiray:
I merged everything, except the part about supporting Podman < 2.0.0. I will leave it up to you to decide what best to do.

One thing that comes to mind is whether we should be skipping the escape sequences when the output is not attached to a terminal device? People using `toolbox list | less` might appreciate it.

Either way, we can use separate PRs to address them.

Thanks again for working on this!

kindermoumoute:
Nice hack 👍

src/cmd/list.go
HarryMichal: I just got back to reading this. I took the comment from the Shell implementation. To be honest, before you explained it, I didn't know 1 stands for bold.

HarryMichal:
> I merged everything, except the part about supporting Podman < 2.0.0. I will leave it up to you to decide what best to do.

I think we can keep things like this. Podman overall moves very quickly forward (all supported versions of Fedora are already on V2) and only the very stable distros (Debian, CentOS, RHEL,..) will have to wait a bit. I don't think it's that important.

> One thing that comes to mind is whether we should be skipping the escape sequences when the output is not attached to a terminal device? People using `toolbox list | less` might appreciate it.

Yes, this is something that should be done.

debarshiray:
> > I merged everything, except the part about supporting Podman < 2.0.0.
> > I will leave it up to you to decide what best to do.
> 
> I think we can keep things like this. Podman overall moves very quickly
> forward (all supported versions of Fedora are already on V2) and only the
> very stable distros (Debian, CentOS, RHEL,..) will have to wait a bit. I don't
> think it's that important.

Ok, that makes sense.

src/cmd/list.go
debarshiray: Yes, sorry about that. In hindsight, the comments in the POSIX shell implementation weren't ideal. :/

PR #498, opened 2020-07-10
cmd/create: Restore the spinner, when pulling an image, to stdout

debarshiray:
This doesn't seem to have any user-visible effect. However, since the
line was inadvertently removed, it's good to put it back.

Fallout from 950f510872404da8089ad7ce5911d8342bf7a116

Fix-up for https://github.com/containers/toolbox/pull/496


HarryMichal:
lgtm

Issue #502, opened 2020-07-14
Listing images panics


HarryMichal:
Podman [changed](https://github.com/containers/podman/pull/6815/) their JSON API again causing Toolbox to panic when listing images (`toolbox list -i` or simply `toolbox list`). The affected field of the JSON is `Created` that used to hold a `string` with a human-readable representation of time (e.g., `5 minutes ago`). That field was changed to an `int` (but the Go code sees it as a `float`) with UNIX time.

Before the discussion in the mentioned PR comes to some result, Toolbox should be overall more resilient when some change in the API happens.

**Output of `toolbox --version`**:
```
toolbox version 0.0.92

```

**Output of `podman version`**:
```
Version:      2.1.0-dev
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

PR #503, opened 2020-07-14
cmd/list: Make listing of containers/images more robust

HarryMichal:
Every time Podman changes their JSON API Toolbox breaks horribly. That
is caused by the complete lack of type guards. With this, the verbosity
of the code goes up a bit but it should be much more resistant to API
changes.

If a field is of a not-expected type, the field is set to "<error>" and
an error message is printed after the rest of the table is printed into
the console.

An alternative to this solution is to unmarshal the JSON response to a
strongly typed structure, as pointed out by @owtaylor.

Closes: #502

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -303,7 +303,7 @@ func (i *toolboxImage) UnmarshalJSON(data []byte) error {
 	var raw struct {
 		ID      string
 		Names   []string
-		Created string
+		Created interface{}
 	}
 
 	if err := json.Unmarshal(data, &raw); err != nil {
@@ -312,7 +312,15 @@ func (i *toolboxImage) UnmarshalJSON(data []byte) error {
 
 	i.ID = raw.ID
 	i.Names = raw.Names
-	i.Created = raw.Created
+	// Until Podman 2.0.x the field 'Created' held a human-readable string in
+	// format "5 minutes ago". Since Podman 2.1 the field holds an integer with
+	// Unix time. Go interprets numbers in JSON as float64.
+	switch value := raw.Created.(type) {
+	case string:
+		i.Created = value
+	case float64:
+		i.Created = utils.HumanDuration(int64(value))
+	}
 
 	return nil
 }
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -27,6 +28,20 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+type toolboxImage struct {
+	ID      string
+	Names   []string
+	Created string
+}
+
+type toolboxContainer struct {
+	ID      string
+	Names   []string
+	Status  string
+	Created string
+	Image   string
+}
 
 var (
 	listFlags struct {
@@ -82,8 +97,8 @@ func list(cmd *cobra.Command, args []string) error {
 		lsImages = false
 	}
 
-	var images []map[string]interface{}
-	var containers []map[string]interface{}
+	var images []toolboxImage
+	var containers []toolboxContainer
 	var err error
 
 	if lsImages {
@@ -104,7 +119,7 @@ func list(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func listContainers() ([]map[string]interface{}, error) {
+func listContainers() ([]toolboxContainer, error) {
 	logrus.Debug("Fetching containers with label=com.redhat.component=fedora-toolbox")
 	args := []string{"--all", "--filter", "label=com.redhat.component=fedora-toolbox"}
 	containers_old, err := podman.GetContainers(args...)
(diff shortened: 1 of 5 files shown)


owtaylor:
I think that the approach that toolbox is taking to parsing json output is generally not optimal - and not the way that pretty much every golang project I've looked at does it. Unmarshaling to a map of interfaces puts the checking (and error handling as is added here) at the point of use, which is both error-prone and hard to maintain. 

We talked to the libpod maintainers some, and their recommendation was to use the libpod bindings that talk over a local http socket. Using this requires the fedora packaging to be updated to automatically create the socket for all users using a systemd user preset, and also would be a pretty major refactoring.

I think a short term fix would be, to, e.g. change GetContainers to return []Container where the Container type has just the fields that you actually care about, and unmarshal directly to that list. Then the json package will do the type checking upfront. It even has some ability to skip fields with a type mismatch:

>  If a JSON value is not appropriate for a given target type, or if a JSON number overflows the target type, Unmarshal skips that field and completes the unmarshaling as best it can. If no more serious errors are encountered, Unmarshal returns an UnmarshalTypeError describing the earliest such error. In any case, it's not guaranteed that all the remaining fields following the problematic one will be unmarshaled into the target object.

src/cmd/list.go
```diff
@@ -179,33 +215,37 @@ func listImages() ([]map[string]interface{}, error) {
 		images = utils.SortJSON(images, "names", true)
 	}
 
-	return images, nil
+	var toolboxImages []toolboxImage
+	for _, image := range images {
+		var i toolboxImage
+		imageJSON, err := json.Marshal(image)
```
owtaylor, line 221: I would have definitely expected the struct to be unmarshal'ed to when podman.GetImages() initially unmarshals the JSON. What's the reason for doing it here instead?

src/cmd/list.go
```diff
@@ -226,35 +266,65 @@ func listOutput(images, containers []map[string]interface{}) {
...
-			var nameString string
-			switch name := container["Names"].(type) {
-			case string:
-				nameString = name
-			case []interface{}:
-				nameString = name[0].(string)
-			}
+// GetCreatedString returns a string with human-readable time when an image was
+// created
+func (i toolboxImage) GetCreatedString() string {
```
owtaylor, line 286: Using getters is definitely an OK approach. I'd probably think about doing something with a custom unmarshal function

``` go
func (t *toolboxImage) UnmarshalJSON(data []byte) error {     
    var raw struct {
	ID      string
	Name    string
	Names   []string
	Status  string
	State   string
	Created interface{}
	Image   string
    }

   if err := json.Unmarshal(data, &raw); err != nil {       
       return err     
    }   

   t.ID = raw.ID
   if raw.Name != "" {
      t.name = c.Name
   } else {
      t.name = c.Names[0]
   }
   [...]
```

This puts all the JSON compatibility hacks into one place - with the getter approach, code that is using toolboxImage has to be aware that some fields are OK to access, and some fields need to use getters to sanitize them.

HarryMichal: Again, you're right. This sounds easier to maintain.

src/cmd/list.go
HarryMichal: Mainly, I wanted to keep the changes only in `src/cmd/list.go` and prevent any major refactors in other parts of the code that already rely on `podman.GetImages()`/`podman.GetContainers`.

HarryMichal: Unless I change `podman.GetImages()`/`podman.GetContainers()`, the additional marshaling needs to take place before unmarshaling into the struct.

src/cmd/list.go
HarryMichal: The latest iteration of the PR uses this approach.

src/cmd/list.go
owtaylor: I understand not wanting to start changing more and more code, so perhaps this is an appropriate workaround for now. But I'd suggest a comment saying that this is a temporary workaround, and the unmarshalling and compatibility code should move to utils/podman.go - after all, having

```
                if podman.CheckVersion("2.0.0") {
                        idKey = "Id"
                } else {
                        idKey = "ID"
                }
```

in cmd/rm.go when it is using `GetContainers()` is not exactly where we want the code to end up either!

src/cmd/list.go
```diff
@@ -226,35 +262,77 @@ func listOutput(images, containers []map[string]interface{}) {
...
+		Created interface{}
+		Image   string
 	}
+
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	c.ID = raw.ID
+	if raw.Name != "" {
```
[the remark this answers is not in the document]
HarryMichal: This should be fixed in the latest interation of the PR.

src/cmd/list.go
```diff
@@ -226,35 +262,77 @@ func listOutput(images, containers []map[string]interface{}) {
...
 
-			fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\n", id, nameString, created, status, imageName)
-		}
+	i.ID = raw.ID
+	i.Name = raw.Names[0]
+	switch value := raw.Created.(type) {
+	case string:
+		i.Created = value
+	case float64:
+		i.Created = units.HumanDuration(time.Since(time.Unix(int64(value), 0))) + " ago"
```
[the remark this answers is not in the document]
HarryMichal: In the latest iteration is https://github.com/containers/toolbox/pull/503/commits/8679f95696fb6d8e1faa6938184d832a38ea623d.

src/cmd/list.go
```diff
@@ -226,35 +262,77 @@ func listOutput(images, containers []map[string]interface{}) {
...
-			status := container[statusKey].(string)
-			imageName := container["Image"].(string)
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
 
-			fmt.Fprintf(writer, "%s\t%s\t%s\t%s\t%s\n", id, nameString, created, status, imageName)
-		}
+	i.ID = raw.ID
+	i.Name = raw.Names[0]
```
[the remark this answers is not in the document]
HarryMichal: In the last iteration where it was relevant I added some comments.

HarryMichal:
System tests are still failing on Fedora Rawhide but that is due to a different issue, not #502 that is fixed by this PR.

HarryMichal:
I just tried this with Podman v1.9.3 and there's a problem with listing containers due to:

```
failed to unmarshal container: json: cannot unmarshal number into Go struct field .State of type string
```

I'll adjust this accordingly.

src/cmd/list.go
HarryMichal: I added a comment before both marshalling + unmarshalling sections.

owtaylor:
Latest version looks good to me!

debarshiray:
Thanks for all your work on this, @HarryMichal

PR #509, opened 2020-07-23
cmd/run: Adjust for changes in 'podman ps' in Podman 2.0

debarshiray:
The key containing the container's name (ie., "Names") is no longer a
string, but a slice of strings.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -194,7 +194,13 @@ func runCommand(container string,
 		} else if containersCount == 1 && defaultContainer {
 			fmt.Fprintf(os.Stderr, "Error: container %s not found\n", container)
 
-			container = containers[0]["Names"].(string)
+			switch value := containers[0]["Names"].(type) {
+			case string:
+				container = value
+			case []interface{}:
+				container = value[0].(string)
+			}
+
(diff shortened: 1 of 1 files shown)


debarshiray:
I wish we had tests for the `enter` command, but I guess it's hard to drive it non-interactively as part of a test-suite. This code path is part of the fallback we have when there's only one toolbox container, but it's not the default, and `toolbox enter` shows a message and continues to use it without requiring any arguments.

I am also looking forward to https://github.com/containers/toolbox/pull/503 which will make the JSON parsing a lot nicer. :)

HarryMichal:
> I wish we had tests for the `enter` command, but I guess it's hard to drive it non-interactively as part of a test-suite. This code path is part of the fallback we have when there's only one toolbox container, but it's not the default, and `toolbox enter` shows a message and continues to use it without requiring any arguments.

It is possible to do it even if it is not interactive. I slimmed the CI considerably in https://github.com/containers/toolbox/pull/375 mainly due to problems with pulling images that caused the CI to flake a lot (https://github.com/containers/toolbox/pull/372). Currently, `bats` does not clean between test cases to save bandwidth and time. We could break that pattern and make the tests more complicated.

PR #510, opened 2020-07-23
cmd/create: Drop redundant label when creating a container

HarryMichal:
The Toolbox repository was moved to the 'containers' organization [some time ago](https://github.com/containers/toolbox/commit/de5e5df9b793edc746cde7125cd437b69f15ccd7) already. The label will remain supported but new containers will not be created with it.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -395,7 +395,6 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 		"--hostname", "toolbox",
 		"--ipc", "host",
 		"--label", "com.github.containers.toolbox=true",
-		"--label", "com.github.debarshiray.toolbox=true",
 	}...)
 
 	createArgs = append(createArgs, devPtsMount...)


debarshiray:
This looks good, but let's wait a bit before landing this because it's only [recently](https://github.com/containers/toolbox/commit/b511a76578555e207be2f17bf787e02b92c612c6) that we switched out the `com.redhat.component` label in favour of the `com.github.containers.toolbox` label.

Those who might be switching back and forth between different Toolbox versions (thanks to OSTree) might be happier.

HarryMichal:
@debarshiray What's the status of this now that https://github.com/containers/toolbox/pull/820 got merged?

olivergs:
I think as we already deprecated the old label we should remove this also for next release.

debarshiray:
Thanks for merging this, @HarryMichal !

I wonder if it's time to completely drop the `com.github.debarshiray.toolbox` label from `list`, `rm`, `rmi`, etc..

As far as I can make out, the current `com.github.containers.toolbox` label was introduced in commit de5e5df9b793edc746cde7125cd437b69f15ccd7 That was right before the release of [Fedora 31](https://fedoraproject.org/wiki/Releases/31/Schedule). Today the oldest supported Fedora is 34.

PR #511, opened 2020-07-23
meson: Add option for installing tests

HarryMichal:
This adds a new option to Meson called `with_tests` making it possible to install whole 'test' subdir into the system's data dir (usually `/share` or `/usr/share`).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -28,6 +28,11 @@ if bash_completion.found()
   )
 endif
 
+install_subdir(
+  'test',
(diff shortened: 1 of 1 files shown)


HarryMichal:
My thinking behind making this a build option was that the tests are not something that is necessary for running Toolbox, thus they don't need to be enabled by default. But I believe you're right about this and the option is not really necessary.

I'll add the link to the PR (I forgot to do that when I added CONTRIBTING.md...).

PR #512, opened 2020-07-23
doc: Match man pages with Go Toolbox and extend them

HarryMichal:
The manual pages were not updated for quite some time already. Several parts are not relevant anymore (e.g., mentions of 'candidate registry') and some could be more accurate/more elaborate.

I mainly expanded toolbox(1), toolbox-create(1) and toolbox-init-container(1) where I tried to explain more of the philosophy of toolbox containers/images, what defines them, what makes them different than common OCI containers/images.

### DIFF
--- a/doc/toolbox-create.1.md
+++ b/doc/toolbox-create.1.md
@@ -26,36 +26,42 @@ the image had a tag, then the tag is included in the name of the container,
 but it's separated by a hyphen, not a colon. A different name can be assigned
 by using the CONTAINER argument.
 
+If the host system is not recognized the used image will be `fedora-toolbox`.
+
+Toolbox containers are primarily created in a way to be tightly integrated with
+the host system. They are not meant to be secure.
+
 ### Entry Point
 
 A key feature of toolbox containers is their entry point, the `toolbox
 init-container` command.
 
-OCI containers are inherently immutable. Configuration options passed through
-`podman create` are baked into the definition of the OCI container, and can't
-be changed later. This means that changes and improvements made in newer
-versions of Toolbox can't be applied to pre-existing toolbox containers
-created by older versions of Toolbox. This is avoided by using the entry point
-to configure the container at runtime.
+Read more about the entry-point in `toolbox-init-container(1)`.
 
-The entry point of a toolbox container customizes the container to fit the
-current user by ensuring that it has a user that matches the one on the host.
-It ensures that configuration files, such as `/etc/host.conf`, `/etc/hosts`,
-`/etc/localtime`, `/etc/resolv.conf` and `/etc/timezone`, inside the container
-are kept synchronized with the host. The entry point also bind mounts various
-subsets of the host's filesystem hierarchy to their corresponding locations
-inside the container to provide seamless integration with the host. This
-includes `/run/libvirt`, `/run/systemd/journal`, `/run/udev/data`,
-`/var/lib/libvirt`, `/var/lib/systemd/coredump`, `/var/log/journal` and others.
+### Toolbox setup
 
-On some host operating systems, important paths like `/home`, `/media` or
-`/mnt` are symbolic links to other locations. The entry point ensures that
-paths inside the container match those on the host, to avoid needless
-confusion.
+`toolbox-create(1)` passes several options to `podman-create(1)` when creating
+toolbox containers to provide the needed functionality. The options have the
+following effects:
+
+- Toolboxes share with the host system:
+    - network stack, including dns
+    - IPC (shared memory, semaphores, message queues,..)
+    - PID namespace
+    - ulimits
+- Toolboxes have access to cherry-picked parts of host filesystem made
+  available under /run/host/
+- Toolboxes are privileged containers
+- SELinux label separation is disabled for toolboxes
+- Toolboxes use as their entry-point `toolbox-init-container(1)`
+
+Despite being privileged, rootless containers cannot have more privileges than
+the user that created them.
+
+Thanks to these options, `toolbox-init-container(1)` can futher set up the
+containers. Read more about the entry-point in `toolbox-init-container(1)`.
 
 ## OPTIONS ##
-
-The following options are understood:
 
 **--distro** DISTRO, **-d** DISTRO
 
@@ -67,6 +73,9 @@ host. Cannot be used with `--image`.
 Change the NAME of the base image used to create the toolbox container. This
 is useful for creating containers from custom-built base images. Cannot be used
 used with `--release`.
+
+If NAME does not contain a domain, the image will be pulled from
+`registry.fedoraproject.org`.
 
 **--release** RELEASE, **-r** RELEASE
 
@@ -95,4 +104,4 @@ $ toolbox create --image bar foo
 
 ## SEE ALSO
 
-`buildah(1)`, `podman(1)`
+`toolbox(1)`, `toolbox-init-container(1)`, `podman(1)`, `podman-create(1)`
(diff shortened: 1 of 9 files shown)


doc/toolbox-create.1.md
```diff
@@ -4,48 +4,144 @@
 toolbox\-create - Create a new toolbox container
 
 ## SYNOPSIS
-**toolbox create** [*--candidate-registry*]
```
debarshiray, line 7: I split the removal of `--candidate-registry` into https://github.com/containers/toolbox/pull/662

doc/toolbox-create.1.md
```diff
@@ -4,48 +4,144 @@
...
 Creates a new toolbox container. You can then use the `toolbox enter` command
 to interact with the container at any point.
 
-A toolbox container is an OCI container created from an OCI image. On Fedora
-the base image is known as `fedora-toolbox`. If the image is not present
-locally, then it is pulled from `registry.fedoraproject.org`. The base image is
-locally customized for the current user to create a second image, from which
-the container is finally created.
+A toolbox container is an OCI container created from an OCI image. Therefore,
+`toolbox create` is analogous to `podman build` executed with specific options.
```
debarshiray, line 15: Did you really mean `podman build`? Or was it meant to be `podman create`?

doc/toolbox-create.1.md
```diff
@@ -4,48 +4,144 @@
...
+On Fedora the base image is known as `fedora-toolbox`. If the image is not
+present locally, then it is pulled from `registry.fedoraproject.org`.
 
-## OPTIONS ##
+If the host system is not recognized the used image will be `fedora-toolbox`.
+
+Toolbox containers are primarily created in a way to be tightly integrated with
+the host system. They are not meant to be secure.
+
+### Options passed to podman build
```
debarshiray, line 25: This should definitely be `podman create`. :)

doc/toolbox-create.1.md
```diff
@@ -4,48 +4,144 @@
...
+  path is just an example one, the number depends on the user's UID)
+
+If a KCM (SSSD Kerberos Cache Manager) socket is detected on the host, it is
+mounted inside of the container automatically. Thanks to this, authentication
+using Kerberos will work even in a container.
+
+- `--volume /run/.heim_org.h5l.kcm-socket:/run/.heim_org.h5l.kcm-socket` (this
+  path is just an example one, the socket may be in a different location)
+
+### Entry-point of toolbox containers
```
debarshiray, line 113: Yes, this is a good idea. I split this out into https://github.com/containers/toolbox/pull/677

doc/toolbox-create.1.md
```diff
@@ -4,48 +4,144 @@
...
+
+A key feature of toolbox containers is their entry-point, the `toolbox
+init-container` command. The entry-point command of an OCI container can not be
+changed, thus limiting their flexibility. The flexibility can be partially
```
debarshiray, line 117: It's not just the entry point of an OCI container that can't be changed. Anything that's specified as part of `podman create` can't be changed, including the bind mounts. We don't necessarily want to change the entry point after the container is created., because it ultimately goes to sleep. What we really want to change, are all the various bits of configuration that we apply to the toolbox containers.

doc/toolbox-init-container.1.md
```diff
@@ -20,13 +13,55 @@ be used as the entry point for all toolbox containers, and must be run inside
...
+
+On some systems some important paths like `/home`, `/media` or `/mnt` are
+symbolic links to different locations. Toolbox detects such scenarios and
+matches such configurations in the container. This behaviour is controlled using
+options `--home-link`, `--media-link` and `--mnt-link`.
+
+Every time a toolbox container is started, `toolbox init-container` does the
+following:
+
+- sets `XDG_RUNTIME_DIR` if it is unset
```
debarshiray, line 33: I wonder if this is straying too far into documenting the details of the implementation. :)

It's definitely a thin line, and a matter of finding some balance between (a) what's actually useful for users of Toolbox versus contributors to Toolbox, and (b) what's meant to be a somewhat stable user-facing characteristic of a Toolbox container versus what's merely an implementation detail and can change over time.

doc/toolbox-create.1.md
```diff
@@ -4,48 +4,144 @@
...
 Assign a different NAME to the toolbox container. This is useful for creating
 multiple toolbox containers from the same base image, or for entirely
 customized containers from custom-built base images.
 
+Either use this option **or** argument <*container*> (the argument <*container*>
+is prefered).
+
+If neither the option or <*container*> are set, the name of a container is based
+on the name of the used image.
+
```
debarshiray, line 136: I have split this out into https://github.com/containers/toolbox/pull/678

I think this text can be simplified by simply not mentioning the older `--container` option. The more we mention something, the more we encourage people to actually use it. Since we consider the older option deprecated, let's just not mention it.

What we should also do is try to emit a deprecation warning from the code when someone does actually use it.

doc/toolbox-list.1.md
```diff
@@ -4,13 +4,33 @@
...
 ## SYNOPSIS
-**toolbox list** [*--containers* | *-c*] [*--images* | *-i*]
+**toolbox list** [*OPTIONS*]
 
 ## DESCRIPTION
 
 Lists existing toolbox containers and images. These are OCI containers and
 images, which can be managed directly with tools like `buildah` and `podman`.
 
+Listed information about images:
```
debarshiray, line 14: Let's wait a bit before we go too much into this part, because it will be influenced by https://github.com/containers/toolbox/issues/579

doc/toolbox-enter.1.md
```diff
@@ -32,6 +33,13 @@ Enter a toolbox container with the given NAME. This is useful when there are
 multiple toolbox containers created from the same base image, or entirely
 customized containers created from custom-built base images.
 
+Either use this option **or** argument <*container*> (the argument <*container*>
```
debarshiray, line 36: I have split this out in https://github.com/containers/toolbox/pull/681

I think this text can be simplified by simply not mentioning the older --container option. The more we mention something, the more we encourage people to actually use it. Since we consider the older option deprecated, let's just not mention it.

What we should also do is try to emit a deprecation warning from the code when someone does actually use it.

doc/toolbox-enter.1.md
```diff
@@ -4,23 +4,24 @@
...
 
 ## DESCRIPTION
 
-Spawns an interactive shell inside a toolbox container. The container should
-have been created using the `toolbox create` command. If there aren't any
-containers, `toolbox enter` will offer to create one for you. When invoked with
-the default parameters, and if there's only one container available, it will
-fall back to it, even if it doesn't match the default name.
+Spawns an interactive shell inside a toolbox container. `toolbox enter` is built
+on top of **toolbox run** by executing a shell (default is the shell used on the
```
debarshiray, line 12: The manuals are meant to be user-facing, so I don't know if we should go so far as to say that `enter` is implemented on top of `run`. It seems like implementation details.

However, it's definitely useful to mention which shell is chosen and the fallback to `/bin/bash`.

debarshiray: I split this out into https://github.com/containers/toolbox/pull/682

doc/toolbox-create.1.md
```diff
@@ -68,6 +74,9 @@ Change the NAME of the base image used to create the toolbox container. This
 is useful for creating containers from custom-built base images. Cannot be used
 used with `--release`.
 
+If NAME does not contain a domain, the image will be pulled from
+`registry.fedoraproject.org`.
```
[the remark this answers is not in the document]
HarryMichal: I was under the impression that's how the tool behaves.

debarshiray:
Fix-ups in https://github.com/containers/toolbox/pull/814

Issue #501, opened 2020-07-13
IPC namespace is different from the host


martinpitt:
toolbox does not change podman's /dev/shm size default, so that stays at 64 MB:

    $ df -h /dev/shm
    shm                      63M     0  63M   0% /dev/shm

I often run browser integration tests in toolbox, and chromium regularly crashes as it runs out of memory.

This can be controlled by `podman create --shm-size=0` (which disables the limit), or e. g.  `podman create --shm-size=512m`. In the spirit of toolbox I'd suggest to simply use `0` as a default and not make this configurable.

masch:
Hi! I don't know what is the best solution, if is better to use unlimited value or a fixed value, I think is better to support the `shm` parameter size from the command arguments because it could handle multiple scenarios.
I could do the PR if you are agree on that solution @HarryMichal

HarryMichal:
This looks alright to me. I'd go with the option to disable the limit (`--shm-size=0`). Go for it, @masch!

Still, we'll have to wait for @debarshiray to make the final call.

masch:
Great! I'll do the PR and you give me feedback on it.

debarshiray:
Interesting. I thought we had a `/dev/shm` just as big as the host's.

The history here is encoded in these two commits:

* fdc00a2778afab061bf165429c4a3f3497dd091d: Back then, `--shm-size` didn't work with rootless. So we used `--tmpfs` to mount a `tmpfs` and calculated the size manually to mimic the kernel's defaults.

* 3e7400836ff6cdfc22bedde76ea21dd504b97463: Then we switched to `--ipc host`to make Qt applications [run without QT_X11_NO_MITSHM](https://github.com/containers/toolbox/issues/163). I confess that I don't really understand exactly what that environment variable is supposed to do, so the fix was a bit of throwing random things at the wall and hoping something would stick. Using `--ipc host` did give us the host's `/dev/shm` - I am quite sure that it wasn't 64M.

So, I am wondering if something changed in Podman. Anyway, I don't mind explicitly using `--shm-size` if that's considered better.

masch:
I've created the #514 PR, let me know what do you thing.

martinpitt:
Thanks @masch! LGTM!

debarshiray:
I wonder if https://github.com/containers/podman/pull/7109 is relevant.

debarshiray:
[...]

PR #514, opened 2020-07-24
Disable shared memory limit size

masch:
The default shared memory used by podman is 64 MB, so in order to prevent regular crashes on integration tests because of that value, the limit is disabled by using the "--shm-size=0" parameter.

This change was requested on #501 issue.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -350,6 +350,7 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 		"--pid", "host",
 		"--privileged",
 		"--security-opt", "label=disable",
+		"--shm-size", "0",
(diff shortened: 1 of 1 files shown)


masch:
Hi! @HarryMichal , thank you for your recommendation. Let me fix it.

masch:
I fixed it, let me know if is OK.

debarshiray:
Thank you for your contribution, @masch but I am going to close this in favour of https://github.com/containers/podman/pull/7109 as mentioned in https://github.com/containers/toolbox/issues/501

Do you want to write a Toolbox test case to detect that the container's IPC namespace matches the host's? That would help us catch such regressions faster.

Issue #505, opened 2020-07-15
Terminal Incompatibility with Alacritty from Arch Linux Installation


mertalpt:
I am running an Arch Linux installation with the [Alacritty](https://github.com/alacritty/alacritty) terminal emulator. I installed the [toolbox package](https://www.archlinux.org/packages/community/x86_64/toolbox/) through pacman, the default Arch package manager.

toolbox itself works fine, I was able to create a container based on fedora-toolbox:30 image. I was also able to enter the container successfully. However, I have issues running some commands inside the container.

Commands such as 'cd', 'ls', 'echo' or 'exit' work fine. But, commands such as 'clear', 'top' result in the message: 

> 'alacritty': unknown terminal type.

and the command is not executed. Also tab, arrow keys and backspace do not work properly within the container. Tab seems to have no effect, while backspace acts like tab.

I also installed the gnome-terminal for a test. It worked fine without issue. But I would rather use the terminal emulator of my preference.

HarryMichal:
Hey @mertalpt! I only used Alacritty once and that's about 2 years ago, so I have zero idea of what could be happening here. My personal guess aims at a possible problem with environmental variables.

@debarshiray and @tpopela on the other hand could.

tpopela:
Hi @mertalpt @HarryMichal .. I'm running Toolbox on Alacritty (manually built from trunk at https://github.com/alacritty/alacritty/commit/f0775b3c89e92b92b74d5c9138a9770af80f589f). No problems here. Most time using tmux, but no problems even with using Alacritty without tmux.

mertalpt:
'echo $TERM' returns 'alacritty'. Any idea what information I can provide to help identify the issue?

tpopela:
@mertalpt If I force the TERM=alacritty (I use xterm-256color) then I see the same problems as you.

HarryMichal:
> Edit: Exitting via 'exit' or Ctrl+D does not kill the container either. I have to manually grep the process and kill it each time.

This is an expected behaviour. Toolbox containers keep running after you exit them. You can stop them with `podman stop <name-of-container>`.

HarryMichal:
@mertalpt, you can try to change the `TERM` variable to something else (e.g., `xterm-256color`, `linux`,..) and see if it "fixes" the issue for now. Still, we need to make this work out of the box.

HarryMichal:
[...]

PR #515, opened 2020-07-26
New function to ensure compatibility with non-standard terminals

mertalpt:
Proposed solution to issue #505 . Maintainers' input is required for some design decisions.

**Problem Summary**
Terminal capabilities are stored in [terminfo](https://linux.die.net/man/5/terminfo) databases and used by libraries such as [ncurses](https://linux.die.net/man/3/ncurses) to give terminal-independent methods of terminal controls.

Default Fedora images pulled by _toolbox_ lack the necessary _terminfo_ entries to support less common terminals that result in unusable shells inside containers. The solution is to provide the _terminfo_ entry of the current terminal described by the _TERM_ environment variable.

**Possible Approaches**

- Add _ncurses-base_ and _ncurses-term_ packages to Fedora images to provide a comprehensive _terminfo_ database that is likely to cover all terminals that can be downloaded through a package manager. It is a simple fix but other images that might be used with _toolbox_ need to include those packages to fix this issue for all possible _toolbox_ users. Also, it will not cover terminals that are not found in _ncurses_ packages. But it might be left as an user responsibility in such edge cases.
- Copy the _terminfo_ entry corresponding to user's terminal described by _TERM_ inside the container _terminfo_ database. This is the approach I took with this pull request. Read below for implementation problems.

**Implementation Problems**

- Standard _terminfo_ database is stored in a directory (/usr/share/terminfo) that requires root privileges for write access. Hence, a read lock cannot be acquired without superuser. Same holds true for every other non-privileged program and it is very unlikely a _terminfo_ entry will ever be changed while _toolbox_ is reading from it. But the possibility is there.
- In the current implementation the host file system is searched at every container entrance in case the necessary _terminfo_ entry gets updated after container creation. This is a tradeoff made in performance against the possibility of a terminal update introducing a change to the corresponding _terminfo_ entry that may cause issues. It can be made so that the host file system is searched only if any _terminfo_ entry for the current terminal is missing from the container.
- I assumed that for any container the host file system (e.g. for "/usr/share/terminfo") can always be accessed through the path "/run/host". I do not know for sure if this assumption is safe in general.

**Implementation Commentary**

1. _terminfo_ entry for the current terminal is searched in the paths with the logic specified by the "Fetching Compiled Descriptions" section of  the [terminfo](https://linux.die.net/man/5/terminfo) manual page.
2. Using "podman-exec" command, necessary directories are created and the _terminfo_ entry is copied into the container.
3. Error cases are checked and reported during the process.
4. The function is called inside _run()_ after _copy\_etc\_profile\_d\_*_ calls.

This process adds around 0.2 seconds to the execution time of "toolbox enter" in my somewhat mid-level laptop.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -40,6 +40,8 @@ environment_variables="COLORTERM \
         SHELL \
         SSH_AUTH_SOCK \
         TERM \
+        TERMINFO \
+        TERMINFO_DIRS \
         TOOLBOX_PATH \
         VTE_VERSION \
         WAYLAND_DISPLAY \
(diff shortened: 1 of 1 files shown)


debarshiray:
Let's keep the discussion in one place in https://github.com/containers/toolbox/issues/505

Either way, the POSIX shell implementation of Toolbox is deprecated. It's fine if you want to keep it updated, but any change has to go into the Go implementation.

debarshiray:
Closing, in favour of https://github.com/containers/toolbox/pull/544

PR #518, opened 2020-07-28
cmd/run: Unset detach keys when executing commands

HarryMichal:
Podman sets keys 'ctrl-p' and 'ctrl-q' as the default detach keys. This behaviour is not desirable in Toolbox. Since Podman v1.8.2[0] it is possible to have set no detach keys by passing an empty string to the --detach-keys option.

Thanks to this the 'Ctrl+p' and 'Ctrl-q' combinations are doing what they should do normally in a terminal.

Fixes: https://github.com/containers/toolbox/issues/394

This PR is a reincarnation of #396 that I mistakenly closed by badly updating the author's branch.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -269,17 +269,31 @@ func runCommand(container string,
 		}
 	}
 
+	logrus.Debug("Checking if 'podman exec' supports disabling the detach keys")
+
+	var detachKeys []string
+
+	if podman.CheckVersion("1.8.1") {
+		logrus.Debug("'podman exec' supports disabling the detach keys")
+		detachKeys = []string{"--detach-keys", ""}
+	}
+
 	envOptions := utils.GetEnvOptionsForPreservedVariables()
 	logLevelString := podman.LogLevel.String()
 
 	execArgs := []string{
 		"--log-level", logLevelString,
 		"exec",
+	}
+
+	execArgs = append(execArgs, detachKeys...)
+
+	execArgs = append(execArgs, []string{
 		"--interactive",
 		"--tty",
 		"--user", currentUser.Username,
 		"--workdir", workingDirectory,
-	}
+	}...)
 
 	execArgs = append(execArgs, envOptions...)
 


src/cmd/run.go
```diff
@@ -279,6 +279,13 @@ func runCommand(container string,
 		"--workdir", workingDirectory,
 	}
 
+	// Podman sets keys 'ctrl-p' and 'ctrl-q' as the default detach keys. This behaviour is not desirable in
+	// Toolbox. Since Podman v1.8.2 it is possible to have set no detach keys by passing an empty string to the
+	// --detach-keys option. https://github.com/containers/podman/pull/5203
+	if podman.CheckVersion("1.8.2") {
```
debarshiray, line 285: The Podman commits are such a mess. :(

The original issue was https://github.com/containers/podman/issues/4208 which led to https://github.com/containers/podman/pull/4441 However, then https://github.com/containers/podman/issues/5166 showed up which seems to say that https://github.com/containers/podman/pull/4441 didn't do what it claimed to do. So, finally we got https://github.com/containers/podman/pull/5203

Anyway, it seems to me that https://github.com/containers/podman/pull/5203 was already in 1.8.1.

toolbox
```diff
@@ -1596,6 +1596,7 @@ run()
             --tty \
             --user "$USER" \
             --workdir "$PWD" \
+            --detach-keys="" \
```
debarshiray, line 1599: Let's not touch the shell script here because doing the version check would be too cumbersome than it's actually worth. So, not changing it at all seems better than a half-hearted change.

src/cmd/run.go
debarshiray: I updated the version to 1.8.1. I hope I got it right.

toolbox

debarshiray:
Thanks for working on this, @EnigmaCurry !

src/cmd/run.go
junjieyuan: Yes, it should be 1.8.1. 
See https://github.com/containers/podman/releases/tag/v1.8.1

debarshiray: Thanks for confirming that, @junjieyuan

Issue #513, opened 2020-07-24
toolbox becomes unusably slow when creating lots of small files in /tmp


hferreiro:
I have a process that from time to time creates lots of log files on /tmp. In the order of tens or hundreds of thousands. When this happens, fuse-overlayfs uses 100% of the CPU and everything works very slowly.

Trying to remove /tmp contents is particularly slow. Restarting the container doesn't help.

A6GibKm:
Have you tested this in other folders inside the sandbox or folder in your home? As far as I know, rootless podman is very slow when it comes to file IO in general.

debarshiray:
Can you reproduce this with:
```
$ podman run --rm -it --env=TERM=$TERM registry.fedoraproject.org/f31/fedora-toolbox:31 /bin/bash
```

`/tmp` is mounted as a `fuse-overlayfs`, both in Toolbox containers and in stock Podman containers, so I'd say this needs to be moved to `fuse-overlayfs`. Although, I do wonder why Podman isn't mounting a `tmpfs` on `/tmp`.

@giuseppe ?

giuseppe:
this is definitely a fuse-overlayfs issue.  Even if we manage to make it faster, dealing with so many files will always be slow using a FUSE file system.

Would it be possible to mount a tmpfs on `/tmp`?  I think that is the best solution.

debarshiray:
> Would it be possible to mount a tmpfs on /tmp?
> I think that is the best solution.

Yes, I think so too. I was wondering if that's something Podman should do by default, or whether it's better if Toolbox explicitly did it. I am fine with either alternative.

Thanks for taking a look, @giuseppe

giuseppe:
> > 
> 
> Yes, I think so too. I was wondering if that's something Podman should do by default, or whether it's better if Toolbox explicitly did it. I am fine with either alternative.

I'd also prefer we had a tmpfs by default, but I think we are just following what Docker does.

hferreiro:
Yes, I confirm this is an issue with podman too. What's interesting is what's keeping fuse-overlayfs busy once the files are created. I'm trying to reproduce the issue again to find out about that.

debarshiray:
> Yes, I confirm this is an issue with podman too. What's interesting
> is what's keeping fuse-overlayfs busy once the files are created. I'm
> trying to reproduce the issue again to find out about that.

[...]

PR #524, opened 2020-08-04
cmd/initContainer: Mount a tmpfs at /tmp to match the host

debarshiray:
Podman doesn't mount a tmpfs at /tmp by default - it needs to be
separately requested. However, doing it as part of 'podman create ...'
won't add a tmpfs at /tmp for existing toolbox containers. Therefore,
it's best done as part of the entry point.

The mount options are the same as used by systemd (see tmp.mount) to
provide a tmpfs at the host's /tmp.

For what it's worth, the mount flags do differ slightly from the host.

The host has:
$ findmnt --output OPTIONS,PROPAGATION /tmp
OPTIONS                  PROPAGATION
rw,nosuid,nodev,seclabel shared

The container has:
$ findmnt --output OPTIONS,PROPAGATION /tmp
OPTIONS                                        PROPAGATION
rw,nosuid,nodev,seclabel,uid=100000,gid=100000 private

This difference was ignored because it doesn't appear to cause any real
problem.

https://github.com/containers/toolbox/issues/513

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -142,6 +142,16 @@ func initContainer(cmd *cobra.Command, args []string) error {
 
 	defer toolboxEnvFile.Close()
 
+	logrus.Debug("Mounting tmpfs at /tmp")
+
+	if err := syscall.Mount("tmpfs",
+		"/tmp",
+		"tmpfs",
+		syscall.MS_NODEV|syscall.MS_STRICTATIME|syscall.MS_NOSUID,
+		"mode=1777"); err != nil {
+		return fmt.Errorf("failed to mount tmpfs at /tmp: %s", err)
+	}
+
 	if initContainerFlags.monitorHost {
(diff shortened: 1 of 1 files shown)


debarshiray:
@giuseppe : Any idea if those `uid=100000,gid=100000` options in the mount flags are a problem? It seems like some behaviour associated with the presence of mount and/or user namespaces. However, they don't seem to cause any problem.

debarshiray:
> @giuseppe : Any idea if those `uid=100000,gid=100000` options in the mount flags are a problem? It seems like some behaviour associated with the presence of mount and/or user namespaces. However, they don't seem to cause any problem.

I learnt from @giuseppe on IRC (in `#podman` on Freenode) that the `uid` and `gid` options are just not shown when they are 0, which is what they are on the host.

I am still a bit surprised because I expected running `findmnt` inside the container to think that `uid` and `gid` are 0, because of the container's user namespace.

PR #533, opened 2020-08-20
initContainer: Only remove password when needed

### LINKED ISSUE #523: sudo(8) inside toolbox containers asks for a password with Podman 2.0.5
When using toolbox on F33 Silverblue rawhide, entering a newly created toolbox results in an error as follows ...

    /usr/bin/id: cannot find name for group ID 1000

Inside of toolbox, when issuing a command sudoer privileges are required for, such as dnf like so ...
sudo dnf install vim-enhanced terminator results in the following prompt to the user ...

    We trust you have received the usual lecture from the local System
    Administrator. It usually boils down to these three things:
    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

then, if I enter the user password the result is password attempt failure ...

    [sudo] password for ssnow:
    Sorry, try again.
    [sudo] password for ssnow:
    Sorry, try again.

This issue was originally reported at https://discussion.fedoraproject.org/t/toolbox-and-root/22123/29 and I originally thought the users system must be broken in some way. That was until I installed a fresh rawhide version of Silverblue.
IMO, it may relate to the first message I get when entering the toolbox container. Perhaps not mapping the user as the root of the container.

HarryMichal:
When a newly created container has already the current user created
inside of it then the user initialization section of command
'init-container' is not triggered. The user init section currently takes
care of:

  1) symlinking /home to /var/home if called with option --home-link

  2) creating user set by --user, --uid, --home and --shell

  3) adds the user to the sudoers group (either sudo or wheel)

  4) removes password of user and of root

This commit does the following:

  - moves 1) before the user init section and makes calling it
    conditional depending on /home being a symlink or not
  - moves 4) after the user init section and depending on the output of
    'passwd --status' (that is expected to be NP; more in 'man
    passwd(1)') calls the said sections

Partially fixes #523

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -348,6 +348,8 @@ func initContainerHelp(cmd *cobra.Command, args []string) {
 func configureUsers(targetUserUid int,
 	targetUser, targetUserHome, targetUserShell string,
 	homeLink, targetUserExists bool) error {
+	var stdout strings.Builder
+
 	if homeLink {
 		if err := redirectPath("/home", "/var/home", true); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


debarshiray (commented):
Oops! Sorry, I failed to notice this PR. :(

Is it still relevant after https://github.com/containers/toolbox/pull/547 ?

HarryMichal:
> Oops! Sorry, I failed to notice this PR. :(
> 
> Is it still relevant after #547 ?

I'll have to adjust it but I'd say "yes". This PR does not touch `useradd` or `usermod`. I should also split it into several commits because now I see I tried to achieve several different things with this.

HarryMichal:
In the end, I only took one part from the original PR: making password removal conditional to prevent log cluttering. What do you think about it, @debarshiray?

debarshiray (commented):
Umm... I am a bit torn about this.

Only deleting the password when some spurious password is set seems like a reasonable thing to do, but it happens only once at container start-up so we don't actually save much. Moreover, the `fork` and `exec` to run the `passwd` process, which is relatively costly, still happens and in the worst case will happen twice.

So, not sure. Seems like more lines of code for no real benefit. :)

HarryMichal:
Fair enough. Let's close this.

Issue #529, opened 2020-08-14
Toolbox at Rawhide (Fedora 33) doesn't run other releases


juanje:
**Describe the bug**

Usually, you can run another Fedora release with toolbox by doing:
```
# On Fedora 32
$ toolbox create --release 31
Created container: fedora-toolbox-31
Enter with: toolbox enter --release 31
$ toolbox enter --release 31
```
But at **Fedora Rawhide** you get the following error:
```
# On Fedora 33
$ toolbox create --release 32
Created container: fedora-toolbox-32
Enter with: toolbox enter --release 32
$ toolbox enter --release 32
Error: invalid entry point PID of container fedora-toolbox-32
```
The full debug output:
```
$ toolbox -v enter -r 32
DEBU Running as real user ID 1000                 
DEBU Resolved absolute path to the executable as /usr/local/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU Checking if /etc/subgid and /etc/subuid have entries for user vagrant 
DEBU TOOLBOX_PATH is /usr/local/bin/toolbox       
DEBU Toolbox config directory is /home/vagrant/.config/toolbox 
DEBU Current Podman version is 2.1.0-dev          
DEBU Old Podman version is 2.1.0-dev              
DEBU Migration not needed: Podman version 2.1.0-dev is unchanged 
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Image: ''                                    
DEBU Release: '32'                                
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-32'               
DEBU Image: 'fedora-toolbox:32'                   
DEBU Release: '32'                                
DEBU Checking if container fedora-toolbox-32 exists 
DEBU Calling org.freedesktop.Flatpak.SessionHelper.RequestSession 
DEBU Starting container fedora-toolbox-32         
DEBU Inspecting entry point of container fedora-toolbox-32 
DEBU Entry point PID is a float64                 
DEBU Entry point of container fedora-toolbox-32 is toolbox (PID=0) 
Error: invalid entry point PID of container fedora-toolbox-32
```

**Steps how to reproduce the behaviour**

1. Enter into a Fedora Rawhide/33 system.
2. Compile and install the current Toolbox versions from the source.
3. Create a new container from the previoues release (32) with:
```
$ toolbox create --release 32
```
4. Try to enter at that container by typing:
```
$ toolbox enter --release 32
```
5. It fails.

**Expected behaviour**
[...]

PR #534, opened 2020-08-21
build: Ensure binaries built on Fedora 33 run on Fedoras 32 & 31

debarshiray:
The /usr/bin/toolbox binary is not only used to interact with toolbox
containers and images from the host. It's also used as the entry point
of the containers by bind mounting the binary from the host into the
container. This means that the /usr/bin/toolbox binary on the host must
also work inside the container, even if they have different operating
systems.

In the past, this worked perfectly well with the POSIX shell
implementation because it got intepreted by whichever /bin/sh was
available.

The Go implementation also mostly worked so far because it's largely
statically linked, with the notable exception of the standard C
library. However, recently glibc-2.32, which is used by Fedora 33
onwards, added a new version of the pthread_sigmask symbol [1] as part
of the libpthread removal project.

This means that /usr/bin/toolbox binaries built against glibc-2.32 on
newer Fedoras pick up the latest version of the symbol and fail to run
against older glibcs in older Fedoras.

One way to fix this is to disable the use of any C code from Go by
using the CGO_ENABLED environment variable [2]. However, this can
negatively impact packages like "os/user" [3] and "net" [4], where the
more featureful glibc APIs will be replaced by more limited
equivalents written only in Go.

Instead, since glibc uses symbol versioning, it's better to tell the
Go toolchain to avoid linking against any symbols from glibc-2.32.

This was accomplished by a few linker tricks:

  * The GNU ld linker's --wrap flag was used when building the Go code
    to divert pthread_sigmask invocations from Go to another function
    called __wrap_pthread_sigmask.

  * A static library was added to provide this __wrap_pthread_sigmask
    function, which forwards calls to the actual pthread_sigmask API in
    glibc. This library itself was not linked with --wrap, and
    specifies the latest permissible version of the pthread_sigmask
    symbol from glibc for each architecture. Currently, the list of
    architectures covers the ones that Fedora builds for.

  * The Go cmd/link linker was switched to external mode [5]. This
    ensures that the final object file containing all the Go code gets
    linked to the standard C library and the wrapper static library by
    the GNU ld linker for the --wrap flag to kick in.

Based on ideas from Ondřej Míchal.

[1] glibc commit c6663fee4340291c
    https://sourceware.org/git/?p=glibc.git;a=commit;h=c6663fee4340291c

[2] https://golang.org/cmd/cgo/

[3] https://golang.org/pkg/os/user/

[4] https://golang.org/pkg/net/

[5] https://golang.org/src/cmd/cgo/doc.go

https://github.com/containers/toolbox/issues/529


debarshiray:
Here's a scratch build for Fedora 34: https://koji.fedoraproject.org/koji/taskinfo?taskID=49760645

I extracted the `/usr/bin/toolbox` binary from the Fedora 34 build (against glibc `master`) and was able to use it on a Fedora 31 host to create and run containers. I also built the code on Fedora 31 (against `glibc-2.31`) and used it on the same host to create and run containers.

PR #536, opened 2020-08-24
libc-wrappers: Tighten the definition of the ppc64le definition

debarshiray:
This is based on the output of 'gcc -dM -E - </dev/null' on a ppc64le
system. For what it's worth, the _CALL_ELF macro is defined as 1 on
the big endian variants of the architecture.

The original intention in commit 6ad9c631806961f3 was to support the
architectures that Fedora builds for, and it doesn't care about PowerPC
variants that aren't ppc64le [1].

[1] https://fedoraproject.org/wiki/Architectures/PowerPC


debarshiray:
Tested with a Fedora scratch build:
https://koji.fedoraproject.org/koji/taskinfo?taskID=50079103

Issue #114, opened 2019-04-13
Containers remain running after exiting


Exalm:
In particular, this means it's impossible to remove a toolbox-created container without first stopping/killing it with `podman`:

```
 ~/toolbox  ./toolbox create -c test
Created container: test
Enter with: toolbox enter --container test
 ~/toolbox  ./toolbox enter -c test 
🔹[exalm@toolbox toolbox]$ logout
 ~/toolbox  ./toolbox rm test
toolbox: failed to remove container test
 ~/toolbox  podman ps
CONTAINER ID  IMAGE                              COMMAND     CREATED         STATUS             PORTS  NAMES
1a08f09ea797  localhost/fedora-toolbox-exalm:30  sleep +Inf  16 seconds ago  Up 10 seconds ago         test
 ~/toolbox  podman stop test
1a08f09ea79710801859bea8dc6a5a85d2031ce1a73dd7d284c3e1fa51a67be0
 ~/toolbox  ./toolbox rm test
 ~/toolbox  
```

debarshiray:
`toolbox rm --force` should also work.

But yes, I'd like to make this properly reference counted, but sadly, I don't know of a way to implement that using the existing Podman command line interface.

imciner2:
Using `toolbox rm --force` is not able to delete the container when it is running on my machine.

Perhaps what would could be done is to detect if the container is running when deleting it and then give the user a prompt such as `This toolbox is currently running, are you sure you wish to delete it [y/N]:` Then call `podman stop` before calling the delete command if they choose to continue.

paul8046:
I am unable to delete a container created by toolbox even after using `toolbox rm --force` and stopping it with podman. I am having to reboot and then `toolbox rm` works.

debarshiray:
> I am unable to delete a container created by toolbox even after using
> `toolbox rm --force` and stopping it with podman. I am having to reboot
> and then `toolbox rm` works.

It will fail if you have currently active `toolbox enter` sessions. We need to improve the error handling there.

Otherwise, if that's not the case and you can reproduce at will, then I suggest trying `podman rm --force <container>` to delete the container. If that also fails, then we might have a Podman bug. In any case, let's use a different issue to discuss this.

Thanks for stopping by!

bam80:
Considering it was reported almost 1.5 years ago, I'm wondering if there was any progress since then.

bam80:
Seems I can't stop the containers left after Toolbox:
```
[...]

PR #541, opened 2020-08-26
cmd/initContainer, cmd/run: Stop container once the last session dies

debarshiray:
Currently, once a toolbox container gets started with 'podman start',
as part of the 'toolbox enter' command, it doesn't stop unless the
host is shut down or someone explicitly calls 'podman stop'. This
becomes annoying if someone tries to remove the container because
commands like 'podman rm' and such don't work without the '--force'
flag, even if all active 'toolbox enter' and 'toolbox run' sessions
have terminated.

A system of reference counting based on advisory file locks has been
used to automatically terminate the container's entry point once all
the active sessions have died.

The 'toolbox enter' and 'toolbox run' sessions acquire shared file
locks, and the container's entry point blocks trying to acquire an
exclusive lock on a common file. The entry point will be unblocked once
all shared locks have been released by the active sessions, and then
it terminates.

Once the container has been started, and the entry point has finished
setting it up, the entry point waits for a while before trying to
acquire its exclusive lock. This is meant to give some time to the
first session to go ahead and acquire its shared lock. A duration of 25
seconds, the same interval as the default for D-Bus method calls, was
chosen for this.


debarshiray:
Note that you should `podman stop` all your containers before trying this out.

HarryMichal:
Maybe I did something wrong, but I couldn't trigger the behavior when I tried this PR. I'll try again this week to make sure we can merge this.

debarshiray:
Note that I am not sure that I got the synchronization right here. There might be loopholes lurking between the entry point starting up and then going away.

debarshiray (commented):
This won't work if a new `enter` or `run` command is invoked, just as the last `podman exec` session terminates to release the last shared lock and the entry point acquires the exclusive lock.

First, there will be a race between the new command invoking `podman start` and the entry point exiting while holding the exclusive lock.  If the new command wins and `podman start` is invoked first, then it won't spawn a new instance of the entry point, and the existing instance will continue to exit.

Second, if the entry point removes the initialization stamp file for its process ID when exiting, then there will be a race between the new command detecting the file and the entry point removing it.  If the new command loses, then it won't be able to distinguish it from a crashed entry point.  The whole sequence can repeat when the new command times out waiting for the file to appear and tries to invoke `podman start` again, because yet another `enter` or `run` might have been invoked that exited just as the command was about try again.

If the entry point doesn't remove the initialization stamp file when exiting, then the new command will be tricked into invoking `podman exec` against a stopped or about to stop container.

PR #546, opened 2020-08-30
Update fallback fedora version from EOL 30

LeonardKoenig:
fedora 30 went EOL this may, thus it makes sense to update the default fallback image to fedora 32.  This only affects distributions w/o a vendor-supplied toolbox image, e.g. ArchLinux.

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -40,7 +40,7 @@ import (
 
 const (
 	idTruncLength          = 12
-	releaseDefaultFallback = "30"
+	releaseDefaultFallback = "31"
 )
 
 const (


src/pkg/utils/utils.go
```diff
@@ -40,7 +40,7 @@ import (
 
 const (
 	idTruncLength          = 12
-	releaseDefaultFallback = "30"
+	releaseDefaultFallback = "32"
```
debarshiray, line 43: Ever since the default release for non-Fedora hosts was added in commit ca9ca9a3ded056408eb4934a6905b2488152b166, we have always pointed it to the oldest supported Fedora release.

I don't think there's any real reason for this. It's just what @andrunko went with, and after that @juhp also stuck with the same convention. So, unless there's a strong reason to change it, let's stick with the pattern.

debarshiray (commented):
Thanks for the patch, @LeonardKoenig !

src/pkg/utils/utils.go
debarshiray: I took the liberty to tweak this, and merged.

debarshiray:
Thanks a lot for your contribution!

src/pkg/utils/utils.go
LeonardKoenig: Very much fine with this, I simply didn't know about that policy.  Thank you for the fast work on this!

juhp:
How about we switch the fallback to the latest stable version about 3 months after its release?

eg If someone installs an f31 toolbox just before it is EOL (like a couple of months from now) it doesn't seem such a good UX.

HarryMichal:
I agree with @juhp.

debarshiray:
Yes, sure. Sounds like a good idea!

LeonardKoenig:
Another thing I've wondered about is how 'transitioning' is done, i.e. say I have a  f30 toolbox/container with my favorite editor etc. installed and now the default switches to f31, thus `toolbox enter` not getting me to the image I'm used to.  Is there some kind of migration mechanism planned?

In any case, this could confuse users if they suddenly are brought to a different toolbox/container than they're used to.

HarryMichal:
> Another thing I've wondered about is how 'transitioning' is done, i.e. say I have a  f30 toolbox/container with my favorite editor etc. installed and now the default switches to f31, thus `toolbox enter` not getting me to the image I'm used to.  Is there some kind of migration mechanism planned?
> 
> In any case, this could confuse users if they suddenly are brought to a different toolbox/container than they're used to.

Currently there's no mechanism that would handle this situation.

LeonardKoenig:
> How about we switch the fallback to the latest stable version about 3 months after its release?
> 
> eg If someone installs an f31 toolbox just before it is EOL (like a couple of months from now) it doesn't seem such a good UX.

Sorry to necrobump, but looking back, in that case we should've switched to **32** (because this PR was opened 3 months after 2020-04). I didn't really check the release date back then, so I missed that.  And as Fedora 31 is EOL in 7 days, I think switching to 32 should happen quite soon -- shall I make a PR or would you just change that?

I mark my calendar for 2021-01-27 to ping/PR for an update to 33 :)

Issue #145, opened 2019-04-30
Provide shims for well-known binaries, which can only run on the host, inside the toolbox containers


debarshiray:
Various commands like `flatpak`, `podman`, `rpm-ostree` etc. can't really be used inside an OCI container. Some of them, like `podman`, might have some limited use, but, by and large, people expect to run them directly on the host.

At the same time, we expect a good percentage of users of the command line interface to spend most of their time inside a toolbox container, and one of the goals of the Toolbox project is to reduce the cognitive burden of using mutable containers on locked-down and immutable host OSes like Silverblue. Commands like `flatpak` and `rpm-ostree` are important tools for interfacing with such OSes.

Therefore, it would be nice if we could provide a user experience that's better than having to switch back and forth between a host and toolbox shell, or having to prefix every command with things like `flatpak-spawn --host`.

The easiest option is to install aliases in the shell running inside the container. However, it doesn't give us things like manuals and shell completion.

Another option is to pre-install the `flatpak`, `podman`, `rpm-ostree`, etc. RPMs in the `fedora-toolbox` base images but remove all the code, leaving behind only the manuals and shell completion. Then the `toolbox` command can place the corresponding shims via bind mounts to forward the calls to the host when starting the toolbox containers.

This will ensure that the `fedora-toolbox` images don't get bigger with useless Go binaries, and the wrappers can be kept updated through the `toolbox` package on the host.

However, I don't know what will happen if one of these packages are updated inside the containers. Would they interfere with the bind mounted shims? It would be nice if we could trim the useless bits from those packages during or after updating the RPM.

cgwalters:
I think it'll be dramatically simpler if we add `/usr/libexec/toolbox/bin` and change the shell startup scripts to inject that into `$PATH` first.

mildred:
Running podman inside the toolbox can be immensely useful. I work with many projects where build scripts is just calling docker images, and being able to run those scripts inside the toolbox can be really useful. There might create hard to detect issues with bind mounts outside the home directory that would reflect the silverblue host and not the toolbox though.

kamalmarhubi:
I'll add `docker-compose` to the list of things useful for dev flows. It has a really low barrier to entry for anything needing a data store or two, so a fair number of web app type things use it—open source and not.

dustymabe:
[...]

PR #553, opened 2020-09-07
WIP: Add binary shims

HarryMichal:
This PR adds support for creating binary shims in toolbox containers for binaries available on the host. This was heavily inspired by code snippets in https://github.com/containers/toolbox/issues/145. There's also very basic support for running the shim binaries with `sudo`.

Closes https://github.com/containers/toolbox/issues/145

### DIFF
--- a/data/libexec/host-runner
+++ b/data/libexec/host-runner
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+#
+# Copyright © 2019 – 2020 Red Hat Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+executable=$(basename $0)
+priviledged=false
+
+if [ $# -ge 1 ]; then
+	case "$1" in
+		--host-runner-sudo )
+			shift
+			priviledged=true
+	esac
+fi
+
+if $priviledged; then
+	executable="sudo $executable"
+fi
+
+exec flatpak-spawn --host --watch-bus --forward-fd=1 --forward-fd=2 --directory=$(pwd) --env=TERM=xterm-256color $executable "$@"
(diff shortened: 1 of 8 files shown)


cjao:
Using the technique of your sudo wrapper, I tried `flatpak-spawn --host --watch-bus --forward-fd=1 --forward-fd=2 --env=TERM=xterm-256color sudo ls /` but got

`sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper`

1player:
Is there any update to this PR?

debarshiray:
Let's see if I can rebase this pull request.

data/libexec/host-runner
```diff
@@ -0,0 +1,34 @@
...
+			shift
+			priviledged=true
+	esac
+fi
+
+if $priviledged; then
+	executable="sudo $executable"
+fi
+
+exec flatpak-spawn --host --watch-bus --forward-fd=1 --forward-fd=2 --directory=$(pwd) --env=TERM=xterm-256color $executable "$@"
```
debarshiray, line 34: Do we really need to explicitly forward file descriptors 1 and 2?  What about 0?

`flatpak-spawn(1)` has always forwarded file descriptors 0, 1 and 2.  We don't do it explicitly elsewhere either.

debarshiray, line 34: We shouldn't hard code `TERM`.  Instead: `--env=TERM=$TERM`  :)

Also, shouldn't we forward all the other environment variables that we usually do?

profile.d/toolbox.sh
```diff
@@ -1,5 +1,9 @@
 # shellcheck shell=sh
 
+# Toolbox provides shim binaries for executing commands on the host from inside
+# of a toolbox container
+export PATH="/usr/libexec/toolbox:$PATH"
```
debarshiray, line 5: Should we really do this unconditionally for both hosts and containers?  I think this should ideally happen only for Toolbx containers, not even generic Podman containers.  Or did I misunderstand something?

data/libexec/host-runner
debarshiray, line 34: I don't think we need the `--directory=$(pwd)` because `flatpak-spawn(1)` has been retaining the current working directory since [before](https://github.com/flatpak/flatpak-xdg-utils/commit/c2266e2bce9d8f72e3f705e04719c7d9a924fd87) the `--directory` option became a thing.

debarshiray, line 34: I am curious about the use of `--watch-bus`.  It's not wrong, but given that `flatpak-spawn(1)` always forwards signals, I am curious if you came across a scenario where the container-side shim would cleanly exit without signals leaving the host-side running.  I am trying to understand if we should use `--watch-bus` more widely.

Issue #183, opened 2019-05-31
Isolated development environments to fend against bugs and malware in code under development


markstos:
One of the reasons to containerize data is to prevent malicious executable inside the container from getting access to my data. It appears that toolbox always mounts the host home directory in the container. That's convenient, but I would prefer if it was optional.

Not all containers need access to my SSH private key, keychain and other things that might be stored there.

juhp:
You want a persistent user container without your /home?  Just trying to understand the requirement/usecase better. - Say compared to just running a standard throwaway Fedora container?

markstos:
Consider a Unix account which is used for both home and work. For a work container, I may prefer to mount only my work directory into the the container. 

Consider an untrusted GUI app. It may only need access to a specific folder to load/save files. It doesn't need access to my .ssh folder and other unrelated files.

For improve security, Chrome OS does not share folders to containers by default from the host. There, if you want to provide data to the containers from your home directory or from a USB drive, you explicitly share it.

Containerization as a security feature loses significant value if you give the untrusted container all your data by default!

markstos:
@juhp I see you work a good deal with Haskell packages.  Presume that one of the Haskell packages  you are install remotely has been compromised. Do Haskell packages need access to your .ssh folder or your Bitcoin wallet? Why share everything into your Haskell dev environment by default?

Fairly recently there was an NPM module that could still local Bitcoin if it was installed:
https://www.trendmicro.com/vinfo/nz/security/news/cybercrime-and-digital-threats/hacker-infects-node-js-package-to-steal-from-bitcoin-wallets

The module was popular and in fact my team had installed it into our tree. We didn't meet the the other criteria for it to trigger a bitcoin theft, but that risk would have been completely mitigated if our dev environment did not have complete access to our home directories by default.  No employee would have chosen to share their Bitcoin wallet into this development environment.

cgwalters:
This is totally valid but is going to be hard.

Practically speaking, I think the best path is to make it easy to run toolboxes as *separate users* (transiently create new uids, each with their own `/home`) - requires a privileged system service to do that.

cgwalters:
Now if you want share *some files* e.g. read-only - that gets messier.

juhp:
Is it possible to exclude files/dirs from a directory bind-mount?

markstos:
[...]

Issue #467, opened 2020-06-11
toolbox failing when there is a container running on another terminal with different $HOME


yilkalargaw:
I sometimes run toolbox in a specific directory to try out things without touching the home directory. To do this I usually change $HOME. When ever I do this and try to open my old toolboxes in another terminal window, toolbox fails with ```toolbox: failed to create /run/.toolboxenv in container dnfy-fedora```.

I think it might have something todo with fuse-overlays because restarting the containers would not help.

How to recreate

```bash
mkdir tmptrial
$HOME=$HOME/tmptrial
toolbox create -c tmp_container
toolbox enter -c tmp_container

# now on a new terminal

cd ~ / toolbox enter -c good_old_container 
```

HarryMichal:
Hi @yilkalargaw! I have a suspicion that this is a problem of Podman and not of Toolbox. I'll try to investigate a bit later on.

juanje:
I saw this issue and I tried to reproduce it. I kinda did.
I couldn't enter into the container after the change of the `$HOME`, but it also happened without other toolbox container running.
Also, I end up with a container in where I can't enter, even from another terminal with the proper `$HOME` value.

Here are my `podman` and `toolbox` versions:
* *podman:* `2.0.2`
* *toolbox:* `0.0.92`

The OS is `Fedora 32.20200712.0 (Silverblue)`.

Steps to reproduce the error:
* **Create a new toolbox's container** (and try it). This is not necessary, it fails with the default container as well, but makes the test cleaner:
```
[juanje@xps ~]$ toolbox create --container poc
Created container: poc
Enter with: toolbox enter --container poc
[juanje@xps ~]$ toolbox enter --container poc
⬢[juanje@toolbox ~]$ 
```

* Create a directory and set the `$HOME` value:
```
[juanje@xps ~]$ mkdir tmphome
[juanje@xps ~]$ cd tmphome
[juanje@xps tmphome]$ export HOME=$PWD
[juanje@xps ~]$ 
```

* Try to enter the `poc` container from that terminal:
```
[juanje@xps ~]$ toolbox enter --container poc
No toolbox containers found. Create now? [y/N] 
A container can be created later with the 'create' command.
Run 'toolbox --help' for usage.
```

NOTE: I pressed `enter`, so it didn't create a new container. It should exist.

* Now nor `podman` nor `toolbox` can see the previously created containers, **at that terminal**:
```
[juanje@xps ~]$ toolbox list
[...]

PR #561, opened 2020-09-20
Added 'boxhome' argument to toolbox create command

onuruluag:
Relates to issues #183, #348, #467
Added optional boxhome argument to create command.
Boxhome is concatenated with homedir in order to be under host's homedir.
Homedir is still mounted. No worries.
Being under host's home enables easy file sharing, and prevents privilege issues.
When boxhome argument is not given, system works as before (uses user's homedir)

User inside the toolbox would have different home path than user at host.
The box home folder is under the host home folder because of privileges.
This enables isolation between host and toolbox.
Host user's home dir is still mounted and does not effect file sharing.
Files created in the container is isolated thus can be easily deleted by user.
Several toolboxes can share same home by supplying the same home path.

Examples:
- Without specifying a home path
```
./toolbox create --container=hometest --image fedora-toolbox:latest && ./toolbox enter hometest       

Created container: hometest
Enter with: toolbox enter hometest

⬢[onurulu@toolbox src]$ echo $HOME
/home/onurulu/
```
- With specifying a home path
```
podman stop hometest && ./toolbox rm hometest

./toolbox create --container=hometest --image fedora-toolbox:latest --boxhome boxhome && ./toolbox enter hometest

Created container: hometest
Enter with: toolbox enter hometest


Welcome to the Toolbox; a container where you can install and run
all your tools.

 - Use DNF in the usual manner to install command line tools.
 - To create a new tools container, run 'toolbox create'.

For more information, see the documentation.

⬢[onurulu@toolbox src]$ echo $HOME
/home/onurulu/boxhome
```

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -46,6 +46,7 @@ var (
 		distro    string
 		image     string
 		release   string
+		boxhome   string
 	}
 
 	createToolboxShMounts = []struct {
@@ -89,6 +90,12 @@ func init() {
 		"r",
 		"",
 		"Create a toolbox container for a different operating system release than the host")
+
+	flags.StringVarP(&createFlags.boxhome,
+		"boxhome",
+		"b",
+		"",
+		"Create a toolbox container with a different home path than the host.")
 
 	createCmd.SetHelpFunc(createHelp)
 	rootCmd.AddCommand(createCmd)
@@ -381,6 +388,7 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 		"init-container",
 		"--gid", currentUser.Gid,
 		"--home", currentUser.HomeDir,
+		"--boxhome", createFlags.boxhome,
 		"--shell", userShell,
 		"--uid", currentUser.Uid,
 		"--user", currentUser.Username,
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -38,6 +38,7 @@ var (
 	initContainerFlags struct {
 		gid         int
 		home        string
+		boxhome     string
 		homeLink    bool
 		mediaLink   bool
 		mntLink     bool
@@ -86,6 +87,12 @@ func init() {
 		"",
 		"Create a user inside the toolbox container whose login directory is HOME")
 	initContainerCmd.MarkFlagRequired("home")
+
+	fmt.Println(initContainerFlags.boxhome)
+	flags.StringVar(&initContainerFlags.boxhome,
+		"boxhome",
+		"",
+		"Create a user inside the toolbox container whose login directory is not HOME.")
 
 	flags.BoolVar(&initContainerFlags.homeLink,
 		"home-link",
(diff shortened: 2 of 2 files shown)


markstos:
@onuruluag Could you check out the build failures?

onuruluag:
> @onuruluag Could you check out the build failures?

@markstos I finally got some time to check. I rebased and re-run the tests. The only failing test is Test-21. As I follow it is a known bug?
```
not ok 21 rm: Try to remove a non-existent container
# (from function `assert_output' in file test/system/libs/bats-assert/src/assert.bash, line 255,
#  in test file test/system/104-rm.bats, line 21)
#   `assert_output "Error: failed to inspect container $container_name"' failed
# 
# -- output differs --
# expected (1 lines):
#   Error: failed to inspect container nonexistentcontainer
# actual (2 lines):
#   
#   Error: failed to inspect container nonexistentcontainer
# --
# 
```
All other tests are passing.

markstos:
Why does this PR have 113 commits? Perhaps it needs to be rebased on to master.

> Homedir is still mounted. No worries.

No worries? If the homedir is still mounted into containers, then malicious containers can still steal my SSH private keys, my crypto wallet and every private document I keep in my home directory.

A "safe by default" design is to share no private data with untrusted containers, and opt-in to what kinds of data you want to share with them.

onuruluag:
> Why does this PR have 113 commits? Perhaps it needs to be rebased on to master.
Because I'm a newbie =). This is my 1st PR.
I tried to rebase. Here is what I did:
`
git remote add upstream https://github.com/containers/toolbox.git
git checkout master
git pull
git checkout boxhome
git rebase master
`
I will try to fix that.

> No worries? If the homedir is still mounted into containers, then malicious containers can still steal my SSH private keys, my crypto wallet and every private document I keep in my home directory.

Yes, you are correct. Host system's home directory is still mounted. But toolbox sees $HOME at a different path. Of course a basic _find_ operation may list all the private files.

> A "safe by default" design is to share no private data with untrusted containers, and opt-in to what kinds of data you want to share with them.

As I understand this behavior is by design in _toolbox_. Because _toolbox_ homepage states this:

> The toolbox environment is based on an OCI image. On Fedora this is the fedora-toolbox image. This image is used to create a toolbox container that seamlessly integrates with the rest of the operating system by providing access to the user's home directory, the Wayland and X11 sockets, SSH agent, etc..

I was not sure to change this behavior.

@markstos Can you also comment about the failing test?

markstos:
Sorry, I don't have an idea about the test failure.

abitrolly:
@onuruluag if original $HOME still can be accessed, then #183 is not addressed and the commit message needs to be edited.

onuruluag:
@abitrolly You are right, this commit does not fix #183. My intention was to relate this commit with that issue because files such as `$HOME/.ssh/id_ecdsa` cannot be accessed that way. So, I thought there is some kind of relation. I may remove this link from the commit message if things work like that.

debarshiray:
As I have explained repeatedly in https://github.com/containers/toolbox/issues/183 and elsewhere, this isn't part of our current to-do list.

Closing.

PR #564, opened 2020-09-22
Fix running of GUI apps from toolboxes

### LINKED ISSUE #562: F33 Silverblue: graphical apps no longer work in toolbox containers
**Describe the bug**

In Fedora 33, graphical applications no longer work from within toolbox containers when in Wayland. This breaks workflows for everyone using Silverblue and is relying on toolbox where graphical applications are needed. (Mainly: development work which may require graphical tools like gitk, VS Code, and so on.)

0. Use Wayland, not X
1. Start a toolbox container
2. Run a graphical app, such as gitk
3. See an error where it cannot connect to X

**Expected behaviour**
Graphical apps should show up.

**Actual behaviour**

When running gitk from within toolbox, the following happens:

```
$ gitk
application-specific initialization failed: couldn't connect to display ":0"
Error in startup script: couldn't connect to display ":0"
    while executing
"load /usr/lib64/libtk8.6.so Tk"
    ("package ifneeded Tk 8.6.10" script)
    invoked from within
"package require Tk"
    (file "/usr/bin/gitk" line 10)
```

toolbox version 0.0.95

toolbox-0.0.95-1.fc33.x86_64

```
Version:      2.1.0-dev
API Version:  1
Go Version:   go1.15
Built:        Tue Aug 25 14:48:48 2020
OS/Arch:      linux/amd64
```

podman-2.1.0-0.179.dev.git43f2771.fc33.x86_64

**Info about your OS**
Fedora Silverblue 33 beta

**Additional context**

This worked fine in the 2020-09-14 build of Silverblue 33 beta. It broke somewhere between then and the latest build of 2020-09-20. I believe it is podman related, especially due to it being a dev build. 

Did a development version of podman slip into F33/SB, breaking toolbox, again?

..Or is it intended to use podman 2.1.x in Fedora 33 (and therefore toolbox would have to workaround or add support for the new version)?

Discussed in https://discussion.fedoraproject.org/t/issues-with-some-graphical-apps-in-f33-containers-toolbox/23456.

HarryMichal:
GNOME Shell 3.38 stopped using an abstract unix socket for communication with X for XWayland. This change causes GUI apps fail to start when ran from a toolbox. To fix this, the unix socket available in a container under `/run/host/tmp` needs to be available under `/tmp`.

Fixes #562 

PS: I'm marking this as WIP until the final solution is found. This seems to work but I'm wondering if we could simply mount `/run/host/tmp` to `/tmp` as a whole instead of cherry-picking what goes in or not.

PPS: I'm not entirely sure which of the symlinked paths are really necessary. I'd appreciate any insight on this because my knowledge of X is very limited.

PPPS: This also relies on some other patches to work/look properly so those will probably need to be merged first.

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -33,6 +33,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type initContainerMount struct {
+	containerPath string
+	source        string
+	flags         string
+}
+
+type initContainerSymlink struct {
+	containerPath string
+	source        string
+	folder        bool
+}
+
 var (
 	initContainerFlags struct {
 		home        string
@@ -45,11 +57,7 @@ var (
 		user        string
 	}
 
-	initContainerMounts = []struct {
-		containerPath string
-		source        string
-		flags         string
-	}{
+	initContainerMounts = []initContainerMount{
 		{"/etc/machine-id", "/run/host/etc/machine-id", "ro"},
 		{"/run/libvirt", "/run/host/run/libvirt", ""},
 		{"/run/systemd/journal", "/run/host/run/systemd/journal", ""},
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -65,6 +65,14 @@ var (
 		{"/var/log/journal", "/run/host/var/log/journal", "ro"},
 		{"/var/mnt", "/run/host/var/mnt", "rslave"},
 	}
+
+	initContainerX11Symlinks = []initContainerSymlink{
+		{"/tmp/.X11-unix", "/run/host/tmp/.X11-unix", true},
+		{"/tmp/.ICE-unix", "/run/host/tmp/.ICE-unix", true},
+		{"/tmp/.font-unix", "/run/host/tmp/.font-unix", true},
+		{"/tmp/.Test-unix", "/run/host/tmp/.Test-unix", true},
+		{"/tmp/.XIM-unix", "/run/host/tmp/.XIM-unix", true},
+	}
 )
 
 var initContainerCmd = &cobra.Command{
@@ -159,6 +167,12 @@ func initContainer(cmd *cobra.Command, args []string) error {
 		syscall.MS_NODEV|syscall.MS_STRICTATIME|syscall.MS_NOSUID,
 		"mode=1777"); err != nil {
 		return fmt.Errorf("failed to mount tmpfs at /tmp: %s", err)
+	}
+
+	if utils.PathExists("/run/host/tmp") {
+		if err := setupX11(); err != nil {
+			return fmt.Errorf("Failed to set up hosts X environment: %w", err)
+		}
 	}
 
 	if initContainerFlags.monitorHost {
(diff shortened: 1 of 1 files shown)


owtaylor:
Background:

https://gitlab.gnome.org/GNOME/mutter/-/issues/1289
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1424

Basically, when mutter was made to auto-start Xwayland, it only autostarted based on the abstract socket, which broke Flatpaks - so the fix was to make it only listen/autostart on the X11 socket - not entirely sure why it couldn't do both.

I'm actually in favor of the 'bind-mount /tmp' approach (and /var/tmp') - I fairly frequently have to work around /tmp being different on the host and toolbox, and can't think of any immediate downsides. 

If we want to go with the mount-by-mount approach, I'd recommend *only* binding in .X11-unix - .font-unix .ICE-unix and .XIM-unix are some very obsolete stuff. As I recall:

 .font-unix - font server - this was to share bitmap fonts between different X servers (X terminals, probably) on a network, I think
 .ICE-unix - X session protocol - support in GNOME and otherwise slowly died between 15-10 years ago
 .XIM-unux - X Input Method protocol. Been replaced by multiple generations of input method framework since

Though there shouldn't be any *harm* in binding in the other ones.

src/cmd/initContainer.go
```diff
@@ -45,18 +57,22 @@ var (
...
+	initContainerMounts = []initContainerMount{
 		{"/etc/machine-id", "/run/host/etc/machine-id", "ro"},
 		{"/run/libvirt", "/run/host/run/libvirt", ""},
 		{"/run/systemd/journal", "/run/host/run/systemd/journal", ""},
 		{"/var/lib/flatpak", "/run/host/var/lib/flatpak", "ro"},
 		{"/var/log/journal", "/run/host/var/log/journal", "ro"},
 		{"/var/mnt", "/run/host/var/mnt", "rslave"},
 	}
+
+	initContainerX11Symlinks = []initContainerSymlink{
```
debarshiray, line 69: It would be better to do it as a bind mount instead. Or is there a reason why they must be symbolic links? Did I miss something?

For individual regular files we prefer symbolic links instead of bind mounts, because symbolic links can track updates to the original file. Whereas, with bound mounts, if the original file was updated atomically, which is what most programs do, then the atomic rename changes the inode, and the two diverge.

However, symbolic links are always weird and introduce various oddities (eg., stat versus lstat), so bind mount is better when possible.

debarshiray (commented):
Thanks for the patch, @HarryMichal !

debarshiray:
On second thoughts, as @owtaylor suggested, let's go with bind mounting the entire `/tmp`. Not only does it avoid having to keep a list of aging locations for X11 support, but it also nice in the sense that you can put throwaway files in `/tmp` and have them shared.

I pushed https://github.com/containers/toolbox/pull/570 while discussing this with Owen in `#gnome-hackers` on GIMPNet.

However, I still like the idea of iterating over a list to create all the symbolic links. We can keep this PR open to pursue that idea.

owtaylor:
> On second thoughts, as @owtaylor suggested, let's go with bind mounting the entire /tmp

To be clear, this was @HarryMichal 's idea

debarshiray:
> > On second thoughts, as @owtaylor suggested, let's go with bind mounting the entire /tmp
> 
> To be clear, this was @HarryMichal 's idea

Oops! Sorry about that.

HarryMichal:
> On second thoughts, as @owtaylor suggested, let's go with bind mounting the entire `/tmp`. Not only does it avoid having to keep a list of aging locations for X11 support, but it also nice in the sense that you can put throwaway files in `/tmp` and have them shared.
> 
> I pushed #570 while discussing this with Owen in `#gnome-hackers` on GIMPNet.
>
> However, I still like the idea of iterating over a list to create all the symbolic links. We can keep this PR open to pursue that idea.

Sounds good to me. I'll try to extend this when I have time to do so.

HarryMichal:
I'm closing this because #570 was merged. Let's pursue the idea with list for symbolic links in a different PR.

PR #565, opened 2020-09-23
cmd/run: Use home folder when $PWD is not in toolbox

HarryMichal:
Since v0.0.91[0] Toolbox throws an error if $PWD is not available in a
toolbox. While this fixes the problem with 'toolbox enter/run' silently
failing to enter/exec in a container, it still requires an action to be
made by the user. I believe it is better to handle such situations more
gracefully by falling back to entering the user's home folder + printing
a warning about doing so.

[0] https://github.com/containers/toolbox/pull/370

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -279,6 +279,15 @@ func runCommand(container string,
 		}
 	}
 
+	if pathPresent, _ := isPathPresent(container, workingDirectory); !pathPresent {
+		fmt.Fprintf(os.Stderr, "Error: path %s not found in container %s\n",
+			workingDirectory, container)
+		fmt.Fprintf(os.Stderr, "Using %s instead.\n",
+			currentUser.HomeDir)
+
+		workingDirectory = currentUser.HomeDir
+	}
+
 	logrus.Debug("Checking if 'podman exec' supports disabling the detach keys")
 
 	var detachKeys []string


HarryMichal:
recheck

src/cmd/run.go
```diff
@@ -279,6 +279,15 @@ func runCommand(container string,
 		}
 	}
 
+	if pathPresent, _ := isPathPresent(container, workingDirectory); !pathPresent {
+		fmt.Fprintf(os.Stderr, "Error: path %s not found in container %s\n",
+			workingDirectory, container)
+		fmt.Fprintf(os.Stderr, "Using %s instead.\n",
+			currentUser.HomeDir)
+
+		workingDirectory = currentUser.HomeDir
```
debarshiray, line 288: The overall idea of falling back to `$HOME` is good, but I think that the implementation should be a bit different.

The first problem is that this adds an extra `podman exec ...`, which can be a bit slow, to every `toolbox enter` and `toolbox run`. People are going to mostly live inside their `$HOME`, so a missing directory is a slightly less common case and increases the cost of the extra call. Given that people think of `enter` as just their shell, this extra weight will start to add up over time. In fact, we should probably try to make `enter` leaner and faster, not heavier. :)

The second problem is that file system operations are inherently racy. Just because the path existed doesn't mean that it still exists when we actually try to use it. It might have gotten removed by the time we try to use it. This might not happen so much under normal circumstances, but depending on this assumption can open us up to all sorts of subtle bugs that a sufficiently motivated user can exploit. In general, with file system operations, it's better to first attempt the main operation, and then handle any errors coming out of it.

So, in this case, we need to reattempt the main `podman exec ...` with `$HOME` as the `workingDirectory` once we receive a 127 exit code. This means that we need to re-organize the code into some form of a loop. Perhaps we need to move the guts of `runCommand` into a different function to make it more manageable. I remember not going down this route because of what might be an intrusive re-organization for a somewhat cosmetic improvement. If it's too intrusive then it's best left as a post-0.1.0 item.

debarshiray (commented):
Isn't there already an issue or something for this? I vaguely remember discussing this before, or it could be that I am mistaken.

src/cmd/run.go
debarshiray: Another thing, since `workingDirectory` is a global variable, it might be a little less surprising to use a separate variable here instead of overriding the global value.

debarshiray: See https://github.com/containers/toolbox/pull/813

PR #567, opened 2020-09-25
test/system: Test that toolboxes start alright

HarryMichal:
This is useful for getting more info if a container fails to start. The current set of tests does not provide that information.


HarryMichal:
recheck

PR #571, opened 2020-10-01
Fix warning for Ubuntu containers: "Error: terminfo entry not found for xterm-256color"

likan999:
On Ubuntu, both /usr/share/terminfo and /lib/terminfo can have contents, and xterm-256color is at /lib/terminfo (provided by ncurses-base package) while xterm+256color is at /usr/share/terminfo (provided by ncurses-term package). This PR checks both directory to suppress the warning.


HarryMichal:
Hi @likan999! Thank you for the patch!

PR #572, opened 2020-10-03
Pass XAUTHORITY environment variable to container

likan999:
On system that uses this instead of ~/.Xauthority, this is need to make sure the container can run X programs.


debarshiray (changes requested):
Thanks for working on this. The patch looks good, but the commit message needs some work:
* Would be better if you used a real email address instead of `...@users.noreply.github.com`
* Prefix the subject of the commit with `pkg/utils: ` and mention *X11 clients* instead of *GUI programs*
* Link to this pull request

debarshiray:
I was playing around with this a bit today, and I realized that we need to add the package that provides `xauth(1)` (ie., the `xorg-x11-xauth` RPM on Fedora) to our default images for this to work out of the box.

Otherwise, if `xauth(1)` is missing, then `pam_xauth.so` won't set the `XAUTHORITY` environment variable inside a session started by `sudo(8)`.

likan999:
Oops I should've addressed the review comments myself. @debarshiray thanks for improving the PR.

Issue #563, opened 2020-09-21
Define hostname when creating toolbox


mvalahtv:
**Is your feature request related to a problem? Please describe.**
I would like to have multiple toolboxes so it would be nice if we can define hostname instead of `toolbox`.

**Describe the solution you'd like**
 It can be done via
```
toolbox create -h mytoolbox
```

meeas:
Related to this, can we also add a method to add a hostname after a toolbox has been created?  I assume we can use podman to add a hostname to a running toolbox, but I suspect that the hostname will not persist trough reboots.

debarshiray:
https://github.com/containers/toolbox/issues/969 has some good ideas on how to chose a better default for the container's host name.

One problem with setting the host name through podman create --host ..., like we do today, is that any attempt to change the hostname later doesn't persist across a container restart. This seems somewhat important, if we want to change the container's host name to uniquely identify them.

Of course, users can hack this up using their shell start-up scripts, but it will be good to come up with a less barbaric solution.  :)

One option could be to leverage the `toolbox init-container` entry point to repeatedly set the intended host name on container start-up.

debarshiray:
Let's first try to improve the default hostname that gets chosen for a Toolbx container before trying to come up with ways to customize it.

I don't want to offer users a way to choose a custom name when creating a Toolbx container, until we have figured out a way in which that name can be modified afterwards.  If a person wants custom names for their containers, it's not a stretch to imagine that they might want to change the name afterwards, and as I mentioned above, it's tricky to change the name later.

I do understand why users want to customize the names, though, and I do accept it as a reasonable thing to do.  It's just that we haven't figured out a way to do that properly.

Improving the default name is still going to be an improvement, and we have a clear way ahead, so let's focus on that first.  Baby steps.  :)

debarshiray:
I am going to close this issue in favour of https://github.com/containers/toolbox/issues/969 so that the list of open issues doesn't balloon out of control, and continues to roughly reflect the items that are on the current to-do list. However, it will still be around for future reference.

debarshiray:
> I don't want to offer users a way to choose a custom name when creating a
> Toolbx container, until we have figured out a way in which that name can be
> modified afterwards. If a person wants custom names for their containers,
> it's not a stretch to imagine that they might want to change the name
> afterwards, and as I mentioned above, it's tricky to change the name later.

[...]

PR #573, opened 2020-10-04
Pass hostname to toolbox container.

likan999:
When SSH to a host, ~/.Xauthority will be used for xauth; the entries in it contains the hostname. If within the toolbox container it uses another hostname, the xauth will fail. Therefore we need to use the same hostname as the host.


vilgotf:
You shouldn't ssh from the toolbox? This (might) break some workflows that depend on the hostname being set to toolbox for the detection of a toolbox environment

vilgotf:
#563 seems like a better solution imo

likan999:
To clarify, I am not ssh from within the toolbox. I ssh from a remote host to my host machine with X forwarding, then run `toolbox enter`; from within the toolbox, if I run any X program, it won't start because of the xauth failure.

likan999:
Addressed comments. Please take another look. Thanks.

likan999:
Addressed comments and resolved conflicts. It is ready for review, please take a look, thanks.

absenth:
This implementation is better than the one I've been trying to put together.  I like that this mimics the `--hostname` command podman accepts.

It would be awesome to see this approved and merged as having a dozen or so toolboxes that  all have the prompt `$user@toolbox` or requiring me to set the hostname after the fact manually is getting old.

sgallagher:
This looks completely reasonable and I'd very much like to see this land. I am in the same situation as absenth above: I have multiple toolboxes for different projects/distros and I'd really like to be able to name them sanely and easily.

sgallagher (approved):
The code changes look completely sane to me.

debarshiray:
> workflows that depend on the hostname being set to toolbox
> for the detection of a toolbox environment

The recommended way to detect a Toolbx environment is to look for the `/run/.toolboxenv` file.  That's also what `/etc/profile.d/toolbox.sh` uses.

debarshiray:
> To clarify, I am not ssh from within the toolbox. I ssh from a remote
> host to my host machine with X forwarding, then run `toolbox enter`;
> from within the toolbox, if I run any X program, it won't start because
> of the xauth failure.

Given the timing, I am tempted to ask if you have seen [this](https://github.com/containers/toolbox/commit/fca24b213e24d637d179b85bbbfd3b3bec70aadd) commit and whether it improves your situation?

Does the suggestion in https://github.com/containers/toolbox/issues/969 to include the host's hostname as a suffix in the container's hostname (eg., toolbox.<hostname> or something similar) help?

src/cmd/create.go
```diff
@@ -390,7 +396,7 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 	createArgs = append(createArgs, xdgRuntimeDirEnv...)
 
 	createArgs = append(createArgs, []string{
-		"--hostname", "toolbox",
+		"--hostname", createFlags.hostname,
```
debarshiray, line 399: We can't directly use `createFlags.hostname` in `createContainer` because `src/cmd/run.go` also uses this function.

debarshiray (commented):
Sorry about the delay.  This is implementing https://github.com/containers/toolbox/issues/563

For what it's worth, https://github.com/containers/toolbox/pull/1007 is a more complete and updated implementation of this idea. It validates the string passed to the `--hostname` option and has tests.

debarshiray:
> This looks completely reasonable and I'd very much like to see
> this land. I am in the same situation as absenth above: I have
> multiple toolboxes for different projects/distros and I'd really
> like to be able to name them sanely and easily.

There are at least two different immediate problems here.

First, there's a need to come up with a better default hostname for the Toolbx containers.  See https://github.com/containers/toolbox/issues/969, and all the issues and pull requests linked from it, for the discussion.  This is often conflated with a desire to make it easier to distinguish one container from another, because the hostname shows up in some very vsible places.  eg., in the shell's prompt, the terminal emulator's window chrome, etc..  However, they aren't exactly the same.

However, they aren't exactly the same thing.  `/run/.containerenv` inside Toolbx containers contain some [metadata](https://github.com/containers/podman/issues/6192) about the containers to help distinguish them better.  This metadata might better than relying on the hostname because it's richer and not constrained by the rules governing valid hostnames.  See https://github.com/containers/toolbox/issues/98, and all the issues and pull requests linked from it, for the discussion.

It's already possible today to use `/run/.containerenv` and `/run/.toolboxenv` to come up with one's own shell prompt without making any changes to the Toolbx code.  Having said that, it will definitely be good to improve the default prompt that Toolbx provides, which is what https://github.com/containers/toolbox/issues/98 is about.

This will probably satisfy a big chunk of users out there.

Then, there's the question of letting users customize the hostname, which is what https://github.com/containers/toolbox/issues/563 is about, but it has [some](https://github.com/containers/toolbox/issues/98#issuecomment-518567193) [problems](https://github.com/containers/toolbox/issues/563#issuecomment-1318532628) that need to be figured out.

debarshiray:
I am going to have to close this pull request.  However, you are welcome to help implement https://github.com/containers/toolbox/issues/969

Either way, thanks for your interest in Toolbx and contributing to the discussion!

debarshiray:
> > To clarify, I am not ssh from within the toolbox. I ssh from a remote
> > host to my host machine with X forwarding, then run `toolbox enter`;
> > from within the toolbox, if I run any X program, it won't start because
> > of the xauth failure.
> 
> Given the timing, I am tempted to ask if you have seen [this](https://github.com/containers/toolbox/commit/fca24b213e24d637d179b85bbbfd3b3bec70aadd) commit and whether it improves your situation?
> 
> Does the suggestion in #969 to include the host's hostname as a suffix in the container's hostname (eg., toolbox. or something similar) help?

I am still interested in solving your problem with SSH and X forwarding.  I am currently inside a Wayland session, which makes it difficult for me to reproduce your problem.  So, feel free to open a new issue with the above details to further investigate it.

PR #582, opened 2020-10-15
docs: fix reference to Fedora CoreOS

crawford:
CoreOS hasn't existed since late 2016 and refers to the now-defunct
Container Linux. Fedora CoreOS is an entirely separate project.


HarryMichal:
Thank you for the patch @crawford!

PR #583, opened 2020-10-15
src/go-build-wrapper: Support setting build options via env var

HarryMichal:
Toolbox uses a build wrapper that is used by Meson becase meson does not
support natively Go. That proved to be troublesome when we tried to
build Toolbox in Fedora's buildsystem that sets a number of CFLAGS,
LDFLAGS and Go specific flags. There was no way to pass those to the
wrapper, hence we hardcoded the most important ones using patches. With
this the patches can be dropped.

To pass an option to the 'go build' command, set GOBUILDFLAGS
environmental variable and export it.

### DIFF
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -27,5 +27,5 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -trimpath -ldflags "-extldflags '-Wl,--wrap,pthread_sigmask $4' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$3" -o "$2"
-exit "$?"
+# shellcheck disable=SC2086
+eval go build -o "$2" -trimpath -ldflags \"-extldflags \'-Wl,--wrap,pthread_sigmask $4\' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$3\" ${GOBUILDFLAGS:-}


src/go-build-wrapper
```diff
@@ -27,5 +27,5 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -trimpath -ldflags "-extldflags '-Wl,--wrap,pthread_sigmask $4' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$3" -o "$2"
-exit "$?"
+# shellcheck disable=SC2086
+eval go build -o "$2" -trimpath -ldflags \"-extldflags \'-Wl,--wrap,pthread_sigmask $4\' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$3\" ${GOBUILDFLAGS:-}
```
debarshiray, line 31: Umm... why do we need the `eval`? If we really do need it, then it would be nice to mention that in the commit message. These things are sometimes really arcane, so good to have an explanation.

Also, it might be better to include the `-trimpath` in `GOBUILDFLAGS`. We use it to [improve the reproducibility](https://github.com/containers/toolbox/pull/448) of the builds but looks like Fedora doesn't have it. Ideally, it would be part of the upstream defaults with distributions being able to override it.

debarshiray (commented):
Thanks for working on this! This should simplify downstream builds quite a bit.

src/go-build-wrapper
debarshiray: Does this actually work? On my laptop it errors out:
```
can't load package: package ": cannot find module providing package "
can't load package: package -ldflags: cannot find module providing package -ldflags
can't load package: package -B: cannot find module providing package -B
can't load package: package 0x517e1e35bdda4278769c0b780f4afbb1cedfb9c0: cannot find module providing package 0x517e1e35bdda4278769c0b780f4afbb1cedfb9c0
can't load package: package -extldflags: cannot find module providing package -extldflags
can't load package: package -Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld ": cannot find module providing package -Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld "
can't load package: package -a: cannot find module providing package -a
can't load package: package -v: cannot find module providing package -v
can't load package: package -x: cannot find module providing package -x
```

The `go build` command probably doesn't like multiple mentions of the `-ldflags` argument.

PR #588, opened 2020-10-21
cmd/run: Run a login shell when falling back to Bash during 'enter'

debarshiray:
'toolbox enter' attempts to run the user's preferred shell as a login
shell by passing it the '-l' argument. However, when the user's
preferred shell is missing from the toolbox container and it falls
back to Bash, then it was no longer a login shell because the '-l'
argument went missing.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -263,7 +263,7 @@ func runCommand(container string,
 				container)
 			fmt.Fprintf(os.Stderr, "Using /bin/bash instead.\n")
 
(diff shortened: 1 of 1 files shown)


HarryMichal:
Yes, I noticed this in the past but forgot to create a PR to fix it. Thanks!

debarshiray:
Oops, I deleted the branch before pushing it out, which confused GitHub. It was merged as d065ca460a85bc04435047f320d6fc6dc7d09b3e

Issue #267, opened 2019-09-20
Support 'sudo toolbox ...'


dustymabe:
One of the use cases from running toolbox on Fedora CoreOS is to debug the system (not development). When running as `root` I see an error:

```
[root@coreos ~]# toolbox 
toolbox: /etc/subgid and /etc/subuid don't have entries for user root
See the podman(1), subgid(5), subuid(5) and usermod(8) manuals for more
information.
[root@coreos ~]# exit
logout
[core@coreos ~]$ sudo toolbox
toolbox: /etc/subgid and /etc/subuid don't have entries for user root
See the podman(1), subgid(5), subuid(5) and usermod(8) manuals for more
information.
[core@coreos ~]$
```

Is this expected?

cgwalters:
This works with https://github.com/cgwalters/coretoolbox

dustymabe:
> This works with https://github.com/cgwalters/coretoolbox

Considering we decided that we don't want to have two implementations and that we'll work together on `toolbox` I don't believe that's a helpful reminder.

cgwalters:
> Considering we decided that we don't want to have two implementations and that we'll work together on toolbox I don't believe that's a helpful reminder.

Where was that decided?

dustymabe:
Well it seems like we've discussed "toolbox" a lot. In https://github.com/coreos/fedora-coreos-tracker/issues/38 and https://github.com/debarshiray/toolbox/issues/8 and https://github.com/coreos/fedora-coreos-tracker/issues/90. @mike-nguyen and @yuqi-zhang have been working together with @debarshiray IIUC to align the efforts. We have not added it to our design document so I guess we are missing a formal agreement, but in the meeting [a week ago](https://github.com/coreos/fedora-coreos-tracker/issues/38#issuecomment-530467441) I didn't hear anyone dissenting on including `toolbox` once the deps were aligned. Outside of the meeting you [spoke up](https://github.com/coreos/fedora-coreos-tracker/issues/38#issuecomment-530929147) and we tried to address the concerns you had in the [next meeting](https://github.com/debarshiray/toolbox/issues/265#issuecomment-532812500). I thought we had addressed those concerns.

cgwalters:
> but in the meeting a week ago I didn't hear anyone dissenting on including toolbox once the deps were aligned.

I dissent until we have resolved the discussion in https://github.com/debarshiray/toolbox/issues/265

debarshiray:
> When running as `root` I see an error:
> 
[...]

PR #591, opened 2020-10-22
cmd/initContainer: Don't rely on user D-Bus to track time zone configuration

debarshiray:
This is one more step towards enabling toolbox(1) to be run as root.
When invoked as 'sudo toolbox ...' there's no user or session D-Bus
instance available for the root user, which prevents the use of D-Bus
services like o.fd.Flatpak.SessionHelper.

https://github.com/containers/toolbox/issues/267

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -216,13 +216,6 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 
 	dbusSystemSocketMountArg := dbusSystemSocket + ":" + dbusSystemSocket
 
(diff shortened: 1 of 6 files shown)


HarryMichal:
The last CI fail is partially caused by a race condition I introduced in #567. The fix for this is in #594. The other part (in Fedora 33) is in my opinion not caused by the CI but a bug in Podman.

debarshiray:
> The last CI fail is partially caused by a race condition I introduced in #567.
> The fix for this is in #594. The other part (in Fedora 33) is in my opinion not
> caused by the CI but a bug in Podman.

Thanks for the insight. I was beginning to get a bit confused.

PR #594, opened 2020-10-24
test/system: Rework check if toolbox started successfully

HarryMichal:
When I added the test that looked at the logs of a toolbox, I did not
realize that the startup of a toolbox takes some time. That caused the
CI to flake even more than usual.

The solution I used was inspired by a helper function in Podman's test
suite (WaitContainerReady()).

### DIFF
--- a/test/system/201-run.bats
+++ b/test/system/201-run.bats
@@ -2,13 +2,10 @@
 
 load helpers
 
-@test "Start the 'running' container" {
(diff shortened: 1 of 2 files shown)


HarryMichal:
recheck

debarshiray:
Thanks for the quick fix, @HarryMichal -- much appreciated.

PR #607, opened 2020-10-29
cmd/run: Pass the UID when notifying the terminal about the container

debarshiray:
This is one more step towards enabling toolbox(1) to be run as root.
When invoked as 'sudo toolbox enter', GNOME Terminal might choose not
to preserve the current toolbox container when opening a new terminal.
This is similar to how the current working directory of a 'su -'
session isn't preserved when opening a new terminal.

https://github.com/containers/toolbox/issues/267


debarshiray:
Oops. This is a duplicate of https://github.com/containers/toolbox/pull/649

PR #609, opened 2020-10-30
pkg/utils, sh: Pass USER environment variable to the container.

likan999:


debarshiray:
Thanks for the patch. What's the use-case for this?

As far as I can make out, [POSIX](https://pubs.opengroup.org/onlinepubs/009696899/basedefs/xbd_chap08.html) doesn't mandate that shells set the `USER` environment variable, but recognizes that many do.

likan999:
Some tools assume the existence of the variable. In my particular case, it is bash-it, some bash library for bash command prompt, that doesn’t show my user name in command prompt. I know $(id -u) or $(whoami) would solve this but I believe there are many cases people (implicitly and maybe mistakenly according to your link) rely on them, for example, the toolbox bash script of this repo uses this variable as well: https://github.com/containers/toolbox/blob/master/toolbox. You can’t just clean up every places that rely on this variable.

debarshiray:
Bash does set the `USER` environment variable, so Bash-it should see it.

The reason we used `USER` in the POSIX shell implementation of Toolbox is that [ShellCheck](https://www.shellcheck.net/) didn't complain about the POSIX compliance of it, and we didn't run into any bug reports where a `/bin/sh` implementation didn't offer `USER`. So, it's a bit of a grey area there.

Ultimately, my point is that we shouldn't try to *enhance* the shell that the user is running. If the shell doesn't offer `USER`, then most (all?) instances of the shell won't have it, and software meant to run on it should be aware of it. If instances of the shell run by Toolbox start to offer `USER`, then it would be a source of confusion because the environment would be different inside and outside the container.

likan999:
I think we both agree different environment inside and outside the toolbox causes confusion. The disagreement comes from what counts as different behavior. As a user I only care about what worked for me outside the toolbox should work for me as well inside the toolbox. The problem is USER environment has nothing to do with bash but login: https://unix.stackexchange.com/a/76356, and toolbox doesn’t simulate logind so USER is missing inside the toolbox even running with bash, thus breaks existing workflows that worked for users outside of toolbox. Can you expand a little bit on your definition of “behavior” and why it is preferred even though it breaks many tools? I understand your point that if the entire world is compliant to POSIX and never rely on the USER environment variable then there won’t be such an issue; however the reality is such grey area existed and will exist for a quite substantial time, especially when SpellCheck doesn’t even warn about its usage, as you observed.

debarshiray:
> The problem is USER environment has nothing to do with bash but
> login: https://unix.stackexchange.com/a/76356

That's an interesting point. To be honest, I haven't dug into the sources to check who sets `USER`. On my Fedora 33 host, I do have `USER` set even inside my Toolbox containers.

likan999:
Yes, I am using Ubuntu host and container and it breaks. On Fedora it is set via /etc/profile, which contains a piece of logic setting USER by `id -un`; while it is not there on Ubuntu.

On Ubuntu host USER is there and in the toolbox the variable is missing, which I regarded as "difference" in behavior. I believe on the Ubuntu host USER is set by login but I haven't verified it.

To address the difference, we can either
1) don't support Ubuntu at all; or
2) support Ubuntu only on "toolbox-official" images that have /etc/profile patched; or require user-built images to patch /etc/profiile; or
3) support the USER variable in general by passing it in, like what it is done in this PR; or
4) simulate some login procedure in toolbox; or
5) ...

I believe 3) is the easiest way to do. I don't think it is as harmful as you previously thought:

For Fedora host & toolbox, since it is in /etc/profile, so both bash and sh has it set in both host and toolbox, w/ or w/o this PR; 

For Ubuntu host & toolbox, it is set by login on host for both bash and sh, and missing in toolbox w/o this PR, and set by toolbox for both sh and bash w/ this PR.

In Fedora it doesn't create any more difference and on Ubuntu it is a pure win. Of course there are a tons of other distributions out there but there won't be many that don't set USER; for any one that really doesn't have USER, then w/ this PR it won't be set in toolbox either, so this PR doesn't create regression in this case either.

Do you still have any other concerns?

debarshiray:
> Yes, I am using Ubuntu host and container and it breaks. On Fedora
> it is set via /etc/profile, which contains a piece of logic setting USER
> by id -un; while it is not there on Ubuntu.

Ok, this makes sense. I agree that it would be good to propagate `USER` into the container.

Thanks for the extensive investigation on this, and my apologies for being a bit thick in the head at times.

debarshiray:
I expanded upon the Git commit message a bit to explain the motivation behind the change.

Thanks for working on this!

Issue #391, opened 2020-03-17
is mlocate useful for toolbox?


juhp:
I am not sure how well mlocate works in Toolbox.

Recently when I tried it, I ended up with a 340MB database, maybe because of ~/.local/share/containers?

Are people using it or should we drop it from the default fedora-toolbox packages?

maxwell-k:
I am not using it and would be happy if you dropped it.

HarryMichal:
Until now I didn't know such tool even existed :)

juhp:
> Until now I didn't know such tool even existed :)

Oh it's useful, on Workstation at least.

juhp:
On Silverblue, `locate` only shows me entries for `/etc` and `/boot` basically... so yeah, but maybe the config could be fixed to index /usr, /home, /var, etc also?

HarryMichal:
I believe for now we can close this since @juhp went ahead and [removed](https://src.fedoraproject.org/container/fedora-toolbox/c/bd035973c9f49a96256951f3e2ec454a0973abfd?branch=master) the tool from the `fedora-toolbox` image and considering that there were no reports complaining about the tool being gone.

debarshiray:
I was playing with this a bit today.

> Recently when I tried it, I ended up with a 340MB database, maybe
> because of ~/.local/share/containers?

How do the sizes of `/var/lib/mlocate/mlocate.db` on the host and inside the container compare?

In my case on a Fedora Workstation system, they were 101M and 100M respectively. Silverblue doesn't have a working `mlocate` on the host, so the comparison there is less meaningful.

> See also https://bugzilla.redhat.com/show_bug.cgi?id=906591

Strangely enough, having `PRUNE_BIND_MOUNTS = "yes"` doesn't seem to prevent `updatedb` from indexing all the bind mounts inside a Toolbox container. eg., `locate Downloads` or `locate pam_xauth.so` returns results from locations that are known to be bind mounts.

PR #612, opened 2020-10-30
cmd/initContainer: Make locate(1) work inside toolbox containers

debarshiray:
... by running updatedb(8) at 24 hour intervals.

https://github.com/containers/toolbox/issues/391


nanonyme:
Is there any way to avoid this? Once a day is *much* too often, I would prefer more like once a week or once a month. This happens basically every time I try to start development work and it makes container unresponsive.

PR #614, opened 2020-11-01
cmd/create, doc/toolbox-create, sh: Allow specifying custom uid/gid mapping.

likan999:
On a system with multiple users, it is needed to setup the proper user and/or group id mapping in order to access some shared mount points.

### DIFF
--- a/doc/toolbox-create.1.md
+++ b/doc/toolbox-create.1.md
@@ -8,6 +8,7 @@ toolbox\-create - Create a new toolbox container
                [*--container NAME* | *-c NAME*]
                [*--image NAME* | *-i NAME*]
                [*--release RELEASE* | *-r RELEASE*]
+               [*--uidmap UIDMAP*] [*--gidmap GIDMAP*]
 
 ## DESCRIPTION
 
@@ -50,6 +51,11 @@ is useful for creating containers from custom-built base images.
 
 Create a toolbox container for a different operating system RELEASE than the
 host.
+
+**--uidmap** UIDMAP, **--gidmap** GIDMAP
+
+The `--uidmap` and `--gidmap` options passed to `podman` when creating the
+container; if both are omit, `--userns=keep-id` will be used.
 
 ## EXAMPLES
 
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -43,8 +43,10 @@ const (
 var (
 	createFlags struct {
 		container string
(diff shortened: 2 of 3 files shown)


debarshiray:
It would be good to know a bit more about the use-case.

The mapping of the user namespace is pretty fundamental to how the environment is set up inside a Toolbox container. Therefore, I am a bit hesitant to add too many knobs to it.

Note that with it's currently possible to [run toolbox(1) as root](https://github.com/containers/toolbox/issues/267). That might also help with.

likan999:
@debarshiray My use case is I have a host that has multiple users, and a mountpoint that is shared by different users. I would like the mountpoint to show correct owners and groups inside the toolbox. Without the manual specified mapping, only current user is recognized, and all other users are mapped to 65534.

My current solution is kind of hacky. I manually added the range 1000-3000 to /etc/subuid and /etc/subgid, and manually tweak the uidmap and gidmap so that the range 1000-3000 is mapped as 1000-3000 within the container, while other uids are mapped to something like 100000-165536. With this, within toolbox the mountpoint is shown correctly.

Theoretically it could be solved by running podman and toolbox as root and keeps the host uid/gid mapping, and I agree it is much cleaner. However, I have tried 0.0.97 and it doesn't work for me. `sudo toolbox create` fails and with `-vvvv` it reports something like `Error: host directory cannot be empty`.

Actually, 0.0.97 even breaks my existing toolbox. I had to revert https://github.com/containers/toolbox/commit/b9a0bd5f0c2a2421ec15eea286ca20f03b7152d2 then it goes back to normal (my test for `sudo toolbox` is not patched).

I guess it is because I am running all of these on Ubuntu and it is not a supported platform currently?

likan999:
I had https://github.com/containers/toolbox/pull/634 that fixes the issue I mentioned previously. I agree `sudo toolbox` would be the way to go, but currently there might be some missing pieces. For example, after entered the toolbox, I want it to switch automatically to my current user, rather than root; meanwhile, I need to environment to inherit from my current user, rather than root, e.g. `XDG_RUNTIME_DIR`. If the parts are already there, please don't hesitate to correct me. Thanks.

likan999:
Will close this one as it is no longer relevant. Changing the uid mapping fixes the mounted volume but it messes up the uids in the rootfs and seems like there is no way to get it right mathematically.

Issue #209, opened 2019-06-28
Avahi resolver does not work inside the toolbox


mildred:
After installing `nss-mdns` inside the toolbox, I have the following in `/etc/nsswitch.conf`:

```
hosts:      files mdns4_minimal [NOTFOUND=return] dns myhostname
```

However, `ssh elyas.local` works outside the toolbox but not inside:

```
mildred@toolbox$ ssh elyas.local
ssh: Could not resolve hostname elyas.local: Name or service not known
```

This is a problem to run ansible on local hosts...

debarshiray:
Yes, I can reproduce with `avahi-resolve --name kolache.local`.

debarshiray:
After playing with `gdbus(1)` and d-feet inside a Flatpak container and a Toolbox container, I see that D-Bus clients inside the latter have problems accessing services on the system D-Bus instance.

This works:
 ```
host$ flatpak run --devel --command=bash org.gnome.dfeet
flatpak$ gdbus introspect --system --dest org.freedesktop.Avahi --object-path /
...
```

This doesn't:
 ```
host$ toolbox enter
flatpak$ gdbus introspect --system --dest org.freedesktop.Avahi --object-path /
Error: Timeout was reached
```

I don't know why this is the case, though.

mildred:
This is also an issue on Silverblue, when using software inside the toolbox that must access network devices (example: simple-scan, not available as a Flatpak, cannot detect network scanners)

Jmennius:
In my case, the dbus introspection does work, but avahi client still doesn't work.

Apparently that's because of SELinux policies:
```log
Nov 08 15:20:07 hostname audit[789]: USER_AVC pid=789 uid=81 auid=4294967295 ses=4294967295 subj=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 msg='avc:  denied  { send_msg } for  scontext=system_u:system_r:avahi_t:s0 tcontext=unconfined_u:system_r:container_runtime_t:s0 tclass=dbus permissive=0
                                        exe="/usr/bin/dbus-broker" sauid=81 hostname=? addr=? terminal=?'
```

I'll see if I can fix that.

PR #625, opened 2020-11-15
Mount Avahi Daemon socket inside the toolbox container

fidencio:
This series mounts the Avahi daemon socket inside the toolbox container and adds the needed dependencies for having Avahi working out of the box (avahi & nss-mdns).

/cc @debarshiray @HarryMichal 

It also fixes https://github.com/containers/toolbox/issues/209

### DIFF
--- a/images/fedora/f28/extra-packages
+++ b/images/fedora/f28/extra-packages
@@ -18,6 +18,7 @@ man-db
 man-pages
 mlocate
 mtr
+nss-mdns
 openssh-clients
 PackageKit-command-not-found
 passwd
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -260,9 +260,20 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 
 	usrMountArg := "/usr:/run/host/usr:" + usrMountFlags + ",rslave"
 
+	var avahiSocketMount []string
+
+	avahiSocket, err := getServiceSocket("Avahi", "avahi-daemon.socket")
+	if err != nil {
+		logrus.Debug(err)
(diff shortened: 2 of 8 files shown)


src/cmd/create.go
```diff
@@ -271,6 +271,17 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 		kcmSocketMount = []string{"--volume", kcmSocketMountArg}
 	}
 
+	var avahiSocketMount []string
+
+	avahiSocket, err := getServiceSocket("Ahavi Daemon", "avahi-daemon.socket")
```
debarshiray, line 276: Nitpick: typo alert!

debarshiray (commented):
Thanks for the patch.

src/cmd/create.go
debarshiray: I took the liberty to fix it.

fidencio: Thanks!

images/fedora/f28/extra-packages
```diff
@@ -1,3 +1,4 @@
+avahi
```
debarshiray: It seems to me that only `nss-mdns` is necessary inside the container because it provides the GNU C Library plugin that hooks into the name resolution mechanism (or NSS) and talks to Avahi over the socket for names in the `.local` mDNS domain.

fidencio: Just for the record, we've discussed this "offline" and `nss-mdns` pulls `avahi` in.
@debarshiray mentioned he'll take care of having the package split for Fedora 34 and onwards.

PR #633, opened 2020-11-28
cmd/create, doc/toolbox-create, sh: Add flag --volume to create command.

likan999:
This allows specifying additional bind mounts when creating the toolbox.


catsout:
Any update?

Issue #622, opened 2020-11-09
Containers don't start when /etc/localtime is an absolute symbolic link on the host (eg., Arch and Ubuntu)


jgresty:
**Describe the bug**
After upgrading to toolbox version 0.0.97 I cannot start any toolboxes anymore. 

Output of `toolbox enter work`:

```
DEBU Running as real user ID 1000                 
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v1 host                 
DEBU Checking if /etc/subgid and /etc/subuid have entries for user x 
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Toolbox config directory is /home/x/.config/toolbox 
DEBU Current Podman version is 2.1.1              
DEBU Old Podman version is 2.1.1                  
DEBU Migration not needed: Podman version 2.1.1 is unchanged 
DEBU Resolving container and image names          
DEBU Container: 'work'                            
DEBU Image: ''                                    
DEBU Release: ''                                  
DEBU Resolved container and image names           
DEBU Container: 'work'                            
DEBU Image: 'fedora-toolbox:31'                   
DEBU Release: '31'                                
DEBU Checking if container work exists            
DEBU Inspecting mounts of container work          
DEBU Starting container work                      
DEBU Inspecting entry point of container work     
DEBU Entry point PID is a float64                 
DEBU Entry point of container work is toolbox (PID=0) 
Error: invalid entry point PID of container work
```

The output is the same for containers created before and after updating. Toolboxes can be started after downgrading to 0.0.96 and they can be entered in 0.0.97 once they have been started.

**Steps how to reproduce the behaviour**
1. Start toolbox

**Expected behaviour**
Toolbox is started

**Actual behaviour**
An error is thrown `Error: invalid entry point PID of container work` and toolbox is not started.

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.97`

**Toolbox package info (`rpm -q toolbox`)**
`toolbox-0.0.97-1-x86_64 (arch)`

**Output of `podman version`**
```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Git Commit:   9f6d6ba0b314d86521b66183c9ce48eaa2da1de2
Built:        Sat Sep 26 16:50:37 2020
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-2.1.1-1-x86_64 (arch)`

[...]

PR #634, opened 2020-11-28
cmd/initContainer: fix /etc/localtime handling for Ubuntu containers.

likan999:
On Ubuntu host and images, the `/run/host/etc/localtime` points to `/usr/share/zoneinfo/SomeTimeZone`; which breaks current code that assumes it point to `/run/host/usr/share/zoneinfo/SomeTimeZone`.


debarshiray:
I updated the commit message to include https://github.com/containers/toolbox/issues/622

Thanks for working on this.

PR #636, opened 2020-12-02
cmd: Remove periods at the end of flag descriptions

HarryMichal:
I noticed we put periods at the end of descriptions of flags while most tools (e.g. git, podman, skopeo, buildah, tools from coreutils,..) do not. Let's not deviate :). This is not very pressing because we don't show the description anyway.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -69,19 +69,19 @@ func init() {
 		"container",
 		"c",
 		"",
-		"Assign a different name to the toolbox container.")
+		"Assign a different name to the toolbox container")
 
 	flags.StringVarP(&createFlags.image,
 		"image",
 		"i",
 		"",
-		"Change the name of the base image used to create the toolbox container.")
+		"Change the name of the base image used to create the toolbox container")
 
 	flags.StringVarP(&createFlags.release,
 		"release",
 		"r",
 		"",
-		"Create a toolbox container for a different operating system release than the host.")
+		"Create a toolbox container for a different operating system release than the host")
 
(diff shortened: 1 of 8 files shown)


src/cmd/create.go
```diff
@@ -69,19 +69,19 @@ func init() {
...
 		"",
-		"Change the name of the base image used to create the toolbox container.")
+		"Change the name of the base image used to create the toolbox container")
 
 	flags.StringVarP(&createFlags.release,
 		"release",
 		"r",
 		"",
-		"Create a toolbox container for a different operating system release than the host.")
+		"Use specific release of the host operating system for toolbox container")
```
debarshiray, line 84: Nitpick: If we are to change the wording for the `release` flag, then it would be nice if the description for the different `release` flags were somehow similar. eg., this change doesn't match with ...

src/cmd/enter.go
```diff
@@ -46,13 +46,13 @@ func init() {
...
 		"",
-		"Enter a toolbox container with the given name.")
+		"Enter a toolbox container with the given name")
 
 	flags.StringVarP(&enterFlags.release,
 		"release",
 		"r",
 		"",
-		"Enter a toolbox container for a different operating system release than the host.")
+		"Enter a toolbox container for a different version of selected operating system")
```
debarshiray, line 55: ... this change.

debarshiray (commented):
Thanks, this makes sense.

debarshiray:
I left out the changes to the `release` description. We can revisit it in another PR.

PR #637, opened 2020-12-02
cmd/create: Make options --image and --release mutually exclusive

HarryMichal:
These two options are incompatible -> make them mutually exclusive.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -102,6 +102,9 @@ func create(cmd *cobra.Command, args []string) error {
 
 	var container string
 	var containerArg string
+	if cmd.Flag("image").Changed && cmd.Flag("release").Changed {
+		return errors.New("options --image and --release cannot be used together")
+	}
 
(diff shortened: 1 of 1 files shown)


src/cmd/create.go
```diff
@@ -102,6 +102,10 @@ func create(cmd *cobra.Command, args []string) error {
 
 	var container string
 	var containerArg string
+	if cmd.Flag("image").Changed && cmd.Flag("release").Changed {
+		fmt.Fprintf(os.Stderr, "Options --image and --release cannot be used together\n")
+		return err
```
debarshiray, line 107: Typo alert? The `err` is undefined. That's also why the test suite failed. :)

debarshiray (changes requested):
Thanks, this makes sense, but ...

src/cmd/create.go
debarshiray: Fixed, and merged. I hope I got it right.

PR #639, opened 2020-12-02
Correctly check validity of container name

HarryMichal:
regexp.MatchString() should only return an error if a wrong pattern is
used. This does not need to be checked outside of the 'utils' package
because the used pattern does not change and it should work. The only
value that needs to be checked is 'matched' -> IsContainerNameValid
only returns a bool.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -81,7 +81,7 @@ func run(cmd *cobra.Command, args []string) error {
 	if runFlags.container != "" {
 		nonDefaultContainer = true
 
-		if _, err := utils.IsContainerNameValid(runFlags.container); err != nil {
+		if !utils.IsContainerNameValid(runFlags.container) {
 			var builder strings.Builder
 			fmt.Fprintf(&builder, "invalid argument for '--container'\n")
 			fmt.Fprintf(&builder, "Container names must match '%s'\n", utils.ContainerNameRegexp)
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -497,10 +497,15 @@ func PathExists(path string) bool {
 }
 
 // IsContainerNameValid checks if the name of a container matches the right pattern
-func IsContainerNameValid(containerName string) (bool, error) {
+func IsContainerNameValid(containerName string) bool {
 	pattern := "^" + ContainerNameRegexp + "$"
 	matched, err := regexp.MatchString(pattern, containerName)
-	return matched, err
+	if err != nil {
+		panicMsg := fmt.Sprintf("failed to parse regular expression for container name: %v", err)
+		panic(panicMsg)
+	}
+
+	return matched
 }
 
 func IsInsideContainer() bool {
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -115,7 +115,7 @@ func create(cmd *cobra.Command, args []string) error {
 	}
 
 	if container != "" {
-		if _, err := utils.IsContainerNameValid(container); err != nil {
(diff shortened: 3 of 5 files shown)


debarshiray (commented):
It's a matter of API design. Ultimately, every API only makes sense if it can actually be used for a real world use-case; but we shouldn't tie it too hard to the users of the API, either.

A `bool` is a rather limited way of indicating an error. It's convenient to use, but doesn't say anything about what actually went wrong. Returning both a value and an error seems to be a pretty widely used idiom in Go. It's also common in C - GLib returns a `GError`, system calls set `errno`, etc..

Also, in terms of the API, it's better to let the caller decide what to do with the error - maybe it wants to log it, maybe it wants to re-throw it further up the call chain, or maybe it ignores it. Especially, for functions that are in different packages, because they are sufficiently decoupled from the rest of the code.

If tomorrow, some part of the code wants to do something else with the error, then having the flexibility would be useful.

In terms of this pull request, I am fine with changing the callers to check the `bool`, but I think `IsContainerNameValid` should continue return `(bool, error)`.

debarshiray:
> It's a matter of API design. Ultimately, every API only makes sense if it can
> actually be used for a real world use-case; but we shouldn't tie it too hard
> to the users of the API, either.
>
> A bool is a rather limited way of indicating an error. It's convenient to use, but
> doesn't say anything about what actually went wrong. Returning both a value
> and an error seems to be a pretty widely used idiom in Go. It's also common
> in C - GLib returns a GError, system calls set errno, etc..
>
> Also, in terms of the API, it's better to let the caller decide what to do with the
> error - maybe it wants to log it, maybe it wants to re-throw it further up the call
> chain, or maybe it ignores it. Especially, for functions that are in different
> packages, because they are sufficiently decoupled from the rest of the code.
>
> If tomorrow, some part of the code wants to do something else with the error,
> then having the flexibility would be useful.
>
> In terms of this pull request, I am fine with changing the callers to check the bool,
> but I think IsContainerNameValid should continue to return (bool, error).

Oops, I got this wrong. My apologies.

I completely missed the fact that [regexp.MatchString](https://golang.org/pkg/regexp/#MatchString) returns an error if the regular expression is invalid. Our regular expression is a constant, so there should never be an error, unless it's a programming mistake.

I should have read your commit message more closely. Sorry.

src/cmd/run.go
```diff
@@ -81,7 +81,7 @@ func run(cmd *cobra.Command, args []string) error {
 	if runFlags.container != "" {
 		nonDefaultContainer = true
 
-		if _, err := utils.IsContainerNameValid(runFlags.container); err != nil {
+		if utils.IsContainerNameValid(runFlags.container) {
```
debarshiray, line 84: Typo alert: missing negation operator.

src/pkg/utils/utils.go
```diff
@@ -497,10 +497,15 @@ func PathExists(path string) bool {
 }
 
 // IsContainerNameValid checks if the name of a container matches the right pattern
-func IsContainerNameValid(containerName string) (bool, error) {
+func IsContainerNameValid(containerName string) bool {
 	pattern := "^" + ContainerNameRegexp + "$"
 	matched, err := regexp.MatchString(pattern, containerName)
-	return matched, err
+	if err != nil {
+		logrus.Warnf("There was an error while matching container name: %v", err)
```
debarshiray, line 504: In this case, I think we should just `panic` because this can only happen due to a programming mistake, and not some likely run-time condition that should be handled or logged. It's similar to an assertion failure -- something that should never happen if the code is sane.

PR #640, opened 2020-12-02
cmd/initContainer: Avoid RPM failures due to unexpected file owners

debarshiray:
When running rootless, files and directories bind mounted from the
host operating system can have their ownership listed as
`nobody:nobody`. This is because the UIDs and GIDs that actually own
those locations are not available inside the container.

Some distribution packages are particular about the file ownerships of
some of these locations. eg., Fedora's `filesystem` RPM. Encountering
`nobody:nobody` as the owner can fail package management transactions
involving such packages leading to unforeseen consequences.

Therefore, configure RPM to leave these locations alone.

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -264,6 +264,24 @@ func initContainer(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	if utils.PathExists("/usr/lib/rpm/macros.d") {
+		logrus.Debug("Configuring RPM to ignore bind mounts")
+
+		var builder strings.Builder
+		fmt.Fprintf(&builder, "# Written by Toolbox\n")
+		fmt.Fprintf(&builder, "# https://github.com/containers/toolbox\n")
+		fmt.Fprintf(&builder, "\n")
+		fmt.Fprintf(&builder, "%%_netsharedpath /dev:/media:/mnt:/proc:/sys:/tmp:/var/lib/flatpak:/var/lib/libvirt\n")
+
+		rpmConfigString := builder.String()
+		rpmConfigBytes := []byte(rpmConfigString)
+		if err := ioutil.WriteFile("/usr/lib/rpm/macros.d/macros.toolbox",
+			rpmConfigBytes,
+			0644); err != nil {
+			return fmt.Errorf("failed to configure RPM to ignore bind mounts: %w", err)
+		}
+	}
+
 	logrus.Debug("Setting up daily ticker")
 
 	daily, err := time.ParseDuration("24h")


src/cmd/initContainer.go
```diff
@@ -252,6 +252,23 @@ func initContainer(cmd *cobra.Command, args []string) error {
...
+# https://github.com/containers/toolbox
+
+%_netsharedpath /dev:/media:/proc:/sys:/tmp:/var/lib/flatpak
+`
+
+		rpmConfigBytes := []byte(rpmConfigString)
+		if err := ioutil.WriteFile("/usr/lib/rpm/toolbox",
+			rpmConfigBytes,
+			0644); err != nil {
+			return errors.New("failed to configure RPM to ignore bind mounts")
```
HarryMichal, line 268:
```suggestion
			return fmt.Errorf("failed to configure RPM to ignore bind mounts: %v", err)
```
I believe, it is important to include the causing error.

HarryMichal (changes requested):
Left a comment with a proposed change. Apart from that, looks good to me. Did not test locally, yet. Great work!

src/cmd/initContainer.go
debarshiray: Good point! It should be `%w`, though, shouldn't it? So that the error can be unwrapped by the caller.

debarshiray:
Note that this doesn't seem to be working as I thought it would.

src/cmd/initContainer.go
```diff
@@ -252,6 +252,23 @@ func initContainer(cmd *cobra.Command, args []string) error {
...
+		logrus.Debug("Configuring RPM to ignore bind mounts")
+
+		rpmConfigString := `# Written by Toolbox
+# https://github.com/containers/toolbox
+
+%_netsharedpath /dev:/media:/proc:/sys:/tmp:/var/lib/flatpak
+`
+
+		rpmConfigBytes := []byte(rpmConfigString)
+		if err := ioutil.WriteFile("/usr/lib/rpm/toolbox",
```
praiskup, line 265: I guess you should use something like `/usr/lib/rpm/macros.d/macros.toolbox` file.

praiskup: Ping, this is the naming pattern that we should follow to actually make RPM load the macros.

debarshiray: Thanks for pointing that out, @praiskup !

debarshiray:
This fixes the `filesystem`, `flatpak` and `libvirt-libs` RPMs.

We might have to tweak the list of paths in `%_netsharedpath` over time as Toolbox evolves and problems with other RPMs come to light.

juhp:
This is really great news!

PR #652, opened 2020-12-21
playbooks: Improve how 'systemd-tmpfiles --create' is invoked

debarshiray:
The most important bit is the use of the 'creates' parameter of the
'command' module [1] because it conditionalizes the invocation of
'systemd-tmpfiles' on the presence of /run/media and self-documents its
purpose.

[1] https://docs.ansible.com/ansible/2.4/command_module.html


HarryMichal:
LGTM!

PR #658, opened 2021-01-05
[WIP] Add support for Red Hat UBI

HarryMichal:
Adding support for [Red Hat UBI] (Universal Base Image) is a requirement for Toolbox to be included in RHEL (and in the process replacing https://github.com/coreos/toolbox).

Adding support for a new image requires implementing an internal "store" of supported images. To be frank, this is something I dreaded quite a bit for some time already :). This implements a very basic image store that allows matching based on a [os-release](https://www.freedesktop.org/software/systemd/man/os-release.html):ID (e.g., `fedora` is Fedora, `rhel` is RHEL - in case of Toolbox UBI,..). This matching is available through a new option `--distro | -d`. The existing `--release | -r` option complements the new option.

The updated workflow should look a bit like this:
- Scenario 1 - host machine is Fedora 33
```
$ toolbox create              <- creates a Fedora 33-based container
$ toolbox create -r 31        <- creates a Fedora 31-based container
$ toolbox create -d rhel -r 7 <- creates a RHEL 7-based container
```
- Scenario 2 - host machine is RHEL 8
```
$ toolbox create           <- creates a RHEL 8-based container
$ toolbox create -r 7      <- creates a RHEL 7-based container
$ toolbox create -d fedora <- creates a Fedora 32-based container (default version of Fedora is used)
```
- Scenario 3 - host machine is Arch Linux
```
$ toolbox create <- creates a Fedora 32-based container
```

On distributions without a matching supported image will Fedora stay as the default.

While this adds only support for UBI, technically speaking, it should be possible to add support for even other images. Another matter is who is going to maintain the images and where they will be stored. Before these questions are answered, we'll continue offering only Fedora and UBI images (but you are always free to use an image of your own making!).

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -27,6 +27,7 @@ import (
 	"github.com/briandowns/spinner"
 	"github.com/containers/toolbox/pkg/podman"
 	"github.com/containers/toolbox/pkg/shell"
+	"github.com/containers/toolbox/pkg/toolbox"
 	"github.com/containers/toolbox/pkg/utils"
 	"github.com/godbus/dbus/v5"
 	"github.com/sirupsen/logrus"
@@ -43,6 +44,7 @@ const (
 var (
 	createFlags struct {
 		container string
+		distro    string
 		image     string
 		release   string
 	}
@@ -54,6 +56,8 @@ var (
 		{"/etc/profile.d/toolbox.sh", "/etc/profile.d/toolbox.sh"},
 		{"/etc/profile.d/toolbox.sh", "/usr/share/profile.d/toolbox.sh"},
 	}
+
+	enterCommand string
 )
 
 var createCmd = &cobra.Command{
@@ -70,6 +74,12 @@ func init() {
 		"c",
 		"",
 		"Assign a different name to the toolbox container.")
+
+	flags.StringVarP(&createFlags.distro,
+		"distro",
+		"d",
+		"",
+		"Use different distribution for the toolbox")
 
 	flags.StringVarP(&createFlags.image,
 		"image",
@@ -88,20 +98,26 @@ func init() {
 }
 
 func create(cmd *cobra.Command, args []string) error {
+	var image toolbox.Image
+	var container, containerArg string
+	var err error
+
 	if utils.IsInsideContainer() {
 		if !utils.IsInsideToolboxContainer() {
 			return errors.New("this is not a toolbox container")
 		}
 
-		if _, err := utils.ForwardToHost(); err != nil {
+		if _, err = utils.ForwardToHost(); err != nil {
 			return err
 		}
 
 		return nil
 	}
 
-	var container string
-	var containerArg string
+	if cmd.Flag("distro").Changed && cmd.Flag("image").Changed {
+		fmt.Fprintf(os.Stderr, "Options --distro and --image cannot be used together\n")
+		return err
+	}
 
 	if len(args) != 0 {
 		container = args[0]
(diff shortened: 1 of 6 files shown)


src/cmd/create.go
```diff
@@ -54,6 +56,8 @@ var (
 		{"/etc/profile.d/toolbox.sh", "/etc/profile.d/toolbox.sh"},
 		{"/etc/profile.d/toolbox.sh", "/usr/share/profile.d/toolbox.sh"},
 	}
+
+	enterCommand string
```
debarshiray, line 60: Do we really need to make this a global variable? Usually, the fewer globals, the better.

src/cmd/create.go
```diff
@@ -123,44 +139,60 @@ func create(cmd *cobra.Command, args []string) error {
...
+	} else {
+		if toolbox.IsHostSystemSupported() {
+			hostID, err := utils.GetHostID()
+			if err != nil {
+				return fmt.Errorf("There was an error while getting host's ID: %v", err)
+			}
+
+			image, _ = toolbox.GetImageForSystem(hostID)
+		} else {
+			image = toolbox.GetFallbackImage()
```
debarshiray, line 162: I wonder if the fallback image should be folded into `GetImageForSystem` itself. All call sites need the fallback, right?

debarshiray (commented):
Thank you for working on this!

Let's see if we can split the commits a bit more. eg., one where `ResolveContainerAndImageNames` gets replaced by the new image backend, followed by one where `--distro` gets added and so on.

Some quick comments:

src/cmd/create.go
```diff
@@ -123,44 +139,60 @@ func create(cmd *cobra.Command, args []string) error {
...
-		createFlags.image,
-		release)
-	if err != nil {
-		return err
+		image, _ = toolbox.GetImageForSystem(createFlags.distro)
+	} else {
+		if toolbox.IsHostSystemSupported() {
+			hostID, err := utils.GetHostID()
+			if err != nil {
+				return fmt.Errorf("There was an error while getting host's ID: %v", err)
```
debarshiray, line 157: It's [better](https://blog.golang.org/go1.13-errors) to use `%w` when wrapping errors.

src/pkg/toolbox/image.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+var (
+	// knownImages holds all known toolbox images
+	// These should work out of the box with all Toolbox commands
+	knownImages = map[string]Image{
+		"fedora-toolbox": {
+			"registry.fedoraproject.org",
+			"fedora-toolbox",
+			"latest",
```
debarshiray, line 18: The `latest` part makes me a bit nervous. :)

I did `toolbox create` on my *F33* host, and got a *F34* container named `fedora-toolbox-latest`. Another point of confusion would be when the host gets upgraded, or the value of `latest` changes over time.

src/pkg/toolbox/image.go
```diff
@@ -0,0 +1,131 @@
...
+func GetFallbackImage() Image {
+	return supportedSystems["fedora"]
+}
+
+// GetImageForSystem returns an image matching the given system ID.
+//
+// System ID should be taken from the ID entry of 'os-release'.
+//
+// https://www.freedesktop.org/software/systemd/man/os-release.html#ID=
+func GetImageForSystem(systemID string) (Image, error) {
```
debarshiray, line 60: It might be good to stick to the same nomenclature. Since we already chose `distro` as the name of our command-line flag, we should probably use the same word elsewhere in the code.

src/pkg/toolbox/image.go
```diff
@@ -0,0 +1,131 @@
...
+		return img, errors.New("unsupported system")
+	}
+
+	img = supportedSystems[systemID]
+
+	return img, nil
+}
+
+// Image holds parts of a full URI of an image
+type Image struct {
```
debarshiray, line 73: I am wondering if we should call it `Distro`, instead of `Image`. Then we could include a function pointer to parse the release string in a distro-specific way.

debarshiray:
I split up the commits, and fixed up some of the above issues in https://github.com/containers/toolbox/pull/667

Most importantly, I had to skip some parts of the new container and image tracking backend to side-step some issues that I couldn't figure out on my own. eg., the one caused by the use of `latest` tags. Let's keep this pull request open to work through those.

travier:

PR #674, opened 2021-01-19
test/system: Remove unused variables

debarshiray:


HarryMichal (approved):
I wonder what these were for... But I don't think they are necessary anymore. Merge!

debarshiray:
Wow! That was fast. Thanks for taking a look.

I think those were terminal escape sequences for producing coloured output that we used to have in the POSIX implementation of `toolbox(1)`.

PR #675, opened 2021-01-22
Match user group when taking ownership of toolboxRuntimeDirectory

### LINKED ISSUE #672: Toolbox errors with `Error: failed to get the current user's default shell`
Toolbox errors with `Error: failed to get the current user's default shell` on running any command when (I think) zsh is the default shell. 

1. Setup `prezto` or maybe just `chsh zsh` works too
2. Run `toolbox create -c abc`
3. See error  `Error: failed to get the current user's default shell` 
4. The same error is also seen on `bash -c 'toolbox create -c abc'`

**Expected behaviour**
It should have created a container

**Actual behaviour**
It failed to create a container with `Error: failed to get the current user's default shell`

**Screenshots**
NA

`toolbox version 0.0.99`

`toolbox-0.0.99-1.fc33.x86_64`

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 14:37:50 2020
OS/Arch:      linux/amd64
```

e.g., `podman-1.9.2-1.fc32.x86_64`

**Info about your OS**
Fedora 33

**Additional context**
I am using Prezto zsh. PFA the log of `toolbox -vv create -c abc &> toolbox-log`

[toolbox-log.txt](https://github.com/containers/toolbox/files/5832755/toolbox-log.txt)

mjlbach:
Closes #672

In initContainer and GetRuntimeDirectory, uid and gid were assumed to match. This is not the case on some systems, causing toolbox to fail to start the container.


mjlbach:
The test failed instantly (0s) and I can't see any logs. The patch works locally on my machine, but if I know what the error is (linting or otherwise) I'll fix it.

src/cmd/initContainer.go
```diff
@@ -115,6 +116,12 @@ func init() {
 		"Create a user inside the toolbox container whose login name is USER")
 	initContainerCmd.MarkFlagRequired("user")
 
+	flags.IntVar(&initContainerFlags.gid,
```
HarryMichal, line 119: The flags are sorted in one way and the flag definition other way.

HarryMichal (changes requested):
I like how this was handled. One minor nitpick and something to discuss.

It is about the `configureUsers()` function in `src/cmd/initContainer.go`. If we want to match the `gid`, we need to apply it to the user (either when creating or modifying). This requires to handle the case when the flag is not defined to maintain compatibility with toolboxes created with previous versions of Toolbox. This is caused by the fact that the entry-point of a container cannot be changed.

HarryMichal:
Hi @mjlbach! Thank you for working on this patch!

> The test failed instantly (0s) and I can't see any logs. The patch works locally on my machine, but if I know what the error is (linting or otherwise) I'll fix it.

Sorry about this, seems like Zuul (the CI system) had some temporary problem. In case you want to rerun the test, you can write a comment with text "recheck". This should trigger a rerun.

HarryMichal:
I wonder if this PR is really related to #672 :).

mjlbach:
Whoop! It totally isn't, I don't know how that happened. Updated the issue.

src/cmd/initContainer.go
mjlbach: Good catch, should be fixed now.

mjlbach:
> It is about the configureUsers() function in src/cmd/initContainer.go. If we want to match the gid, we need to apply it to the user (either when creating or modifying). This requires to handle the case when the flag is not defined to maintain compatibility with toolboxes created with previous versions of Toolbox. This is caused by the fact that the entry-point of a container cannot be changed.

Sorry, I may need a little bit more explanation. Currently only the UID seems to be matched, do you also want to match the gid?

anthr76:
Any updates on this PR? Would like to have this functionality!

HarryMichal:
> > It is about the configureUsers() function in src/cmd/initContainer.go. If we want to match the gid, we need to apply it to the user (either when creating or modifying). This requires to handle the case when the flag is not defined to maintain compatibility with toolboxes created with previous versions of Toolbox. This is caused by the fact that the entry-point of a container cannot be changed.
> 
> Sorry, I may need a little bit more explanation. Currently only the uid seems to be matched, do you also want to match the gid? The only place the gid is currently used is here: https://github.com/containers/toolbox/pull/675/files#diff-deeeece906034038075c67156564a9484e6d19694887a9ecb8309aca0ed56402R481

Toolbox adds a user or modifies the existing one to match the user that created the toolbox. Currently we only match the `uid` but with this we can also match the `gid`. https://github.com/containers/toolbox/blob/master/src/cmd/initContainer.go#L349

src/cmd/initContainer.go
```diff
@@ -109,6 +110,12 @@ func init() {
 		"Create a user inside the toolbox container whose numerical user ID is UID")
 	initContainerCmd.MarkFlagRequired("uid")
 
+	flags.IntVar(&initContainerFlags.gid,
+		"gid",
+		0,
+		"Create a user inside the toolbox container whose numerical group ID is GID")
+	initContainerCmd.MarkFlagRequired("gid")
```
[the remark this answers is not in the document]
debarshiray: I took the liberty to address this.

debarshiray:
I took the liberty to update the manual and tweaked the commit message to mention that we aren't using the specified GID to actually configure the user inside the toolbox container.

Thank you for your contribution.

mjlbach:
Thank you @debarshiray! I'm sorry I put this on the back burner due to school, but thank for your following through for me 😆

PR #676, opened 2021-01-22
Check if /mnt exists before adding it as a volume mount

mjlbach:
Some machine don't have /mnt. This just checks if it exists before trying to bind mount it into the container.

cc @tfmoraes

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -313,17 +313,19 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 		}
 	}
 
-	logrus.Debug("Checking if /mnt is a symbolic link to /var/mnt")
-
 	var mntLink []string
 	var mntMount []string
 
-	mntPath, _ := filepath.EvalSymlinks("/mnt")
-	if mntPath == "/var/mnt" {
-		logrus.Debug("/mnt is a symbolic link to /var/mnt")
-		mntLink = []string{"--mnt-link"}
-	} else {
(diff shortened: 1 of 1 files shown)


src/cmd/create.go
```diff
@@ -318,13 +318,15 @@ func createContainer(container, image, release string, showCommandToEnter bool)
...
 	var mntMount []string
 
-	mntPath, _ := filepath.EvalSymlinks("/mnt")
-	if mntPath == "/var/mnt" {
-		logrus.Debug("/mnt is a symbolic link to /var/mnt")
-		mntLink = []string{"--mnt-link"}
-	} else {
-		mntMount = []string{"--volume", "/mnt:/mnt:rslave"}
-	}
+	if utils.PathExists("/mnt") {
```
debarshiray, line 321: We should move the log message from a few lines above inside this `if` branch.

src/cmd/create.go
```diff
@@ -318,13 +318,15 @@ func createContainer(container, image, release string, showCommandToEnter bool)
...
+    mntPath, _ := filepath.EvalSymlinks("/mnt")
+    if mntPath == "/var/mnt" {
```
debarshiray, line 323: Nitpick: the indentation is off; `go fmt` might help.

debarshiray (changes requested):
Thanks for working on this!

debarshiray:
I took the liberty to make the above adjustments and merged to `master`. I hope I got it right.

Once again, thanks for your contribution.

mjlbach:
Thank you!

PR #692, opened 2021-02-10
Make profile.d script more generic

andrunko:
Make it more generic, to avoid having to patch it downstream (e.g. in the Endless OS packaging).


andrunko:
Force pushed to mention this PR in the commit msgs, not planning other changes atm.

andrunko:
@debarshiray Any thoughts on this one?

profile.d/toolbox.sh
```diff
@@ -10,7 +10,11 @@ toolbox_welcome_stub="$toolbox_config/toolbox-welcome-shown"
 # shellcheck disable=2046
 # shellcheck disable=SC1091
 eval $(
-          . /usr/lib/os-release
+          if [ -f /usr/lib/os-release ]; then
+              . /usr/lib/os-release
+          else
+              . /etc/os-release
```
debarshiray, line 16: Strictly speaking, we should check `/etc/os-release` first and then fall back to `/usr/lib/os-release`.  The `os-release(5)` [manual](https://www.freedesktop.org/software/systemd/man/os-release.html) says:
```
The file /etc/os-release takes precedence over /usr/lib/os-release.
Applications should check for the former, and exclusively use its data
if it exists, and only fall back to /usr/lib/os-release if it is missing.
```

That's also what we do in `test/system/libs/helpers.bash`.

profile.d/toolbox.sh
```diff
@@ -44,13 +48,18 @@ if [ -f /run/.containerenv ] \
...
+            echo " - To create a new tools container, run 'toolbox create'."
+            echo ""
+            printf "For more information, see the "
+            # shellcheck disable=SC1003
+            printf '\033]8;;https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/\033\\documentation\033]8;;\033\\'
+            printf ".\n"
+        else
+            echo " - To create a new tools container, run 'toolbox create'."
+        fi
         echo ""
```
debarshiray, line 63: Nitpick: an extra newline above the `echo` would be good.

debarshiray (changes requested):
Thanks for the fixes, @andrunko ; and apologies for the delay.

Our `os-release(5)` handling has been wrong since the very early days, but it worked, so we never got around to correcting it.  It's great to see it happen finally.

debarshiray:
I took the liberty to update the pull request with the above tweaks.

debarshiray:
Damn!  The unit tests don't run for changes to `profile.d/*`.

PR #693, opened 2021-02-12
test/system: Refactor tests using bats-support and bats-assert

HarryMichal:
This has been split from #517.

The bats-support[0] and bats-assert[1] libraries extend the
capabilities of bats[2]. Mainly, bats-assert is very useful for clean
checking of values/outputs/return codes.

Apart from updating the cases to use the libraries, the test cases have
been restructured in a way that they don't depend on each other anymore.
This required major changes in the helpers.bats file.

Overall, the tests are cleaner to read and easier to extend due to the
test cases being independent.

There will be a follow up commit that will take care of downloading of
the tested images locally and caching them using Skopeo to speedup the
tests and try to resolve network problems when pulling the images that
we experienced in the past.

[0] https://github.com/bats-core/bats-support
[1] https://github.com/bats-core/bats-assert
[2] https://github.com/bats-core/bats-core


HarryMichal:
Fedora 34 was branched quite recently and GPG keys are not OK, yet. The rawhide test suite can be ignored for a little while.

PR #698, opened 2021-02-18
Add deprecation notices to Shell Toolbox

HarryMichal:
Shell Toolbox has been replaced by the Go implementation a quite long
time ago. People on several ocassions created PRs that still update it.
It was not clear that the Shell implementation has been deprecated and
is no longer maintained.

### DIFF
--- a/toolbox
+++ b/toolbox
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+# THIS IMPLEMENTATION OF TOOLBOX IS NO LONGER MAINTAINED.
+# Please, use the Go implementation instead.
 
 exec 3>/dev/null
 
@@ -2186,6 +2188,10 @@ update_container_and_image_names()
 
 
 arguments=$(save_positional_parameters "$@")
+
+echo "THIS IMPLEMENTATION OF TOOLBOX IS NO LONGER MAINTAINED."
+echo "Please, use the Go implementation instead."
+echo
 
 host_id=$(get_host_id)
 if [ "$host_id" = "fedora" ] 2>&3; then


toolbox
```diff
@@ -2264,6 +2268,8 @@ fi
 
 echo "$base_toolbox_command: TOOLBOX_PATH is $TOOLBOX_PATH" >&3
 
+echo "THIS IMPLEMENTATION OF TOOLBOX IS NO LONGER MAINTAINED."
+
```
debarshiray, line 2272: Would be good to also include the advice to use the Go implementation as in the above comment, and a trailing newline character to make the message stand out from the other messages.

debarshiray (commented):
Nice, thanks!

toolbox
HarryMichal: Good point!

heyakyra:
Why is the deprecated installation still in the repo, and still packaged in the archives published under the github releases?

The warning says to switch to another implementation, but doesn't provide any guidance on doing so. Shouldn't only the recommended executable be packaged and released?

debarshiray:
> Why is the deprecated implementation still in the repo, and still
> packaged in the archives published under the github releases?

It's there to preserve the Git history associated with that file.  The history is still interesting when developing Toolbx.

> The warning says to switch to another implementation, but doesn't
> provide any guidance on doing so.

Both the GitHub [project](https://github.com/containers/toolbox) and the [website](https://containertoolbx.org/) link to [instructions](https://containertoolbx.org/install/) for building Toolbx.

heyakyra:
Removing the file would not remove the git history, just make it an extra step or two to get to, right?

debarshiray:
> Removing the file would not remove the git history, just make it
> an extra step or two to get to, right?

Yes, but that extra step or two would be annoying enough.

debarshiray:
I added a link to the installation guide in the deprecation notice in https://github.com/containers/toolbox/pull/1094

PR #699, opened 2021-02-18
Update bug-report.md

HarryMichal:
A lot of errors are caused by toolboxes not starting up. In such cases we require
the output of `podman start --attach` to see what is going on. It would be easier
if users provided this information right when they are filling a bug.


debarshiray:
Thanks! Merged after rewrapping the commit message to fit within 72 characters per line.

PR #702, opened 2021-02-19
.zuul, playbooks: Unify the system-test*.yaml for all Fedora versions

debarshiray:
The playbooks were too fragmented and were hard to maintain. With this
the number of files is significantly lowered.

https://github.com/containers/toolbox/pull/517


debarshiray:
recheck

PR #705, opened 2021-02-19
test/system: Don't block waiting for user input, if the test fails

debarshiray:


debarshiray:
recheck

HarryMichal:
Is this still relevant? Looking back, I think the problem is tied to the fact that the command `podman exec`, used by `toolbox run` under the hood, are unconditionally given the `-it` flags which cause hiccups with the output "redirection". I haven't checked but I have a feeling the flaking is currently happening for the "Hello World" tests before this one. But maybe all of them are flaking.

debarshiray:
> Is this still relevant? Looking back, I think the problem is tied to the fact that
> the command `podman exec`, used by `toolbox run` under the hood, are
> unconditionally given the `-it` flags which cause hiccups with the output
> "redirection". I haven't checked but I have a feeling the flaking is currently
> happening for the "Hello World" tests before this one. But maybe all of them
> are flaking.

Yes, you are right.

I wrote this in an attempt to fix the flakes - to see how the CI reacts to it.  Obviously, it didn't help. :)

Let's close this.

PR #707, opened 2021-02-23
Also bind mount systemd-resolved /run directories

nanonyme:
This directory structure contains in distros that have systemd247
or higher (eg F34) systemd-resolved varlink socket which toolbox
guest can use


nanonyme:
Hasn't yet been gone through end-to-end testing.

debarshiray:
Thanks for working on this, @nanonyme !

nanonyme:
Retroactively this should fix https://bugzilla.redhat.com/show_bug.cgi?id=1933433 for programs using glibc NSS in F34 toolbox.

PR #520, opened 2020-07-31
Add Arch Linux toolbox image

erazemk:
This PR is a continuation of #328.

Tested with `podman build --tag archlinux:latest ./Dockerfile`, but with a few problems:

* Image is 1.87 GB, possibly because of base-devel or kernel (Arch's image from Dockerhub is 426 MB)
* Can't enter the container with `podman run --name arch-test --rm localhost/archlinux` (even though it is listed when running `podman images`)

The following packages were installed when building the image:
```
dnssec-anchors-20190629-2  elfutils-0.180-1  gc-8.0.4-4  guile-2.2.6-2  icu-67.1-1  iproute2-5.7.0-1
iputils-20190709-2  ldns-1.7.1-2  libcroco-0.6.13-2  libedit-20191231_3.1-1  libmicrohttpd-0.9.71-1
libmpc-1.1.0-2  libpipeline-1.5.2-1  libxml2-2.9.10-2  mkinitcpio-28-1  mkinitcpio-busybox-1.31.1-1
pciutils-3.7.0-1  perl-error-0.17029-2  perl-mailtools-2.21-4  perl-timedate-2.33-2  psmisc-23.3-2
systemd-sysvcompat-245.7-1  xxhash-0.8.0-1  autoconf-2.69-7  automake-1.16.2-3  base-2-2
bash-completion-2.10-2  binutils-2.34-5  bison-3.6.4-1  curl-7.71.1-1  diffutils-3.7-3  fakeroot-1.24-2
file-5.39-1  findutils-4.7.0-2  flex-2.6.4-3  gawk-5.1.0-1  gcc-10.1.0-2  gettext-0.20.2-1
git-2.28.0-1  grep-3.4-1  groff-1.22.4-3  gzip-1.10-3  keyutils-1.6.3-1  less-551-3
libtool-2.4.6+42+gb88cebd5-13  linux-5.7.11.arch1-1  linux-firmware-20200721.2b823fc-1  lsof-4.93.2-3
m4-1.4.18-3  make-4.3-3  man-db-2.9.3-1  man-pages-5.07-1  mlocate-0.26.git.20170220-3  mtr-0.93-3
openssh-8.3p1-3  pacman-5.2.2-1  patch-2.7.6-8  pkgconf-1.7.3-1  rsync-3.2.2-2  sed-4.8-1  sudo-1.9.2-1
[...]

PR #710, opened 2021-02-25
Add Arch Linux toolbox image

erazemk:
*This PR builds on the previous tries (#328 and #520) of including an Arch Linux image in toolbox.*

The PR adds an Arch Linux image, created from the upstream base-devel branch.

It includes most of the packages that the Fedora-based images do, but currently excludes flatpak (required for flatpak-spawn), since it adds ~690 MB of extra packages/dependencies.

The base-devel image (without any extra packages) is 729 MB, when syncing repositories 741 MB and with all the packages (excluding flatpak) 831 MB, which should be the minimum if based on archlinux:base-devel.

If the final size is very important I can try and base it off of `archlinux:latest` (which includes only the base packages, not base-devel), but that would still require adding some of the missing packages - although the base image size is 416 MB, which is quite an improvement..

Another option could be to separate latest and base-devel versions into separate images, similarly to what is done with Fedora versions.

I can also add the [yay AUR helper](https://github.com/Jguer/yay) if there is a need for it.


xPMo:
Didn't work for me, but I figured it out.
I saw that a mount was failing by manually starting the container: `podman container start  --interactive --attach archlinux-toolbox`

Adding this line to the Dockerfile fixed that.

```
RUN touch /etc/machine-id
```

I also added a sudoers entry for the `%wheel` for mine container.

erazemk:
I just now realized that the proper way to create and test custom images is by building them and using `toolbox create --image` to use them with toolbox.

I tried running the container that way, but got the error `Error: invalid entry point PID of container`, but that was fixed by adding `/etc/machine-id` to the image, like you recommended, but I got the following error:

```sh
bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;'
bash: PROMPT_COMMAND: line 0: ` ; printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"'
```

Do you not get this error and if so, which OS are you using?

xPMo:
I'm not getting that error, I'm not sure where that is coming from (it doesn't match any code in `toolbox` afaik). Maybe that's coming from your ~/.bashrc? That may still be something that can be fixed in the container config, but more likely is just some assumption that your bashrc is making.

> which OS are you using?

[Latest toolbox on an Arch host.](https://archlinux.org/packages/community/x86_64/toolbox/)

erazemk:
> Maybe that's coming from your ~/.bashrc?

It seems the problem was indeed with this, I use the fish shell, so I don't have a local bashrc file, just the system one (which I haven't changed). This is weird though, since I based the Dockerfile off of the Fedora files, which work normally with toolbox. Copying the root user's bashrc doesn't help either, I still get the same error.

> That may still be something that can be fixed in the container config, but more likely is just some bad assumption that your bashrc is making. What do you get from `grep -A5 PROMPT_COMMAND ~/.bash*`?

When replacing `~/.bash*` with `/etc/bashrc` I get the following output:
```bash
   if [ -z "$PROMPT_COMMAND" ]; then
      case $TERM in
      xterm*|vte*)
        if [ -e /etc/sysconfig/bash-prompt-xterm ]; then
            PROMPT_COMMAND=/etc/sysconfig/bash-prompt-xterm
        elif [ "${VTE_VERSION:-0}" -ge 3405 ]; then
            PROMPT_COMMAND="__vte_prompt_command"
        else
            PROMPT_COMMAND='printf "\033]0;%s@%s:%s\007" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"'
        fi
        ;;
      screen*)
        if [ -e /etc/sysconfig/bash-prompt-screen ]; then
            PROMPT_COMMAND=/etc/sysconfig/bash-prompt-screen
        else
            PROMPT_COMMAND='printf "\033k%s@%s:%s\033\\" "${USER}" "${HOSTNAME%%.*}" "${PWD/#$HOME/\~}"'
        fi
        ;;
      *)
        [ -e /etc/sysconfig/bash-prompt-default ] && PROMPT_COMMAND=/etc/sysconfig/bash-prompt-default
        ;;
      esac
    fi
    # Turn on parallel history
    shopt -s histappend
```

xPMo:
> I use the fish shell, so I don't have a local bashrc file

Yeah, I use Zsh.

I tried replacing my ~/.bashrc /etc/bashrc from my Fedora toolbox and couldn't reproduce the issue.
Let's try tracing:

```bash
PS4='${FUNCNAME[*]}:$BASH_LINENO> ' bash -xi 2> >(grep -C4 PROMPT_COMMAND)
```

xPMo:
https://bugs.archlinux.org/task/71002

xPMo:
@erazemk: I closed #716 about the profile errors, Arch's image should add the `vte-common` package to support VTE terminals on hosts. I updated [my `arch-img` branch to include it](https://github.com/erazemk/toolbox/compare/master...xPMo:arch-img?expand=2).

PR #712, opened 2021-02-25
src/create: Handle read-only /boot

HarryMichal:
CoreOS recently made /boot read-only[0]. This caused an issue with starting containers because /boot was mounted only with option rslave but missed the ro option. This caused a permission issue.

This scenario is very similar to the one with /usr on Fedora Silverblue. The solution for this is to check mount options of the path and check if it uses the rw option or ro and then add it to the mount options in the --volume option in 'podman create'.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/734

https://github.com/coreos/fedora-coreos-config/commit/1de21ffa98bb22995e5b059501e1955bf52b562c

## Situation on CoreOS

This, sadly, does not fix existing containers on CoreOS because the configuration of existing containers can not be changed. The only way to get this fix is to recreate toolboxes.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -265,8 +265,19 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 	logrus.Debugf("%s canonicalized to %s", currentUser.HomeDir, homeDirEvaled)
 	homeDirMountArg := homeDirEvaled + ":" + homeDirEvaled + ":rslave"
 
+	bootMountFlags := "rw"
+	isBootReadWrite, err := isPathReadWrite("/boot")
+	if err != nil {
+		return err
+	}
+	if !isBootReadWrite {
+		bootMountFlags = "ro"
+	}
+
+	bootMountArg := "/boot:/run/host/boot:" + bootMountFlags + ",rslave"
+
 	usrMountFlags := "ro"
-	isUsrReadWrite, err := isUsrReadWrite()
+	isUsrReadWrite, err := isPathReadWrite("/usr")
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


HarryMichal:
In the future, we will likely take inspiration from https://github.com/cgwalters/coretoolbox and instead of cherry-picking parts of the host filesystem mount the whole `/`. I have a local branch where I already tried this and it seems to work without any refactoring of the code. But let's stick with the current approach until v0.1.0 is out, then we can start experimenting more :).

To explain, this change would allow us to make the code for the `create` command much more simple and give us more control over the mounts because they could be completely handled by the entry-point.

src/cmd/create.go
```diff
@@ -265,8 +265,19 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 	logrus.Debugf("%s canonicalized to %s", currentUser.HomeDir, homeDirEvaled)
 	homeDirMountArg := homeDirEvaled + ":" + homeDirEvaled + ":rslave"
 
+	bootMountFlags := "rw"
```
debarshiray, line 268: I am curious why the logic here for `/boot` is subtly different from that of `/usr` below. Is there something inherently different in the two cases, or is it just a cosmetic difference?

If it's just a cosmetic issue, then it will be good to keep them consistent. The code around mount points and options sometimes requires careful attention while reading, so there's some value in avoiding superficial differences. :)

(For what it's worth, in the absence of a more profound argument, starting from a more conservative "default" value of `ro` and then conditionally switching to `rw` feels safer.)

debarshiray (commented):
Thanks for fixing this and getting it merged! One minor observation:

src/cmd/create.go
debarshiray: See https://github.com/containers/toolbox/pull/804

HarryMichal: You've got a point. Good catch! I'm not sure after such a long time what made me write it like so. I'll keep it in mind in the future.

PR #715, opened 2021-02-27
README: /etc/machine-id is required

xPMo:
See discussion on #710. ArchLinux with a missing machine-id fails, but with a machine-id succeeds.

If this is not supposed to be required, I can open a bug report instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Paths:
 * `/etc/hosts`: optional, if present not a bind mount
 * `/etc/krb5.conf.d`: directory, not a bind mount
 * `/etc/localtime`: optional, if present not a bind mount
+* `/etc/machine-id`: not a bind mount
 * `/etc/resolv.conf`: optional, if present not a bind mount
 * `/etc/timezone`: optional, if present not a bind mount
 


HarryMichal:
Hi @xPMo! Thank you for the patch.

I don't think the file is intended to be a hard requirement. But the requirement is currently set by the way the tool is written. For now I'd add this to the list and in the future we can remove it.

README.md
```diff
@@ -133,6 +133,7 @@ Paths:
 * `/etc/hosts`: optional, if present not a bind mount
 * `/etc/krb5.conf.d`: directory, not a bind mount
 * `/etc/localtime`: optional, if present not a bind mount
+* `/etc/machine-id`: not a bind mount
```
debarshiray, line 136: I think this is a case where we should rather fix the code than add to the list of requirements in the documentation. :)

Specifically in `mountBind` we should add a branch for `fi.Mode().IsRegular()` and `os.Create()` the `containerPath` there, ignoring the error in case the file already existed before.

debarshiray (commented):
Thanks for bringing this up!

README.md
debarshiray: See https://github.com/containers/toolbox/pull/794

PR #716, opened 2021-02-27
profile.d: Make PROMPT_COMMAND a complete command

xPMo:
See discussion on #710. Other users may also append a prompt command in a similar way.
Making this a `:` instead ensures that an appended command will not cause an error.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -60,7 +60,7 @@ if [ -f /run/.containerenv ] \
     if ! [ -f /etc/profile.d/vte.sh ] && [ -z "$PROMPT_COMMAND" ] && [ "${VTE_VERSION:-0}" -ge 3405 ]; then
         case "$TERM" in
             xterm*|vte*)
-                [ -n "${BASH_VERSION:-}" ] && PROMPT_COMMAND=" "
+                [ -n "${BASH_VERSION:-}" ] && PROMPT_COMMAND=":"
                 ;;
         esac
     fi


HarryMichal:
Hi @xPMo! Thank you for the patch.

What do you think, @debarshiray? I don't really understand this shell magic.

xPMo:
It's been a while since I looked at this, and I remember that it's set to a non-empty string to prevent a half-broken workaround in Fedora's bashrc from triggering. (iirc this workaround is now more reliable as of F34). The issue was with it causing problems in Arch's bashrc, which tries to append to a non-empty `PROMPT_COMMAND` by first adding a semicolon to terminate the existing shell command, which does not work if there is no command. Setting `PROMPT_COMMAND=':'` or `PROMPT_COMMAND='true'` would cover both cases.

fwiw, Bash 5.1+ now allows `PROMPT_COMMAND` to be an array parameter explicitly to avoid this kind of mess.

xPMo:
https://bugs.archlinux.org/task/71002

debarshiray:
> It's been a while since I looked at this, and I remember that
> it's set to a non-empty string to prevent a half-broken workaround
> in Fedora's bashrc from triggering. (iirc this workaround is
> now more reliable as of F34).

The workaround was primarily for supporting UBI Toolboxes. It was not aimed at Fedora itself, even though one of the root causes for the problem seen on UBI Toolboxes was due to Fedora's `/etc/bashrc` (and hence all its downstream derivatives) doing [something](https://pagure.io/setup/pull-request/23) that it shouldn't.

The workaround only triggers if `/etc/profile.d/vte.sh` is absent. Is that really absent in Arch's Toolbox image? That file should be present so that VTE-based terminals can continue to track the current working directory.

The workaround is pretty gross, so the only thing that we should be doing is to remove it, once we have an UBI that contains `/etc/profile.d/vte.sh`.

xPMo:
> The workaround only triggers if `/etc/profile.d/vte.sh` is absent. Is that really absent in Arch's Toolbox image? That file should be present so that VTE-based terminals can continue to track the current working directory.

Ah.

```
❯ pacman -F /etc/profile.d/vte.sh
etc/profile.d/vte.sh is owned by extra/vte-common 0.64.2-2
```

I don't use a VTE-based terminal, so I don't have it on my host. It is absent in #710, so I will see about adding it there.

> The workaround is pretty gross, so the only thing that we should be doing is to remove it, once we have an UBI that contains `/etc/profile.d/vte.sh`.

I'm in agreement here, now. Closing in favor of waiting until that happens. Thanks for helping me track this down.

PR #717, opened 2021-02-28
Allow override of tmpfilesdir folder

ClashTheBunny:
* in situations where that directory isn't accessable, it would be nice
  to support installing everything else to a prefix and then instructing
  the user to get those those files installed somehow else.

* I'm not sure if systemd is required if we don't use tmpfilesdir, but
  I'm leaving it outside of the if

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,11 +15,15 @@ go_md2man = find_program('go-md2man')
 shellcheck = find_program('shellcheck', required: false)
 skopeo = find_program('skopeo', required: false)
 
-systemd_dep = dependency('systemd')
 bash_completion = dependency('bash-completion', required: false)
 
 profiledir = get_option('profile_dir')
-tmpfilesdir = systemd_dep.get_pkgconfig_variable('tmpfilesdir')
+
+tmpfilesdir = get_option('tmpfiles_dir')
+if tmpfilesdir == ''
+  systemd_dep = dependency('systemd')
+  tmpfilesdir = systemd_dep.get_pkgconfig_variable('tmpfilesdir')
+endif
 
 if bash_completion.found()
   install_data(
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -4,3 +4,9 @@ option(
   type: 'string',
   value: '/usr/share/profile.d'
 )
+
+option(
+  'tmpfiles_dir',
+  description: 'Directory for system-wide tmpfiles.d(5) files',
+  type: 'string',
+)


ClashTheBunny:
Retest please

meson.build
```diff
@@ -17,7 +17,11 @@ shellcheck = find_program('shellcheck', required: false)
 profiledir = get_option('profile_dir')
 
 systemd_dep = dependency('systemd')
```
debarshiray, line 19: Like you suspected, `systemd_dep` can moved inside the `if` branch because it's otherwise not used.

debarshiray (changes requested):
Thanks for working on this, @ClashTheBunny

meson_options.txt
```diff
@@ -4,3 +4,8 @@ option(
   type: 'string',
   value: '/usr/share/profile.d'
 )
+option(
+  'tmpfilesdir',
```
debarshiray, line 8: It will be good to call it `tmpfiles_dir` for the sake of consistency with the other option (ie., `profile_dir`) in `meson_options.txt`.

In fact, `tmpfiles_dir` is a thing in `systemd.pc`. I wonder why.

meson_options.txt
```diff
@@ -4,3 +4,8 @@ option(
...
+  description: 'Directory for tmpfilesdir files to be installed by root user',
```
debarshiray, line 9: I am curious about the *installed by root user* part. I thought that this option is meant to be useful when installing to a non-system prefix, and to avoid requiring `root` access while doing that.

debarshiray:
I took the liberty to rebase the branch on top of `main` and tweaked it as above before merging.

Thanks for your contribution, @ClashTheBunny !

meson_options.txt
ClashTheBunny: My impression was that there is no current way to configure user specific tmpfiles.d.  They will always need root to be effective.

If that's not the case, this can obviously be updated.  I didn't look into it a ton, so it may totally be possible.

ClashTheBunny: Oh, never mind, it's right there in the synopsis.  I wonder if I was blind to it, or if it's since been better documented:
```
       /etc/tmpfiles.d/*.conf
       /run/tmpfiles.d/*.conf
       /usr/lib/tmpfiles.d/*.conf

       ~/.config/user-tmpfiles.d/*.conf
       $XDG_RUNTIME_DIR/user-tmpfiles.d/*.conf
       ~/.local/share/user-tmpfiles.d/*.conf
       ...
       /usr/share/user-tmpfiles.d/*.conf
```

debarshiray:
Another positive outcome of this commit is that it makes it possible to [dodge having a build dependency](https://src.fedoraproject.org/rpms/toolbox/c/9bebde5bb60f36e38324c1e230e457c495d2d864) on the entirety of systemd during the build.  It can potentially speed up builds.

Issue #718, opened 2021-03-06
Erratic behaviour with toolbox /etc/resolv.conf when changing its symlink target in host


nanonyme:
**Describe the bug**
toolbox seems to resolve the final target of /etc/resolv.conf when creating the image. This is causing a set of weird side-effects

* If /etc/resolv.conf is a real file when creating toolbox and is changed to point to /run/systemd/resolve/stub-resolv.conf after creating toolbox, there will be a broken symlink chain /etc/resolv.conf -> /run/host/etc/resolv.conf -> /run/systemd/resolve/stub-resolv.conf (this should be fixed through https://github.com/containers/toolbox/pull/707)
* If /etc/resolv.conf points to /run/systemd/resolve/stub-resolv.conf and it's change to /run/systemd/resolve/resolv.conf, the change is not reflected inside toolbox at all

**Steps how to reproduce the behaviour**
1. sudo rm /etc/resolv.conf && sudo touch /etc/resolv.conf
2. toolbox enter
3. readlink -f /etc/resolv.conf 
4. exit
5. sudo ln -sf /run/systemd/resolve/stub-resolv.conf /etc/resolv.conf
6. toolbox enter
7. readlink -f /etc/resolv.conf 

**Expected behaviour**
Step 3 outputs /run/host/etc/resolv.conf, step 7 outputs /run/systemd/resolv/stub-resolv.conf

**Actual behaviour**
Step 3 outputs /run/host/etc/resolv.conf, step 7 outputs nothing because /etc/resolv.conf is now a symlink to a broken symlink

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.99.1`

**Toolbox package info (`rpm -q toolbox`)**
`toolbox-0.0.99.1-1.fc34.x86_64`

**Output of `podman version`**
```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.16
Built:        Mon Feb 22 16:08:57 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-3.0.1-1.fc34.x86_64`

**Info about your OS**
Fedora SIlverblue 34

nanonyme:
@debarshiray any thoughts? What I would suggest after the bind mount is that we stop resolving /etc/resolv.conf further. Just allow it to always be a symlink to /run/host/resolv.conf and things should work consistently.

debarshiray:
[...]

PR #719, opened 2021-03-06
Always make /etc/resolv.conf point to host /etc/resolv.conf

nanonyme:
Attempt to fix https://github.com/containers/toolbox/issues/718

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -176,12 +176,10 @@ func initContainer(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			if _, err := os.Readlink("/etc/resolv.conf"); err != nil {
-				if err := redirectPath("/etc/resolv.conf",
-					"/run/host/etc/resolv.conf",
-					false); err != nil {
-					return err
-				}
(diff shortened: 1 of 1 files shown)


src/cmd/initContainer.go
```diff
@@ -176,12 +176,10 @@ func initContainer(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			if _, err := os.Readlink("/etc/resolv.conf"); err != nil {
-				if err := redirectPath("/etc/resolv.conf",
-					"/run/host/etc/resolv.conf",
-					false); err != nil {
-					return err
-				}
+			if err := redirectPath("/etc/resolv.conf",
```
debarshiray, line 179: One problem with doing this is that it doesn't leave users with any way to locally tweak the `/etc/resolv.conf` of their containers.

debarshiray (commented):
Thanks for working on this!

I left some comments on the original issue at https://github.com/containers/toolbox/issues/718

debarshiray:
Tentatively closing this pull request, because I think the original [issue](https://github.com/containers/toolbox/issues/718) went away.

PR #723, opened 2021-03-15
images: Use a regular file, not a symbolic link, for README.md

debarshiray:
A while ago, 'podman build' stopped supporting COPY with relative
symbolic links [1]. Therefore, these image definitions can't be used
without first temporarily removing the symbolic links, which is
annoying.

The downside is that the copies of README.md now has to be separately
updated, which isn't that big of a hassle compared to the problem that
it fixes.

[1] https://github.com/containers/buildah/issues/1952


HarryMichal (approved):

debarshiray:
Thanks, @HarryMichal

PR #724, opened 2021-03-19
cmd/initContainer: Ignore unknown flags

HarryMichal:
The init-container command uses several flags. In the future we'd like
to minimize their number. In order to be able to do that without
breaking systems with older versions of Toolbox, the command can't error
out due to usage of unknown flags.

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -122,6 +122,7 @@ func init() {
 		"Create a user inside the toolbox container whose login name is USER")
 	initContainerCmd.MarkFlagRequired("user")
 
+	initContainerCmd.FParseErrWhitelist.UnknownFlags = true
 	initContainerCmd.SetHelpFunc(initContainerHelp)
 	rootCmd.AddCommand(initContainerCmd)
 }


HarryMichal:
recheck

src/cmd/initContainer.go
```diff
@@ -122,6 +122,7 @@ func init() {
 		"Create a user inside the toolbox container whose login name is USER")
 	initContainerCmd.MarkFlagRequired("user")
 
+	initContainerCmd.FParseErrWhitelist.UnknownFlags = true
```
debarshiray, line 125: Umm... is this really necessary?

We always bind mount the same `toolbox` binary that's being used on the host into the container. We also don't support online updates of Toolbox. That means that a change in the Toolbox version will always be accompanied by a reboot (and restart of existing containers).

So, if or when we reduce the number of `init-container` flags, we can always update the rest of the command line parsing logic as necessary (eg., by using `FParseErrWhitelist.UnknownFlags`), and both the host and existing containers will get updated in tandem.

The thing that's immutable about a container is its JSON configuration with the invocation of the entry-point (as specified during `podman create`) embedded in it, not the `toolbox` binary that's present inside the container.

debarshiray: See https://github.com/containers/toolbox/pull/807

PR #727, opened 2021-03-20
cmd/create: Start spinner earlier

HarryMichal:
Showing spinner after a lot of work on creating a toolbox is done (even
though not really time consuming) does not make much sense.

When a spinner is started successfully, a stop command is deferred.
There's no need to stop it additionally.


debarshiray:
We ended up reverting this in https://github.com/containers/toolbox/pull/746

PR #731, opened 2021-03-21
playbooks: Properly separate builddir setup & build

HarryMichal:
Call "meson builddir" makes Meson create a build directory called
"builddir". It does not make it build the project. A subsequent call to
"meson compile" or "ninja" needs to be made. This subtle detail causes
a minor (purely visual) discrepancy in the CI output. Fix this for both
unit-test & system-test job definitions.

### DIFF
--- a/playbooks/system-test.yaml
+++ b/playbooks/system-test.yaml
@@ -5,13 +5,18 @@
     toolbox_bin: '/usr/local/bin/toolbox'
 
   tasks:
-    - name: Build toolbox
+    - name: Set up build directory
       command: meson builddir
+      args:
+        chdir: '{{ zuul.project.src_dir }}'
+
+    - name: Build Toolbox
+      command: ninja -C builddir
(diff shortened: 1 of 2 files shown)


playbooks/system-test.yaml
```diff
@@ -5,13 +5,18 @@
...
+    - name: Set up build directory
       command: meson builddir
       args:
         chdir: '{{ zuul.project.src_dir }}'
+
+    - name: Build Toolbox
+      command: ninja -C builddir
+      args:
+        chdir: '{{ zuul.project.src_dir }}'
         creates: builddir/src/toolbox
```
debarshiray, line 17: I am a bit surprised that this `creates` wasn't causing problems so far. Since `meson builddir` doesn't generate the binary, I would have expected it to complain.

debarshiray (commented):
Thanks for cleaning this up!

PR #732, opened 2021-03-21
meson: Rearrange lines & check for Skopeo

HarryMichal:
Skopeo is now a dependency for running system tests[0]. It is not a hard
dependency but at least give the user a heads-up.


meson.build
```diff
@@ -13,10 +13,12 @@ add_project_link_arguments('-pthread', language: 'c')
 go = find_program('go')
 go_md2man = find_program('go-md2man')
 shellcheck = find_program('shellcheck', required: false)
-
-profiledir = get_option('profile_dir')
+skopeo = find_program('skopeo', required: false)
 
 systemd_dep = dependency('systemd')
+bash_completion = dependency('bash-completion', required: false)
```
debarshiray, line 19: Nitpick: if the intention was to ensure alphabetical ordering, then shouldn't `bash_completion` be above `systemd` ? :)

PR #733, opened 2021-03-21
build: Drop ShellCheck on Shell Toolbox

HarryMichal:
Shell Toolbox has been replaced by the Go implementation a quite while
ago. It is kept in the repository but is no longer actively developed.
There is no need to continue checking it with ShellCheck.


debarshiray:
Sadly, we need to continue running ShellCheck on the POSIX shell implementation.  See https://github.com/containers/toolbox/pull/1094

PR #735, opened 2021-03-22
cmd/root: Streamline preRun of CLI root

HarryMichal:


src/cmd/root.go
```diff
@@ -100,26 +95,24 @@ func init() {
 }
 
 func preRun(cmd *cobra.Command, args []string) error {
+	var toolboxPath string
```
debarshiray, line 98: Does `toolboxPath` really need to be defined separately?

debarshiray (commented):
This is a nice clean-up. Thanks.

src/cmd/root.go
debarshiray: See https://github.com/containers/toolbox/pull/797

src/cmd/root.go
```diff
@@ -62,11 +62,6 @@ var (
 )
 
 func Execute() {
-	if err := setUpGlobals(); err != nil {
```
debarshiray, line 65: I think this does actually need to be here. The reason is a bit convoluted and not obvious at all, but has got to do with the quirks of Cobra, which probably can't be fixed without breaking API.

Even though we set `SilenceUsage` to `true` to have full control over what gets shown in the case of an error, there is still (at least?) one occasion in which we use the usage function set using `SetUsageFunc` (ie., `rootUsage`) - when an unknown flag is used. eg., `toolbox --foo`. Oddly enough, an unknown command won't lead to `rootUsage`. eg., `toolbox foo`.

Since `rootUsage` uses `executableBase`, that variable needs to be set earlier, which means that `setUpGlobals` needs to run before `rootUsage`. It turns out that the `PersistentPreRunE` hook (ie., `preRun`) doesn't get invoked when an unknown flag is encountered. Therefore, we can't put `setUpGlobals` inside `preRun`.

I guess, we should write a test for this now. :)

HarryMichal: Yes, you are right about this being convoluted. I did not realize at the time I merged this the possible problem. I'm having problems with this while wiring `toolbox` to `toolbox enter`.

The problem with `executableBase` not being set until `setUpGlobals` is run is solved by https://github.com/containers/toolbox/pull/775.

debarshiray: See https://github.com/containers/toolbox/pull/802

PR #737, opened 2021-03-22
cmd/rm,rmi: Use function for getting list of images/containers

HarryMichal:
Instead of using own ways of getting all toolbox images/containers (and
thus unnecessary duplication of code) use a dedicated function.

### DIFF
--- a/src/cmd/rm.go
+++ b/src/cmd/rm.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/containers/toolbox/pkg/podman"
 	"github.com/containers/toolbox/pkg/utils"
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


src/cmd/rm.go
```diff
@@ -70,31 +69,15 @@ func rm(cmd *cobra.Command, args []string) error {
...
-		containers_old, err := podman.GetContainers(args...)
-		if err != nil {
-			return fmt.Errorf("failed to list containers with label=com.github.containers.toolbox=true: %w", err)
-		}
+		var toolboxContainers []toolboxContainer
 
-		logrus.Debug("Fetching containers with label=com.github.debarshiray.toolbox=true")
-		args = []string{"--all", "--filter", "label=com.github.debarshiray.toolbox=true"}
-		containers_new, err := podman.GetContainers(args...)
+		toolboxContainers, err := listContainers()
```
debarshiray, line 74: Does `toolboxContainers` really have to be defined separately?

src/cmd/rmi.go
```diff
@@ -70,33 +69,15 @@ func rmi(cmd *cobra.Command, args []string) error {
...
-		images_old, err := podman.GetImages(args...)
-		if err != nil {
-			return fmt.Errorf("failed to list images with label=com.github.containers.toolbox=true: %w", err)
-		}
+		var toolboxImages []toolboxImage
 
-		logrus.Debug("Fetching images with label=com.github.debarshiray.toolbox=true")
-		args = []string{"--filter", "label=com.github.debarshiray.toolbox=true"}
-		images_new, err := podman.GetImages(args...)
+		toolboxImages, err := listImages()
```
debarshiray, line 74: Ditto about `toolboxImages`.

debarshiray (commented):
This is a nice clean-up indeed.

src/cmd/rm.go
debarshiray: See https://github.com/containers/toolbox/pull/798

src/cmd/rmi.go
debarshiray: See https://github.com/containers/toolbox/pull/798

PR #739, opened 2021-03-23
utils: bump Fedora fallback version to 33

juhp:
32 is approaching EOL

see also #546


HarryMichal:
Thank you for the patch, @juhp!

PR #742, opened 2021-03-24
Add a basic Debian 10 toolbox image

andrewaylett:
This is largely a copy-and-paste of the Fedora environment.  I've not used it particularly extensively, but `toolbox create` and `toolbox enter` both work without reporting errors.


hyperreal64:
> This is largely a copy-and-paste of the Fedora environment. I've not used it particularly extensively, but `toolbox create` and `toolbox enter` both work without reporting errors.

Thank you for creating this! I've tried it and found that the `capsh` program needs to be available in the container, otherwise you can't enter it and you get the error "/bin/bash not found in container". Adding the libcap2-bin package to extra-packages fixes this.

andrewshadura:
See also #371.

travier:

PR #757, opened 2021-04-30
Rename Dockerfile to Containerfile

runlevel5:
## Context

Because the container images are being built with `podman` instead of `docker`. I think it is a good practice to rename `Dockerfile` to `Containerfile`.


HarryMichal:
Hi @runlevel5! Thank you for the patch. Could you, please, add a link to this PR into the commit message? Otherwise the patch looks good.

runlevel5:
@HarryMichal I've just amended this PR

HarryMichal:
Great. Thank you!

Issue #758, opened 2021-05-01
Easy way to have multiple toolboxes for different purposes.


emiljoha:
**Is your feature request related to a problem? Please describe.**
My one single toolbox container becomes what I want to get away from by using an immutable desktop, a big mess.

**Describe the solution you'd like**
A way to have multiple reproducible toolboxes for different purposes.

**Describe alternatives you've considered**
The different toolboxes could be defined in the configuration file and point to either repository images or Containerfiles.

**Additional context**
This would be an important feature if I would pitch the developer team I work on to start using toolbox. The IT department can manage the underlying OS and run the same thing on every computer in the company, while the developer teams can standardize within the team on a common environment matching production. But still keeping the flexibility of individual team members to make change to their setup while still making it possible to "nuke and pave" and get back to the "official" dev environment.

pierreprinetti:
I think that a corporate-managed ostree image is a bad idea. It's a security nightmare (who signs the ca-cert package?), a bad practice (will IT choose my editor? and what if our customers use a different browser that what's mandated by the corp?) and a bit dystopian if I may.

In order to give your engineers a base set of common tools, you could instead reverse your design and distribute a Toolbox container image for your project. Everyone would still be using the publicly audited ostree image of your OS of choice, but they would have the choice to use your Toolbox image for development.

You can create a repo containing a Dockerfile here on Github (or your corporate git server), build it in Quay (or Docker Hub, or a private registry), and let your engineers use it.

Example Dockerfile:

```Dockerfile
FROM registry.fedoraproject.org/f34/fedora-toolbox:34

LABEL maintainer="You <you@yourcompany.example.com>"

RUN dnf install -y zsh tmux autojump-zsh fd-find ripgrep

RUN echo 'export ZDOTDIR="${XDG_CONFIG_HOME:-${HOME}/.config}/zsh"' >> /etc/zshenv
```

Individual engineers can then:
```
toolbox create -i quay.io/your-company/project-toolbox:34 project
toolbox enter project
```

emiljoha:
> A corporate-managed ostree image can become a security nightmare (who signs the ca-cert package?)

Not sure I follow. Are not most installations on company owned computers corporate-managed in some form or another? What would be the different from a security perspective in a immutable system?

> a bad practice (will IT choose my editor? and what if our customers use a different browser that what's mandated by the corp?) and overall a maintenance burden for your IT department.

[...]

Issue #759, opened 2021-05-06
enable project wiki?


juhp:
https://github.com/containers/toolbox/wiki doesn't seem to work?
It could be useful for documenting various tweaks/hacks/OSes etc

HarryMichal:
Maybe we could even use it as the main wiki... It's just a matter of writing the docs :)

juhp:
Thank you for making this available finally!

debarshiray:
*Summarizing my earlier discussion with @HarryMichal and @olivergs from a few months ago.  I realized that I never commented here.  I apologize for that.*

What's the purpose of having a wiki?

I think that a wiki is not a good idea at this point, because it adds yet another source of truth for the documentation.  In short, we should polish our existing documentation instead of adding more and more sources for it that diverge over time and we can no longer keep track off.

Over time, our documentation has fractured across multiple wikis and platforms.  For a project that's relatively new and small, this is a problem, because it prevents us from quickly changing things, whether it's our branding, our command line interface, the direction of the project, etc..

Here are the other wikis and documentation platforms that we have used in the past:
* https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/
* https://fedoraproject.org/wiki/Workstation/Desktop_Containers

... and I think we should take them down after finding ways to absorb them into our documentation.

This is why, we had decided in one of our calls that we will consolidate as much as possible (if not everything) in our manuals.  Our manuals are written in Markdown, so they are web-friendly, and are shipped as part of Toolbox itself.  The `README.md` is the landing page for the GitHub repository, so we can find ways to lead the reader from further into the documentation.  We also have a `CONTRIBUTING.md`.

This is not without precedent. eg., look at systemd.  They seem to be able to pack a significant amount of information in their [manuals](https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html), and systemd is a much bigger project than Toolbox.  In fact, as per their [README.md](https://github.com/systemd/systemd/) systemd seems to have deprecated their [wiki](https://github.com/systemd/systemd/).

Another example is Flatpak, which is also a much bigger project than Toolbox.

On top of that, Toolbox now has a [website](https://containertoolbx.org/), and right now, our wiki only repeats what's already there.

[...]

PR #767, opened 2021-05-13
Add more usage examples to README.md

emiljoha:
These are things that I did not know was possible until it was pointed out to
me by @pierreprinetti in https://github.com/containers/toolbox/issues/758

I actually looked at the `--help` output and played around a while but gave up thinking this kind of use was not possible.

Use the opensuse toolbox image in the examples partly as that is what I am comfortable with but
also because I suspect that the mention of another distro sets the feeling that this
is not just for Fedora.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ This will create a container called `fedora-toolbox-<version-id>`.
 
 ### Define toolbox container with a Containerfile:
 
-The parent image must satisfy the [image requirements](##Image-requirements).
+The parent image must satisfy the [image requirements](#Image-requirements).
 
 ```Dockerfile
 FROM registry.opensuse.org/opensuse/toolbox
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ This will create a container called `fedora-toolbox-<version-id>`.
 
 ### Specify image for toolbox container:
 ```console
-[user@hostname ~]$ toolbox create --image registry.opensuse.org/opensuse/toolbox
+[user@hostname ~]$ toolbox create --image registry.fedoraproject.org/f34/fedora-toolbox:34 fedora34-toolbox
 [user@hostname ~]$
 ```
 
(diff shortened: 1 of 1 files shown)


emiljoha:
Just realized that it seems like the toolbox utility I am using is this one https://github.com/kubic-project/microos-toolbox :innocent:  The two utilities are very similar so I still think that this type of functionality should be better advertised. Will have to look over if my examples actually work for this version of the tool.

emiljoha:
Tested my changes in a Boxes virtual machine of fedora sliverblue and made appropriate changes.

HarryMichal:
Hi @emiljoha! Thank you for the PR. We try to keep the README as short as possible. Aren't man pages of Toolbox a better place for usage examples? Or an online wiki (#759)?

I'd personally go in the direction of adding usage examples to the man pages. What do you think?

emiljoha:
> We try to keep the README as short as possible. Aren't man pages of Toolbox a better place for usage examples? Or an online wiki (#759)?

Personally the README is the first place I look for the general idea of what the software can be used for. Then the man-pages/--help I use to remember how to do that thing I know I can do with the tool.

But I of course I will happily respect the direction of the project. I have seen examples in man pages/--help and that can be very helpful, and there is value in having documentation shipped with the tool.

Maybe move the actual examples to the man-page/--help but add a mention that more usage examples are provided there?

HarryMichal:
> Personally the README is the first place I look for the general idea of what the software can be used for.

I'd say the README already gives a rough idea of what the usage is like. I don't think putting in more than 3/4 examples is good because then the file could get too long. But adding in a reference to other places (man pages/wiki) would users like you give an idea where to go next instead of searching for clues.

debarshiray (commented):
Thanks for the patch!

Most of this has been part of the manuals for the `create` and `enter` commands for a while.  Unlike the `README.md` the manuals are installed on user machines as part of Toolbx, and that's historically where people look for documentation on command line tools.

debarshiray:
I am closing this pull request as per the above comments.

However, thanks for stopping by and trying out Toolbx!

PR #771, opened 2021-05-20
Make toolboxes use the container name as hostname.

runfalk:
This allows the user to easily see which container they are currently in instead of the generic @toolbox.

I use multiple toolboxes, tied to the projects I'm working on. When I have many open at the same time it's hard for me to know which one I'm actually in since they all have the same hostname (`toolbox`). This changes things so that the container re-uses the container name as host name.

An alternative solution would be to prefix the prompt like Python virtualenvs do: `(fedora-toolbox-34) user@toolbox:`. Since this change wasn't very invasive I opted to do it without discussion.


runfalk:
It looks something like this:

```
runfalk@hostname:~/dev/toolbox (main)$ ./builddir/src/toolbox create --distro fedora --release f34
Download registry.fedoraproject.org/fedora-toolbox:34 (500MB)? [y/N]: y
Created container: fedora-toolbox-34
Enter with: toolbox enter
runfalk@hostname:~/dev/toolbox (main)$ toolbox enter
runfalk@fedora-toolbox-34:~/dev/toolbox (main)$
```

README.md
```diff
@@ -47,14 +47,14 @@ systems it's just a matter of installing the `toolbox` package.
 [user@hostname ~]$ toolbox create
 Created container: fedora-toolbox-33
 Enter with: toolbox enter
-[user@hostname ~]$
+[user@fedora-toolbox-33 ~]$
```
HarryMichal, line 50: In this case the hostname does not change.

images/fedora/f32/README.md
```diff
@@ -47,14 +47,14 @@ systems it's just a matter of installing the `toolbox` package.
 [user@hostname ~]$ toolbox create
 Created container: fedora-toolbox-33
 Enter with: toolbox enter
-[user@hostname ~]$
+[user@fedora-toolbox-33 ~]$
```
HarryMichal, line 50: Ditto.

images/fedora/f33/README.md
```diff
@@ -47,14 +47,14 @@ systems it's just a matter of installing the `toolbox` package.
 [user@hostname ~]$ toolbox create
 Created container: fedora-toolbox-33
 Enter with: toolbox enter
-[user@hostname ~]$
+[user@fedora-toolbox-33 ~]$
```
HarryMichal, line 50: Ditto.

images/fedora/f34/README.md
```diff
@@ -47,14 +47,14 @@ systems it's just a matter of installing the `toolbox` package.
 [user@hostname ~]$ toolbox create
 Created container: fedora-toolbox-33
 Enter with: toolbox enter
-[user@hostname ~]$
+[user@fedora-toolbox-33 ~]$
```
HarryMichal, line 50: Ditto.

HarryMichal (changes requested):
Found a mistake in changes of examples in README.

runfalk:
Good spotting. Fixed!

HarryMichal:
Thank you, @runfalk, for the patch! I'm not entirely sure about this change as it is not as minor as it would seem. I'd like to know what @debarshiray thinks about this. I'm inclined to accepting this but @debarshiray may have some more insight.

runfalk:
I don't think this requires toolbox recreation (label 6). I tested this change with a pre-change toolbox and everything I tested worked just fine. This only changes new toolboxes. Is the hostname essential to anything in toolbox itself?

HarryMichal:
I used the label because the change does not affect existing containers. Those already have the hostname set. Therefore, containers have to be recreated with the version of Toolbox that has this patch.

runfalk:
The point I wanted to make is that they don't have to be recreated, as they work perfectly fine even with the patched `toolbox`.

HarryMichal:
I found myself thinking about this today and realized this may be incompatible with #784. It is a classic problem with containers: the configuration is baked in and can not be changed. In order to make this work across container renames, some other mechanism should exist.

mattweidner:
Perhaps a compromise would be to add a toolbox create option (--hostname) that is passed through to podman to set the container hostname to an arbitrary string. If no --hostname option is provided, the default string "toolbox" could be used. The user could choose to set the hostname the same as the container name or something else, but there would be no implicit expectation for the hostname to change when the container is renamed. This would mirror the operation of podman, allow users to have a choice in their container's hostname, and potentially deconflict #784.

dabrain34:
would be good to have a prefix with toolbox always because you name your container without toolbox in the name, you might be confused too.
"--hostname", "toolbox-" + container,

cig0:
I'm happy to see a whole lot of other people would love to see this feature implemented :)

Today I quickly put together the following workaround:

```shell
C=(ansible aws gcp python)

for i in "${C[@]}";
    do
        toolbox create -c $i \
        && toolbox run -c $i sudo hostname $i \
        && toolbox run -c $i hostname \
        && echo -e "\n\n";
    done
```

Which provides some relief (!?) to me:

image

shoop:
I have the opposite request, I don't understand why the **hostname** needs to be set. I fully support a mechanism to show which container is active, but I have issues with the hostname set to something different than the original host:

1. Configuration software like [chezmoi](https://github.com/twpayne/chezmoi) expects to be able to distinguish on hostname.
2. Network software now suddenly needs to be able to resolve a not-really-existing hostname. Even the workaround proposed to use `containername.hostname` is fabricating a non-existing network name, really, as the toolbox itself does not have a separate network.

So I'd like to have the containername in `/run/.toolboxenv` and then have bash prompts etc be updated to use that, instead of messing around with what is in my view a system property.

buck-ross:
@shoop I have an alternative proposition, but first, I'd like to take the time to address the concerns you've enumerated:

> Configuration software like chezmoi expects to be able to distinguish on hostname

This is true, but many software products with this requirement tend to require only a distinct hostname, not necessarily a DNS resolvable one. As such, it can become problematic if multiple toolbox containers, running the same software package, are given the same hostname (I.E, the host's hostname), meanwhile it seems entirely unnecessary to place the burden on the toolbox devs to ensure that the hostname is always DNS resolvable.

> Network software now suddenly needs to be able to resolve a not-really-existing hostname

What about this is sudden? All toolbox containers currently have the hostname `toolbox`, which is presumably never network-resolvable. What we really need is a way to give some containers the same hostname as the host (or other resolvable names), and others may have other hostnames, which may or may not be resolvable.

> So I'd like to have the containername in /run/.toolboxenv

This is already provided in `/run/.containerenv`. No need to modify `/run/.toolboxenv`.

buck-ross:
I propose that instead of having the hostname always set to the container name, or always set to the host's hostname, we should instead give the user the option of setting the hostname to some (sane) value when invoking `toolbox create` with an optional flag, like `--hostname`. This was already proposed in #210, but was recently closed due to the author seemingly being unwilling to re-write their changes in Go.

I've already isolated the exact line which needs to be changed ([create.go#L395](https://github.com/containers/toolbox/blob/main/src/cmd/create.go#L395)), and have a plan to add this new flag on my own fork as soon as I get the chance. Assuming this idea is well received, I'll make my own PR in a few weeks, once I've had a chance to write the new feature & all associated test-cases.

debarshiray:
> The point I wanted to make is that they don't have to be
> recreated, as they work perfectly fine even with the
> patched `toolbox`.

What do you mean?  As far as I can see, this pull request only changes the `podman create ...` invocation for creating new Toolbx containers.  So how will it automatically update the hostnames of existing containers?

src/cmd/create.go
```diff
@@ -401,7 +401,7 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 	createArgs = append(createArgs, xdgRuntimeDirEnv...)
 
 	createArgs = append(createArgs, []string{
-		"--hostname", "toolbox",
+		"--hostname", container,
```
debarshiray, line 404: I suspect that the rules for a valid hostname are different from the rules for a valid OCI container name.

debarshiray (commented):
To improve the default hostname of the containers, we should implement https://github.com/containers/toolbox/issues/969

When it comes to offering better ways to visually distinguish Toolbx containers from one another, see https://github.com/containers/toolbox/issues/98 for some ideas.  eg., there's metadata in `/run/.containerenv` that can be used by users to customize their shell prompts without requiring changes in Toolbx itself.

debarshiray:
I am going to have to close this pull request.

However, thanks for your interest in Toolbx and contributing to the discussion.

debarshiray:
> I'm happy to see I'm not the only one that would love to see this feature implemented :)
> 
> Today I quickly put together the following workaround:
> 
> ```shell
> C=(ansible aws gcp python)
> 
> for i in "${C[@]}";
>     do
>         toolbox create -c $i \
>         && toolbox run -c $i sudo hostname $i \
>         && toolbox run -c $i hostname \
>         && echo -e "\n\n";
>     done
> ```
> 
> Which provides some relief (!?) to me:
> 
> image

@mildred has [another way](https://github.com/containers/toolbox/issues/98#issuecomment-518567193) to do this via a start-up script.

debarshiray:
> I am going to have to close this pull request.
> 
> However, thanks for your interest in Toolbx and contributing to the discussion.

You are also welcome to help implement https://github.com/containers/toolbox/issues/969

PR #773, opened 2021-05-24
Use buildah for building images

runfalk:
This allows us to build all Fedora images using the same script. Run it by calling `images/build-fedora` from the repository root.

This also gets rid of the duplicated README files.

I marked this as draft since I haven't considered what other places that might need updating.


runfalk:
If you have a look and think this is something you want I'm happy to put more work into it. I'll keep it on pause until then.

debarshiray (commented):
What's the motivation for this?

The `images` sub-directory is meant to offer the sources for the Toolbox images, so that anybody hacking on Toolbox has a single reference point to all the images and can quickly rebuild an image locally, if necessary. Fedora's OCI image infrastructure only works with Dockerfiles, even though `buildah` does offer other ways to build images, and that's why the images are defined the way they are.

A significant change in the way the OCI images are defined in Fedora's build infrastructure and in the upstream `toolbox.git` will make it very difficult to keep them synchronized as they evolve.

runfalk:
The main motivation was #771, where I needed to update all existing README files because they're duplicated for every image. The current dockerfiles also has lots of `RUN` commands which each produces a new layer as I understand. For instance the `RUN dnf clean all` will delete the dnf cache for the image but it'll still get downloaded as part of the previous layers, so there is litte benefit in removing it. Using `buildah` makes it simple to avoid that since everything gets put in one layer.

Ideally I'd like to have more buildah scripts in there for other distros like Ubuntu. Even if they aren't officially distributed it'd be very helpful for people looking to set up their own. I personally use toolbox with Ubuntu for work since the official dev environment is Ubuntu 18.04.

> A significant change in the way the OCI images are defined in Fedora's build infrastructure and in the upstream toolbox.git will make it very difficult to keep them synchronized as they evolve.

I made this a draft precisely because it's pretty invasive. I didn't know this would have big implications to the existing workflow, or that there was an upstream repository for this. If you don't think it should go in here feel free to close it.

debarshiray:
> I needed to update all existing README files because they're
> duplicated for every image.

Well, there's a [reason](https://github.com/containers/toolbox/commit/a8c6fd19d8cb5f35567cb8a94e79e20a1b9b19f0) for that. :)

The `README.md` is present as part of the Fedora image definitions because the [Fedora Container Guidelines](https://docs.fedoraproject.org/en-US/containers/guidelines/guidelines/) used to require them. So, we started by including the main `README.md`, because ... why not? :)

If the current iteration of the guidelines don't require a `README.md`, then we can simply remove them.

> The current dockerfiles also has lots of RUN commands which each
> produces a new layer as I understand. For instance the RUN dnf
> clean all will delete the dnf cache for the image but it'll still get
> downloaded as part of the previous layers, so there is litte benefit in
> removing it.

Not if `podman build --squash ...` is used. The Fedora OCI image infrastructure also squashes all the layers into a single one.

> Ideally I'd like to have more buildah scripts in there for other distros like
> Ubuntu. Even if they aren't officially distributed it'd be very helpful for
> people looking to set up their own. I personally use toolbox with Ubuntu
> for work since the official dev environment is Ubuntu 18.04.

For other operating system distributions we should use whichever format is preferred by the infrastructure that's used to distribute the images.

runfalk:
Fair enough! Since it doesn't seem like the desired approach I'll close this.

debarshiray:
Ok, thanks for understanding. I still appreciate your participation in the project, though.

Thanks again!

PR #774, opened 2021-05-25
test/system: Decouple image caching from Zuul

HarryMichal:
See commit for description.

Fixes #751

PR #776, opened 2021-05-26
cmd/list: Recognize UBI images

### LINKED ISSUE #753: UBI images not listed as toolbox images
**Describe the bug**
When a [UBI]() image is present on a host machine, it is not listed in the output of `toolbox list`.

1. Pull an UBI image - `podman pull registry.access.redhat.com/ubi8/ubi`
2. `toolbox list`

**Expected behaviour**

UBI image is shown in the output.

**Actual behaviour**

UBI image is not shown in the output.

`toolbox version 0.0.99.1`

`toolbox-0.0.99.1-1.fc35.x86_64`

```
Version:      3.1.0-dev
API Version:  3.0.0
Go Version:   go1.16
Built:        Mon Feb 22 23:40:29 2021
OS/Arch:      linux/amd64
```

`podman-3.1.0-0.42.dev.gitd999328.fc35.x86_64`

**Info about your OS**
Fedora Silverblue Rawhide

HarryMichal:
Requires #774 

Fixes #753

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -52,11 +52,18 @@ var (
 		onlyImages     bool
 	}
 
-	// toolboxLabels holds labels used by containers/images that mark them as compatible with Toolbox
-	toolboxLabels = map[string]string{
+	// toolboxImageLabels holds labels used by images that mark them as compatible with Toolbox
+	toolboxImageLabels = map[string]string{
 		"com.github.debarshiray.toolbox": "true",
 		"com.github.containers.toolbox":  "true",
+        "com.redhat.component": "ubi8-container",
 	}
+
+    // toolboxContainerLabels holds labels used by container that mark them as compatible with Toolbox
+    toolboxContainerLabels = map[string]string{
+        "com.github.debarshiray.toolbox": "true",
+        "com.github.containrs.toolbox": "true",
+    }
 )
 
 var listCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


src/cmd/list.go
```diff
@@ -52,11 +52,18 @@ var (
...
+	toolboxImageLabels = map[string]string{
 		"com.github.debarshiray.toolbox": "true",
 		"com.github.containers.toolbox":  "true",
+        "com.redhat.component": "ubi8-container",
 	}
+
+    // toolboxContainerLabels holds labels used by container that mark them as compatible with Toolbox
+    toolboxContainerLabels = map[string]string{
+        "com.github.debarshiray.toolbox": "true",
+        "com.github.containrs.toolbox": "true",
```
debarshiray, line 65: Nitpick: typo alert! `containrs`.

debarshiray (commented):
Let's hold this back for a bit.

I understand that not listing the corresponding image for UBI Toolbox containers is a rough edge. However, the whole UBI feature is a bit experimental. The concept of Toolbox, even if it's only a Fedora container on a Fedora host, itself is quite new. With UBI we are talking about a free RHEL environment created just like that on any Linux distribution.

So it's ok, if we take a release or two to hammer out the details. Especially since we are very likely going to get a Toolbox-specific image that's created out of the UBI RPM repositories.

There's also the issue that we really want to get 0.1.0 done, and we really need to keep the churn down for that. Changing the labels can very likely lead to backwards compatibility issues in the future, where we get stuck with a non-ideal label for a while. We managed to get rid of some our past labels, so let's not add more to the list.

In short, this bug will likely fix itself in the coming months. So let's not get ourselves into a corner trying to rush through a fix.

debarshiray:
> Let's hold this back for a bit.

See https://github.com/containers/toolbox/pull/792

PR #777, opened 2021-05-26
test/system: Use env var for invoking Toolbox

HarryMichal:
The system test refactor[0] replaced the 'run_toolbox' helper function
with 'run toolbox', which is a normal invocation of Toolbox. This makes
it impossible to override Toolbox used during the tests using env var.

[0] https://github.com/containers/toolbox/pull/693

### DIFF
--- a/test/system/001-version.bats
+++ b/test/system/001-version.bats
@@ -2,9 +2,10 @@
 
 load 'libs/bats-support/load'
 load 'libs/bats-assert/load'
+load 'libs/helpers.bash'
 
 @test "version: Check version using option --version" {
-  run toolbox --version
+  run $TOOLBOX --version
 
   assert_output --regexp '^toolbox version [0-9]+\.[0-9]+\.[0-9]+(\.[0-9]+)?$'
 }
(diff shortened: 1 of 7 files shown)


debarshiray (commented):
Thanks for cleaning this up!

The commit message is missing a link to this PR, though. It's possible to find it by looking up the commit in GitHub's web interface, but that's one extra step that starts to become a hassle after some time.

Also, when referring to other commits in the commit message, it's convenient to include the actual commit ID. Only having the link to the pull request or the issue makes one go through a few more hoops to look at the commit - open the PR or issue in the browser, and look at the commit in the GitHub UI. Given that the reader was looking at the original commit in the terminal, having to switch back and forth between the terminal and browser becomes annoying - the GitHub UI and raw Git formats the commits differently, etc.. It's not impossible but the minor irrations begin to add up. :)

Here are some patterns that I find a bit more convenient:
* https://github.com/containers/toolbox/commit/7973181136494a4ba147808c9ad51ace9aa4305f
* https://github.com/containers/toolbox/commit/6543ff6a0665bea93a1869b6a830c17e85eb8f61
* https://github.com/containers/toolbox/commit/0c08a42ac846cf29fd026f5f0c1499bfa8220978

In short, we write the commit message only once; but in case someone looks it up later, that's usually because the person is either trying to quickly audit the changes in a new release or hunting a bug, and in those cases it's really worth taking the extra trouble to make it easy for the reader.

PR #778, opened 2021-05-26
test/system: Don't use sudo for executing commands as root

HarryMichal:
The use of sudo when executing commands in a toolbox using 'toolbox run'
causes problem in the Zuul environment. I don't know what is causing
that but it makes the CI unreliable. Replace the use of sudo with
'su -c'.

Supersedes #705


HarryMichal:
recheck

HarryMichal:
So that didn't solve the issue :/.

PR #780, opened 2021-05-26
test/system: Bump secondary fedora image from 29 to 32

HarryMichal:
The fedora-toolbox:32 image is the first of images in the renamed
toolbox image repository[0]. With the change we can drop the
pull_image_old() function because it was kept only for the old image.

[0] https://github.com/containers/toolbox/pull/615


HarryMichal:
recheck

HarryMichal:
recheck

HarryMichal:
recheck

TristanCacqueray:
recheck

PR #786, opened 2021-06-07
pkg/podman: Add error parsing method

HarryMichal:
This new helper function parses the output of Podman in stderr into
a dedicated structure with defined methods for working with the error.

The most significant part is the method Is() that wraps around
errors.Is() and if that fails to find a match resorts to string
matching. This is needed because there are no defined errors to catch
errors coming from Podman.

The structure is only meant to be used internally.

### DIFF
--- a/src/pkg/podman/error.go
+++ b/src/pkg/podman/error.go
@@ -1,0 +1,124 @@
+/*
+ * Copyright © 2019 – 2021 Red Hat Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package podman
+
+import (
+	"bytes"
+	"strings"
+)
+
+// internalError serves for representing errors printed by Podman to stderr
+type internalError struct {
+	errors []string
+}
+
+func (e *internalError) Error() string {
+	if e.errors == nil || len(e.errors) == 0 {
+		return ""
+	}
+
+	var builder strings.Builder
+	for i, part := range e.errors {
+		if i != 0 {
+			builder.WriteString(": ")
+		}
+		builder.WriteString(part)
+	}
+	return builder.String()
+}
+
+// Is lexically compares errors
+//
+// The comparison is done for every part in the error chain not across.
+func (e *internalError) Is(target error) bool {
+	if target == nil {
+		return false
+	}
+
+	if e.errors == nil || len(e.errors) == 0 {
+		return false
+	}
+
+	for {
+		if e.errors[0] == target.Error() {
+			return true
+		}
+
+		err := e.Unwrap()
+		if err == nil {
+			return false
+		}
+		e = err.(*internalError)
+	}
+}
+
+func (e internalError) Unwrap() error {
+	if e.errors == nil || len(e.errors) <= 1 {
+		return nil
+	}
+
+	return &internalError{e.errors[1:]}
+}
+
+// parseErrorMsg serves for converting error output of Podman into an error
+// that can be further used in Go
+func parseErrorMsg(stderr *bytes.Buffer) error {
+	errMsg := stderr.String()
+	errMsg = strings.TrimSpace(errMsg)
+	if errMsg == "" {
+		return nil
+	}
+
+	// Stderr is not used only for error messages but also for e.g. loading
+	// bars. Also, several errors can be present (e.g., when 'podman pull' retries
+	// to pull an image several times after erroring). Get all of them.
+	errMsgSplit := strings.Split(errMsg, "Error: ")
+	// We're only interested in the part with the last error
+	errMsg = errMsgSplit[len(errMsgSplit)-1]
+	// Sometimes an error contains a newline (e.g., responses from Docker
+	// registry). Normalize them into further parseable error message.
+	// See an example bellow.
+	errMsg = strings.ReplaceAll(errMsg, "\n", ": ")
+	// Wrapped error messages are usually separated by a colon followed by
+	// a single space character
+	errMsgPartsRaw := strings.Split(errMsg, ": ")
+	// The parts of the err message still can have a whitespace or a colon at the
+	// beginning or at the end. Trim them.
+	var errMsgParts []string
+	for _, part := range errMsgPartsRaw {
+		part = strings.TrimSpace(part)
+		part = strings.Trim(part, ":")
+		// Podman error messages are usually prepended with the "Error:" string.
+		// Sometimes the error contains errors in a bullet list. This list is
+		// usually prepended with a message equal to "errors:".
+		//
+		// The colons at the end don't have to be checked since they've been
+		// trimmed away.
+		//
+		// Example:
+		//   Error: Error initializing source docker://foobar:latest: Error reading manifest latest in docker.io/library/foobar: errors:
+		//   denied: requested access to the resource is denied
+		//   unauthorized: authentication required
+		if part == "Error" || part == "errors" {
+			continue
+		}
+
+		errMsgParts = append(errMsgParts, part)
+	}
+
+	return &internalError{errMsgParts}
+}
--- a/src/pkg/podman/error_test.go
+++ b/src/pkg/podman/error_test.go
@@ -1,0 +1,197 @@
+/*
+ * Copyright © 2019 – 2021 Red Hat Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package podman
+
+import (
+	"bytes"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInternalError(t *testing.T) {
+	type expect struct {
+		IsNil  bool
+		Error  string
+		Search string
+		Wrap   []string
+	}
+
+	testCases := []struct {
+		name   string
+		input  string
+		expect expect
+	}{
+		{
+			name:  "Empty error message",
+			input: "",
+			expect: expect{
+				IsNil: true,
+				Error: "",
+			},
+		},
+		{
+			name:  "Text with no prolog before an error message",
+			input: "There is no prolog before the error message.",
+			expect: expect{
+				Error: "There is no prolog before the error message.",
+			},
+		},
+		{
+			name:  "Text with prolog before an error message",
+			input: "There is the prolog.Error: an error message",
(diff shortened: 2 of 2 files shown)


src/pkg/podman/podman.go
```diff
@@ -332,3 +334,67 @@ func SystemMigrate(ociRuntimeRequired string) error {
...
+}
+
+// parseErrorMsg serves for converting error output of Podman into an error
+// that can be further used in Go.
+//
+// The returned error can be examined using methods Unwrap, Is & As.
+func parseErrorMsg(stderr *bytes.Buffer) internalError {
+	var err internalError
+	var errMsg string
+	var errMsgParts []string
```
debarshiray, line 381: Why not use the more idiomatic `:=` to define the variables?

src/pkg/podman/podman.go
```diff
@@ -332,3 +334,67 @@ func SystemMigrate(ociRuntimeRequired string) error {
...
+	errMsg := fmt.Sprintf(format, a...)
+	if e.Err == nil {
+		e.Err = errors.New(errMsg)
+	} else {
+		e.Err = fmt.Errorf("%v: %w", errMsg, e.Err)
+	}
+}
+
+// Is wraps around errors.Is() and add support for string comparison of errors
+func (e internalError) Is(err error) bool {
```
debarshiray, line 359: Is this really the idiomatic way of unwrapping a chain of errors? The [convention](https://blog.golang.org/go1.13-errors) seems to be to offer an `Unwrap` method.

debarshiray: Reading the rest of the code and thinking about it more, I am wondering if there are two different approaches to implementing `internalError`.

One is to have it wrap an inner `error`, that could be another `internalError` wrapping yet another `error`, and so on. In that case `internalError` would implement an `Unwrap` method which would return the inner `error`, and `errors.Is` would either match the innermost `error` or a suffix of the chain of errors.

Second is to keep a slice of strings, each representing a layer in Podman's call chain that adds its own context to the error string that gets written to `stderr`. In that case `internalError` would implement an `Unwrap` method that would create another `internalError` by dropping the first element from the slice, and an `Is` method that can match any single member of the slice (ie., any specific layer in Podman) like `grep`.

Notice that there's a subtle difference in the behaviour of `errors.Is` between the two options. Given that parsing `stderr` of any command is a very fragile operation, I think, the second option is more robust. We can't rely on a suffix of the chain of errors remaining unchanged, and it looks like ultimately that's how the code is using `internalError`.

I think what's bothering me about the current implementation is that we are first parsing Podman's `stderr` by splitting the different layers, constructing the chain of errors and then parsing (or splitting) them again in `Is`. We should do it only once.

HarryMichal: I updated the implementation to match the second option of the two you proposed.

src/pkg/podman/podman.go
```diff
@@ -332,3 +333,83 @@ func SystemMigrate(ociRuntimeRequired string) error {
...
+			return true
+		}
+
+		if e = e.Unwrap(); e == nil {
+			return false
+		}
+	}
+}
+
+func (e internalError) Unwrap() *internalError {
```
debarshiray, line 380: Is this the [conventional](https://blog.golang.org/go1.13-errors) signature for `Unwrap`? Isn't it:
```
func (e *internalError) Unwrap() error {
    ...
}
```

Honest question. I could be wrong.

debarshiray (commented):
Thanks for updating the branch, @HarryMichal !  This looks really good.  One small question:

src/pkg/podman/podman.go
```diff
@@ -332,3 +333,83 @@ func SystemMigrate(ociRuntimeRequired string) error {
...
+	if e.errors == nil || len(e.errors) == 0 {
+		return false
+	}
+
+	for {
+		if e.errors[0] == target.Error() {
+			return true
+		}
+
+		if e = e.Unwrap(); e == nil {
```
debarshiray, line 374: Can't this be implemented as a `for ... range ...` loop? That could help with the signature of `Unwrap`.

src/pkg/podman/podman.go
```diff
@@ -332,3 +333,83 @@ func SystemMigrate(ociRuntimeRequired string) error {
 
 	return nil
 }
+
+// internalError serves for representing errors printed by Podman to stderr
+type internalError struct {
```
debarshiray, line 338: It seems like the `internalError` implementation is the kind of thing that should live in a different file. It's self-contained and somewhat separate from the rest of `podman.go`, which is mostly about wrapping `podman` invocations.

debarshiray (commented):
We should add some unit tests for the `internalError` implementation.

src/pkg/podman/podman.go
HarryMichal: It is not, the method should return `error`. Nice catch!

src/pkg/podman/podman.go
HarryMichal: I wanted to follow the original implementation of `Is()`. Using a `for ... range ...` loop would work, too.

src/pkg/podman/podman.go
HarryMichal: Sure, sounds good!

HarryMichal:
> We should add some unit tests for the `internalError` implementation.

Yes, we should. The setup will be a bit tricky since the type is not exported but Go should have a way to go around the fact.

src/pkg/podman/error_test.go
```diff
@@ -0,0 +1,144 @@
...
+			err := parseErrorMsg(bytes.NewBufferString(tc.input.Msg))
+
+			if tc.expect.IsNil {
+				assert.Nil(t, err)
+				return
+			} else {
+				assert.NotNil(t, err)
+			}
+
+			errInternal := err.(*internalError)
```
debarshiray, line 118: I understand that the cast to an `internalError` is necessary to look at the slice of error strings and such, but shouldn't the public interface (ie., `Is`, `Unwrap`, etc.) be tested without the cast?

src/pkg/podman/error.go
```diff
@@ -0,0 +1,114 @@
...
+	// The parts of the err message still can have a whitespace at the
+	// beginning ar at the end. Trim them.
+	var errMsgParts []string
+	for _, part := range errMsgPartsRaw {
+		part = strings.TrimSpace(part)
+		part = strings.Trim(part, ":")
+		// Podman error messages are usually prepended with the "Error: " string
+		// Sometimes the error contains errors in a bullet list. This list is
+		// usually prepended with a message equal to "errors:"
+		if part == "Error" || part == "errors:" {
```
debarshiray, line 106: Can the bullet list also be prepended with `Error`? Or just `errors:`?

It's good to document such odd cases with concrete examples in comments and/or test cases. The whole error parsing logic is quite tricky and sensitive, which makes it even more important.

src/pkg/podman/error.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// Stderr is not used only for error messages but also for e.g. loading
+	// bars
+	errMsgSplit := strings.SplitN(errMsg, "Error: ", 2)
+	// We're only interested in the part with an error
+	errMsg = errMsgSplit[len(errMsgSplit)-1]
+	// Sometimes an error contains a newline (e.g., responses from Docker
+	// registry). Normalize them into further parseable error message.
+	errMsg = strings.ReplaceAll(errMsg, "\n", ": ")
```
debarshiray, line 93: Is this related to the bullet lists?

HarryMichal: Yes, that's right.

src/pkg/podman/error_test.go
HarryMichal: The `internalError` type is not public and thus will never be used outside of the `podman` package. Casting to `error` to `*internalError` will be a common practice in the package.

HarryMichal: Also, `Is` and `Unwrap` are not a public interface in the classic `error` type, thus the cast is needed to invoke them.

src/pkg/podman/error.go
HarryMichal: Elaborated and added an example.

src/pkg/podman/error.go
HarryMichal: I added a hint to look a few lines down for a concrete example.

src/pkg/podman/podman.go
debarshiray: One of the main advantages of implementing `internalError` as a slice of error strings is that the `Is` implementation, where we are matching against each element in the slice, becomes very straightforward. So it's good to take advantage of that.

src/pkg/podman/error_test.go
debarshiray: `Is` and `Unwrap` are not part of the `error` interface, but they are very much [public](https://blog.golang.org/go1.13-errors). That's how `errors.Unwrap` and `errors.Is` work. Errors created using the `%w` verb in `fmt.Errorf` have an `Unwrap` method. Etc..

The public interface is how almost every (all?) user of `internalError` will use it. So, we should use those in our test suite as much as possible. A cast to `internalError` should only be used if we are specifically looking inside the guts of the `internalError` implementation.

src/pkg/podman/error_test.go
```diff
@@ -0,0 +1,197 @@
...
+time="2021-07-04T01:08:25+02:00" level=warning msg="failed, retrying in 1s ... (2/3). Error: Error initializing source docker://registry.redhat.io/foobar:latest: Error reading manifest latest in registry.redhat.io/foobar: unknown: Not Found"
+time="2021-07-04T01:08:27+02:00" level=warning msg="failed, retrying in 1s ... (3/3). Error: Error initializing source docker://registry.redhat.io/foobar:latest: Error reading manifest latest in registry.redhat.io/foobar: unknown: Not Found"
+Error: Error initializing source docker://registry.redhat.io/foobar:latest: Error reading manifest latest in registry.redhat.io/foobar: unknown: Not Found
+`,
+			expect: expect{
+				Error: "Error initializing source docker://registry.redhat.io/foobar:latest: Error reading manifest latest in registry.redhat.io/foobar: unknown: Not Found",
+			},
+		},
+		{
+			name:  "Error message from 'podman login' - no such host",
```
debarshiray, line 145: Let's restrict ourselves to parsing only `podman pull` errors.

src/pkg/podman/error_test.go
```diff
@@ -0,0 +1,197 @@
...
+			name:  "Error message with newline and with errors in \"bullet\" list",
+			input: "Error: foobar:\n  err1\n  err2\n  err3",
+			expect: expect{
+				Error:  "foobar: err1: err2: err3",
+				Search: "err2",
+				Wrap:   []string{"foobar", "err1", "err2", "err3"},
+			},
+		},
+		{
+			name: "Error message from 'podman pull' - unknown registry",
```
debarshiray, line 90: Let's restrict ourselves to only parsing `podman pull` errors related to authentication. It's going to be a struggle to keep that working over time, so let's not make it worse for no real benefit.

src/pkg/podman/error_test.go
```diff
@@ -0,0 +1,197 @@
...
+			name: "Error message from 'podman pull' - unauthorized (Red Hat Registry)",
+			input: `Trying to pull registry.redhat.io/foobar:latest...
+Error: Error initializing source docker://registry.redhat.io/foobar:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication
+`,
+			expect: expect{
+				Error: "Error initializing source docker://registry.redhat.io/foobar:latest: unable to retrieve auth token: invalid username/password: unauthorized: Please login to the Red Hat Registry using your Customer Portal credentials. Further instructions can be found here: https://access.redhat.com/RegistryAuthentication",
+			},
+		},
+		{
+			name: "Error message from 'podman pull' - unknwon image (Fedora Registry)",
```
debarshiray, line 117: Ditto. Only authentication errors.

debarshiray:
Let's fold this pull request into https://github.com/containers/toolbox/pull/787

Otherwise it's becoming impossible to manage all the branches that include each other and are evolving in different ways.

PR #787, opened 2021-06-07
cmd/create, pkg/podman: Add capabilities for logging to registries

### LINKED ISSUE #754: Registry authentication is not supported
**Describe the bug**

Some registries (e.g., registry.redhat.io) require the user to be authenticated before they can pull an image from it. Toolbox should notice when a registry requires authentication and ask the user for their credentials instead of failing.

1. `toolbox create -i registry.redhat.io/ubi8/ubi`

**Expected behaviour**
Toolbox allows the user to authenticate.

**Actual behaviour**
Toolbox fails to pull an image and exits

`toolbox version 0.0.99.1`

`toolbox-0.0.99.1-1.fc35.x86_64`

```
Version:      3.1.0-dev
API Version:  3.0.0
Go Version:   go1.16
Built:        Mon Feb 22 23:40:29 2021
OS/Arch:      linux/amd64
```

`podman-3.1.0-0.42.dev.gitd999328.fc35.x86_64`

**Info about your OS**
Fedora Silverblue Rawhide

**Additional context**
This is required for getting Toolbox into RHEL.

HarryMichal:
Closes #754


HarryMichal:
recheck

HarryMichal:
I may have gotten an idea on how to test this reliably. Until I have it figured out or I give up, let's postpone merging this.

src/pkg/podman/podman.go
```diff
@@ -226,13 +234,75 @@ func IsToolboxImage(image string) (bool, error) {
...
+		}
+
+		err := parseErrorMsg(&stderr)
+		return fmt.Errorf("failed to login to registry %s: %w", registry, err.Err)
+	}
+
+	return nil
+}
+
+// Pull pulls an image. Wraps around command 'podman pull'.
```
tpopela, line 256: Extra pulls

HarryMichal: I think that is intentional since Go tooling requires function comments to start with the name of the function. But I'd reword the function anyway :).

HarryMichal:
recheck

HarryMichal:
recheck

HarryMichal:
This should finally be good to go.

doc/toolbox-create.1.md
```diff
@@ -63,6 +67,25 @@ containers. Read more about the entry-point in `toolbox-init-container(1)`.
 
 ## OPTIONS ##
 
+**--authfile** AUTHFILE
```
debarshiray, line 70: This option is inherently insecure because it expects people to put their very possibly *secret* credentials in an unencrypted file. Dangerous. :)

Yes, `podman create ...` offers this option, but it doesn't necessarily mean we should offer it. Unless someone shows up with a very valid use-case that can't be achieved in any other way, we shouldn't be offering it.

doc/toolbox-create.1.md
```diff
@@ -63,6 +67,25 @@ containers. Read more about the entry-point in `toolbox-init-container(1)`.
...
+
+Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json
+
+Note: You can also override the default path of the authentication file by
+setting the REGISTRY\_AUTH\_FILE environment variable. `export REGISTRY_AUTH_FILE=path`
+
+Note: This option is equal to one in `podman-create(1)`.
+
+**--creds** USERNAME:PASSWORD
```
debarshiray, line 79: Ditto, as `--authfile`.

Generally, it's considered insecure to pass *secret* credentials through the command line because any other user on the system can look at the command lines of all processes on the system, including the credentials.

The better alternative is to feed such credentials through the standard input of the target program by somehow indirecting it. In the case of `podman pull`, I am not sure what would be the best approach. Worst case, we will have to use `podman pull --creds ...`, until we find a better way.

Regardless, let's not add this flag to the public interface of Toolbox unless someone comes up with a really good use-case for it.

src/pkg/podman/podman.go
```diff
@@ -332,3 +401,67 @@ func SystemMigrate(ociRuntimeRequired string) error {
 
 	return nil
 }
+
+type internalError struct {
```
debarshiray, line 405: Is this is the same as https://github.com/containers/toolbox/pull/786 ?

src/cmd/create.go
```diff
@@ -718,20 +743,89 @@ func pullImage(image, release string) (bool, error) {
 		return false, nil
 	}
 
+pull_image:
```
debarshiray, line 746: We should handle the authentication errors in the same way we handle fallbacks in `run`. ie., refactor the guts of `pullImage` into smaller functions and then run a loop.

Using `goto` to loop across such a big chunk of code with some `return`s in between is soon going to turn into a nightmare.

src/cmd/create.go
```diff
@@ -718,20 +743,89 @@ func pullImage(image, release string) (bool, error) {
...
+					args = append(args, "--authfile", createFlags.authfile)
+				}
+
+				if createFlags.creds != "" {
+					creds := strings.Split(createFlags.creds, ":")
+					args = append(args, "--username", creds[0], "--password", creds[1])
+				}
+
+				if err = podman.Login(domain, args...); err != nil {
+					return false, err
```
debarshiray, line 819: It would be nice to sanitize this error. Currently it contains duplicated text:
```
Error: failed to login to registry.redhat.io: error logging into "registry.redhat.io": invalid username/password
```

src/cmd/create.go
```diff
@@ -718,20 +743,89 @@ func pullImage(image, release string) (bool, error) {
...
+				}
+			}
+
+			if !rootFlags.assumeYes {
+				s.Stop()
+				fmt.Fprintf(os.Stderr, builder.String())
+
+				if !utils.AskForConfirmation("Do you want to log into the registry and try to pull the image again? [y/N]") {
```
debarshiray, line 803: This seems like a needless step. Since we know it requires authentication, we should just ask for it, instead of asking the user if we should ask for it. :)

`coreos/toolbox` has this prompt, but their error handling is also rudimentary at best, if not non-existent, but we can do better.

doc/toolbox-create.1.md
HarryMichal: I added the option to match [coreos/toolbox](https://github.com/coreos/toolbox). I also did not realise this particular option can be changed by setting the `REGISTRY_AUTH_FILE` environmental variable. I'll drop the option.

doc/toolbox-create.1.md
HarryMichal: > Worst case, we will have to use podman pull --creds ..., until we find a better way.

I do not understand what you mean by this.

src/pkg/podman/podman.go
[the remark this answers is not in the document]
HarryMichal: Yes. This PR depends on #786.

.zuul.yaml
```diff
@@ -12,7 +12,7 @@
 - job:
     name: system-test-fedora-33
     description: Run Toolbox's system tests in Fedora 33
-    timeout: 1200
+    timeout: 1500
```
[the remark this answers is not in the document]
HarryMichal: Yes, the tests with the new test cases and enhanced setup started to timeout. Mainly on Rawhide where the performance of Podman/Skopeo is somewhat crippled for some reason. But the timeout already occurred even on stable releases of Fedora.

src/cmd/create.go
HarryMichal: You are absolutely right. I was not sure how to properly separate this, so I went with the `goto` option. I'll try to think of something.

src/cmd/create.go
HarryMichal: > This seems like a needless step.

True. Good point.

> coreos/toolbox has this prompt...

That's why I added it :)

doc/toolbox-create.1.md
debarshiray: Oh, my apologies about the rambling comment.

I wrote it before fully reading the entire pull request. I thought we are using `podman pull --creds username:password ...`, not `podman login`. So, I meant that if we really can't have a nice way to feed the credentials to Podman, then we can use `podman pull --creds ...` in whichever way that works until a better solution emerges.

In other words, I don't want Toolbox to encourage unsafe user behaviour (unless there's a compelling use-case, of course), but we have to work with the current realities of Podman. At least, folks can point the finger at us. :)

src/cmd/create.go
```diff
@@ -502,6 +553,23 @@ func createHelp(cmd *cobra.Command, args []string) {
...
+func ensureImageName(image, release string) string {
+	hasDomain := utils.ImageReferenceHasDomain(image)
+
+	if hasDomain {
+		return image
+	}
+
+	image, err := utils.GetFullyQualifiedImageFromDistros(image, release)
+	if err != nil {
+		return ""
```
debarshiray, line 568: We shouldn't cut out the `error` from inner utility functions. Generally speaking, it's a good convention to always offer an `error` from functions that can possibly fail. It provides an explanation for the failure, which can be really useful when debugging.

Sometimes it can happen that an `error` thrown by one of the lower layers doesn't make sense in the context of the upper layers. In such cases, the `error` can either be wrapped or re-written, but an `error` should still be there.

src/cmd/create.go
```diff
@@ -502,6 +553,23 @@ func createHelp(cmd *cobra.Command, args []string) {
 	}
 }
 
+// ensureImageName translates an image URI into a "pullable" reference
+//
+// On error is returned an empty string
+func ensureImageName(image, release string) string {
```
debarshiray, line 559: Umm... why not call it `ensureFullyQualifiedImage` or something like that? Otherwise, it's a bit confusing as to why we would ensure an image name when we already have `image`.

debarshiray (commented):
Thanks for the updates, @HarryMichal

doc/toolbox-create.1.md
debarshiray: I see that we are using `--creds` for testing. I wonder if we can solve that some other way.

src/cmd/create.go
```diff
@@ -190,12 +207,46 @@ func createContainer(container, image, release string, showCommandToEnter bool)
...
+		if image == "" {
+			var builder strings.Builder
+			fmt.Fprintf(&builder, "image %s not found in local storage and known registries\n", image)
+			fmt.Fprintf(&builder, "Make sure the image URI is correct")
+
+			errMsg := builder.String()
+			return errors.New(errMsg)
+		}
+
+		domain := utils.ImageReferenceGetDomain(image)
```
debarshiray, line 221: I was hoping that we wouldn't be significantly increasing the length of `createContainer`, and keep the details of authentication encapsulated within `pullImage`. :)

In terms of creating a container, whether we need to authenticate or not to pull an image, is an implementation detail.

src/pkg/podman/podman.go
```diff
@@ -226,18 +237,80 @@ func IsToolboxImage(image string) (bool, error) {
...
+// for their handling. The provided errors can not be relied on completely as
+// they are created based on error message parsing that differs based on used
+// registry.
+func Pull(imageName string) error {
+	var stderr bytes.Buffer
+
+	logrus.Debugf("Pulling image %s", imageName)
+
+	logLevelString := LogLevel.String()
+	args := []string{"--log-level", logLevelString, "pull", imageName, "--quiet"}
```
debarshiray, line 278: Umm... why `--quiet`? We do expose the entire spew from `podman pull ...` when `--verbose` is enabled.

src/pkg/podman/podman.go
```diff
@@ -226,18 +237,80 @@ func IsToolboxImage(image string) (bool, error) {
...
 }
 
+// Pull pulls an image. Wraps around command 'podman pull'.
+//
+// The image pull can fail for many reasons. The recognized reasons are:
+//   - manifest unknown (ErrManifestUnknown)
+//   - name unknown (ErrNameUnknown)
+//   - unauthorized (ErrUnauthorized)
+//
+// To learn more about OCI API error codes, see: https://github.com/opencontainers/distribution-spec/blob/main/spec.md
```
debarshiray, line 266: I am a bit confused about what I can learn from this URL. :)

Is it about the section titled *Error Codes*? It talks about the `code` field which must contain only uppercase characters and underscores, but that's not what we are looking for in the code below. The error codes might be defined by OCI, but that doesn't help us much since we are just scraping `stderr` for known sub-strings.

src/pkg/podman/podman.go
```diff
@@ -226,18 +237,80 @@ func IsToolboxImage(image string) (bool, error) {
...
+		// This is returned if the name used during an operation is unknown to the
+		// registry.
+		// The "unknown" error message is an exception applying to registry.redhat.io
+		if err.Is(errors.New("name unknown")) || err.Is(errors.New("unknown")) {
+			return ErrNameUnknown
+		}
+
+		// The access controller was unable to authenticate the client. Often this
+		// will be accompanied by a Www-Authenticate HTTP response header
+		// indicating how to authenticate.
```
debarshiray, line 302: This comment seems redundant, unless we can somehow leverage the HTTP response header to our benefit. Otherwise, the error message makes it clear that this is about authentication.

src/pkg/podman/podman.go
debarshiray: Oh, because the progress information from `podman pull` interferes with the screen scraping. :(

This is one reason why screen scraping is horrible. We need to find a way to do this without using `--quiet` because having the actual progress bar show up when using `--verbose` is quite useful for debugging. eg., you get to see if things are just slow or stuck.

src/cmd/create.go
HarryMichal: Yes, sure. `ensureFullyQualifiedImage` sounds better.

src/cmd/create.go
HarryMichal: Would a debug print be enough? Or you'd rather have the function to return the error?

src/cmd/create.go
HarryMichal: The context also did not fit into `pullImage` :). I'll shuffle the code around a bit.

src/pkg/podman/podman.go
HarryMichal: The progress bar won't show up regardless of the use of `--quiet` since `stderr` is redirected into a bytes buffer.

Is there anything else we can do instead of screen scraping? We could either print the content of the bytes buffer once the execution of Pull is done or somehow pass it through to Toolbox's `stderr` when debug logging is enabled. This will still interfere with the parsing though.

One way around could be to update the error parsing method to start parsing upon reading `Error:` in the output. That usually signals the presence of an error.

HarryMichal: I have a potential solution in #823.

HarryMichal: I tried to address this issue. Try it.

src/pkg/podman/podman.go
HarryMichal: Removed the part mentioning the response header.

src/pkg/podman/podman.go
HarryMichal: You're right. I removed the link.

src/pkg/podman/podman.go
```diff
@@ -33,6 +35,13 @@ var (
 
 var (
 	LogLevel = logrus.ErrorLevel
+	ShowLogs = false
+)
+
+var (
+	ErrUnauthorized    = errors.New("not authorized to access resources in registry")
+	ErrManifestUnknown = errors.New("registry does not know the requested manifest")
+	ErrNameUnknown     = errors.New("registry does not know the requested repository name")
```
debarshiray, line 44: Umm... no. :)

Scraping the `stderr` of `podman` invocations to reconstruct the errors is a stop-gap measure for implementing authentication, and that's a very nice way to put it. Currently, I am not sure if even that much is acceptable as an upstream change.

Redirecting the `stderr` to a non-terminal file descriptor disables Podman's progress bars when using `--verbose`. The `stderr` parsing logic in `parseErrorMsg` gets more and more fragile as we try to cover more and more scenarios with it, and it will all collapse as soon as something gets tweaked in the many layers of code lying underneath `podman`.

So, no, just no. :) Let's not go overboard and try to do more of this, for some cosmetic niceties. We really should be pushing for proper exit codes, instead of going down this route.

src/cmd/create.go
debarshiray: In this case, I'd return the error, because it's directly tied to the success or failure of this function.

Sometimes, the success or failure of inner functions don't reflect directly on the function itself. In such cases, one could dump the error into the debug logs, if the error is being ignored or worked around.

doc/toolbox-create.1.md
debarshiray: For the test suite, we should aim to make this work:
```
$ echo -e "username\npassword" | toolbox -y create --image registry.redhat.io/ubi8/ubi:8.4
```

That way we can still feed the toy credentials to the test suite without adding a `--creds` flag to our public user interface.

The other thing is to avoid passing the credentials, particularly the password, through the command line of the `podman login` invocation or any other spawned binary. So that a malicious user on the system can't sit there staring at the process list and steal the credentials.

In order to satisfy both constraints, we need to somehow get `podman login` to transparently pick up the credentials even when the standard input has been redirected to a non-terminal device, as will be the case for the test suite. Sadly, `podman login` insists on using the `--password-stdin` and `--username` flags when the input has been redirected, which makes it a bit annoying.

Therefore, I think `toolbox` should take care of reading the username and password itself in a way that works transparently with both terminal devices and otherwise. Then, we should use `podman login --password-stdin` to feed the password by putting it in a `bytes.Buffer` acting as the standard input.

test/system/000-setup.bats
```diff
@@ -1,11 +1,91 @@
...
+load 'libs/bats-support/load'
+load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 @test "test suite: Setup" {
     # Cache the default image for the system
     _pull_and_cache_distro_image $(get_system_id) $(get_system_version) || die
     # Cache all images that will be needed during the tests
     _pull_and_cache_distro_image fedora 32 || die
+    _pull_and_cache_distro_image rhel 8.4 || die
```
debarshiray, line 12: Can't we use a Fedora (or anything that's not UBI) image for testing?

test/system/000-setup.bats
```diff
@@ -1,11 +1,91 @@
...
+        -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
+        -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
+        -p 50001:443 \
+        docker.io/library/registry:2
+    assert_success
+    echo "# Podman logs - docker-registry-auth"
+    $PODMAN --root "$PODMAN_REG_ROOT" logs docker-registry-auth
+
+    # Add UBI8 to created registries
+    run $SKOPEO copy "dir:${IMAGE_CACHE_DIR}/fedora-toolbox-32" "docker://localhost:50000/fedora-toolbox:32"
```
debarshiray, line 83: Typo in the comment? Is the comment really necessary? We didn't use one for the similar statement below. :)

debarshiray (changes requested):
A couple of things about the test suite:

(You did a really good job setting up the registries as part of the test suite! Well done.)

debarshiray:
Unless `podman pull` grows a dedicated exit code to denote authorization errors, we can't merge this upstream. However, downstream distributors who want this feature can use this patch-set at their own risk.

In the absence of a dedicated exit code from `podman pull`, this patch-set tries to parse the standard error stream of `podman pull` to guess whether the image requires logging into the registry or not. This sort of screen scraping is very fragile and is a bad idea.

First, the messages that get dumped in the error stream don't have any fixed structure or format, and can change from one Podman version to another. Moreover, the error stream is also used to show things like progress bars, so those need to be skipped when parsing it. This makes it very fragile to test whether the parsing works as expected across several Podman versions in several operating systems.

Second, since the error stream is also used to show progress bars for `podman pull`, redirecting it away from a terminal device (to intercept its contents) prevents the progress bar from being shown. We show the progress bar when `toolbox --verbose ...` is used because they are useful for checking the status of the pulls. Ultimately, we are downloading hundreds of megabytes as part of a critical operation, so having access to a working progress bar is important.

So, long story short, we either need `podman pull` to have a [dedicated exit code](https://github.com/containers/podman/issues/10858) for authorization failures, or we need to re-implement `podman pull` directly inside Toolbox so that we have better access to the errors.

HarryMichal:
I'll close this since this approach is not acceptable long-term-wise. If the work is attempted again, it can be done in a separate PR.

PR #792, opened 2021-06-18
cmd/list, test/system: Ignore the problem of UBI not being listed

debarshiray:
Not having the corresponding image for UBI toolbox containers show up
in 'toolbox list' is a rough edge. However, the whole UBI feature is
a bit experimental. It's about a gratis RHEL environment getting
created in a jiffy on any host, which is something that hasn't been
done before, and those containers also suffer from various shortcomings
because of the limited package set of UBI.

So it's not that big of a problem if it takes a release or two to
hammer out the details. Especially since it's likely that there will
be a special Toolbox-specific image that's created out of the UBI RPM
repositories, which will likely have the com.github.containers.toolbox
label.

There's also the issue that 0.1.0 needs to be finished, and for that
the the churn needs to be kept down. Changing the labels can very
likely lead to compatibility issues in the future, because of which it
either can't be removed for a while or the wrong images start to get
listed. Some of the older labels have finally been removed, so it's
better not to add more to the list.

In short, this problem will likely fix itself in the coming months, so
it's wise not to create complications trying to rush through a fix.

This reverts commits 1df36591d02d4fe32c2579c8b4d6c96fca9a406f and
e09de9f3e53dee80eec33619b8773a85a1fb863d.

https://github.com/containers/toolbox/issues/753

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -52,18 +52,11 @@ var (
 		onlyImages     bool
 	}
 
-	// toolboxImageLabels holds labels used by images that mark them as compatible with Toolbox
-	toolboxImageLabels = map[string]string{
+	// toolboxLabels holds labels used by containers/images that mark them as compatible with Toolbox
+	toolboxLabels = map[string]string{
 		"com.github.debarshiray.toolbox": "true",
 		"com.github.containers.toolbox":  "true",
-        "com.redhat.component": "ubi8-container",
 	}
-
-    // toolboxContainerLabels holds labels used by container that mark them as compatible with Toolbox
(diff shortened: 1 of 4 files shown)


HarryMichal:
Thank you for working on this!

> However, the whole UBI feature is a bit experimental.

While I agree with you on this being a bit experimental, the feature has been [advertised](https://debarshiray.wordpress.com/2021/01/14/toolbox-after-a-gap-of-15-months/) to some extent already and I guess some users already use these images. It might be worth using the approach I mentioned offline for listing the images if there is an UBI-based toolbox present.

PR #794, opened 2021-06-21
README.md, cmd/initContainer: Don't require /etc/machine-id in images

debarshiray:
Since `/etc/machine-id` is bind mounted into the toolbox container from
the host operating system, it doesn't make sense to make it mandatory
for images to have that file. Apparently, (some?) Arch Linux images
don't have `/etc/machine-id`.

Since a missing `containerPath` for a directory is handled the same way,
there's no reason not to do the same for regular files. It will make
life a bit easier for those creating toolbox images for different
distributions.

https://github.com/containers/toolbox/pull/710


HarryMichal (approved):
Looks good to me.

debarshiray:
> Looks good to me.

Thanks for the review!

PR #796, opened 2021-06-21
pkg/podman: Replace go-version with go-semver

HarryMichal:
In the past I decided to use HarryMichal/go-version[0] (which is a fork
with a small patch of mcuadros/go-version[1]) because I did not find
a different solution. I noticed that Viper pulls in coreos/go-semver[2]
that more-or-less does the same job. Let's use it to reduce the
footprint of the tool.

[0] https://github.com/HarryMichal/go-version
[1] https://github.com/mcuadros/go-version
[2] https://github.com/coreos/go-semver


debarshiray (changes requested):
Thanks, this seems worth doing!

However, I would be a lot less stressed if we had unit tests for our `CheckVersion` function to demonstrate that this doesn't introduce any regressions.

Could you please add some unit tests before the commit that makes the switch?

PR #800, opened 2021-06-22
cmd/list: Support images without names

HarryMichal:
Some people create images manually. If such created images are recognize
as toolbox images (they have the proper labels) but do not have
a name/tag then 'toolbox list' will panic due to index being out of
range.


olivergs:
Works for me. Tests running Ok.

Thanks!

debarshiray:
Thanks! Extra brownie points for adding a test. :)

PR #801, opened 2021-06-22
Flatten source file structure

martymichal:
With the Go rewrite of Toolbox, I introduced the new "src" directory. At
the time it seemed like the right choice but overtime I came to realize
that:

  - CI tools don't work well with source code in subdirectories
  - pkg.go.dev wants go.mod in the root directory

This moves all the content of "src" to the project root, updates
meson.build in the project root with the content of meson.build from
"src" and renames the Shell implementation of Toolbox from toolbox to
toolbox-old because it conflicted with the built Go binary.

I realize this affects the git history. This issue can be overcome by
using the --follow option in git log.


debarshiray:
It might be better to do this post 0.1.0.

PR #802, opened 2021-06-22
cmd/root: Unbreak handling of unknown flags

debarshiray:
This reverts commit 6bbbedf675a2fcfc03e720a252d501e1507561a4.

### DIFF
--- a/test/system/002-help.bats
+++ b/test/system/002-help.bats
@@ -33,3 +33,11 @@ load 'libs/helpers.bash'
   assert_line --index 0 "Error: unknown command \"foo\" for \"toolbox\""
   assert_line --index 1 "Run 'toolbox --help' for usage."
 }
+
+@test "help: Try to run toolbox with non-existent flag (shows usage screen)" {
+  run $TOOLBOX --foo
+
+  assert_failure
+  assert_line --index 0 "Error: unknown flag: --foo"
+  assert_line --index 1 "Run 'toolbox --help' for usage."
+}
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -62,6 +62,11 @@ var (
 )
 
 func Execute() {
+	if err := setUpGlobals(); err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s\n", err)
+		os.Exit(1)
+	}
+
 	if err := rootCmd.Execute(); err != nil {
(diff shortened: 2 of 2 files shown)


debarshiray:
The commit message needs to be elaborated. I submitted the PR to see if the new test case is working as expected.

HarryMichal:
The issue is not necessarily about handling unknown flags. It's about the fact that we rely on a certain variable being set and the said variable is used in usage help text.

I suggest applying #775 instead.

debarshiray:
> The issue is not necessarily about handling unknown flags.
> It's about the fact that a certain variable is not set before parsing
> flags. Do we necessarily need the variable being set?

The `executableBase` variable needs to be set because it's needed by the code that handles unknown flags. Otherwise, the user-facing problem is that the code for handling unknown flags shows the wrong output.

> I suggest applying #775 instead.

I really want to roll a new release first. #775 looks like a more intrusive change. I just want to quickly address the regression for the new release.

HarryMichal:
> I really want to roll a new release first. #775 looks like a more intrusive change. I just want to quickly address the regression for the new release.

I'm not entirely sure about #775 being too intrusive but there's no point in arguing :). Let's postpone it for the next cycle.

debarshiray:
I was referring to this comment from you:
> Before this can be merged some test cases have to be written
> to make sure the behaviour is correct.

test/system/002-help.bats
```diff
@@ -33,3 +33,11 @@ load 'libs/helpers.bash'
...
   assert_line --index 1 "Run 'toolbox --help' for usage."
 }
+
+@test "help: Try to run toolbox with non-existent flag (shows usage screen)" {
+  run $TOOLBOX --foo
+
+  assert_failure
+  assert_line --index 0 "Error: unknown flag: --foo"
+  assert_line --index 1 "Run 'toolbox --help' for usage."
+}
```
HarryMichal, line 43: Having tests for unknown flags in all other commands could also be beneficial.

HarryMichal (commented):
The change is ok. The comment is related to testing. Feel free to merge without addressing it.

test/system/002-help.bats
debarshiray: That's a good idea. I added some tests in https://github.com/containers/toolbox/pull/812

PR #804, opened 2021-06-23
cmd/create: Style fixes

debarshiray:
It's good to avoid subtle variations in the logic for /boot and /usr,
unless actually necessary, because it makes the code easier to read.

Fallout from 7ec26a27df7199777b94a17b2935e7ff35757937


HarryMichal (approved):
Looks good.

PR #806, opened 2021-06-23
completion/bash: Completely drop flag --very-verbose

HarryMichal:
Follow up to 7fafcd271e519df4e8e2f073bfe84921dcb685d7


debarshiray:
Good catch!

PR #807, opened 2021-06-23
cmd/initContainer: Don't ignore unknown flags, yet

debarshiray:
The reason for setting FParseErrWhitelist.UnknownFlags to 'true' was to
prepare for a future when the 'init-container' command would have fewer
options than it does now.

However, there's no need to prepare for it, because the version of
toolbox(1) that's bind mounted into the container is the same as the
one on the host. So, FParseErrWhitelist.UnknownFlags can be set in
future if, or when, the number of flags do get reduced.

This reverts commit 5c2086e9ea045f73bdc574734d659807a2d68d1e.

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -122,7 +122,6 @@ func init() {
 		"Create a user inside the toolbox container whose login name is USER")
 	initContainerCmd.MarkFlagRequired("user")
 
-	initContainerCmd.FParseErrWhitelist.UnknownFlags = true
 	initContainerCmd.SetHelpFunc(initContainerHelp)
 	rootCmd.AddCommand(initContainerCmd)
 }


HarryMichal:
Is this revert really needed? Yes, the motivation behind the original change was flawed but I don't see a problem in keeping it.

debarshiray:
> Is this revert really needed? Yes, the motivation behind the original
> change was flawed but I don't see a problem in keeping it.

It's not really necessary in the sense that it doesn't cause any user-visible breakage.

However, leaving it in will be confusing to anybody reading the code because they will wonder why `FParseErrWhitelist.UnknownFlags` is set and that too only for the `init-container` command. `git blame` will lead them to commit 5c2086e9ea045f73bdc574734d659807a2d68d1e which will only add to the confusion.

Also, I didn't understand this sentence in the commit message:
> In the future we'd like to minimize their number.

Are the current command line flags for `init-container` creating any problems? Do we already have a plan as to what we want to do with them? Wouldn't we need to keep accepting the current flags for a while to not break existing containers, even after we have introduced a new set of flags?

Finally, *not* ignoring unknown flags in `init-container` does have the benefit of catching coding errors in `create`. If by mistake someone mistypes the invocation of `init-container` in the code, containers will fail to start and will draw attention to the problem. If unknown flags were being silently ignored, then things might appear to continue to work, but be broken in subtle ways which might be harder to detect.

PR #811, opened 2021-06-25
cmd/root: Wire root command to 'enter' command

HarryMichal:
One of requirements to get Toolbox to RHEL is to make the root command
('toolbox') have similar functionality as 'toolbox enter'.

Making 'toolbox' behave exactly the same as 'toolbox enter', making it
effectively an alias is not exactly desirable as the root command has
its own behaviour too. Instead, `toolbox` only translates to `toolbox
enter` when no arguments or flags (excluding the global ones) are
provided.

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -177,17 +177,62 @@ func rootHelp(cmd *cobra.Command, args []string) {
 }
 
 func rootRun(cmd *cobra.Command, args []string) error {
-	var builder strings.Builder
-	fmt.Fprintf(&builder, "missing command\n")
-	fmt.Fprintf(&builder, "\n")
-	fmt.Fprintf(&builder, "create    Create a new toolbox container\n")
-	fmt.Fprintf(&builder, "enter     Enter an existing toolbox container\n")
-	fmt.Fprintf(&builder, "list      List all existing toolbox containers and images\n")
-	fmt.Fprintf(&builder, "\n")
-	fmt.Fprintf(&builder, "Run '%s --help' for usage.", executableBase)
+	if len(args) != 0 {
+		panic("unexpected argument: commands known or unknown shouldn't reach here")
+	}
 
-	errMsg := builder.String()
-	return errors.New(errMsg)
+	if utils.IsInsideContainer() {
+		if !utils.IsInsideToolboxContainer() {
+			return errors.New("this is not a toolbox container")
+		}
+
+		if _, err := utils.ForwardToHost(); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	container, image, release, err := utils.ResolveContainerAndImageNames("", "", "", "")
+	if err != nil {
+		return err
+	}
+
+	userShell := os.Getenv("SHELL")
+	if userShell == "" {
+		return errors.New("failed to get the current user's default shell")
+	}
+
+	command := []string{userShell, "-l"}
+
+	hostID, err := utils.GetHostID()
+	if err != nil {
+		return fmt.Errorf("failed to get the host ID: %w", err)
+	}
+
+	hostVariantID, err := utils.GetHostVariantID()
+	if err != nil {
+		return errors.New("failed to get the host VARIANT_ID")
+	}
+
+	var emitEscapeSequence bool
+
+	if hostID == "fedora" && (hostVariantID == "silverblue" || hostVariantID == "workstation") {
+		emitEscapeSequence = true
+	}
+
+	if err := runCommand(container,
+		true,
+		image,
+		release,
+		command,
+		emitEscapeSequence,
+		true,
+		false); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func rootUsage(cmd *cobra.Command) error {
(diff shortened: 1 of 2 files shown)


src/cmd/root.go
```diff
@@ -177,6 +178,13 @@ func rootHelp(cmd *cobra.Command, args []string) {
 }
 
 func rootRun(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		rootFlags.assumeYes = true
```
debarshiray, line 182: Umm... we shouldn't set `assumeYes` to `true` in this code path, because it can be surprising. eg., someone who has just run into Toolbox and casually types in `toolbox` will suddenly be downloading 100s of MBs without having a way to bow out of it.

I don't think this is necessary for compatibility with `coreos/toolbox`.

src/cmd/run.go
```diff
@@ -205,6 +205,21 @@ func runCommand(container string,
 			fmt.Fprintf(os.Stderr, "Entering container %s instead.\n", container)
 			fmt.Fprintf(os.Stderr, "Use the 'create' command to create a different toolbox.\n")
 			fmt.Fprintf(os.Stderr, "Run '%s --help' for usage.\n", executableBase)
+		} else if defaultContainer && rootFlags.enterShortcut {
+			fmt.Fprintf(os.Stderr, "Default container does not exist. Creating it.\n")
```
debarshiray, line 209: Umm... we shouldn't add a separate criterion for the `toolbox` with no commands and flags case. We should stick to the existing set of criteria that's used to make `toolbox enter` a bit more flexible.

Our objective isn't to have exact message by message compatibility with `coreos/toolbox`. That's fundamentally impossible because Toolbox has a lot many more features and users using it across very different environments, whereas `coreos/toolbox` is a relatively small shell script that's used in a very specific way. eg., we specifically don't want entering just `toolbox` to potentially start downloading 100s of MBs, because that will be very surprising to those who have just stumbled across `toolbox(1)`.

I think the idea is to have some rough compatibility, where the two things aren't completely different.

In this case, I'd expect `toolbox` to behave like `toolbox enter`. For absolute beginners, who might have read about `coreos/toolbox`, Toolbox would still behave in somewhat the same way, even though it might prompt them for various things. As long as the prompts are self-explaining, that's fine. One can also imagine `coreos/toolbox` growing such prompts in the future.

Compatibility or not, ultimately, we need to ensure that the tool continues to behave in a coherent, understandable and easy to document way. The more special cases we add, the less coherent and understandable it becomes, and the harder it gets to document it.

src/cmd/root.go
```diff
@@ -177,6 +178,13 @@ func rootHelp(cmd *cobra.Command, args []string) {
...
+		rootFlags.enterShortcut = true
+
+		return enter(enterCmd, args)
+	}
+
 	var builder strings.Builder
 	fmt.Fprintf(&builder, "missing command\n")
```
debarshiray, line 189: Is there a reason to keep the `missing command` message anymore?

src/cmd/root.go
```diff
@@ -177,6 +178,13 @@ func rootHelp(cmd *cobra.Command, args []string) {
 }
 
 func rootRun(cmd *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		rootFlags.assumeYes = true
+		rootFlags.enterShortcut = true
+
+		return enter(enterCmd, args)
```
debarshiray, line 185: It would be better not to re-use the top-level entry points for a command, even when functionally one command includes other. eg., `enter()` doesn't actually call `run()` or `create(), even though it does include both those two commands.

Each of the top-level entry points, even if largely the same, are still subtly different (eg., when a container is created as part of `enter` we don't show the hint about how to use it, and `run` is a lot more strict about what to run than `enter`), and it feels icky to re-use the `cobra.Command` of a different command in this way (the `root` command isn't `enter`).

So, directly re-using the top-level entry point with a fake `cobra.Command` will only cause confusion. In this case, let's handle the top-level details separately and invoke `runCommand`.

We might notice some new patterns that are repeating between `root`, `enter` and `run`, and be able to shuffle or refactor the internal functions to better accommodate them, but the top-level functions are also not that big and the code in them is mostly catering to the actual differences in those commands.

debarshiray:
recheck

debarshiray:
recheck

src/cmd/root.go
HarryMichal: The goal was to be compatible with `coreos/toolbox`. Seems like that was unneeded.

src/cmd/root.go
HarryMichal: The intent was for this to be configurable. As that is not the intent, it is redundant.

src/cmd/root.go
HarryMichal: > `enter()` doesn't actually call run() or `create(), even though it does include both those two commands.

True but `enter()` provides extra functionality on top of `run()`. The root command was meant as an alias for `create` hence the call.

Still, you've got a point with this not being a nice way to go about it. This just makes my desire for clearly splitting the command-line code from the actual Toolbox functionality grow.

HarryMichal:
Thanks @debarshiray for the review and for the update. Appreciated!

PR #813, opened 2021-06-25
cmd/run: Optimize 'enter' and 'run' in the non-fallback case

debarshiray:
Currently, the `enter` command involves two extra invocations of
`podman exec` to detect if the user's chosen shell and current working
directory are present inside the toolbox container. Each invocation is
sufficiently expensive to add a noticeable overhead to the `enter` and
`run` commands. Moreover, file system operations being inherently racy,
it's always better to detect errors and handle them instead of trying
to pre-emptively avoid them.

Therefore, this shuffles the code around to attempt the non-fallback
invocation, and then handle the errors by attempting a series of
fallbacks for the command and the current working directory.

Unfortunately, in case of a missing command, `capsh(1)` adds an extra
error message that seems difficult to get rid of:
```
  $ toolbox enter
  /bin/sh: /bin/zsh: No such file or directory
  Error: command /bin/zsh not found in container fedora-toolbox-34
  Using /bin/bash instead.
```

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -36,6 +36,9 @@ var (
 		distro    string
 		release   string
 	}
+
+	runFallbackCommands = [][]string{{"/bin/bash", "-l"}}
+	runFallbackWorkDirs = []string{"" /* $HOME */}
 )
 
 var runCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


src/cmd/run.go
```diff
@@ -299,48 +287,82 @@ func runCommand(container string,
...
+						workDir = currentUser.HomeDir
+					}
+
+					fmt.Fprintf(os.Stderr, "Using %s instead.\n", workDir)
+					runFallbackWorkDirsIndex++
+				} else {
+					err = fmt.Errorf("directory %s not found in container %s", workDir, container)
+					return err
+				}
+			} else {
```
debarshiray, line 344: We can't assume that an exit code of 127 from `podman exec` means that either the working directory or the command was absent. The given shell might also return with 127, which will get forwarded by `podman exec`, if the last command that was attempted inside the shell was absent. In such cases this will create a loop.

See https://github.com/containers/toolbox/pull/872

PR #814, opened 2021-06-25
Various miscellaneous fixes to the manuals

debarshiray:


doc/toolbox-run.1.md
```diff
@@ -13,13 +13,9 @@ toolbox\-run - Run a command in an existing toolbox container
...
 been created using the `toolbox create` command.
 
-`toolbox run` wraps around `podman exec` and by default passes several options
-to it. It allocates a tty, connects to stdin, runs the passed command as the
-current user in the current directory and shares common environmental
-variables.
-
-The executed command is wrapped in `capsh` that gets rid of all extra
-capabilities that could negatively affect the experience.
+On Fedora, the default container is known as `fedora-toolbox-N`, where N is
```
HarryMichal, line 16: This part is already mentioned in `tooöbox-create(1)`. Is it necessary to mention it here, too?

doc/toolbox-create.1.md
```diff
@@ -17,19 +17,21 @@ to interact with the container at any point.
...
+The container is created with `podman create`, and its entry point is set to
+`toolbox init-container`.
 
-If the host system is not recognized the used image will be `fedora-toolbox`.
+By default, a toolbox container is named after its corresponding image. If the
+image had a tag, then the tag is included in the name of the container, but
+it's separated by a hyphen, not a colon. A different name can be assigned by
+using the CONTAINER argument.
 
-Toolbox containers are primarily created in a way to be tightly integrated with
```
HarryMichal, line 31: I wouldn't drop completely the mention of toolbox containers not being secure.

HarryMichal (changes requested):
Thanks for the review. I highlighted a few points:

doc/toolbox-run.1.md
debarshiray: It's mentioned in `toolbox-create(1)` from the point of view of `create` - how the name of the image maps to that of the container. For a reader new to Toolbox that might still be a bit of a mental exercise.

A user will encounter the default naming scheme for containers very often, and the concept of a *default container* is very pervasive in Toolbox. So it's important to clearly explain it in the context of creating and using a container.

Generally, for technical documentation it's sometimes good to repeat the same thing if it's sufficiently important. Either to refresh the user's memory or to draw their attention to it. Having to traverse too many references can get disorienting.

HarryMichal: Very true. Thank you for the explanation!

doc/toolbox-create.1.md
debarshiray: First, if we really want to talk about security, then it's not nice to slip in a sentence like *not meant to be secure* in the middle of a second-level manual like `toolbox-create(1)`. It should really be mentioned a lot more clearly in somewhere more visible. Like `README.md` or `toolbox(1)`.

Second, words like *privileged* and *secure* are very loaded and have no meaning, unless they are clearly explained at some length. Casually mentioning those words in passing do no good, because everybody interprets them differently.

What's the definition of secure? How is it insecure? How is it insecure than any other software running on the system? Is Bash secure - it lets me run `rm -rf *`? Why are we even creating something that's insecure and hence harmful to our users?

We clearly mention what Toolbox does, the context in which it exists, and people can draw their own conclusions if that fits their definition of *secure*.

We can definitely have a section about what role Toolbox plays alongside things like OSTree-based OSes and Flatpak, because only then we can say something meaningful about security. However, that should be somewhere top-level like `README.md` or `toolbox(1)`, not `toolbox-create(1)`.

PR #818, opened 2021-06-29
tests: Made system tests non destructive

olivergs:
The tests are now non destructive.

What I've done is to setup the store configuration for podman and skopeo to save the data in a directory in the project dir when running the system tests.


test/system/000-setup.bats
```diff
@@ -2,6 +2,10 @@
 
 load 'libs/helpers'
 
+setup() {
+    _setup_containers_store
```
HarryMichal, line 6: I'd call `_setup_containers_store` as a part of the other setup functions since it is a requirement for all the tests to be non-destructible. Also, it is not pretty to call "private" functions from test files (of course, there are some exceptions - e.g., `000-setup.bats`, `000-teardown.bats`).

test/system/999-teardown.bats
```diff
@@ -2,6 +2,8 @@
 
 load 'libs/helpers'
 
+
```
HarryMichal, line 5: Nitpick: stray newline

test/system/libs/helpers.bash
```diff
@@ -2,15 +2,18 @@
 
 load 'libs/bats-support/load'
 
+# Helpful globals
+readonly PROJECT_DIR=${PWD}
+readonly IMAGE_CACHE_DIR="${PROJECT_DIR}/image-cache"
+readonly TEMP_STORAGE_DIR="${PROJECT_DIR}/test-storage-root"
+readonly ROOTLESS_PODMAN_STORE_DIR="$TEMP_STORAGE_DIR/storage"
```
HarryMichal, line 9: Nitpick: missing braces

test/system/libs/helpers.bash
```diff
@@ -2,15 +2,18 @@
 
 load 'libs/bats-support/load'
 
+# Helpful globals
+readonly PROJECT_DIR=${PWD}
```
HarryMichal, line 6: I wonder if going forward a better location to use for out temp files/image cache would be somewhere else than `$PWD` or `tmpfs` (which [cannot](https://github.com/containers/podman/issues/10693#issuecomment-863007516) be used for Podman store). The use of `$PWD` could be problematic when having the tests installed in a system-location (e.g., being installed via a a package manager into `/usr/share/`).

Maybe we could use a location in `XDG_RUNTIME_DIR`? Of course, a default has to be set in case the tests are run as root or a user without the variable set.

HarryMichal (changes requested):
Great work! This is something we sorely need to be able to test Toolbox more easily. I left a few comments to be addressed before we merge this.

test/system/libs/helpers.bash
```diff
@@ -61,6 +77,7 @@ function _pull_and_cache_distro_image() {
   fi
 
   for ((i = ${num_of_retries}; i > 0; i--)); do
+    echo $PODMAN pull ${image} >> test.log
```
HarryMichal, line 80: I'm unsure what this line is supposed to do. Could you explain, please?

test/system/000-setup.bats
olivergs: I've just removed the "private"  underscore. The reason it is called in each setup is because the export statement in the last line of _setup_containers_store is needed to be executed in each test for podman to read the variable CONTAINERS_STORAGE_CONF and point it to the correct file. I did as you told, but the test were failing because of that export not taking effect in the rest of the tests, so I just removed the "private" marking for that function.

test/system/libs/helpers.bash
olivergs: It is a good point, but I think this would be done in a different issue/pr. We need to think about the pros/cons of different places to store temporary data and the possible use cases.

test/system/libs/helpers.bash
olivergs: That was something I did to take a look at the output of podman pull and I forgot to reset to previous state.

:shame:

test/system/libs/helpers.bash
HarryMichal: Fair enough. Let's stick to what you have then.

test/system/libs/helpers.bash
```diff
@@ -27,6 +30,20 @@ function cleanup_containers() {
 }
 
 
+function setup_containers_store() {
+  mkdir -p ${TEMP_STORAGE_DIR}
+  # Setup a storage config file for PODMAN
+  echo -e "[storage]\n  driver = \"overlay\"\n  rootless_storage_path = \"${ROOTLESS_PODMAN_STORE_DIR}\"\n" > ${PODMAN_STORE_CONFIG_FILE}
```
HarryMichal, line 36: Question: `ROOTLESS_PODMAN_STORE_DIR` will be created by Podman or does it have to be created manually?

test/system/000-setup.bats
HarryMichal: Fair enough.

test/system/libs/helpers.bash
```diff
@@ -27,6 +30,20 @@ function cleanup_containers() {
...
+  export CONTAINERS_STORAGE_CONF=${PODMAN_STORE_CONFIG_FILE}
+}
+
+
+function _clean_containers_store() {
+  cleanup_all
```
HarryMichal, line 42: The environment is not set up properly so this will blast the host machine of all Podman state.

test/system/000-setup.bats
HarryMichal: Looking back, making the function a "private" one makes more sense as it is required for proper execution of the tests.

HarryMichal:
This should be ready to go now.

test/system/libs/helpers.bash
HarryMichal: Seems like it is created by Podman itself.

HarryMichal:
Thank you for the groundwork on this, @olivergs! This will make testing Toolbx so much easier!

PR #819, opened 2021-06-29
pkg/utils: Move cmd utility functions to cmd

HarryMichal:
pkg/utils has been in Go Toolbox since its birth. Along the way it
accumulated a number of functions where a few of them are purely CLI
related. Since the majority of functions in the package are related to
some "deeper" functionality in Toolbox, it makes more sense to move the
selected few to package cmd. This will make pkg/utils a bit leaner and
create a dedicated space for cmd utility functions to live in.

In the process the error creation functions no longer require the
executableBase argument to be passed to them.

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -1,0 +1,92 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+)
+
+// askForConfirmation prints prompt to stdout and waits for response from the
+// user
+//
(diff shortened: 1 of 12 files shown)


HarryMichal:
recheck

src/cmd/utils.go
```diff
@@ -0,0 +1,92 @@
+package cmd
```
debarshiray, line 1: We are missing the copyright and license notices. See https://github.com/containers/toolbox/pull/924

HarryMichal: Good point! Forgot about it.

Issue #821, opened 2021-06-30
Error: invalid entry point PID of container


kuba3351:
Hello

I am using an Fedora Silverblur Rawhide 35 and after toolbox update to version 0.0.99.2 I cannot start any container created before the upgrade, because of error:  invalid entry point PID of container

I am attaching a logs with verbose:

```
[jakub@NBJSIERZEGA sbpwsdlvalidator]$ toolbox enter -vvv
DEBU Running as real user ID 1000                 
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU Checking if /etc/subgid and /etc/subuid have entries for user jakub 
DEBU Validating sub-ID file /etc/subuid           
DEBU Validating sub-ID file /etc/subgid           
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Migrating to newer Podman                    
DEBU Toolbox config directory is /var/home/jakub/.config/toolbox 
INFO[0000] podman filtering at log level debug          
DEBU[0000] Called version.PersistentPreRunE(podman --log-level debug version --format json) 
DEBU[0000] overlay storage already configured with a mount-program 
DEBU[0000] Merged system config "/usr/share/containers/containers.conf" 
DEBU[0000] overlay storage already configured with a mount-program 
DEBU[0000] Using conmon: "/usr/bin/conmon"              
DEBU[0000] Initializing boltdb state at /var/home/jakub/.local/share/containers/storage/libpod/bolt_state.db 
DEBU[0000] Using graph driver overlay                   
DEBU[0000] Using graph root /var/home/jakub/.local/share/containers/storage 
DEBU[0000] Using run root /run/user/1000/containers     
DEBU[0000] Using static dir /var/home/jakub/.local/share/containers/storage/libpod 
DEBU[0000] Using tmp dir /run/user/1000/libpod/tmp      
DEBU[0000] Using volume path /var/home/jakub/.local/share/containers/storage/volumes 
DEBU[0000] overlay storage already configured with a mount-program 
DEBU[0000] Set libpod namespace to ""                   
DEBU[0000] [graphdriver] trying provided driver "overlay" 
DEBU[0000] overlay: mount_program=/usr/bin/fuse-overlayfs 
DEBU[0000] backingFs=extfs, projectQuotaSupported=false, useNativeDiff=false, usingMetacopy=false 
DEBU[0000] Initializing event backend journald          
[...]

PR #822, opened 2021-06-30
Use installed toolbox instead of binary from host

kuba3351:
I hit an issue (https://github.com/containers/toolbox/issues/821) that was caused by GLIBC mismatch between host and container, so toolbox binary didn't worked inside container. I have an idea to prevent this issue by not using toolbox binary from host, but install toolbox binary in the container using dnf. It will install the binary that is always able to run in the container, so I think this issue should not happen again.

This breaks compatibility with images that does not have installed toolbox, so I think a fllback logic should be implemented that checks if there is an `/usr/bin/toolbox` in the container, and if not, will mount the binary from host like before to have backward compatibility.

What do you think about?

### DIFF
--- a/images/fedora/f33/extra-packages
+++ b/images/fedora/f33/extra-packages
@@ -32,6 +32,7 @@ shadow-utils
 sudo
 tcpdump
 time
+toolbox
 traceroute
 tree
 unzip
--- a/images/fedora/f34/extra-packages
+++ b/images/fedora/f34/extra-packages
@@ -32,6 +32,7 @@ shadow-utils
 sudo
 tcpdump
 time
+toolbox
 traceroute
 tree
 unzip
--- a/images/fedora/f35/extra-packages
+++ b/images/fedora/f35/extra-packages
@@ -32,6 +32,7 @@ shadow-utils
 sudo
 tcpdump
 time
+toolbox
 traceroute
 tree
 unzip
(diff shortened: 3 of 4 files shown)


debarshiray:
I already commented on https://github.com/containers/toolbox/issues/821 about the overall problem, and the way to address it.

As for this specific pull request ...

I am apprehensive of involving RPM transactions as part of the container creation process. It makes things non-deterministic. Among other things, RPM packages have scriptlets, which get run as part of the transaction on user machines. That opens up a huge possibility for things to go wrong.

That's why OSTree-based OSes don't run RPM scriptlets on user machines. Scriptlets are discouraged, and if really necessary are run on the server, so that the outcome is the same for everybody.

Secondly, using the `toolbox` binary from the host was an explicit design choice. It means that the communication between the host and the container doesn't need to be a stable protocol, which is a huge benefit. This is also why online updates are implicitly unsupported. It means that we can change Toolbox, and expect that the user will reboot the machine as part of the update, which will restart all containers with the new version of Toolbox, keeping both sides synchronized.

Finally, *fallback logic* comes with the downside of exploding the test matrix or, even worse, being impossible to test. In this case, we will end up creating two subtly different kinds of Toolbox containers. Those that have `toolbox` mounted from the host, and those that have it installed natively inside the container. Over time, they will have different backwards compatibility requirements - backwards compatibility is tricky enough, this will be a lot worse.

Go does have the downside of having huge binaries, but the upside of that is that the binaries are mostly self-contained. So let's try to take advantage of that as much as possible. With a programming language like C, it would be a lot harder to mount in the binary across OSes from the host to the container, like we do now.

debarshiray:
Anyway, thanks for stopping by, and taking the time to work on this problem. I really appreciate that.

PR #823, opened 2021-07-03
pkg/podman: Print stderr of execs to system stderr when asked for

HarryMichal:
With this the logs (resp. stderr) of all invocations of Podman are shown
when the '--log-podman' option is set.

This change affects the behaviour of spinner which is now hidden when
the logging is turned on.

Depends on https://github.com/containers/toolbox/pull/826

### DIFF
--- a/src/pkg/podman/podman.go
+++ b/src/pkg/podman/podman.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/HarryMichal/go-version"
 	"github.com/containers/toolbox/pkg/shell"
@@ -33,6 +34,7 @@ var (
 
 var (
 	LogLevel = logrus.ErrorLevel
+	ShowLogs = false
 )
 
 // CheckVersion compares provided version with the version of Podman.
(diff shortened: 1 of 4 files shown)


src/pkg/podman/podman.go
```diff
@@ -320,15 +338,31 @@ func Start(container string, stderr io.Writer) error {
...
 
-	if err := shell.Run("podman", nil, nil, nil, args...); err != nil {
+	if err := shell.Run("podman", nil, nil, stderr, args...); err != nil {
 		return err
 	}
 
 	return nil
 }
+
+func setupStderrWriter(buffer io.Writer) io.Writer {
```
debarshiray, line 356: It will be better to put this inside `pkg/shell`, because we want the same behaviour for all spawned binaries. As a nice side-effect, it will make the change more self-contained and smaller.

debarshiray (commented):
Nice trick with `io.MultiWriter`!

Sadly, this will won't unbreak the progress bar from `podman pull` when run as `toolbox --verbose` because the standard error stream won't be connected to a file descriptor that's a terminal device. We should mention that in the commit message.

debarshiray:
Let's fold this into https://github.com/containers/toolbox/pull/787

PR #825, opened 2021-07-03
pkg/utils, cmd/create: Properly check reference being an ID

HarryMichal:
Turns out the braces do not need to be escaped. Let's fix it!

Method for matching strings against a regular expressions return
an error when the regular expressions is faulty. The first return value
tells if a match was found. So, we should check for a match and not for
an error.

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -501,7 +501,7 @@ func HumanDuration(duration int64) string {
 
 // ImageReferenceCanBeID checks if 'image' might be the ID of an image
 func ImageReferenceCanBeID(image string) bool {
-	matched, err := regexp.MatchString("^[a-f0-9]\\{6,64\\}$", image)
+	matched, err := regexp.MatchString("^[a-f0-9]{6,64}$", image)
 	if err != nil {
 		panic("regular expression for ID reference matching is invalid")
 	}
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -500,9 +500,12 @@ func HumanDuration(duration int64) string {
 }
 
 // ImageReferenceCanBeID checks if 'image' might be the ID of an image
-func ImageReferenceCanBeID(image string) (bool, error) {
+func ImageReferenceCanBeID(image string) bool {
 	matched, err := regexp.MatchString("^[a-f0-9]\\{6,64\\}$", image)
-	return matched, err
+	if err != nil {
+		panic("regular expression for ID reference matching is invalid")
+	}
+	return matched
 }
 
 func ImageReferenceGetBasename(image string) string {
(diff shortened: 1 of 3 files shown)


src/pkg/utils/utils.go
```diff
@@ -500,9 +500,12 @@ func HumanDuration(duration int64) string {
 }
 
 // ImageReferenceCanBeID checks if 'image' might be the ID of an image
-func ImageReferenceCanBeID(image string) (bool, error) {
-	matched, err := regexp.MatchString("^[a-f0-9]\\{6,64\\}$", image)
-	return matched, err
+func ImageReferenceCanBeID(image string) bool {
+	matched, err := regexp.MatchString("^[a-f0-9]{6,64}$", image)
+	if err != nil {
+		logrus.Debugf("failed to match regexp for ID reference: %v", err)
```
debarshiray, line 506: I'd use a `panic` here. An error can only be thrown if the regular expression is faulty, which would be a programmer error, not a run-time condition.

debarshiray (changes requested):
Good catch!

I wish https://golang.org/pkg/regexp/#MatchString said this clearly upfront. However, the sample code seems to indicate that an error is only thrown if the regular expression is invalid.

src/pkg/utils/utils.go
HarryMichal: Fixed.

src/pkg/utils/utils_test.go
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright © 2019 – 2021 Red Hat Inc.
```
debarshiray, line 2: This should just be 2021. The range starts from the year the file (or the code in it) was created.

src/pkg/utils/utils.go
```diff
@@ -500,9 +500,12 @@ func HumanDuration(duration int64) string {
 }
 
 // ImageReferenceCanBeID checks if 'image' might be the ID of an image
-func ImageReferenceCanBeID(image string) (bool, error) {
-	matched, err := regexp.MatchString("^[a-f0-9]\\{6,64\\}$", image)
-	return matched, err
+func ImageReferenceCanBeID(image string) bool {
+	matched, err := regexp.MatchString("^[a-f0-9]{6,64}$", image)
```
debarshiray, line 504: Ah, I missed the change in the escaping.

This is potentially an actual behavioral change, so we should put it in a separate commit. Especially because regular expressions are really weird. eg., `grep` differentiates between so-called basic and extended regular expressions where the characters `?`, `+`, `{`, `|`, `(` and `)` need or need not be escaped.

debarshiray (changes requested):
Thanks for updating the pull request!

src/pkg/utils/utils_test.go
HarryMichal: Ah, right.

src/pkg/utils/utils.go
HarryMichal: I moved the change to a separate commit.

PR #826, opened 2021-07-03
cmd: Add function to determine if a spinner can be started

HarryMichal:
The logic for determining if a spinner can be started is non-trivial and
hard to replicate.

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -1,0 +1,33 @@
+/*
+ * Copyright © 2019 – 2021 Red Hat Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


src/cmd/utils.go
```diff
@@ -0,0 +1,33 @@
...
+package cmd
+
+import (
+	"os"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+func canStartSpinner() bool {
```
debarshiray, line 26: The spinner is only used in `src/cmd/create.go`, so this function should go in that file.

debarshiray (changes requested):
Is this still important with the current state of https://github.com/containers/toolbox/pull/787 ?  At least it needs a rebase.

src/cmd/utils.go
HarryMichal: I'd rather put it into `src/cmd/utils.go` because it isn't strictly related to the `create` command.

HarryMichal:
> Is this still important with the current state of #787 ? At least it needs a rebase.

I'd merge it anyway. There's always the possibility of us adding a new spinner.

PR #827, opened 2021-07-03
cmd/create: Mount whole host instead of its subsets

HarryMichal:
Having whole host filesystem mounted inside of a toolbox is a very handy
feature that makes the toolboxes much more complete environment and
covers systems with diverse root directories (e.g., Fedora Silverblue).

Thanks to the chnge we can drop several separate mounts which are being
handled now by mounting the whole host filesystem.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -265,28 +265,6 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 	logrus.Debugf("%s canonicalized to %s", currentUser.HomeDir, homeDirEvaled)
 	homeDirMountArg := homeDirEvaled + ":" + homeDirEvaled + ":rslave"
 
-	bootMountFlags := "ro"
-	isBootReadWrite, err := isPathReadWrite("/boot")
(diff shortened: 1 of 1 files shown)


src/cmd/create.go
```diff
@@ -265,28 +265,6 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 	logrus.Debugf("%s canonicalized to %s", currentUser.HomeDir, homeDirEvaled)
 	homeDirMountArg := homeDirEvaled + ":" + homeDirEvaled + ":rslave"
 
-	bootMountFlags := "ro"
-	isBootReadWrite, err := isPathReadWrite("/boot")
```
debarshiray, line 269: I wonder if there are host operating systems out there where we will have to check the read/write status of `/`?

I guess it's good to keep it simple unless we know for sure. :)

debarshiray (commented):
Thanks for working on this, @HarryMichal 

I merged it after tweaking the commit message to emphasize why the *more complete environment* is useful.

PR #828, opened 2021-07-03
cmd/root: Add support for configuration files

HarryMichal:
Some aspects of Toolbox could be configured (e.g., the default used
image). Let's enable that!


data/config/toolbox.conf
```diff
@@ -0,0 +1,13 @@
+[general]
+# Default used distribution (e.g., fedora, rhel)
+# Taken from ID field in /etc/os-release
```
debarshiray, line 3: This is the *fallback* distribution, in case the host OS isn't supported, right? The default should still the same distribution as the host as long as it's recognized.

We should explain the distinction better in the comment. Otherwise it can be confusing.

data/config/toolbox.conf
```diff
@@ -0,0 +1,13 @@
...
+## default_distro = "fedora"
+
+# Default version of a distribution (e.g., 34, 8.4)
+# Taken from VERSION_ID field in /etc/os-release
+## default_release = "34"
+
+# Default used image
+# Overrides 'default_distro' and 'default_release'
+# E.g., registry.fedoraproject.org/fedora-toolbox
+## default_image = "registry.fedoraproject.org/fedora-toolbox"
```
debarshiray, line 13: The `:34` tag should probably be included. Otherwise it looks incomplete.

src/cmd/root.go
```diff
@@ -354,6 +362,34 @@ func newSubIDFileError() error {
 	return errors.New(errMsg)
 }
 
+func setUpConfiguration() error {
+	logrus.Debug("Setting up configuration")
+
+	configFiles := []string{
+		"/etc/containers/toolbox.conf",
+		fmt.Sprintf("%s/.config/containers/toolbox.conf", currentUser.HomeDir),
```
debarshiray, line 370: We should use `os.UserConfigDir` as in the rest of the code (and handle the error).

debarshiray (changes requested):
Thanks for working on this, @HarryMichal !  Some quick comments:

debarshiray (changes requested):

debarshiray:
I split out the break-up of `ResolveContainerAndImageNames` into https://github.com/containers/toolbox/pull/838

data/config/toolbox.conf
HarryMichal: > This is the fallback distribution, in case the host OS isn't supported, right?

No, I meant it as an override that supersedes even the host OS recognition. Does that sound like a bad option to have?

data/config/toolbox.conf
HarryMichal: Right

src/cmd/root.go
HarryMichal: Ah! I did not know about that one.

src/cmd/root.go
```diff
@@ -143,6 +144,11 @@ func preRun(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	err := setUpConfiguration()
+	if err != nil {
```
debarshiray, line 148: Umm... wouldn't it be more idiomatic to put this on the same line? We don't need `err` outside the error check.

data/config/toolbox.conf
debarshiray: Ok, I see. Thinking about this some more, what you are saying makes more sense because it makes it identical to the `--distro` option.

Generally speaking, the configuration file options should mimic the command line ones. If something can't be tweaked through the command line, then it shouldn't be possible through the configuration file either. There might be valid exceptions, but as a general guideline, this makes things more manageable (ie., easier to document, explain, etc.).

src/pkg/utils/utils.go
```diff
@@ -288,7 +292,7 @@ func GetContainerNamePrefixForImage(image string) (string, error) {
 
 func GetDefaultImageForDistro(distro, release string) string {
 	if _, supportedDistro := supportedDistros[distro]; !supportedDistro {
-		distro = "fedora"
+		distro = viper.GetString("general.default_distro")
```
debarshiray, line 295: Umm... is this correct? The `fedora` here is meant as a hard-coded *fallback* that will always work. That's why it also can't be changed through the command line, because otherwise we can't guarantee it. Hence it shouldn't be influenced by the configuration file either.

data/config/toolbox.conf
debarshiray: In that same spirit, we should also use the same blurbs that we use to document the command line options in the manuals. That would reiterate the fact that they are the same options, but with a different priority (ie., command line overrides configuration file).

Any deviation has the potential to confuse a user in unexpected ways.

data/config/toolbox.conf
```diff
@@ -0,0 +1,13 @@
+[general]
```
debarshiray, line 1: We need a `meson.build` to actually install this file, so that users can tweak it easily.

debarshiray:
I took the liberty to address some of the above comments and updated the branch.

src/pkg/utils/utils.go
```diff
@@ -707,24 +711,29 @@ func ResolveImageName(distro, image, release string) (string, string, error) {
 	logrus.Debugf("Image: '%s'", image)
 	logrus.Debugf("Release: '%s'", release)
 
+	if viper.IsSet("general.image") {
+		image = viper.GetString("general.image")
```
debarshiray, line 715: This doesn't look right, because it will override the `--image` command line option with the `image` option from the configuration file.

Explicitly specified command line options should always have priority over the configuration file.

HarryMichal: You're right.

src/cmd/root.go
```diff
@@ -359,6 +364,31 @@ func newSubIDFileError() error {
...
+		if err := viper.MergeInConfig(); err != nil {
+			if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+				logrus.Debugf("Setting up configuration: file %s not found", configFile)
+			} else {
+				logrus.Debugf("Setting up configuration: failed to merge file %s: %s", configFile, err)
+			}
+		}
+	}
+
+	return true, nil
```
HarryMichal, line 389: There's no need to return `error` since we handle errors gracefully and do not fail early in any case.

src/pkg/utils/utils.go
HarryMichal: I'm unsure about this exact behaviour. Somebody may request some unsupported distribution but what they get is Fedora. In such a case I'd expect the tool to yell at me that the requested distro is not supported instead falling back to Fedora.

HarryMichal: Hmm... maybe I went a bit overboard here. Still, I'm not entirely sure we want to have a hard-coded fallback to Fedora. Yes, Fedora is our default but the user may want something else. But I may be wrong here.

If we keep the current behaviour, we could drop the following check with the panic because we can be certain the fallback will save the day.

debarshiray: Whether we make the hard-coded fallback configurable or not is a different discussion. Currently it's not tweakable through the command line interface, so there's no need to expose it through the configuration file.

The way things stand at the moment, `fedora` is the only feasible hard fallback. It's the only fully supported community driven image. So if somebody runs `toolbox create` on Arch or Debian, we really don't have any other option than to fall back to `fedora`. :)

Once supported images for Arch or Debian are in place, the code won't need to hit the fallback.

src/cmd/root.go
debarshiray: Yes, you are right. I will drop it.

src/cmd/root.go
```diff
@@ -359,6 +364,35 @@ func newSubIDFileError() error {
...
+		configFiles = append(configFiles, fmt.Sprintf("%s/containers/toolbox.conf", userConfigDir))
+	}
+
+	viper.SetConfigType("toml")
+
+	for _, configFile := range configFiles {
+		viper.SetConfigFile(configFile)
+
+		if err := viper.MergeInConfig(); err != nil {
+			if _, ok := err.(viper.ConfigFileNotFoundError); ok {
```
debarshiray, line 385: Does it actually work? It doesn't seem like Viper actually throws this error when a file is missing.

HarryMichal: Yes, it does on my machine.

src/cmd/root.go
```diff
@@ -359,6 +364,35 @@ func newSubIDFileError() error {
...
+				logrus.Debugf("Setting up configuration: file %s not found", configFile)
+			} else {
+				logrus.Debugf("Setting up configuration: failed to merge file %s: %s", configFile, err)
```
debarshiray, line 388: We should probably error out for other errors. eg., if the configuration file can't be parsed.

src/pkg/utils/utils.go
```diff
@@ -708,23 +710,27 @@ func ResolveImageName(distro, image, release string) (string, string, error) {
...
+		if viper.IsSet("general.image") {
+			image = viper.GetString("general.image")
+		} else {
+			image = getDefaultImageForDistro(distro, release)
+		}
 	} else {
 		release = ImageReferenceGetTag(image)
 		if release == "" {
-			release = releaseDefault
+			release = viper.GetString("general.release")
```
debarshiray, line 733: Once we potentially read `image` from the configuration file, the `release` needs to be extracted from it.

Earlier, `getDefaultImageForDistro` generated the `image` from the default `release`, but with a configuration file the release will be embedded in the `image`.

src/cmd/root.go
```diff
@@ -359,6 +364,35 @@ func newSubIDFileError() error {
 	return errors.New(errMsg)
 }
 
+func setUpConfiguration() error {
```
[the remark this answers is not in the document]
HarryMichal: Good point.

src/cmd/root.go
HarryMichal: Yes, I agree.

src/cmd/root.go
HarryMichal: I was wrong :)

src/pkg/utils/utils.go
```diff
@@ -708,23 +710,27 @@ func ResolveImageName(distro, image, release string) (string, string, error) {
 	logrus.Debugf("Release: '%s'", release)
 
 	if distro == "" {
-		distro = distroDefault
+		distro = viper.GetString("general.distro")
 	}
 
 	if distro != distroDefault && release == "" {
```
debarshiray, line 716: This condition needs to be updated.

The idea is that if we have a non-default `distro` (ie., one that doesn't match `/etc/os-release`), then we need the user to specific a `release` either through the command line or the configuration file. Currently, this isn't checking the presence of a release in the configuration file.

src/pkg/utils/utils.go
HarryMichal: Isn't this covered by `ImageReferenceGetTag()` called a bit earlier?

HarryMichal: Oops, now I see that function is not called in case the value is set in the config.

HarryMichal: Brain food: I wonder whether this exact line behaves correctly. If you run `podman pull registry.fedoraproject.org/fedora`, it'll try to pull `registry.fedoraproject.org/fedora:latest`. Note the `latest` tag. Maybe we should do the same instead of using the default.

src/cmd/create.go
```diff
@@ -505,10 +505,10 @@ func getDBusSystemSocket() (string, error) {
 
 func getEnterCommand(container, release string) string {
 	var enterCommand string
-	containerNamePrefixDefaultWithRelease := utils.ContainerNamePrefixDefault + "-" + release
+	containerNamePrefixDefaultWithRelease := utils.GetDefaultContainerNamePrefix() + "-" + release
 
 	switch container {
-	case utils.ContainerNameDefault:
+	case utils.GetDefaultContainerNamePrefix():
```
debarshiray, line 511: This doesn't look right. We don't want only the prefix here, but the entire name including the release.

src/pkg/utils/utils.go
debarshiray: Yes, I thought about `latest` too. Not sure why we don't handle that. Let's ignore it, because it's an issue of its own.

src/pkg/utils/utils.go
```diff
@@ -285,9 +283,13 @@ func GetContainerNamePrefixForImage(image string) (string, error) {
 	return basename, nil
 }
 
-func GetDefaultImageForDistro(distro, release string) string {
+func GetDefaultContainerNamePrefix() string {
+	return containerNamePrefixDefault
```
debarshiray, line 287: This doesn't seem to address the problem that the hint shown by `create` is influenced by the configuration file.

I think the main complication is that the configuration file lets us specify an image. Earlier we would just pick the host and release from `/etc/os-release`, map that to a fixed image, and go. Now, that mapping can be overridden by the configuration file.

This means that `toolbox create` without any parameters might still use a custom image, and lead to a custom container name.

src/pkg/utils/utils.go
```diff
@@ -700,23 +702,26 @@ func ResolveImageName(distro, image, release string) (string, string, error) {
...
+		return "", "", fmt.Errorf("release not found for non-default distribution %s", distro)
 	}
 
-	if image == "" {
-		image = GetDefaultImageForDistro(distro, release)
+	if image == "" && viper.InConfig("general.image") {
+		image = viper.GetString("general.image")
+		release = ImageReferenceGetTag(image)
+	} else if image == "" {
+		image = getDefaultImageForDistro(distro, release)
```
debarshiray, line 720: Umm... on second thoughts, this is also not quite correct.

The `--distro` and `--release` options on the command line should override the configuration file's `image` option.

test/system/101-create.bats
```diff
@@ -69,3 +69,32 @@ teardown() {
 
   assert_output --regexp "Created[[:blank:]]+fedora-toolbox-32"
 }
+
+@test "create: Create a container with overriden distro option (system-wide)" {
+    skip_if_system_config_unwritable
```
debarshiray, line 74: Do these tests actually do anything?

test/system/libs/helpers.bash
```diff
@@ -11,6 +11,11 @@ readonly SKOPEO=$(command -v skopeo)
 readonly PROJECT_DIR=${PWD}
 readonly IMAGE_CACHE_DIR="${PROJECT_DIR}/image-cache"
 
+readonly TOOLBOX_SYSTEM_CONFIG="/etc/containers/toolbox.conf"
+readonly TOOLBOX_USER_CONFIG="~/.config/containers/toolbox.conf"
+readonly BACKUP_SYSTEM_CONFIG="$BATS_TMPDIR/system-toolbox.conf"
+readonly BACKUP_USER_CONFIG="$BATS_TMPDIR/user-toolbox.conf"
```
debarshiray, line 17: Are we really moving around the actual system-wide and user-specific configuration files for the test suite? If so, then it would make it difficult to isolate the test suite and make it non-destructive.

We should probably override `$HOME` for the user-specific one. If it's hard to redirect the system-wide file, then we could skip it.

debarshiray:
I split out everything, other than the tests in https://github.com/containers/toolbox/pull/851

I left out the tests because they seem to be incomplete and not do anything.

HarryMichal:
Closing. The work on tests can be done in a separate PR.

test/system/libs/helpers.bash
HarryMichal: Maybe we should offer an env var for the tests switching on the tests working with system-wide configuration files.

PR #830, opened 2021-07-07
cmd, pkg/utils: Drop release arg from some function signatures

martymichal:
The release string does not have to be passed as an argument in the cmd
package since flag values are globally available and the image argument
is already fully qualified (contains tag -> release).

GetFullyQualifiedImageFromDistros() function already receives an image
reference with tag included. There's no need to pass the release
separately.

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -320,7 +320,7 @@ func GetEnvOptionsForPreservedVariables() []string {
 	return envOptions
 }
 
-func GetFullyQualifiedImageFromDistros(image, release string) (string, error) {
+func GetFullyQualifiedImageFromDistros(image string) (string, error) {
 	logrus.Debugf("Resolving fully qualified name for image %s from known registries", image)
 
 	if ImageReferenceHasDomain(image) {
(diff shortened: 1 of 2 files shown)


src/pkg/utils/utils.go
```diff
@@ -332,6 +332,11 @@ func GetFullyQualifiedImageFromDistros(image, release string) (string, error) {
 		return "", fmt.Errorf("failed to get the basename of image %s", image)
 	}
 
+	release := ImageReferenceGetTag(image)
+	if release == "" {
```
debarshiray, line 336: Commit [262c90e06fdb91e0](https://github.com/containers/toolbox/commit/262c90e06fdb91e0b693fae33a519eb2756de75b) added some assertions to ensure that the `image` and `release` values stay coherent and don't drift apart, and that needs `release`.

I wonder if the assertion can be moved higher up the call chain to `createContainer()` to drop the `release` argument, like you did here, but then we will have to parse it at runtime and handle the potential error.

I don't know what's the best way forward without digging deeper into the code and its history.  Please let me know if you have some thoughts.

debarshiray (commented):
Oops!  This has been lying unattended for a long time.  My apologies, @martymichal !

PR #831, opened 2021-07-07
tests: Added a verbose test for container starting

olivergs:


test/system/103-run.bats
```diff
@@ -19,6 +19,29 @@ teardown() {
   cd "$CURDIR" || return 1
 }
 
+@test "run: containerstart: Check container starts without issues" {
```
HarryMichal, line 22: I am not a fan of the `context: context: testname` syntax. Maybe we could be a bit radical and shuffle the tests a bit to give more flexibility for tests like this.

test/system/103-run.bats
```diff
@@ -19,6 +19,29 @@ teardown() {
...
+
+  create_default_container
+
+  run $PODMAN start "$(get_system_id)-toolbox-$(get_system_version)"
```
HarryMichal, line 26: Maybe use a variable for the container name?

test/system/103-run.bats
```diff
@@ -19,6 +19,29 @@ teardown() {
...
+
+  CONTAINER_INITIALIZED=0
+
+  for TRIES in 1 2 3 4 5
+  do
+    run $PODMAN logs "$(get_system_id)-toolbox-$(get_system_version)"
+    CONTAINER_OUTPUT=$output
+    run grep 'Listening to file system and ticker events' <<< $output
```
HarryMichal, line 34: Replace `$output` with `$CONTAINER_OUTPUT`.

HarryMichal (changes requested):
I like the solution. Left a few comments.

test/system/103-run.bats
olivergs: I can remove the "run:" and just leave the containerstart? The other option is to create a new file for container tests and put it before the run ones

test/system/103-run.bats
olivergs: Cool. Adding this to code :+1:

test/system/103-run.bats
olivergs: Yes. Nice catch. Forgot to change there. :+1:

test/system/103-run.bats
HarryMichal: Maybe `start` would be enough. And I'd put it into its own file.

olivergs: Done!

HarryMichal (approved):
I think this is good to go. What do you think @debarshiray?

test/system/103-run.bats
HarryMichal: After looking at https://github.com/containers/toolbox/pull/839 I think using `container` instead of `start` will be better. That way the two tests could live in the same file since they are both related to correct behaviour and state of toolbox containers.

olivergs: Makes sense :)

HarryMichal:
I took the liberty to update the PR. Fixed some a couple of newlines and added reference to the PR to the commit message.

PR #834, opened 2021-07-08
playbooks/setup-env: Show version of glibc

HarryMichal:
An upgrade of glibc has caused an issue on Fedora Rawhide[0]. We need a
clear indicator that a change in glibc could cause it.

[0] https://github.com/containers/toolbox/issues/821


olivergs:
LGTM. :+1:

PR #837, opened 2021-07-13
utils, help: Show basic help when "man" command is not available

### LINKED ISSUE #713: RFE: Support systems without man installed
Fedora CoreOS systems do not have the man command installed. Running `toolbox --help` on such a system results in:
```
$ toolbox --help
Error: man(1) not found
```
which is not a great user experience.

toolbox uses the https://github.com/spf13/cobra module for argument parsing and this module support generating help text from argument parsing code. I have not yet dig into the code to figure out what's needed to use that in toolbox.

olivergs:
Fixes #713

Fedora CoreOS systems do not have the man command installed. Running
toolbox --help on such a system results in a "man(1) not found" error.

As a compromise for systems without man, we added a simple help text
showing the most commonly used toolbox commands and an URL that direct
users to the Toolbox website where they can find the manuals in Markdown
format.

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -67,7 +67,16 @@ func showManual(manual string) error {
 	manBinary, err := exec.LookPath("man")
 	if err != nil {
 		if errors.Is(err, exec.ErrNotFound) {
-			return errors.New("man(1) not found")
+			fmt.Print(`toolbox - Tool for containerized command line environments on Linux
+
+Common commands are:
+create    Create a new toolbox container
+enter     Enter an existing toolbox container
+list      List all existing toolbox containers and images
+
+Go to https://github.com/containers/toolbox for further information.
+`)
+			return nil
 		}
 
 		return errors.New("failed to lookup man(1)")
--- a/test/system/002-help.bats
+++ b/test/system/002-help.bats
@@ -9,10 +9,14 @@ setup() {
 }
 
 @test "help: Run command 'help'" {
+  if ! hash man 2>/dev/null; then
+    skip "Test works only if man is in PATH"
+  fi
+
   run $TOOLBOX help
 
   assert_success
-  assert_output --partial "toolbox - Tool for containerized command line environments on Linux"
+  assert_line --index 0 --partial "toolbox(1)()"
 }
 
 @test "help: Run command 'help' with no man present" {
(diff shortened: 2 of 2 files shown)


HarryMichal (commented):
The change is correct and fixes the issue but I'm wondering whether the `utils` package is the correct place for this text to live in. But if https://github.com/containers/toolbox/pull/819 gets merged in the future, my complaint will cease to exist since the function will move to the `cmd` package.

olivergs:
As this fix is "easy" to resolve, we can wait for #819 merge and then put this ontop.

HarryMichal:
I took the liberty of updating the PR. I tested it on a FCOS host and it works. If we ever get FCOS into our CI the newly written test will also cover this scenario. Thank you for your work on this!

src/cmd/utils.go
```diff
@@ -67,7 +67,16 @@ func showManual(manual string) error {
...
 		if errors.Is(err, exec.ErrNotFound) {
-			return errors.New("man(1) not found")
+			fmt.Print(`toolbox - Tool for containerized command line environments on Linux
+
+Common commands are:
+create    Create a new toolbox container
+enter     Enter an existing toolbox container
+list      List all existing toolbox containers and images
+
+Go to https://github.com/containers/toolbox for further information.
```
debarshiray, line 77: It will be better to use `strings.Builder`, like we do elsewhere in the code.  It avoids the odd indentation.

test/system/002-help.bats
```diff
@@ -9,10 +9,30 @@ setup() {
 }
 
 @test "help: Run command 'help'" {
+  if ! hash man 2>/dev/null; then
```
debarshiray, line 12: `command -v man` might be a better option than `hash man`.  While `hash` does get the job done, it doesn't seem to be the most appropriate tool for the job, and `command` is more obvious when reading.

debarshiray (commented):
Thanks for working on this!

debarshiray:
Thanks for adding the tests! :)

test/system/002-help.bats
HarryMichal: Ah, right. Forgot about the existence of `command`.

src/cmd/utils.go
HarryMichal: Good point :+1:

debarshiray: Submitted https://github.com/containers/toolbox/pull/922

test/system/002-help.bats
debarshiray: Submitted https://github.com/containers/toolbox/pull/922

Issue #836, opened 2021-07-12
RFE: Bash & Zsh completion


travier:
Provide Bash & Zsh completion for command line options.

This is already supported by the argument parsing library currently used in toolbox (https://github.com/spf13/cobra/blob/master/shell_completions.md).

olivergs:
@travier I've created this PR:

https://github.com/containers/toolbox/pull/840

It is still a WIP and needs some work, but it is able to generate the Zsh completions by using

`toolbox completion zsh`

That will output the completions to stdout.

If you could check them works as expected from a Zsh user POV would be great.

PR #840, opened 2021-07-16
WIP cmd: Added shell completion command

olivergs:
Initial work for fixing #836

I need to dig a bit more in the command completion generation to enhance the completions for args like distro.

### DIFF
--- a/src/cmd/completion.go
+++ b/src/cmd/completion.go
@@ -1,0 +1,71 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var completionCmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate completion script",
+	Long: `To load completions:
+
+Bash:
+
+  $ source <(yourprogram completion bash)
+
+  # To load completions for each session, execute once:
+  # Linux:
+  $ yourprogram completion bash > /etc/bash_completion.d/yourprogram
+  # macOS:
+  $ yourprogram completion bash > /usr/local/etc/bash_completion.d/yourprogram
+
+Zsh:
+
+  # If shell completion is not already enabled in your environment,
+  # you will need to enable it.  You can execute the following once:
+
+  $ echo "autoload -U compinit; compinit" >> ~/.zshrc
+
+  # To load completions for each session, execute once:
+  $ yourprogram completion zsh > "${fpath[1]}/_yourprogram"
+
+  # You will need to start a new shell for this setup to take effect.
+
+fish:
+
+  $ yourprogram completion fish | source
+
+  # To load completions for each session, execute once:
+  $ yourprogram completion fish > ~/.config/fish/completions/yourprogram.fish
+
+PowerShell:
+
+  PS> yourprogram completion powershell | Out-String | Invoke-Expression
+
+  # To load completions for every new session, run:
+  PS> yourprogram completion powershell > yourprogram.ps1
+  # and source this file from your PowerShell profile.
+`,
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
+	Args:                  cobra.ExactValidArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		switch args[0] {
+		case "bash":
+			cmd.Root().GenBashCompletion(os.Stdout)
+		case "zsh":
+			cmd.Root().GenZshCompletion(os.Stdout)
+			// case "fish":
+			// 	cmd.Root().GenFishCompletion(os.Stdout, true)
+			// case "powershell":
+			// 	cmd.Root().GenPowerShellCompletionWithDesc(os.Stdout)
+			// }
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(completionCmd)
+}
--- a/src/meson.build
+++ b/src/meson.build
@@ -5,6 +5,7 @@ go_build_wrapper_program = find_program('go-build-wrapper')
 
 sources = files(
   'toolbox.go',
+  'cmd/completion.go',
   'cmd/create.go',
   'cmd/enter.go',
   'cmd/help.go',


travier:
This looks like a great start and works nicely for general options. Here are some suggestions for next steps:

- Add "extra completion" options as manually specified in https://github.com/containers/toolbox/blob/main/completion/bash/toolbox for existing toolbox containers, images and supported distros
- Remove manually generated Bash completion and generate & install completion for Bash & Zsh in RPM spec to skip the need for manual user installation on systems

olivergs:
I was working on that today. I'm fixing some bits here and there and update the branch.

About the generated bash and rpm install, I will work on that after I get the completion feature finished.

src/cmd/completion.go
```diff
@@ -0,0 +1,100 @@
...
+		distros = append(distros, key)
+	}
+	return distros
+}
+
+func getImageCompletionNames() []string {
+	var image_names []string
+	if images, err := getImages(); err == nil {
+		for _, image := range images {
+			image_names = append(image_names, image.Names[0])
```
HarryMichal, line 86: Be careful here because `image.Names[0]` may not contain a single string. See https://github.com/containers/toolbox/pull/800.

HarryMichal:
I rebased this on top of https://github.com/containers/toolbox/pull/917 to benefit from the work on updating dependencies (required for the Cobra features this PR makes use of).

HarryMichal:
This should be good for testing. @debarshiray @olivergs @travier :pray:

HarryMichal:
Darn you Zuul! This works locally :/.

travier:
I gave it a try today by doing:
```
$ meson -Dprofile_dir=/etc/profile.d builddir
$ ninja -C buillddir
$ ./builddir/src/toolbox completion zsh > ~/.path/to/zsh/completion/zsh_completions/_toolbox
```
but as far as I could see, the comments from https://github.com/containers/toolbox/pull/840#issuecomment-884149688 still apply. This is however still progress as at least the commands are completed so addressing my comments could be for a later PR.

Thanks!

HarryMichal:
> but as far as I could see, the comments from [#840 (comment)](https://github.com/containers/toolbox/pull/840#issuecomment-884149688) still apply. This is however still progress as at least the commands are completed so addressing my comments could be for a later PR.

Which part of the comment do you mean? If the "extra completion" then could you elaborate a bit more? I don't quite understand the suggestion in its current form. The later part should be taken care of at build (resp. install) time. The completion is generated using the build system.

> Thanks!

Also thanks!

travier:
Yes, it's about the extra completion. Being able to type `toolbox run <TAB>` and to get a list of existing toolbox containers to use. If this is supposed to work then I might have something during testing.

HarryMichal:
> Yes, it's about the extra completion. Being able to type `toolbox run <TAB>` and to get a list of existing toolbox containers to use.

That's what my work on top of Oliver's was mainly about.

> If this is supposed to work then I might have something during testing.

Your way of setting up the completion seems just fine. It also depends on the context where you execute this. In a toolbx container the container/image completions won't work unless you have a Podman shim in-place (the completion leverages Toolbox's code). And the `toolbox run <TAB>` example won't work as that is not the command's signature. Try `toolbox run -c <TAB>` or `toolbox create -d <TAB>`.

travier:
Hum, do I also need to install the new toolbox binary for it to work? Have not tried that yet.

olivergs:
Yes. You need the binary with the updated code as the completion depends on a "private subcommand" in the binary to "discover" the options for the command completion feature.

travier:
OK, I've done that now, this works for the `create -c` and `run -c` commands!
Thanks and sorry for the confusion.

travier (approved):
LGTM 🙂

HarryMichal:
recheck

HarryMichal:
From Matrix chat:
```
martymichal: starlighthastro, I managed to reproduce the build error when building the cobra-completions branch. Will try to fix it going forward.
martymichal: Peculiar... according to 'ls' and 'stat' the file exists but executing it from the command line (e.g., './build/src/toolbox') yields 'bash: ./build/src/toolbox: No such file or directory'.
martymichal: Any ideas what the problem could be?
martymichal: Ok, found it. The problem is in `/run/host` not existing on the host.
```

debarshiray:
> This should be good for testing. @debarshiray @olivergs @travier

I haven't looked at this, yet.

We urgently need to get a new release out, and that requires:
* Fixes to our build system to not require `patchelf(1)` (ie., https://github.com/containers/toolbox/pull/942)
* Conditionalizing some of the compatibility code that we have for github.com/coreos/toolbox

I will get to this once that's done.

oturpe:
I took a look at this to test if it fixes #975.
So I built from this branch and installed on my system.
That issue is fixed,
but I found two points that could be improved:

If the subcommand is invalid,
Bash completion suggests any subcommand as the next argument.
Like this:
```
$ toolbox invalid-command <Tab><Tab>
create  enter   help    list    rm      rmi     run
```
I would expect no suggestions there,
because nothing but _Error: unknown command "invalid-command" for "toolbox"_
can result if the command starts like that.

If the subcommand is valid,
options are not suggested before a hyphen is inserted.
I would expect to given them even without prefix.
Like this:
```
$ toolbox create <Tab><Tab>
(nothing happens)
$ toolbox create -<Tab><Tab>
--assumeyes   --container=  --distro=     --image=      --log-podman  --release=    -y            
-c            -d            -i            --log-level   -r            -v            
--container   --distro      --image       --log-level=  --release     --verbose     
```

oturpe:
On closer look,
the two issues I listed above do not seem to be specific to this pull request.
I ǵet the same behavior if I use the `$ toolbox completion bash` completions.
This is on Fedora, package version `toolbox-0.0.99.2^3.git075b9a8d2779-4.fc35.x86_64`.

HarryMichal:
Thank you @olivergs for the initial work on this! It helped me tremendously!

debarshiray (commented):
I didn't realize that this one was ready to go in, because of the *WIP* in the title.

HarryMichal:
Forgot to change the title. Fixed it just now.

completion/meson.build
```diff
@@ -0,0 +1,36 @@
...
+  command: [generate_completions_program, meson.global_source_root() / 'src', 'bash'],
+  install: true,
+  install_dir: bash_comp_dir,
+  output: 'toolbox')
+
+completion_zsh = custom_target('zsh-completion',
+  capture: true,
+  command: [generate_completions_program, meson.global_source_root() / 'src', 'zsh'],
+  install: true,
+  install_dir: get_option('datadir') / 'zsh' / 'site_functions',
```
debarshiray, line 28: Isn't it `site-functions`, not `site_functions`?  At least that's what I have on Fedora 35.

meson_options.txt
```diff
@@ -17,3 +17,10 @@ option(
   description: 'Directory for system-wide tmpfiles.d(5) files',
   type: 'string',
 )
+
+option(
+  'install_completions',
```
debarshiray, line 22: Umm... do we need a new build flag for completions?

With this pull request, we now have several subtly different knobs for having shell completions.  First, there's the `install_completions` build flag, then we have the optional dependencies on `bash-completion` and `fish`, but we have some fallbacks in case they are missing.  Documenting those three layers is going to be needlessly more difficult in comparison to what [we had before](https://containertoolbx.org/install/):
```
The following dependencies enable various optional features:
  * bash-completion
```

Then we would have to test those layers.

Generally speaking, we shouldn't be adding build options and fallbacks without well-defined use-cases.

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+
+  $ toolbox completion fish | source
+
+  # To load completions for each session, execute once:
+  $ toolbox completion fish > ~/.config/fish/completions/toolbox.fish
+
+`,
+	Hidden:                true,
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "zsh", "fish"},
```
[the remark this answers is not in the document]
debarshiray: https://github.com/containers/podman/issues/14110 says this about PowerShell:
*on windows you would need to manually load it into powershell*

... which makes me think that we should blindly support it without signing up to testing it.

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+
+func completionEmpty(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return nil, cobra.ShellCompDirectiveNoFileComp
+}
+
+func completionCommands(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	commandNames := []string{}
+	commands := cmd.Root().Commands()
+	for _, command := range commands {
+		if strings.Contains(command.Name(), "complet") {
```
debarshiray, line 81: *complet*, not *completion* ?

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+	"strings"
+
+	"github.com/containers/toolbox/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var completionCmd = &cobra.Command{
+	Use:   "completion [bash|zsh|fish|powershell]",
+	Short: "Generate completion script",
+	Long: `To load completions:
```
debarshiray, line 15: Note that this help string is likely not going to show up anywhere.  We explicitly try to consolidate our documentation in the manuals, to avoid fragmentation, and those don't come from here.  The strings in the Cobra `Command` instances are purely for developer documentation.

Secondly, there are no stability guarantees for the command line interface of this command.  It's purely an internal implementation detail that is liable to change anytime.  Any non-trivial documentation can give the wrong impression.

Third, the `completion` command comes with its own shell completion that's self-documenting and self-explanatory.  The documentation here is about how completion generally works in the shells, which isn't really Toolbx's domain.

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+
+Bash:
+
+  $ source <(toolbox completion bash)
+
+  # To load completions for each session, execute once:
+  # Linux:
+  $ toolbox completion bash > /etc/bash_completion.d/toolbox
+  # macOS:
```
debarshiray, line 24: Toolbx is Linux-only.  :)

src/cmd/create.go
```diff
@@ -91,6 +92,14 @@ func init() {
 		"Create a toolbox container for a different operating system release than the host")
 
 	createCmd.SetHelpFunc(createHelp)
+
+	if err := createCmd.RegisterFlagCompletionFunc("distro", completionDistroNames); err != nil {
+		logrus.Panicf("failed to register flag completion function: %v", err)
```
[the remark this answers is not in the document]
HarryMichal: `logrus.Panicf` is a nicely formatted `panic`. But it also introduces unnecessary fluff in the call trace. Replacing it with pure `panic` combined with `fmt.Sprintf` is probably a better solution simply for readability purposes.

src/cmd/completion.go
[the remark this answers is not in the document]
HarryMichal: I believe the comment part is a fall out. I wouldn't worry about supporting PowerShell unless somebody from the community requests it. It's use on Linux is possible but I've yet to see anybody who actively uses it.

> ... which makes me think that we shouldn't blindly support it without signing up to testing it.

Agreed.

I'll make a follow up PR that removes the references to PowerShell.

completion/meson.build
HarryMichal: Good catch. My mistake.

meson_options.txt
HarryMichal: My aim with providing fallbacks was to make sure we're always installing in the "correct" locations. But looking at what Podman [does](https://github.com/containers/podman/blob/main/Makefile#L76-L78), we could probably just drop the fallbacks and hard-code the paths reducing the number of layers.

I'd rather go with the approach of having an explicit switch for completions instead of having to require said shells.

src/cmd/completion.go
HarryMichal: Nice catch.

src/cmd/completion.go
HarryMichal: I agree with `Long` being redundant and should be removed.

While I like us relying mostly on manuals, I also think the use of items like `Use`, `Short` in the Cobra `Command` instances is also the right thing to do. We may not use them now but that might change if we decide to tackle [Project pickle](https://github.com/containers/toolbox/projects/4). Also manuals don't tend to be localized while code strings yes (and it's much easier to do).

HarryMichal:
Opened https://github.com/containers/toolbox/pull/1055 to address the raised concerns.

completion/meson.build
```diff
@@ -0,0 +1,36 @@
+generate_completions_program = find_program('generate_completions.py')
+
+if bash_completion.found()
+  bash_comp_dir = bash_completion.get_pkgconfig_variable('completionsdir')
```
debarshiray, line 4: This introduces a Meson warning.

completion/meson.build
```diff
@@ -0,0 +1,36 @@
...
+else
+  bash_comp_dir = get_option('datadir') / 'bash-completion' / 'completions'
+  message('bash-completion not found: using', get_option('prefix') / bash_comp_dir, 'as a falback install directory')
+endif
+
+if fish.found()
+  fish_comp_dir = fish.get_pkgconfig_variable('completionsdir')
```
debarshiray, line 11: Ditto.

completion/meson.build
debarshiray: Fixed in https://github.com/containers/toolbox/pull/1096

completion/meson.build
debarshiray: Fixed in https://github.com/containers/toolbox/pull/1096

src/cmd/completion.go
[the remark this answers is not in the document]
debarshiray: I see that you covered this in https://github.com/containers/toolbox/pull/1055 (and https://github.com/containers/toolbox/pull/1113)

src/cmd/completion.go
debarshiray: Resolved in https://github.com/containers/toolbox/pull/1055 (and https://github.com/containers/toolbox/pull/1113)

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+func completionContainerNames(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	containerNames := []string{}
+	if containers, err := getContainers(); err == nil {
+		for _, container := range containers {
+			containerNames = append(containerNames, container.Names[0])
+		}
+	}
+
+	if len(containerNames) == 0 {
+		return nil, cobra.ShellCompDirectiveNoFileComp
```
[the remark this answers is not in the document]
debarshiray: Addressed in https://github.com/containers/toolbox/pull/1114

src/cmd/completion.go
debarshiray: > I also think the use of items like Use, Short in the Cobra Command
> instances is also the right thing to do.

Like I said, they are fine for developer documentation and a few other specific things.  However, not when it comes to any non-trivial amount of text.  Those are harder to keep synchronized across multiple places.

> We may not use them now but that might change if we decide to
> tackle [Project pickle](https://github.com/containers/toolbox/projects/4).

We will cross that bridge when we come to it.  :)

> Also manuals don't tend to be localized while code strings yes

```
$ ls /usr/share/man
ca  es  id  ...  pt  ...
cs  fr  it  ...  pt_BR  ...
```

> Also, the text in Short is used in Cobra-generated completions scripts.

Yes, I agree.  That's the string in `Short` matches the title of the corresponding manual.  :)

debarshiray: Resolved, since `Long` was dropped in https://github.com/containers/toolbox/pull/1055 (and https://github.com/containers/toolbox/pull/1113).

meson_options.txt
debarshiray: We can't be correct if we are guessing.  If a shell's completion machinery offers an API for the location, then we should be using only that.  eg., how can we know that the right location for fish is `/usr/share/fish/completions` or `/usr/share/fish/vendor_completions.d`?  Any random downstream distributor or a new release of the shell can switch the location and we would be in trouble.

In other words, I don't think Podman is doing the right thing.  :)

> I'd rather go with the approach of having an explicit switch for completions
> instead of having to require said shells.

There are three groups of people building Toolbx:
* Downstream distributors shipping `toolbox` binaries
* Continuous integration
* Developers hacking on the sources

For the first, it's acceptable to require the shells as a build dependency so that we can use the appropriate API to get the location for completions.  However, it doesn't mean that the shells need to be run-time dependencies of `toolbox`.  eg., I have fish and zsh completions installed for `flatpak` and `podman` on my Fedora 36 even though I don't have the `fish` and `zsh` RPMs.

The continuous integration case seems to be the same as the first.

Developers hacking on Toolbx, who aren't specifically working on the completions, don't need to worry about the completions at all.  They likely don't even want to, because their running shells won't pick up the newly installed completions from a random prefix.  Even more so if their host OS is Fedora Silverblue because they would be hacking inside a Toolbx container.  So, it's fine if the build skips over the completions because the necessary build dependencies were missing.

If a developer is indeed hacking on the completions, then it's reasonable to expect them to have the necessary shells installed.  Otherwise how would they test their changes?  :)

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(completionCmd)
+}
+
+func completionEmpty(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return nil, cobra.ShellCompDirectiveNoFileComp
```
debarshiray, line 74: Is this really needed?  See below.

src/cmd/create.go
```diff
@@ -58,9 +58,10 @@ var (
 )
 
 var createCmd = &cobra.Command{
-	Use:   "create",
-	Short: "Create a new toolbox container",
-	RunE:  create,
+	Use:               "create",
+	Short:             "Create a new toolbox container",
+	RunE:              create,
+	ValidArgsFunction: completionEmpty,
```
debarshiray, line 64: Isn't it enough to have `ValidArgs: nil` ?  Is it really necessary to have the dynamic variant?

src/cmd/completion.go
```diff
@@ -0,0 +1,209 @@
...
+
+  # To load completions for each session, execute once:
+  $ toolbox completion fish > ~/.config/fish/completions/toolbox.fish
+
+`,
+	Hidden:                true,
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "zsh", "fish"},
+	Args:                  cobra.ExactValidArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
```
debarshiray, line 51: We should use `RunE` as elsewhere and return an `error`.

meson_options.txt
debarshiray: I did some more digging.

> In other words, I don't think Podman is doing the right thing. :)

Another incorrect example is Flatpak.  :)

A good example to emulate would be systemd:
* https://github.com/systemd/systemd/blob/main/shell-completion/bash/meson.build
* https://github.com/systemd/systemd/blob/main/shell-completion/zsh/meson.build

We should default to using the shell's API to detect the location for installing the completions, but have a build flag to override that.  Just like we do for `tmpfiles_dir`.

This is particularly relevant for Z shell, because it doesn't have an API and the Internet says that Debian and Fedora use [different locations](https://unix.stackexchange.com/questions/607809/where-should-i-install-the-zsh-completion-file-as-a-package-maintainer).  A build flag for the location can also [optimize the build](https://src.fedoraproject.org/rpms/toolbox/c/9bebde5bb60f36e38324c1e230e457c495d2d864) when there's an alternate API for the location.

debarshiray: I have tried this in https://github.com/containers/toolbox/pull/1123

src/cmd/create.go
HarryMichal: Yes, it is. If a custom completion function is not set, the default completion behaviour is then used -> listing of files/dirs in the current directory. `completionEmpty` ensures there is no completion shown. In the case of `create` that is the correct behaviour since it does not accept files/dirs as an argument but only an arbitrary string that the user has to choose.

src/cmd/completion.go
debarshiray: Fixed in https://github.com/containers/toolbox/pull/1055

src/cmd/create.go
debarshiray: I see.  Thanks for explaining that!

PR #842, opened 2021-07-16
test/system: Track bats libs as submodules & install them better

HarryMichal:
This will make it easier to work with system tests.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
   'c',
   version: '0.0.99.2',
   license: 'ASL 2.0',
-  meson_version: '>= 0.40.0',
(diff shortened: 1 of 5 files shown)


meson.build
```diff
@@ -38,7 +38,25 @@ endif
 
 install_subdir(
   'test',
-  install_dir: join_paths(get_option('datadir'), meson.project_name())
+  install_dir: join_paths(get_option('datadir'), meson.project_name()),
+  exclude_files: [
+    'system/libs/bats-assert/.git',
```
debarshiray, line 43: Is it necessary to list `.git` here when it's listed in `exclude_directories` below?

PR #845, opened 2021-07-20
utils: bump the fallback Fedora version to 34

juhp:
[skip ci]


HarryMichal (approved):
I agree with the change.

HarryMichal:
Thank you for the PR, @juhp!

PR #850, opened 2021-07-22
test/system: Use BATS_RUN_TMPDIR for image cache

HarryMichal:
Using the current working directory for cache is not a good solution
since the test files may reside in a location that is unwritable (e.g.,
/usr/share). The `BATS_RUN_TMPDIR` variable should point to a location
that is sure to be writeable from the test suite.

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -8,8 +8,7 @@ readonly TOOLBOX=${TOOLBOX:-toolbox}
 readonly SKOPEO=$(command -v skopeo)
 
 # Helpful globals
-readonly PROJECT_DIR=${PWD}
-readonly IMAGE_CACHE_DIR="${PROJECT_DIR}/image-cache"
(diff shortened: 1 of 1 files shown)


olivergs:
I've been reading some docs about the BATS_TMPDIR. Maybe could be useful to use BATS_RUN_TMPDIR instead BATS_TMPDIR.

The reference for this is: https://github.com/bats-core/bats-core/issues/226

HarryMichal:
recheck

HarryMichal:
> I've been reading some docs about the BATS_TMPDIR. Maybe could be useful to use BATS_RUN_TMPDIR instead BATS_TMPDIR.
> 
> The reference for this is: [bats-core/bats-core#226](https://github.com/bats-core/bats-core/issues/226)

Adjusted the PR to reflect this :).

olivergs (approved):
Looks good to me.

After checkout and running test all worked perfectly.

Thanks for the changes!

PR #857, opened 2021-08-03
tests: Fixed tests to setup the XDG_RUNTIME_DIR variable when empty

olivergs:
Toolbox can't run properly if the XDG_RUNTIME_DIR is not set properly.

Now the tests can run in environments without it by setting it to /run/user/$UID if empty.

### DIFF
--- a/test/system/002-help.bats
+++ b/test/system/002-help.bats
@@ -4,6 +4,10 @@ load 'libs/bats-support/load'
 load 'libs/bats-assert/load'
 load 'libs/helpers.bash'
 
+setup() {
+  check_xdg_runtime_dir
+}
+
 @test "help: Run command 'help'" {
   run $TOOLBOX help
 
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -5,6 +5,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  check_xdg_runtime_dir
   cleanup_containers
 }
 
--- a/test/system/102-list.bats
+++ b/test/system/102-list.bats
@@ -5,6 +5,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  check_xdg_runtime_dir
   cleanup_all
 }
 
--- a/test/system/103-container.bats
+++ b/test/system/103-container.bats
@@ -5,6 +5,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  check_xdg_runtime_dir
   cleanup_containers
 }
 
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -5,6 +5,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  check_xdg_runtime_dir
   cleanup_containers
 }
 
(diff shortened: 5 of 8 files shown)


HarryMichal:
I understand the motivation behind this but is it really the job of the test suite to setup `XDG_RUNTIME_DIR`? Maybe it is but the root cause of the issue in https://github.com/containers/podman/issues/10296#issuecomment-891003308 is in Toolbox incorrectly handling `XDG_RUNTIME_DIR` being unset.

I guess adding the check to the test suite is fine but I'd just leave it in the setup job (the first one in the test suite) and make it fail in case the test fails.

@olivergs what do you think?

olivergs:
As Toolbox relies in XDG_RUNTIME_DIR to be correctly set to work, we need to ensure that directory is correctly set. The problem we had with the fedora gating is that the tests were run with a user that had not a runtime dir set (the machine runs as root and then changed the user using "su - fedora", so the variable is not set to the fedora user's runtime dir).

So, only in the case of no XDG_RUNTIME_DIR variable is set, we set it.

About the root of the problem, yes, toolbox is handling wrongly XDG_RUNTIME_DIR because It needs that directory to be set always, as we rely in the mount of that dir inside the container to detect the container is already running during the initialization command, and now, if the directory does not exist, just tries to do an empty mount.

I was working in a patch to fox that to bailout with a warning if not XDG_RUNTIME_DIR is set, but I've not finished it yet.

About leaving the change just in the setup job will not work, as we need to export the variable each time we need it, and it needs to be set in every test. If you only set it in setup, when you run the next test, the variable is gone.

HarryMichal:
> About leaving the change just in the setup job will not work, as we need to export the variable each time we need it, and it needs to be set in every test. If you only set it in setup, when you run the next test, the variable is gone.

I was thinking more along the lines of: the variable is not set, bail out and let the user know to fix their environment. I wanted to evade having to set the dir in the tests. Thinking more about it, it probably is not much of a problem.

> I was working in a patch to fox that to bailout with a warning if not XDG_RUNTIME_DIR is set, but I've not finished it yet.

Do you have that work pushed somewhere?

olivergs:
Not until now. I pushed it here: https://github.com/olivergs/toolbox/tree/toolbox-xdg-runtime-dir-bailout, but it is not working because I found that the initialization command is not playing fairly with the changes I did. I've to spend a bit more time into fix that.

HarryMichal:
Just noticed the function is not called in all test files. `000-setup.bats`, `001-version.bats` and `999-teardown.bats` do not contain it. Is that intentional or not?

olivergs:
It is intentional. We do not run toolbox neither in setup nor teardown, and the version test do not require the XDG_RUNTIME_DIR to work

HarryMichal:
And Podman does not require the variable to be set?

olivergs:
Don't know if pofman requires the variable to be set for certain operations, but at least for our use case it seems not be needed. The tests run correctly if the variable is not set.

HarryMichal:
Ok. Could you add the link to this PR to the commit message and adjust the title of the commit to be in imperative? From `tests: Fixed ...` to `tests: Fix ...`. After that we can merge.

olivergs:
Done. Rebased and wairing for the CI to pass.

olivergs:
recheck

PR #858, opened 2021-08-05
tests: Changed container image source for busybox

olivergs:
Due to docker rate limiting we can not rely in docker.io for retrieving the images.

This was detected when executing our tests for podman fedora gating pipeline. Our busybox image was not downloaded and
one of the list tests was failing.

### DIFF
--- a/images/test/busybox/Containerfile
+++ b/images/test/busybox/Containerfile
@@ -1,0 +1,2 @@
+FROM busybox
+
--- a/images/test/busybox/README.md
+++ b/images/test/busybox/README.md
@@ -1,0 +1,4 @@
+This image is just a clone of the base busybox image used for testing purposes.
+
+The Containerfile is used to build the quay.io/toolbox_tests/busybox image that
+is used in the toolbox test suite.
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -11,7 +11,7 @@ readonly SKOPEO=$(command -v skopeo)
 readonly IMAGE_CACHE_DIR="${BATS_RUN_TMPDIR}/image-cache"
 
 # Images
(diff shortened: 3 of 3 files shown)


olivergs:
recheck

images/test/busybox/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM busybox
```
HarryMichal, line 1: Is this file really needed? It is not particularly hard to get a busybox image and there is nothing special about the image (any other image that is not a toolbox image would be sufficient).

HarryMichal (changes requested):
The change looks alright. The reasoning behind switching from Docker to Quay is also fine. Apart from a single comment this looks good to go.

images/test/busybox/Dockerfile
olivergs: I added this file to build the hosted image in the quay.io repository from it. When I created the quay.io image I just uploaded this Dockerfile. We can remove it from our code but I considered to add it to "subliminally" add the information about what image are we using at quay.io and where it comes from.

I know any other image could do the work, but as busybox is the smaller we can find out there, I think busybox is ok for what we need.

HarryMichal: > I know any other image could do the work, but as busybox is the smaller we can find out there, I think busybox is ok for what we need.

Yes, that is undoubtedly true.

> I added this file to build the hosted image in the quay.io repository from it.

Fair enough. Maybe just add a `README.md` explaining the purpose of the Dockerfile.

One more thing. Maybe rename the file to `Containerfile`. The name is a bit more "generic".

HarryMichal (approved):
Looks alright. The test failure in Rawhide is unrelated.

PR #860, opened 2021-08-10
test/system: Use quay.io/libpod/busybox instead of rolling our own

debarshiray:
Commit 09fb2377272b196f switched away from docker.io/library/busybox
because of rate limiting on the docker.io registry. Fortunately,
there's already a quay.io/libpod/busybox image that's used by Podman's
test suite, and all that the Toolbox test suite needs is a non-Toolbox
image. Hence there's no need to maintain a separate busybox image.

### DIFF
--- a/images/test/busybox/Containerfile
+++ b/images/test/busybox/Containerfile
@@ -1,2 +1,0 @@
-FROM busybox
-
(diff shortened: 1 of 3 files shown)


olivergs:
Before committing 09fb2377272b196f6422f85f54ea2c60be474c22 @edsantiago told me to use quay.io/libpod/busybox image for testing, but don't rely on it. After making the tests work I created our own hosted image to avoid problems in the future if the image is removed, renamed or changed.

Issue #568, opened 2020-09-27
tty device used inside toolbox gets odd permissions on host causing pinentry/gpg-agent to not work


dustymabe:
**Describe the bug**

My setup includes running gpg-agent on my host and doing development (git commit, etc) inside my toolbox container. When I git-commit I sign the commits, which means the process talks to the agent that then runs a pinentry program to ask the user for the passphrase.

When running in the terminal I prefer for the passphrase to be asked to me in the terminal window and not in a popup. So I'll `unset DISPLAY` to force it to use the terminal window. This stopped working sometime in the past few months. I finally tracked down some more information on why.

When running inside toolbox the permissions on my tty device are as expected:

```
$ ls -l $(tty)
crw--w----. 1 root nobody 136, 14 Sep 27 13:37 /dev/pts/14
```

However outside of toolbox (from the perspective of the host) the
permissions look odd:

```
[dustymabe@media ~]$ id
uid=1001(dustymabe) gid=1001(dustymabe) groups=1001(dustymabe) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
[dustymabe@media ~]$ ls -l /dev/pts/14
crw--w----. 1 100000 tty 136, 14 Sep 27 14:39 /dev/pts/14
```

I would expect the owner of the tty to be the user who launched the toolbox container (`1001`). If I change the ownership to the expected UID then my pinentry program appropriately asks for the passphrase on the same terminal window (inside toolbox) where I am running `git commit`. 

I'm running Fedora 33 with:

```
$ rpm -q toolbox podman
toolbox-0.0.95-1.fc33.x86_64
podman-2.1.0-0.187.rc2.fc33.x86_64
```

Indeed if I go back to an old machine I haven't updated in a while I can see that the permissions are what I would expect them to be. This out of date machine has:

```
[dustymabe@hattop ~]$ rpm -q toolbox podman 
toolbox-0.0.18-1.fc31.noarch
podman-1.8.0-2.fc31.x86_64
```

**Steps how to reproduce the behaviour**

1. Launch toolbox container
2. Inspect owner inside the container: `ls -l $(tty)`. Should show up as `root` since you are inside a user namespace.
3. Inspect the owner of that same device outside the container and verify the permissions looks weird.

debarshiray:
Thanks for narrowing it down to those *working* and *broken* versions.

[...]

PR #861, opened 2021-08-11
images: Add Arch Linux image

erazemk:
*This commit intends to add Arch Linux support to toolbox. It was built based on work from my three previous PRs (#328, #520 and #710), hopefully this one will be the last.*

Compared to previous PRs I've added the `vte-common` package, as recommended by @xPMo, which should fix the problems mentioned in #710 and I've worked in a separate branch instead of in `main` to avoid merging issues.

The image is built from the upstream `archlinux:base-devel` tag.

It includes most of the packages that the Fedora-based images do, but currently excludes `flatpak` (required for `flatpak-spawn`), since it adds ~690 MB of extra packages/dependencies.

If the final size is very important I can try and base it off of `archlinux:latest` (which includes only the base packages, not also development-related ones), but that would still require adding some of the missing packages back.

Another option could be to separate `latest` and `base-devel` versions into separate images, similarly to what is done with Fedora versions.

If this PR is merged into toolbox I also plan to add documentation on how to create an Arch Linux-based container and which distributions are available, since this would be the first image that is not Fedora-based (for now).


Foxboron:
Needs `flatpak-xdg-utils` for `toolbox` support inside the container. I'm not super clear why packages like `mlocate` is installed though?

I'm a package maintainer in Arch and maintain most of the container stack there. If there is anything missing please do poke me and I'll try solve it.

erazemk:
> I'm not super clear why packages like `mlocate` is installed though?

I added `mlocate` and others to have a similar set of packages to official Fedora versions. If you think some of them aren't needed, please tell me which ones and I'll gladly remove them to make the image smaller.

Foxboron:
I'm not sure what upstream expects frankly. But if they want a similar base set of packages thats fine. Arch doesn't lend itself to small disk footprints though :)

>It includes most of the packages that the Fedora-based images do, but currently excludes flatpak (required for flatpak-spawn), since it adds ~690 MB of extra packages/dependencies.

Just to clarify, `flatpak` never provided `flatpak-spawn`. It's provided by a separate project which I packages as `flatpak-xdg-utils` recently.

erazemk:
> I'm not sure what upstream expects frankly. But if they want a similar base set of packages thats fine. Arch doesn't lend itself to small disk footprints though :)

Besides the required files listed in the readme, I didn't come across any specific minimal set of packages described anywhere, so I don't think they have an official list. The Fedora version of the package list looks like common programs developers would use, so if there are more useful alternatives that more Arch users use, we can add those instead.

> > It includes most of the packages that the Fedora-based images do, but currently excludes flatpak (required for flatpak-spawn), since it adds ~690 MB of extra packages/dependencies.
> 
> Just to clarify, `flatpak` never provided `flatpak-spawn`. It's provided by a separate project which I packaged as `flatpak-xdg-utils` recently.

That's great, not needing flatpak installed inside the container reduces the image size significantly.

anthr76:
Are man pages working in here? I have a similar image and doesn't seem like so..

Foxboron:
We explicitly remove all the man pages and a lot of fluff such as translations, locales to keep the image size down.

https://github.com/archlinux/archlinux-docker/blob/master/pacman-conf.d-noextract.conf#L8

images/archlinux/Containerfile
```diff
@@ -0,0 +1,24 @@
...
+# Install extra packages
+COPY extra-packages /
+RUN pacman -Syu --needed --noconfirm - < extra-packages
+RUN rm /extra-packages
+
+# Clean up cache
+RUN pacman -Scc --noconfirm
+
+# Enable sudo permission for wheel users
+RUN sed -i 's/# %wheel ALL=(ALL) NOPASSWD: ALL/%wheel ALL=(ALL) NOPASSWD: ALL/' /etc/sudoers
```
Foxboron, line 22:
```suggestion
RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/toolbox
```

Would be simpler.

erazemk:
> We explicitly remove all the man pages and a lot of fluff such as translations, locales to keep the image size down

While on that note, I will probably update the readme to be Arch-specific (although I'm not sure why every image even has to have its own readme if they all look the same as the main readme) and I wanted to list the steps required to re-add the manpages if so desired. Can you please describe the process or commands needed to re-add the manpages or point me to the proper wiki article.

images/archlinux/Containerfile
erazemk: Thanks for the suggestion :)

Didn't realize Github allowed committing suggestions, I'll use that next time.

anthr76:
I think just removing that like at the end of the build should be sufficent (https://github.com/archlinux/archlinux-docker/blob/master/pacman-conf.d-noextract.conf#L8). I just removed that line and did something like

`sudo pacman -S $(sudo pacman -Qq $(sudo pacman -Fq /usr/share/man/ | cut -d/ -f2) 2> /dev/null)`

and alas had manpages. Like @Foxboron said it's probaly a good idea not to ship the image with them though and instead let them get pulled in as the user updates and installs packages within the toolbox

erazemk:
> Like @Foxboron said it's probaly a good idea not to ship the image with them though and instead let them get pulled in as the user updates and installs packages within the toolbox

Yup, this is what I meant, not shipping manpages with the image or having to create your own image with manpages, but rather allowing users to re-add the manpages inside the container if needed. Is reinstalling / updating the packages the only way to re-add them?

anthr76:
A little side note, Arch's glibc is out of date so being blocked by:

https://archlinux.org/packages/core/x86_64/glibc/ || https://github.com/containers/toolbox/issues/821

ybsci:
I'm not very knowledgeable but I think you can find patched glibc here. i am using this image
https://hub.docker.com/layers/bugthing/archlinux-for-toolbox/latest/images/sha256-59682ce4a6ad22bbece1528284881fb790a24bead7fe7a46b8e9672a9e5a22f5?context=explore

`/bin/sh -c patched_glibc=glibc-linux4-2.33-4-x86_64.pkg.tar.zst &&     curl -LO "https://repo.archlinuxcn.org/x86_64/$patched_glibc" &&     bsdtar -C / -xvf "$patched_glibc"`

juhp:
Just for reference/FYI: https://github.com/toolbox-images/toolbox-images/tree/main/images/arch

erazemk:
> Just for reference/FYI: https://github.com/toolbox-images/toolbox-images/tree/main/images/arch

Didn't realize how easy it was to set that up, but in any case, the PR adds extra toolbox-specific packages so I think it's still useful.

I have no idea why it fails for Rawhide though, any info on that is very welcome.

Foxboron:
That container is a "hack" and doesn't conform to the toolbox specification defined in this repo.

anthr76:
+1 on @Foxboron thought and including some critical packages to have in the toolbox. It's a good workaround but much preferred to have a tailored image.

juhp:
BTW mlocate does work in the fedora toolboxes now.

erazemk:
@HarryMichal are you interested in including the Arch image in toolbox? The PR is complete, there are no more changes planned for now.

juligreen:
Could this be merged then? :)

anthr76:
For the folks that been testing a similar Arch toolbox has xdg-open been working correctly?

TheEvilSkeleton:
No.

anthr76:
Right, flatpak-spawn proplely works but not `xdg-open` we might need to communicate some package tweaks to Arch packagers to get that working.

juhp:
Well, xdg-open doesn't work for fedora-toolbox either, unless you mean something different.
I use `flatpak-spawn --host xdg-open ...` as a workaround.

anthr76:
https://github.com/containers/toolbox/issues/568 I'm now facing this. Anyone else? Is there a good community on matrix to talk about and squash these things?

TheEvilSkeleton:
Maybe https://matrix.to/#/#toolbx:matrix.org is what you're looking for.

TheEvilSkeleton:
> Well, xdg-open doesn't work for fedora-toolbox either, unless you mean something different. I use `flatpak-spawn --host xdg-open ...` as a workaround.

Thanks for sharing this. We can create a script file that automatically aliases `xdg-open` to `flatpak-spawn --host xdg-open` and writes it to `/etc/profile.d/*.sh` to make it as painless as possible.

I implemented a little script that checks all the `xdg-*` commands and aliases accordingly: https://gitlab.com/TheEvilSkeleton/archlinux-toolbox/-/blob/5b774c780c2abe442b3465ac6b3774c649908679/xdg-commands.sh, not sure if it's relevant or not so I figured I should just share it.

Foxboron:
I've continued the work in this branch over here.

https://github.com/toolbx-images/images/pull/2

Trying my best to retain maintainer credit so please do tell me if there are any issues :)

travier:

Foxboron:
@debarshiray Do you intend to merge this branch? There are some changes that needs to be done for `toolbx` to understand the Arch `os-release` file as we need to translate `BUILD_ID=rolling` to the `latest` tag.

I have some local patches I need to rebase for this that allows for it to be overriden.

debarshiray:
> @debarshiray Do you intend to merge this branch?

Yes, it's high time.  :)

I finally have some time to focus on pushing Toolbx forward again, as opposed to barely keeping the lights on, and it looks like we are on our way to ironing out the final wrinkles around built-in support for [Ubuntu](https://github.com/containers/toolbox/pull/483).  So, it's a good time to try to land the Arch Linux support.

> There are some changes that needs to be done for `toolbx` to
> understand the Arch `os-release` file as we need to translate
> `BUILD_ID=rolling` to the `latest` tag.

Yes, exactly.  I was going to write a comment about that when I noticed yours.

I am in the process of installing Arch Linux in GNOME Boxes to see how Toolbx behaves on an Arch Linux host (eg., whether an Arch Linux host will default to the Arch Linux image) when I noticed the `BUILD_ID=rolling` in `os-release(5)`.

*I finally managed to get `pacman -S gnome` to complete after figuring out that I needed to turn on `systemd-networkd` and `systemd-resolved` in order to have IP addresses and resolve domain names.  :)*

> I have some local patches I need to rebase for this that allows for
> it to be overriden.

Cool.  How do you want to handle them?  I saw that you have been maintaining the Arch Linux images at [github.com/toolbox-images/images](https://github.com/toolbx-images/images/pull/4), which is fantastic.  Do you want to synchronize the latest version of the image before merging it here, like we did for Ubuntu?

Foxboron:
> Cool. How do you want to handle them?

I'm not really sure if I'll have time for that the next couple of days. I can take a quick look tonight.

>I saw that you have been maintaining the Arch Linux images at https://github.com/toolbx-images/images/pull/4, which is fantastic. Do you want to synchronize the latest version of the image before merging it here, like we did for Ubuntu?

You can probably merge this. I don't mind being listed as `CODEOWNER` for this if you want. I can come with followup changes if there is a need for it.

erazemk:
> You can probably merge this. I don't mind being listed as CODEOWNER for this if you want. I can come with followup changes if there is a need for it.

I think this might be the safer option, I haven't used Arch Linux in a while and even when making this PR I wasn't totally sure which base packages are necessary. So I don't mind not being the code owner.

debarshiray (changes requested):
Thanks for persisting with this, @erazemk !  You have made some great progress here.

As @Foxboron [mentioned](https://github.com/containers/toolbox/pull/861#issuecomment-1551170977), we need to handle the lack of version numbers in Arch Linux due to it's rolling release model.

Secondly, we should synchronize this with the image on [github.com/toolbx-images/images](https://github.com/toolbx-images/images/tree/main/archlinux), including the GitHub Action to build and publish it on quay.io.

Finally, the `.github/CODEOWNERS` file needs to be updated.

images/archlinux/Containerfile
```diff
@@ -0,0 +1,24 @@
+FROM docker.io/library/archlinux:base-devel
+
+ENV NAME=archlinux-toolbox VERSION=base-devel
+LABEL com.github.containers.toolbox="true" \
+      name="$NAME" \
+      version="$VERSION" \
+      usage="This image is meant to be used with the toolbox command" \
+      summary="Base image for creating Arch Linux toolbox containers" \
+      maintainer="Erazem Kokot <mail@erazem.eu>"
```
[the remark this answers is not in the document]
erazemk: Oh sorry, I meant I don't mind not being the maintainer, not codeowner, Foxboron can be the single maintainer, since he has so much more knowledge.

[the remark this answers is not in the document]
debarshiray: Answered in https://github.com/containers/toolbox/pull/861#issuecomment-1553501696

debarshiray:
> > Cool. How do you want to handle them?
> 
> I'm not really sure if I'll have time for that the next couple of days.
> I can take a quick look tonight.

Okay!

> > I saw that you have been maintaining the Arch Linux images at [toolbx-images/images#4](https://github.com/toolbx-images/images/pull/4), which is fantastic. Do you want to synchronize the latest version of the image before merging it here, like we did for Ubuntu?
> 
> You can probably merge this. I don't mind being listed as `CODEOWNER` for this if you want. I can come with followup changes if there is a need for it.

It's getting a bit late for today.  Once I have finished installing Arch Linux, I will play with the image a bit on different hosts, and then merging it.

The one big thing that I might not be able to do myself is the GitHub Action for building and publishing the image.  Could you and/or @travier take care of that?

Foxboron:
> The one big thing that I might not be able to do myself is the GitHub Action for building and publishing the image. Could you and/or @travier take care of that?

I can probably take a look at that basing it off on the work we did in the other repository.

.github/CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 *    @HarryMichal @debarshiray
 /data/gfx/*.gif    @jimmac
+/images/archlinux  @Foxboron
```
debarshiray, line 3: I see that `ID=arch` in `os-release(5)` on Arch Linux.  To avoid confusion, I think we should also name the directory `arch`.

debarshiray:
> > The one big thing that I might not be able to do myself is the GitHub Action for building and publishing the image. Could you and/or @travier take care of that?
> 
> I can probably take a look at that basing it off on the work we did in the other repository.

Okay!

Foxboron:
Here is a diff with the changes I had locally.

But looking at it now, if we moved `ParseRelease` functions to resolve in `init()` it would be simpler.

```diff
diff --git a/src/pkg/utils/utils.go b/src/pkg/utils/utils.go
index 7a6b29a..1c8316f 100644
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -47,6 +47,7 @@ type Distro struct {
 	ImageBasename          string
 	GetFullyQualifiedImage GetFullyQualifiedImageFunc
 	ParseRelease           ParseReleaseFunc
+	OverrideRelease        string
 }

 const (
@@ -102,18 +103,28 @@ var (
 			"fedora-toolbox",
 			getFullyQualifiedImageFedora,
 			parseReleaseFedora,
+			"",
 		},
 		"rhel": {
 			"rhel-toolbox",
 			"toolbox",
 			getFullyQualifiedImageRHEL,
 			parseReleaseRHEL,
+			"",
 		},
 		"ubuntu": {
 			"ubuntu-toolbox",
 			"ubuntu-toolbox",
 			getFullyQualifiedImageUbuntu,
 			parseReleaseUbuntu,
+			"",
+		},
+		"arch": {
+			"archlinux-toolbox",
+			"archlinux-toolbox",
+			getFullyQualifiedImageArchLinux,
+			parseReleaseArchLinux,
+			"latest",
 		},
 	}
 )
@@ -144,6 +155,9 @@ func init() {
 			if err == nil {
 				containerNamePrefixDefault = distroObj.ContainerNamePrefix
 				distroDefault = hostID
+				if distroObj.OverrideRelease != "" {
+					release = distroObj.OverrideRelease
+				}
 				releaseDefault = release
 			}
 		}
@@ -355,6 +369,11 @@ func getFullyQualifiedImageUbuntu(image, release string) string {
 	return imageFull
 }

+func getFullyQualifiedImageArchLinux(image, release string) string {
+	imageFull := "quay.io/toolbx-images/" + image
+	return imageFull
+}
+
 // GetGroupForSudo returns the name of the sudoers group.
 //
 // Some distros call it 'sudo' (eg. Ubuntu) and some call it 'wheel' (eg. Fedora).
@@ -727,6 +746,10 @@ func parseReleaseUbuntu(release string) (string, error) {
 	return release, nil
 }

+func parseReleaseArchLinux(release string) (string, error) {
+	return "latest", nil
+}
+
 // PathExists wraps around os.Stat providing a nice interface for checking an existence of a path.
 func PathExists(path string) bool {
 	if _, err := os.Stat(path); !os.IsNotExist(err) {
```

debarshiray:
Do you think we should make hosts running Arch derivates like Manjaro use the Arch images?

I don't know how similar Arch and Manjaro are?  Are they like Fedora/RHEL and Debian/Ubuntu, where the derived distribution might have significant changes?  Or are they like RHEL and its clones?  If they are sufficiently similar, then it might be worth making Manjaro hosts use the Arch image.

I did create a Manjaro virtual machine, but I know too little about the Arch universe to be able to figure it out with certainty.  I did notice differences like Arch using Wayland but not Manjaro, differences in GNOME versions (44 versus 43), different terminal emulation applications, etc..

Foxboron:
>I don't know how similar Arch and Manjaro are. Are they like Fedora/RHEL and Debian/Ubuntu, where the derived distribution might have significant changes? Or are they like RHEL and its clones? If they are sufficiently similar, then it might be worth making Manjaro hosts use the Arch image.

It would be similar to forcing Linux Mint to use Ubuntu, or Ubuntu to use Debian. At least in the Manjaro example.

debarshiray:
> > I don't know how similar Arch and Manjaro are. Are they like Fedora/RHEL and Debian/Ubuntu, where the derived distribution might have significant changes? Or are they like RHEL and its clones? If they are sufficiently similar, then it might be worth making Manjaro hosts use the Arch image.
> 
> It would be similar to forcing Linux Mint to use Ubuntu, or Ubuntu to use Debian. At least in the Manjaro example.

Ok, understood!.

images/archlinux/Containerfile
```diff
@@ -0,0 +1,24 @@
+FROM docker.io/library/archlinux:base-devel
+
+ENV NAME=archlinux-toolbox VERSION=base-devel
```
debarshiray, line 3: We should replace this `ENV` with `ARG`, like @travier [did](https://github.com/containers/toolbox/commit/9506173f88dc26bfb0229edea0d39699d636d9e2) for the Fedora images, to prevent `NAME` and `VERSION` from leaking into the interactive command line environment inside the container.

Secondly, I think the image should be called `arch-toolbox`, not `archlinux-toolbox`, to match the `ID` in `os-release(5)` on Arch Linux.  Toolbx uses `os-release(5)` in various different places and the concept of `ID` is embedded very deeply in the logic.  I think it will be simpler if we stuck to the `os-release(5)` values, instead of adding exceptions to the logic.

debarshiray (changes requested):

.github/CODEOWNERS
debarshiray: I took the liberty to change the directory name to `arch`.

images/archlinux/Containerfile
Foxboron: Fine with me :)

debarshiray: I saw that @travier [dropped](https://github.com/toolbx-images/images/commit/46eb11775cf608bfa764698efb888c37a780675d) the `ENV` from `github.com/toolbx-images/images`.  So, I took the liberty to do the same here.

I filed https://github.com/toolbx-images/images/pull/82 to rename the image on the registry to avoid user-visible breakage.

debarshiray:
I merged the definition of the `arch-toolbox` image.  Let's figure out how to wire it to `--distro` in a different pull request.  This one has gone on for too long already.  :)

> @debarshiray Fwiw, with the email thread and the issues I have lost a bit of track what needs to be done to merge this. Please be explicit if you need me to look into anything :)

The most important thing is to [rename](https://github.com/toolbx-images/images/pull/82) the image on the registry.  Otherwise, it won't work with `--distro`.

Other than that, issues that might need changes directly in Toolbx:

* Figure out how to customize the shell prompt inside a Toolbx container on Arch Linux.  This is due to the order in which the default Arch Linux host prompt is set by the Bash start-up scripts, specifically `~/.bashrc`.  I am open to ideas on how to address this one.

* Extending the *welcome to Toolbox banner* to include `pacman(8)`, just like DNF.  This is easy.

* Figure out how to make locales other than `C`, `C.UTF-8` and `POSIX` available inside the Toolbx container for Arch.  I might already have a solution for it.  See https://github.com/toolbx-images/images/issues/76

It will be good if `/etc/localtime` can be changed into a relative symbolic link on Arch Linux hosts.  However, that doesn't need changes in Toolbx and isn't causing any immediate problems today.  It's something for the Arch distribution.

debarshiray:
> Let's figure out how to wire it to `--distro` in a different pull request.

I have some work-in-progress code here: https://github.com/containers/toolbox/pull/1303

It needs some more testing, test cases, and we need to wait for the image to get renamed on the registry.

Foxboron:
> Figure out how to customize the shell prompt inside a Toolbx container on Arch Linux. This is due to the order in which the default Arch Linux host prompt is set by the Bash start-up scripts, specifically ~/.bashrc. I am open to ideas on how to address this one.

I had a quick look at this.

`.bashrc` doesn't actually play into it since toolbox is running with `bash -l`.

The issue is that the the following snippet is missing from `/etc/bash.bashrc`.

```bash
for i in /etc/profile.d/*.sh; do
    if [ -r "$i" ]; then
        if [ "$PS1" ]; then
            . "$i"
        else
            . "$i" >/dev/null
        fi
    fi
done
```

And quickly looking it seems this is done twice in fedore. One in `/etc/profile` and once in `/etc/bashrc`. Meanwhile Arch has a simpler `/etc/bash.bashrc` file which does not do this and unconditionally set `PS1`.

https://gitlab.archlinux.org/archlinux/packaging/packages/bash/-/blob/main/system.bashrc

The question is if I should poke the Arch `bash` maintainers to figure out why we overwrite `$PS1`?

travier:
I've set up the https://quay.io/repository/toolbx/arch-toolbox repo on Quay.io. You can mirror the GitHub Action from the Ubuntu PR to make one for this image and push to the repo.

debarshiray:
> I've set up the https://quay.io/repository/toolbx/arch-toolbox repo on Quay.io. You can mirror the GitHub Action from the Ubuntu PR to make one for this image and push to the repo.

Thanks, @travier !  Done: https://github.com/containers/toolbox/pull/1308

Foxboron:
For the prompt, I think there are two issues:

1) How to show the prompt on Arch hosts

2) How to show the prompt on non-Arch hosts.

I think getting the first one solved is less of a priority as we can include documentation in the ArchWiki, while the second point should probably be fixed.

debarshiray:
`> For the prompt, I think there are two issues
> 
>     1. How to show the prompt on Arch hosts
> 
>     2. How to show the prompt on non-Arch hosts.
> 
> 
> I think getting the first one solved is less of a priority as we can include documentation in the ArchWiki, while the second point should probably be fixed.

Okay.

I think the common (sub-)problem across those cases is that Arch Linux's `/etc/bash.bashrc` is sourced after `/etc/profile.d`.  Since `/etc/bash.bashrc` sets `PS1`, Toolbx's customization in `/etc/profile.d` gets overridden.

Solving this will retain the customized shell prompt inside Arch containers running on all Fedora hosts where the user didn't change the shell start-up files in `$HOME` to set `PS1`.  This is because Fedora's default `~/.bash_profile` and `~/.bashrc` don't set `PS1`.  Only `/etc/bashrc` sets `PS1` (to the default) on Fedora.

*I think we should really move the discussion about customizing the shell prompt on Arch Linux somewhere else other than this merged pull request.  :)*

PR #867, opened 2021-08-30
Added error info on non 0 and unknown exit code

olivergs:
When a command is executed with toolbox run and it returns a non zero exit code, if that exit code is not handled it is just ignored.

This prevents users to identify errors when executing commands in toolbox.

With this fix, the unhandled exit codes are returned as errors to give the user a better error info.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -363,7 +363,13 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 				}
 			}
 		default:
-			return nil
+			var newerr error
+			if err != nil {
+				newerr = fmt.Errorf("failed to invoke command %s in container %s (Exit code: %d): %s", command[0], container, exitCode, err)
+			} else {
+				newerr = fmt.Errorf("failed to invoke command %s in container %s (Exit code: %d)", command[0], container, exitCode)
+			}
+			return newerr
 		}
 	}
 


src/cmd/run.go
```diff
@@ -363,7 +363,13 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 				}
 			}
 		default:
-			return nil
+			var newerr error
+			if err != nil {
+				newerr = fmt.Errorf("failed to invoke command %s in container %s (Exit code: %d): %s", command[0], container, exitCode, err)
+			} else {
+				newerr = fmt.Errorf("failed to invoke command %s in container %s (Exit code: %d)", command[0], container, exitCode)
```
debarshiray, line 370: Wouldn't it be better if `enter` and/or `run` forwarded the exit code of the requested command? eg., `toolbox run false` would have an exit code of 1 and `toolbox run true` would have 0.

I suppose, most people use the exit code programmatically because they can already see the standard error and output streams as a human. In that case, not forwarding the exit code breaks the programmability of it.

But it's hard to say without knowing the exact use-case.

Part of the complication in forwarding the exit code is that Cobra, the Go framework we use for our command line handling, only lets us return an error. See `Execute` in `src/cmd/root.go`. So, I think we will need to provide our own custom `error` implementation that can encapsulate an exit code.

debarshiray (commented):
Thanks for converting the discussion with @fmuellner into a pull request, @olivergs !

HarryMichal:
I propose an alternative implementation of this which fully propagates the return codes of the invoked commands (including Podman): https://github.com/HarryMichal/toolbox/tree/cmd/run/return-code-propagation

HarryMichal:
Created https://github.com/containers/toolbox/pull/1013 with the alternative implementation.

HarryMichal:
Closing in favor of #1013

Issue #134, opened 2019-04-26
Ask for confirmation before doing downloads


allanday:
If I run `toolbox create`, toolbox immediately starts downloading a rather large image file. This is undesirable for two reasons:

 1. It is bad behaviour to do large downloads without asking first - it could eat someone's data cap, or interfere with other network activity
 2. It's surprising: to the uninitiated, there's nothing about `toolbox create` that suggests that a download will be involved

The simple way to fix this would be to ask for permission before doing the download, like:
```
A toolbox container image is required. Download now? (400MB) [y/N]
```

allanday:
Or you could do two lines:

```
Image required to create toolbox container.
Download fedora-bla-bla-bla-30 (400 MB) ? [y/N]
```

debarshiray:
Yes, I agree. One problem is that there's no easy way to get the size of the impending download. :(

We could hard code the size of the `fedora-toolbox` images until we figure out a better way to get the numbers. Those are the things that most people are going to download, and the figure doesn't need to be very accurate.

allanday:
> Those are the things that most people are going to download, and the figure doesn't need to be very accurate.

Yes, showing an approximate figure would be fine for now.

PR #875, opened 2021-09-13
create: always offer to pull

juhp:
This PR addresses 2 common user issues with `toolbox create`:

- users need to manually download newer images themselves
- ~~the hardcoded "500MB download" is misleading~~

I have tested this locally and it seems to work okay for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ Toolbx and [Linux distro support](https://containertoolbx.org/distros/).
 
 ##
 
+[![Star History Chart](https://api.star-history.com/svg?repos=containers/toolbox&type=Date)](https://star-history.com/#containers/toolbox&Date)
+
+
+##
+
 [![Zuul](https://zuul-ci.org/gated.svg)](https://softwarefactory-project.io/zuul/t/local/builds?project=containers/toolbox)
 [![Daily Pipeline](https://softwarefactory-project.io/zuul/api/tenant/local/badge?project=containers/toolbox&pipeline=periodic)](https://softwarefactory-project.io/zuul/t/local/builds?project=containers%2Ftoolbox&pipeline=periodic)
 


juhp:
Any feedback on this?
It is quite small :)

src/cmd/create.go
```diff
@@ -693,9 +689,11 @@ func pullImage(image, release string) (bool, error) {
 	}
 
 	if promptForDownload {
-		fmt.Println("Image required to create toolbox container.")
+		if _, err := podman.ImageExists(imageFull); err != nil {
+			fmt.Println("Image required to create toolbox container.")
+		}
 
-		prompt := fmt.Sprintf("Download %s (500MB)? [y/N]:", imageFull)
+		prompt := fmt.Sprintf("Download %s ? [y/N]:", imageFull)
```
debarshiray, line 696: It was a [deliberate](https://github.com/containers/toolbox/issues/134) [choice](https://github.com/containers/toolbox/commit/b718fbdcda74b0f98ec5035e9a41809132f2da05) to show the size of the image.

While the number doesn't exactly match the size of the image, I don't think it's misleading. The images have always been around the 500MB mark. The Fedora 33 image is 525MB and the F35 image is 497MB. In the past they were slightly smaller, but still above 450MB.

It doesn't matter that much if the image is exactly 500MB or 400MB or whatever. The main point is that they are big, and it's important for the user to know how big before they commit to that amount of network I/O.

Everybody loves to complain about the hard-coded number, but nobody ever wants to sit down and get the right size with Skopeo. :)

debarshiray (changes requested):

src/cmd/create.go
juhp: Okay that is fair. I guess it is not that important, but it does bother me each time I see it.

I think it is misleading in the sense that I thought the compressed download is maybe in the order of ~200MB??
I am not sure, am I way off here?  So how about writing `Image size` instead of `Download`?
(I just checked with skopeo manually and the current fedora-toolbox:35 `Size` is 339,350,428 bytes.)

Anyway this is just a cosmetic thing, so I am also okay to drop this commit and we could revisit it later.

juhp:
So how about this `Download <<image>> (image size ~400MB)` then.  Any better?

juhp:
I could suggest for the pull part of this to make it an option instead:

ie `toolbox create --pull ...`

That would make it opt-in which should alleviate the storage space concerns with always pulling.
(Personally I would prefer to pull by default since it is too easy to forget,
but at least having the command option makes it more visible and easier to update.
It might be nice to add a `purge` command eventually specifically for removing obsolete toolbox images.)

debarshiray:
> So how about this `Download <<image>> (image size ~400MB)` then.
> Any better?

Well, the best option would be to comment on https://github.com/containers/skopeo/issues/641 and push for it to get prioritized. It's not entirely trivial to get the exact size of the image.

src/cmd/create.go
debarshiray: Those numbers that I were quoting were from `podman images`.

As another data point, the new `registry.access.redhat.com/ubi8/toolbox:8.5` image for RHEL 8.5 is 558MB, and the equivalent for RHEL 9 Beta is 591MB.

It's likely that the actual amount of data downloaded over the wire is less than the size reported by `podman images`, because it might be more compressed over the wire than when stored.  The numbers reported by `podman pull` seem much less than the `podman images` numbers, but I am not sure how to interpret them. I never dug into the code to convince myself exactly what's happening here.

What are the semantics of the size reported by Skopeo? Is it the amount being sent over the wire?

It will be good if someone could dig into this a bit and figure out exactly what's going on here.

As long as we are stuck with an estimate, it's better if that estimate is on the higher side. eg., if we say 300MB and then the download ends up being 450MB, and it blows the data limit of something by that extra 150MB, then the user won't be happy. :)

Hence, 400MB or 500MB doesn't matter that much, because as we begin to support more and more distributions, the image sizes will vary by that much anyway. However, if we can convince ourselves that we are only downloading in the ball park of 100MB or 150MB, then that might be something, because then it would be five times smaller. :)

debarshiray: `Image size` versus `Download` seems confusing.

The only thing the user really needs to know is that we are going to download a lot (for some vague definition of *lot*) of data over the network. The approximate number is there to give some idea of what *lot* means. Everything else is only going to be confusing. eg., see all the confusion around Flatpak download sizes, progress bars, etc..

Of course, it would be good to improve our estimate, but that seems to need some digging. :)

Other than, we are also tight on screen real estate. We should try to fit most strings into a 80 character width terminal, and this string is already quite wide, and the length of the URL is outside our control:
```
Download registry.fedoraproject.org/fedora-toolbox:34 (500MB)? [y/N]:
```

It's not the most important thing, and in some cases we really don't have an option (eg., `toolbox list`), but still.

juhp:
Okay, sure, thank you for the detailed replies and clarity: let's drop the size part of this PR completely.

src/cmd/create.go
juhp: Okay, thanks, I agree with you - it makes sense

sgallagher:
Any hope of reviving this PR? It would vastly simplify the documentation on how to upgrade the Developer Toolbox I maintain, for instance.

juhp:
I made my main branch current which caused this PR to close - not that it was going anywhere...

Anyway you can still find my original proposal in https://github.com/containers/toolbox/commit/8bfbbf4c5e58e2f002f517df8deed8bf1ba8effd in (https://github.com/juhp/toolbox/tree/always-pull)
(I think that was it basically)

PR #878, opened 2021-09-19
images: add Ubuntu 20.04 image

francesconicoletta:
This should add an Ubuntu 20.04 LTS image, the packages are pretty much the same as Fedora, plus libcap2-bin, needed to fix "/bin/bash not found in container".


creatorofthesky:
I have noticed with all of these images that none of theme seem to work with apt / apt-get. The package manager is completely missing on every single image I have tried.

Jmennius:
> I have noticed with all of these images that none of theme seem to work with apt / apt-get. The package manager is completely missing on every single image I have tried.

Did you try images from #483?

andrewshadura:
See also #298, #371, #742.

debarshiray:
https://github.com/containers/toolbox/pull/483 supersedes this.

debarshiray:
Closing.

PR #881, opened 2021-09-21
Make welcome message appear on Fedora Kinoite

kuba3351:
I would like to make the toolbox.sh welcome message appear in Fedora Kinoite as Fedora Kinoite is an official variant of Silverblue and also using toolbox.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -20,9 +20,10 @@ eval $(
 if [ -f /run/ostree-booted ] \
    && ! [ -f "$host_welcome_stub" ] \
    && [ "${ID}" = "fedora" ] \
-   && { [ "${VARIANT_ID}" = "workstation" ] || [ "${VARIANT_ID}" = "silverblue" ]; }; then
+   && { [ "${VARIANT_ID}" = "workstation" ] || [ "${VARIANT_ID}" = "silverblue" ] || [ "${VARIANT_ID}" = "kinoite" ]; }; then
     echo ""
-    echo "Welcome to Fedora Silverblue. This terminal is running on the"
+    # shellcheck disable=SC3059
+    echo "Welcome to Fedora ${VARIANT_ID^}. This terminal is running on the"
     echo "host system. You may want to try out the Toolbox for a directly"
     echo "mutable environment that allows package installation with DNF."
     echo ""


HarryMichal:
Hi @kuba3351! Thanks for the PR. I'm willing to merge this PR, though I'd like this message to be handled by the distributions themselves :).

The case modification is most likely not a problem as we require bash or zsh (don't know why zsh) for running the welcome script.

Could you, please, wrap the body of the commit messages to 72 characters and include a link to this PR in them?

kuba3351:
Made commit message shorter and included link to MR in them. Thank you for review.

HarryMichal:
Thanks!

HarryMichal:
Still gonna ask you, could you merge the two commits together and in the commit title use `profile.d:` as a prefix and the title itself write in imperative form? In the commit message also mention why disabling the check is ok (I mentioned it above). The two commits touch the same area and one will cause errors without the other. Thanks!

kuba3351:
Hello,

I squashed these commits into one and rewrite commit message. I am sorry about that. I think it will be userful to describe how to write correct commit messages in CONTRIBUTING.md

HarryMichal:
Few more nits. The title of the commit has a typo: `Skow` -> `Show` and the body of the message has over 72 characters on a single line. No need to make the message shorter, just break it over few lines. Some editors track this and break the line for you. If the message is too long it gets hard to read it in some environment (e.g., git cli).

kuba3351:
Fixed typo and formatted the commit message. I have to apply your hints to another PR that I also opened :)

HarryMichal:
I just remembered there is also `VARIANT` in os-release. That one would work even better than this solution (if they key is present) but I don't mind sticking to this for the time being (definitely better than to write `Silverblue` on `Workstation`).

debarshiray:
> The case modification is most likely not a problem as we require
> bash or zsh (don't know why zsh) for running the welcome script.

Never say *most likely* about shell scripts. :)

profile.d/toolbox.sh
```diff
@@ -20,9 +20,10 @@ eval $(
 if [ -f /run/ostree-booted ] \
    && ! [ -f "$host_welcome_stub" ] \
    && [ "${ID}" = "fedora" ] \
-   && { [ "${VARIANT_ID}" = "workstation" ] || [ "${VARIANT_ID}" = "silverblue" ]; }; then
+   && { [ "${VARIANT_ID}" = "workstation" ] || [ "${VARIANT_ID}" = "silverblue" ] || [ "${VARIANT_ID}" = "kinoite" ]; }; then
     echo ""
-    echo "Welcome to Fedora Silverblue. This terminal is running on the"
+    # shellcheck disable=SC3059
+    echo "Welcome to Fedora ${VARIANT_ID^}. This terminal is running on the"
```
debarshiray, line 26: This sort of case modification with a `^` is a Bash-ism.  It doesn't work with Z shell.  See https://github.com/containers/toolbox/issues/1017 or try this:
```
% . /etc/os-release
% echo "${VARIANT_ID^}"
zsh: bad substitution
```

So, yes, ShellCheck was right.

debarshiray:
> I'm willing to merge this PR, though I'd like this message to be
> handled by the distributions themselves :).

There are some things in `profile.d/toolbox.sh` that can be argued to be details specific to the downstream distributor and some (like `PS1` and Terminfo handling) that people can expect the upstream Toolbx project to offer out-of-the-box.

Having everything in one place does have the advantage of avoiding bugs and feature requests from ping-ponging across projects.

If the conditionals based on the downstream keep growing a lot, we could consider splitting the file up.  Maybe have separate snippets based on the `ID` in `os-release(5)` and install the one that's relevant to the host operating system.  Testing those separate files is going to require expanding our test suite and CI, though.

debarshiray:
> I just remembered there is also `VARIANT` in os-release. That
> one would work even better than this solution (if they key is
> present) but I don't mind sticking to this for the time being
> (definitely better than to write `Silverblue` on `Workstation`).

I think, these days we should use `PRETTY_NAME` and let the downstream distributor decide how the host operating system should be called.  `PRETTY_NAME` has the advantage of being effectively a mandatory field.  The `os-release(5)` [manual](https://www.freedesktop.org/software/systemd/man/os-release.html) says:
```
If not set, a default of "PRETTY_NAME="Linux"" may be used.
```

Note that the [original](https://github.com/containers/toolbox/commit/79f59b667b22b647253cb047a9c04cd60d1db03a) welcome banner was from a time when on Fedora Silverblue `VARIANT` and `VARIANT_ID` were set to the same values as on Fedora Workstation.  That's why it started off hard-coded as *Fedora Silverblue*.

Using `PRETTY_NAME` will also reduce the amount of distribution-specific information that we have to carry around.

debarshiray:
> Note that the [original](https://github.com/containers/toolbox/commit/79f59b667b22b647253cb047a9c04cd60d1db03a) welcome banner was from a time when on Fedora Silverblue `VARIANT` and `VARIANT_ID` were set to the same values as on Fedora Workstation. That's why it started off hard-coded as _Fedora Silverblue_.

For what it's worth, Fedora Silverblue [got its own](https://pagure.io/workstation-ostree-config/c/c18ef957d11862d32f362722931dbfdf1f5beb0d) `VARIANT` and `VARIANT_ID` only as recently as Fedora 32.

PR #882, opened 2021-09-22
Implemented new flag into toolbox create to let change dot color

kuba3351:
I have an idea of a small but userful feature about changing the dot color that is displayed inside toolbox. The color can be assigned during toolbox creation using a new `--color` flag. I think it is a userful when I have a few toolboxes and if I remember the colors assigned to a toolboxes I can quickly check in what toolbox I am without scrolling up to see what name I typed into a `toolbox enter` command.

I have no experience in Go before, but was able to implement it, so please review my code and tell what do you think about this feature. Thanks


kuba3351:
I added a bash completion for the new flag, and asking if anyone can help with review and merge?

kuba3351:
Squashed into one commit and corrected commit message, and friendly pinging for review.

debarshiray (changes requested):
From a very high-level, I am reluctant to add a new command line flag to `toolbox create` to change the colour of the dot in the shell's command prompt, because it moves in a direction where the colour cannot be changed once the container has been created. eg., note that parameters passed to `podman create` are very difficult, if not impossible, to change afterwards.

This is part of the bigger conversation about making it easier to identify different containers, and there are several ideas and suggestions floating around about that.

kuba3351:
Thanks for your reply.

I agree that making an easier to identify different containers can be achieved by different ways, and this is just one of a lot of possible solutions.

But I am not agree that color of the dot cannot be changed after the container is created, because it is implemented on top of an environment variable, and change the value of `TOOLBOX_COLOR` variable in the running container is very easy, just run `TOOLBOX_COLOR=<new_value>` inside container and the dot color should change instantly. So, it is possible to implement the same flag on the `toolbox enter` command, read the provided value in a `toolbox init` that is running inside the container, and set the new value of `TOOLBOX_COLOR` variable inside the container.

But If you think that assign a color of the dot is not a good way to identify containers and this feature will be useless, you can close this PR.

debarshiray:
> I am not agree that color of the dot cannot be changed after the container is
> created, because it is implemented on top of an environment variable, and
> change the value of TOOLBOX_COLOR variable in the running container is
> very easy, just run TOOLBOX_COLOR=<new_value> inside container and
> the dot color should change instantly.

First of all, environment variables are notoriously tricky to manage.  We shouldn't be introducing one lightly.  :)

When implementing this through `toolbox enter` we need to be careful so that all the different shell instances running in the same container continue to have the same colour.  So, if the colour is changed for one, then it gets changed for all.

kuba3351:
Thank for your reply. So if you don't recommend to implement it on top of environemnt variable, what I should use instead to store information about selected dot color in the container?

debarshiray:
> if you don't recommend to implement it on top of environemnt variable,
> what I should use instead to store information about selected dot color
> in the container?

I don't have any concrete ideas off-hand.  I need to sit down and carefully go through all the different issues and PRs open around this topic.

One idea is to automatically generate a colour for a container.

Another idea is to enable a graphical utility (like Settings) to assign colours to each container, like in [Chrome OS](https://www.aboutchromebooks.com/news/chrome-os-98-adds-management-of-multiple-chromebook-linux-containers/).

kuba3351:
> One idea is to automatically generate a colour for a container.

I think automatically generate colour of the dot is not a good idea, because I think users should have control of what colours they want to assign to a container. For example, I am opening two toolboxes to work with the same project and one of another, and I would like two toolboxes to have the same color and one to have another, if assigning of the color will be automatic, it will not be possible.

And second, it will be a problem what to do if user opens more toolboxes than possible colors of the dot

I think that we can treat assigning of the dot color like labeling. Different colors of the dot are different labels of the toolboxes, like labels of issues or Merge Requests. The similat concept is in Qubes OS, this is an OS in which apps are running in the VMs instead of the containers, and you can assign a window colour to each VM during creation and every app that is running inside a VM have a chosen window color, so you can quickly identify VMs

So I think it should be implemented in a way that let users to chose colors themselves

castedo:
I'm probably biased, but i think you should do what I propose in #990. The file system is a great place to store all kinds of data. Pick whatever emoji/symbol of whatever color you want to put it in a file in a standard path under /etc/. You can do that after container creation, during container creation, before container creation during image building. Whatever tool can change the file system and there are lots of those tools.

kuba3351:
ok, I can change the implementation to be based on that file instead of an environment variable. I will work on it.

PR #899, opened 2021-10-22
test/system: Test startup on Rawhide with supported versions

HarryMichal:
We need to know if the latest changes in the libc (that is dynamically
linked to the binary) causes problems in containers based on older
releases of Fedora.

The estimate of the version numbers ís very crude and does not follow
the upstream schedule. That should not be a problem.

A part of an existing test has been reused an made into a helper
function to implement this.

### DIFF
--- a/test/system/103-container.bats
+++ b/test/system/103-container.bats
@@ -19,23 +19,30 @@ teardown() {
 
   create_default_container
 
-  run $PODMAN start $CONTAINER_NAME
+  res="$(container_started $CONTAINER_NAME)"
 
-  CONTAINER_INITIALIZED=0
-
-  for TRIES in 1 2 3 4 5
-  do
-    run $PODMAN logs $CONTAINER_NAME
-    CONTAINER_OUTPUT=$output
-    run grep 'Listening to file system and ticker events' <<< $CONTAINER_OUTPUT
-    if [[ "$status" -eq 0 ]]; then
-      CONTAINER_INITIALIZED=1
-      break
-    fi
-    sleep 1
-  done
-
-  echo $CONTAINER_OUTPUT
-  assert [ "$CONTAINER_INITIALIZED" -eq 1 ]
+  assert [ "$res" -eq 1 ]
 }
 
+@test "container(Fedora Rawhide): Containers with supported versions start without issues" {
+  local os_release="$(find_os_release)"
+  local system_id="$(get_system_id)"
+  local system_version="$(get_system_version)"
+  local rawhide_res="$(awk '/rawhide/' $os_release)"
+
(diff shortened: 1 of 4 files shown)


HarryMichal:
> Do we want the test to be only for Fedora Rawhide or should I adjust it even for older releases? The bottom limit would still be the "eldest" supported release.

Let's stick to Rawhide for now as it is the main motivation behind this test and other releases may be added in the future.

HarryMichal:
This slowness on Rawhide is making me angry.

test/system/103-container.bats
```diff
@@ -19,23 +19,30 @@ teardown() {
...
+  res="$(container_started $CONTAINER_NAME)"
+
+  assert [ "$res" -eq 1 ]
 }
 
+@test "container(Fedora Rawhide): Containers with supported versions start without issues" {
+  local os_release="$(find_os_release)"
+  local system_id="$(get_system_id)"
+  local system_version="$(get_system_version)"
+  local rawhide_res="$(awk '/rawhide/' $os_release)"
```
debarshiray, line 31: I am curious. Can't we just use `grep`? It's easier to understand than `awk`.

debarshiray (commented):
Thanks for the new test.

This line in the commit message made me curious:
```
    The estimate of the version numbers is very crude and does not follow
    the upstream schedule. That should not be a problem, though.
```

Why is it crude? How does it not follow the upstream schedule? And why it shouldn't be a problem? :)

Are you referring to how we are testing N, N-1 and N-2 (where N is the version of Rawhide)?

debarshiray (commented):
Seems like there's no commit `da6b6a7c5a6e5e284e6642c29159a872ceec94e2`:
```
[rishi@kolache toolbox]$ git show da6b6a7c5a6e5e284e6642c29159a872ceec94e2
fatal: bad object da6b6a7c5a6e5e284e6642c29159a872ceec94e2
```

I suppose it's meant to be 54a2ca1ead343c3db2c7443cc58b9d9e896d3c46

HarryMichal:
> Are you referring to how we are testing N, N-1 and N-2 (where N is the version of Rawhide)?

Yes, that's exactly it.

HarryMichal:
> Seems like there's no commit `da6b6a7c5a6e5e284e6642c29159a872ceec94e2`:
> 
> ```
> [rishi@kolache toolbox]$ git show da6b6a7c5a6e5e284e6642c29159a872ceec94e2
> fatal: bad object da6b6a7c5a6e5e284e6642c29159a872ceec94e2
> ```
> 
> I suppose it's meant to be [54a2ca1](https://github.com/containers/toolbox/commit/54a2ca1ead343c3db2c7443cc58b9d9e896d3c46)

Ah, snap. Seems like I mistakenly pointed that to my fork :/.

PR #915, opened 2021-11-07
test/system: Add new tests for failures and test placeholders for entering containers

HarryMichal:
There's never enough tests.

Moved some test files around to keep the tested commands together.


debarshiray:
Thanks for the tests, @HarryMichal !

PR #916, opened 2021-11-07
src: Decouple CLI specifics from runCommand

HarryMichal:
From the commit message:
```
This is a major surgery that makes the footprint of runCommand
smaller by removing the CLI related functionality from it and
moving it to its callers. This move causes some duplication but
that can be solved in the future by introducing a CLI-specific
abstraction.

A new error for non-existent containers had to be added to make
it possible to handle the case from callers of the function.

The behaviour of all the affected commands ('root', enter, run)
should remain the same.
```

This is part of some ground-work I needed to do to work on some new functionality in Toolbx.

This is based on https://github.com/containers/toolbox/pull/915 which contains tests that should catch any errors that I could have made.


debarshiray (commented):
Umm... I am struggling to see the point of this change.

What's this *new functionality* that you mention?

> This move causes some duplication but
> that can be solved in the future by introducing a CLI-specific
> abstraction.

The duplication is quite a bit, and it's not clear to me that we need even more abstraction because Toolbox code is small enough that we shouldn't need too much.

The code here really should remain the same as much as possible unless there are other things getting in the way. So I am hesitant to walk in the other direction without seeing the full picture.

PR #917, opened 2021-11-11
src: Update dependencies and fix issue with usage message

### LINKED ISSUE #962: Doubled usage line
**Describe the bug**
The usage output is doubled when a newer version of Cobra is used to build Toolbx.

1. Build Toolbx with latest dependencies (mainly Cobra 1.2.1)
2. Run system tests

**Expected behaviour**
```
; toolbox --foo
Error: unknown flag: --foo
```

**Actual behaviour**
```
; toolbox --foo
Error: unknown flag: --foo
Run 'toolbox --help' for usage.Error: Run 'toolbox --help' for usage.
```

`toolbox version 0.0.99.3`

not relevant; build from source

not relevant

not relevant

**Info about your OS**
not relevant

**Additional context**
Connected upstream issue: https://github.com/spf13/cobra/issues/1532

HarryMichal:
We haven't updated the dependencies of Toolbox since the rewrite in Go. There was no real need since our main target is Fedora that takes care of updating the different packages. But because of that we did not catch a regressions in Cobra (see the commit messages).

The dependency update has been done first for patch versions and then even for major versions. Both steps were filled with a call to `go mod tidy`.

Fixes #962


HarryMichal:
recheck

debarshiray:
> But because of that we did not catch a regressions in Cobra (see the commit messages).

Umm... I don't quite understand much from the commit messages.

Is there an actual regression in Cobra that we are addressing?  But since we haven't updated the dependencies, how can it be a recent regression?  Or are these just a periodic update of the dependencies?

test/system/002-help.bats
```diff
@@ -63,61 +63,61 @@ setup() {
 
   assert_failure
   assert_line --index 0 "Error: unknown flag: --foo"
-  assert_line --index 1 "Run 'toolbox --help' for usage."
+  assert_line --index 1 "Run 'toolbox create --help' for usage."
```
debarshiray, line 66: I actually want to keep it as:
`Run 'toolbox --help' for usage`

There can be several reasons why the command line parsing failed.  Some of these failures can be quite subtle in terms of the code paths they take and not obvious at first sight.  Hence, it's a lot easier to reason about the code if they simply suggest `toolbox --help`.  It's not worth dealing with the odd bug report arising out of a slightly wrong suggestion.

Secondly, I'd like to consolidate the mental model followed by documentation.  It's a lot easier to manage the content if we are sure that people generally start at the top-level and then dive into each command.  That's why the top level `toolbox(1)` manual is deliberately brief and prominently points out each command.

debarshiray (changes requested):
It will be easier to understand if the commit message included a clear example of the change in behaviour of Cobra. Like in https://github.com/spf13/cobra/issues/1532

debarshiray:
> > But because of that we did not catch a regressions in Cobra (see the commit messages).
> 
> Umm... I don't quite understand much from the commit messages.
> 
> Is there an actual regression in Cobra that we are addressing? But since we
> haven't updated the dependencies, how can it be a recent regression? Or are
> these just a periodic update of the dependencies?

Ah, I see that it's a change in behaviour in newer Cobra versions that distributors might be using for their builds!

In that case, does the newer approach of using the templates also work with older versions of Cobra?  I am asking because different versions of Fedora might have different versions of Cobra.

HarryMichal:
> In that case, does the newer approach of using the templates also work with older versions of Cobra? I am asking because different versions of Fedora might have different versions of Cobra.

The template has been available from the very first version of Cobra, so yes.

test/system/002-help.bats
HarryMichal: Personally, I find it more useful for the usage to mention the specific command. But I don't have a strong opinion about this, so I'll revert that :+1:.

HarryMichal:
@debarshiray?

HarryMichal:
This supersedes https://github.com/containers/toolbox/pull/775

src/go.mod
```diff
@@ -3,17 +3,22 @@ module github.com/containers/toolbox
 go 1.13
 
 require (
-	github.com/HarryMichal/go-version v1.0.0
+	github.com/HarryMichal/go-version v1.0.1
 	github.com/acobaugh/osrelease v0.0.0-20181218015638-a93a0a55a249
-	github.com/briandowns/spinner v1.10.0
+	github.com/briandowns/spinner v1.16.0
```
debarshiray, line 8: Shouldn't this be updated to `v1.17.0` to include your fix for https://github.com/briandowns/spinner/issues/123 ?

src/go.mod
```diff
@@ -3,17 +3,22 @@ module github.com/containers/toolbox
 go 1.13
 
 require (
-	github.com/HarryMichal/go-version v1.0.0
+	github.com/HarryMichal/go-version v1.0.1
```
debarshiray, line 6: Thanks for splitting the dependency bumps into two steps. It might came in handy in case we hit any regressions.

debarshiray (commented):
Thanks. One quick comment:

src/go.mod
```diff
@@ -3,17 +3,25 @@ module github.com/containers/toolbox
...
+	github.com/magiconair/properties v1.8.5 // indirect
+	github.com/mattn/go-colorable v0.1.11 // indirect
+	github.com/mattn/go-isatty v0.0.14
 	github.com/sirupsen/logrus v1.4.2
-	github.com/spf13/cobra v0.0.5
-	github.com/spf13/viper v1.3.2
+	github.com/spf13/cast v1.3.1 // indirect
+	github.com/spf13/cobra v0.0.7
+	github.com/spf13/pflag v1.0.5 // indirect
+	github.com/spf13/viper v1.4.0
```
debarshiray: I wonder why the Viper version went from v1.3.2 to v1.4.0. It doesn't look like a patch update. I suppose it was caused by Cobra?

src/go.mod
HarryMichal: Sure. The change has been made before the release.

src/go.mod
HarryMichal: I might have made a mistake while rebasing. I got a bit confused at one moment. But it is possible a dependency bump caused it.

debarshiray:
Some more observations ...

As far as I understand, the wrong behaviour isn't this, as mentioned in the commit message:
```
$ toolbox --foo
Error: unknown flag: --foo
Error: Run 'toolbox --help' for usage.
```

It's actually this:
```
$ toolbox --foo
Error: unknown flag: --foo
Run 'toolbox --help' for usage.Error: Run 'toolbox --help' for usage.
```

ie., the usage string gets strangely duplicated, with an `Error:` prefix in the middle.

We could have avoided the duplication by not printing the text ourselves, if it wasn't for the extra prefix, etc.. These are details that we don't have to explain in detail, but the example is important. A wrong example will confuse people even more because it won't match their observation.

Secondly, we should bump the dependencies with `go get -u` *after* working around the Cobra issue. Otherwise it will break `git bisect`.

src/go.mod
```diff
@@ -3,17 +3,25 @@ module github.com/containers/toolbox
...
+	github.com/HarryMichal/go-version v1.0.1
 	github.com/acobaugh/osrelease v0.0.0-20181218015638-a93a0a55a249
 	github.com/briandowns/spinner v1.10.0
 	github.com/docker/go-units v0.4.0
-	github.com/fsnotify/fsnotify v1.4.7
-	github.com/godbus/dbus/v5 v5.0.3
-	github.com/mattn/go-isatty v0.0.8
+	github.com/fsnotify/fsnotify v1.4.9
+	github.com/godbus/dbus/v5 v5.0.6
+	github.com/konsorten/go-windows-terminal-sequences v1.0.3 // indirect
```
debarshiray: We don't need to highlight the changes in the indirect dependencies in the commit message. There are just too many of those and they keep coming and going. It's better to only highlight our direct dependencies in the commit message, unless there's one that we really need to be careful about.

HarryMichal: Fair enough.

src/go.mod
HarryMichal: Just tried it again and the version got bumped again, so does not seem like a problem on my part.

debarshiray:
recheck

debarshiray:
Wow! Software Factory is really struggling today.

Issue #920, opened 2021-11-12
Provide a graphic for the README


jimmac:
Remove logos from the source repo as they are being moved to the [website repo](https://github.com/containers/containertoolbx.org/commit/12e3203f7b579f0d7af208cfea2846dc541a2836). Provide a graphic for the README.

PR #921, opened 2021-11-12
replace outdated logos with pixels

jimmac:
Fixes https://github.com/containers/toolbox/issues/920


debarshiray:
Thanks, @jimmac

PR #922, opened 2021-11-13
Style fixes

debarshiray:
Minor fix-ups to https://github.com/containers/toolbox/pull/837

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -67,15 +67,14 @@ func showManual(manual string) error {
 	manBinary, err := exec.LookPath("man")
 	if err != nil {
 		if errors.Is(err, exec.ErrNotFound) {
-			fmt.Print(`toolbox - Tool for containerized command line environments on Linux
-
-Common commands are:
-create    Create a new toolbox container
-enter     Enter an existing toolbox container
-list      List all existing toolbox containers and images
-
-Go to https://github.com/containers/toolbox for further information.
-`)
+			fmt.Printf("toolbox - Tool for containerized command line environments on Linux\n")
+			fmt.Printf("\n")
+			fmt.Printf("Common commands are:\n")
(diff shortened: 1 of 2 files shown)


src/cmd/utils.go
```diff
@@ -67,15 +67,14 @@ func showManual(manual string) error {
...
-Go to https://github.com/containers/toolbox for further information.
-`)
+			fmt.Printf("toolbox - Tool for containerized command line environments on Linux\n")
+			fmt.Printf("\n")
+			fmt.Printf("Common commands are:\n")
+			fmt.Printf("create    Create a new toolbox container\n")
+			fmt.Printf("enter     Enter an existing toolbox container\n")
+			fmt.Printf("list      List all existing toolbox containers and images\n")
+			fmt.Printf("\n")
+			fmt.Printf("Go to https://github.com/containers/toolbox for further information.\n")
```
HarryMichal, line 77: Didn't you want to use `fmt.Builder` instead? Having several prints in a row is non-ideal.

debarshiray: I mainly wanted to avoid using a backtick-quoted raw string literal.  Other than the cosmetics, it makes it difficult to know if there are trailing whitespaces or not, which can be important for user-visible text.

Writing to a `strings.Builder` with `fmt.Fprintf` is the same as writing to `os.Stdout`, which is the same as using `fmt.Printf`. :) We already have some prior examples in the code.

My thoughts were a bit muddled last night. Sorry about that.

HarryMichal: Is it the same, though? Several prints to `os.Stdout` could get theoretically "out of sync" (flushing) while using `strings.Builder` results in a single print. Also, it could happen that this text would be generated by a function and to test in in unit tests, it would have to return the text -> `strings.Builder` is the most efficient to build it.

Anyway, you already merged, so there's not much point in discussing this further.

PR #923, opened 2021-11-13
build: Don't assume that libc.so is always in /usr/lib or /usr/lib64

debarshiray:
The location for public shared libraries can change from one operating
system distribution to another. eg., while Fedora uses /usr/lib and
/usr/lib64, depending on the hardware architecture, Debian uses paths
like /usr/lib/x86_64-linux-gnu. Therefore, it's best not to assume
anything and ask the toolchain.

### DIFF
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -16,9 +16,9 @@
 #
 
 
-if [ "$#" -ne 3 ]; then
+if [ "$#" -ne 4 ]; then
     echo "go-build-wrapper: wrong arguments" >&2
-    echo "Usage: go-build-wrapper [SOURCE DIR] [OUTPUT DIR] [VERSION]" >&2
+    echo "Usage: go-build-wrapper [SOURCE DIR] [OUTPUT DIR] [VERSION] [C COMPILER]" >&2
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


src/go-build-wrapper
```diff
@@ -27,7 +27,22 @@ if ! cd "$1"; then
     exit 1
 fi
 
-go build -trimpath -ldflags "-extldflags '-Wl,-rpath,/run/host/usr/lib -Wl,-rpath,/run/host/usr/lib64' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$3" -o "$2/toolbox"
+if ! libc_dir=$("$4" --print-file-name=libc.so); then
```
debarshiray, line 30: I wish there was a way to do this with Meson's [compiler.find_library](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerfind_library).

HarryMichal:
Thanks for this!

Testing on Debian would be helpful in this case. One solution is to add support to SoftwareFactory or just use GitHub Actions since we mainly want to test the build and that can be done in a container, too. Further testing would require support in SoftwareFactory since we require a VM for testing purpose and not a container.

PR #925, opened 2021-11-14
README graphic: adjustments

jimmac:
- shadow was too prominent for a white background
- properly right align
- filesize considerations. The old graphic was 700x160 to avoid scale
  filtering, but resulted in a large file.
- sadly the only way to have proper control is html+css


jimmac:
It's all great, but doesn't work on github :)

PR #930, opened 2021-11-18
Add capsh as a container requirement in the docs

acminor:
Without capsh a toolbox might fail to be entered into with a message about not finding /bin/bash.


debarshiray:
Thanks, @acminor 

Merged after tweaking the commit message to not be longer than 72 characters per line.

PR #932, opened 2021-11-21
github: Add pull request template

HarryMichal:
The pull request template has been overdue. The contribution guidelines need to be updated to reflect this, too. Git hooks could also be helpful in the contribution process.

https://github.com/containers/toolbox/pull/932

### DIFF
--- a/.github/pull_request.md
+++ b/.github/pull_request.md
@@ -1,0 +1,20 @@
+---
+name: Pull request
+about: Toolbx's pull request template
+title: ''
+assignees: ''
+---
+
+<!--
+Thank you for contributing to Toolbx!
+
+Describe what your pull request does. Either use your own words or use the text in some of your commit messages if applicable.
+
+If your pull request fixes an issue, mention it in the description like this:
+
+Fixes: #xxx
+
+'xxx' stands for the issue number
+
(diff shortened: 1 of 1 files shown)


debarshiray:
Personally, I care more about the format of the Git commit messages.

This is because when I am hacking on Toolbox, especially when I am debugging existing code and trying to figure out why some things are done the way they are, I am not looking at the Git history through GitHub's web UI.  The web UI is quite limiting in the Git commands that it exposes.  I am mainly using Git's command line interface, because it's right next to my text editor.

For me, the GitHub issue or pull request carry extra secondary context. They are where people discuss the commits, the different trade-offs and they add extra secondary context to the changes that landed. So, *sometimes*, if the commit message isn't enough, then I end up reading through the issue or pull request to understand the different view points, pros & cons, who said what, etc..

In other words, the Git commits are *primary*, and the GitHub issues and pull requests are *secondary*.

So, I wish we could have some Git hooks that would ensure that:
* Commit messages don't exceed 72 characters per line.
* Commit messages contain a link to an issue, if there's one, or the GitHub pull request.
* Prevent trailing white spaces.

... because I think sometimes, whether for valid or invalid reasons, contributors will override the GitHub templates, which is fine, but I don't want them to play fast and loose with the actual Git commits.

PR #935, opened 2021-11-25
cmd/create: Add option --authfile

### LINKED ISSUE #689: RFE: Migration path for users of coreos/toolbox
In RHCOS, we are shipping https://github.com/coreos/toolbox in the OS and we want to move to shipping `containers/toolbox` in its place.  In order to do that, we need to ensure that the upgrade path is smooth, as `toolbox` is heavily utilized by the Red Hat Support organization for gathering debug information.

I see two requirements for RHCOS users:
  - support for `.toolboxrc` files
  - retention of the same behavior from `coreos/toolbox`

**Describe alternatives you've considered**
The CoreOS team is currently maintaining the bash script implementation over at `coreos/toolbox` and we could continue to do so.

**Additional context**
FWIW, I don't expect the `containers/toolbox` team to work on this alone.  Please reach out to the CoreOS team to help with the work.

HarryMichal:
The option accepts a path to a file that is passed to an internal call
to 'podman pull' via the '--authfile' option. This will make it easier
to pull images from registries with authentication in-place.

The commit repurposes the test work I did for a rejected feature on
registry authentication. A registry is started locally and the use of
auth file is tested against it.

Fixes #689

Depends on: #818 & #1024

Requires a Docker registry image at quay.io/toolbox_tests/registry for running tests.

### DIFF
--- a/doc/toolbox-create.1.md
+++ b/doc/toolbox-create.1.md
@@ -4,7 +4,8 @@
 toolbox\-create - Create a new toolbox container
 
 ## SYNOPSIS
-**toolbox create** [*--distro DISTRO* | *-d DISTRO*]
+**toolbox create** [*--authfile AUTHFILE*]
+               [*--distro DISTRO* | *-d DISTRO*]
                [*--image NAME* | *-i NAME*]
                [*--release RELEASE* | *-r RELEASE*]
                [*CONTAINER*]
@@ -79,6 +80,14 @@ confusion.
 
 ## OPTIONS ##
 
+**--authfile** FILE
+
+Path to a FILE with credentials for authenticating to the registry for private
+images. The FILE is usually set using `podman login`, and will be used by
+`podman pull` to get the image.
+
+The default location for FILE is `$XDG_RUNTIME_DIR/containers/auth.json`.
+
 **--distro** DISTRO, **-d** DISTRO
 
 Create a toolbox container for a different operating system DISTRO than the
@@ -120,6 +129,12 @@ $ toolbox create --distro fedora --release f36
 $ toolbox create --image bar foo
 ```
 
+### Create a toolbox container from a custom image needing authentication
+
+```
+$ toolbox create --authfile ~/auth.json --image registry.example.com/bar
+```
+
 ## SEE ALSO
 
-`toolbox(1)`, `toolbox-init-container(1)`, `podman(1)`, `podman-create(1)`
+`toolbox(1)`, `toolbox-init-container(1)`, `podman(1)`, `podman-create(1)`, `podman-login(1)`, `podman-pull(1)`
--- a/playbooks/setup-env.yaml
+++ b/playbooks/setup-env.yaml
@@ -12,8 +12,10 @@
           - flatpak-session-helper
           - golang
           - golang-github-cpuguy83-md2man
+          - httpd-tools
           - meson
           - ninja-build
+          - openssl
           - podman
           - skopeo
           - systemd
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -42,6 +42,7 @@ const (
 
 var (
 	createFlags struct {
+		authFile  string
 		container string
 		distro    string
 		image     string
@@ -66,6 +67,11 @@ var createCmd = &cobra.Command{
 
 func init() {
 	flags := createCmd.Flags()
+
+	flags.StringVar(&createFlags.authFile,
+		"authfile",
+		"",
+		"Path to a file with credentials for authenticating to the registry for private images")
 
 	flags.StringVarP(&createFlags.container,
 		"container",
@@ -122,6 +128,12 @@ func create(cmd *cobra.Command, args []string) error {
 
 	if cmd.Flag("image").Changed && cmd.Flag("release").Changed {
 		return errors.New("options --image and --release cannot be used together")
+	}
+
+	if cmd.Flag("authfile").Changed {
+		if !utils.PathExists(createFlags.authFile) {
+			return fmt.Errorf("file %s not found", createFlags.authFile)
+		}
 	}
 
 	var container string
@@ -717,7 +729,7 @@ func pullImage(image, release string) (bool, error) {
 		defer s.Stop()
 	}
 
-	if err := podman.Pull(imageFull); err != nil {
+	if err := podman.Pull(imageFull, createFlags.authFile); err != nil {
 		var builder strings.Builder
 		fmt.Fprintf(&builder, "failed to pull image %s\n", imageFull)
 		fmt.Fprintf(&builder, "If it was a private image, log in with: podman login %s\n", domain)
--- a/src/pkg/podman/podman.go
+++ b/src/pkg/podman/podman.go
@@ -227,9 +227,18 @@ func IsToolboxImage(image string) (bool, error) {
 }
 
 // Pull pulls an image
-func Pull(imageName string) error {
+//
+// authfile is a path to a JSON authentication file and is internally used only
+// if it is not an empty string.
+func Pull(imageName string, authfile string) error {
 	logLevelString := LogLevel.String()
-	args := []string{"--log-level", logLevelString, "pull", imageName}
+	args := []string{"--log-level", logLevelString, "pull"}
+
+	if authfile != "" {
+		args = append(args, []string{"--authfile", authfile}...)
+	}
+
+	args = append(args, imageName)
 
 	if err := shell.Run("podman", nil, nil, nil, args...); err != nil {
 		return err
--- a/test/system/000-setup.bats
+++ b/test/system/000-setup.bats
@@ -19,4 +19,6 @@ load 'libs/helpers'
         _pull_and_cache_distro_image fedora "$((system_version-1))" || false
         _pull_and_cache_distro_image fedora "$((system_version-2))" || false
     fi
+
+    _setup_docker_registry
 }
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -117,3 +117,37 @@ teardown() {
   assert_line --index 0 "Error: release not found for non-default distribution $distro"
   assert [ ${#lines[@]} -eq 1 ]
 }
+
+@test "create: Try to create a container and pass a non-existent file to the --authfile option" {
+  local file="$BATS_RUN_TMPDIR/non-existent-file"
+
+  run $TOOLBOX create --authfile "$file"
+
+  assert_failure
+  assert_output "Error: file $file not found"
+}
+
+@test "create: Create a container based on an image from locked registry using an authentication file" {
+  local authfile="$BATS_RUN_TMPDIR/authfile"
+  local image="fedora-toolbox:32"
+
+  run $PODMAN login --authfile "$authfile" --username user --password user "$DOCKER_REG_URI"
+  assert_success
+
+  run $TOOLBOX --assumeyes create --image "$DOCKER_REG_URI/$image"
+
+  assert_failure
+  assert_line --index 0 "Error: failed to pull image $DOCKER_REG_URI/$image"
+  assert_line --index 1 "If it was a private image, log in with: podman login $DOCKER_REG_URI"
+  assert_line --index 2 "Use 'toolbox --verbose ...' for further details."
+  assert [ ${#lines[@]} -eq 3 ]
+
+  run $TOOLBOX --assumeyes create --authfile "$authfile" --image "$DOCKER_REG_URI/$image"
+
+  rm "$authfile"
+
+  assert_success
+  assert_line --index 0 "Created container: fedora-toolbox-32"
+  assert_line --index 1 "Enter with: toolbox enter fedora-toolbox-32"
+  assert [ ${#lines[@]} -eq 2 ]
+}
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 load 'libs/bats-support/load'
+load 'libs/bats-assert/load'
 
 # Helpful globals
 readonly TEMP_BASE_DIR="${XDG_CACHE_HOME:-$HOME/.cache}/toolbox"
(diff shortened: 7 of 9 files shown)


src/pkg/podman/podman.go
```diff
@@ -227,9 +227,18 @@ func IsToolboxImage(image string) (bool, error) {
 }
 
 // Pull pulls an image
-func Pull(imageName string) error {
+//
+// authfile is a path to a JSON authentication file and is internally used only
+// if it is not an empty string. The file at the path should exist
+//
+// see podman-pull(1) manual page for more information
```
debarshiray, line 234: Isn't this a bit redundant? :)

Anybody who is reading the code will know that `podman pull` is involved, etc..

debarshiray (changes requested):
Thanks for working on this, @HarryMichal !

I am happy that some of your work from https://github.com/containers/toolbox/pull/787 came in useful.  However, since those commits aren't part of the Git history, it's not worth referring to them in the commit message without any further context:
```
The commit repurposes the test work I did for a rejected feature on
registry authentication.
```

The reader will wonder: *what work*, *what feature*, *why was it rejected*, etc..  Either leave out the sentence; or if you think the changes to the tests aren't easy to understand, then feel free to describe them, even if it repeats some text from https://github.com/containers/toolbox/pull/787

test/system/100-root.bats
```diff
@@ -5,8 +5,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
-  check_xdg_runtime_dir
-  setup_containers_store  
+  _setup_environment
```
debarshiray, line 8: Cool!

doc/toolbox-create.1.md
```diff
@@ -119,6 +127,12 @@ $ toolbox create --distro fedora --release f30
 $ toolbox create --image bar foo
 ```
 
+### Create a toolbox container using an image from a registry with authentication
+
+```
+$ toolbox create --image quay.io/foo/bar --authfile $HOME/auth.json
```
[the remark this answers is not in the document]
HarryMichal: Fixed.

doc/toolbox-create.1.md
```diff
@@ -79,6 +80,13 @@ confusion.
 
 ## OPTIONS ##
 
+**--authfile** AUTHFILE
+
+Path to a file used for registry authentication. Exposes option '--authfile' of
```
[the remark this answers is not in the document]
HarryMichal: Fixed.

doc/toolbox-create.1.md
```diff
@@ -79,6 +80,13 @@ confusion.
...
+command 'podman pull' that is used internally.
+
+See `podman-pull(1)` for more information.
```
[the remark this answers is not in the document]
HarryMichal: Fixed and also added `podman-login(1)` to the `SEE ALSO` section.

src/pkg/podman/podman.go
```diff
@@ -227,9 +227,18 @@ func IsToolboxImage(image string) (bool, error) {
 }
 
 // Pull pulls an image
-func Pull(imageName string) error {
+//
+// authfile is a path to a JSON authentication file and is internally used only
+// if it is not an empty string. The file at the path should exist
```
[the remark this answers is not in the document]
HarryMichal: Fixed.

src/pkg/podman/podman.go
HarryMichal: True. Fixed.

test/system/100-root.bats
HarryMichal: Right?

HarryMichal:
> I am happy that some of your work from #787 came in useful. However, since those commits aren't part of the Git history, it's not worth referring to them in the commit message without any further context:
> 
> ```
> The commit repurposes the test work I did for a rejected feature on
> registry authentication.
> ```
> 
> The reader will wonder: _what work_, _what feature_, _why was it rejected_, etc.. Either leave out the sentence; or if you think the changes to the tests aren't easy to understand, then feel free to describe them, even if it repeats some text from #787

That's a valid point and I fixed that in the iteration. Thanks!

I tried to address all the raised concerns. Anything else?

playbooks/setup-env.yaml
```diff
@@ -12,8 +12,11 @@
           - flatpak-session-helper
           - golang
           - golang-github-cpuguy83-md2man
+          - httpd-tools
           - meson
           - ninja-build
+          - openssl
+          - patchelf
```
HarryMichal, line 19: Not needed since #942.

src/cmd/create.go
```diff
@@ -115,6 +121,12 @@ func create(cmd *cobra.Command, args []string) error {
 		return errors.New("options --image and --release cannot be used together")
 	}
 
+	if cmd.Flag("authfile").Changed {
+		if !utils.PathExists(createFlags.authFile) {
+			return fmt.Errorf("file %s does not exist", createFlags.authFile)
```
debarshiray, line 126: Nitpick: *not found* seems to be more common.

src/pkg/podman/podman.go
```diff
@@ -227,9 +227,16 @@ func IsToolboxImage(image string) (bool, error) {
...
+// authfile is a path to a JSON authentication file and is internally used only
+// if it is not an empty string. The file at the path should exist.
+func Pull(imageName string, authfile string) error {
 	logLevelString := LogLevel.String()
-	args := []string{"--log-level", logLevelString, "pull", imageName}
+	args := []string{"--log-level", logLevelString, "pull"}
+    if authfile != "" {
+        args = append(args, []string{"--authfile", authfile}...)
+    }
+    args = append(args, imageName)
```
debarshiray, line 239: The alignment is off.

*This was caught by `gofmt` in our test suite.  I have no idea why it eventually timed out.*

src/pkg/podman/podman.go
```diff
@@ -227,9 +227,16 @@ func IsToolboxImage(image string) (bool, error) {
 }
 
 // Pull pulls an image
-func Pull(imageName string) error {
+//
+// authfile is a path to a JSON authentication file and is internally used only
+// if it is not an empty string. The file at the path should exist.
```
debarshiray, line 232: Strictly speaking, nobody can ensure that *the file at the path should exist*.  All filesystem operations are inherently racy - files can come and go behind under our feet at any moment.  That's why the only option is to go ahead and try the operation and be ready to handle the errors.

`podman pull` does the same, but it doesn't offer an unique exit code for us to detect the specific error and show a meaningful message to the user.  We should try to see if we can get it to do that.

In the meantime, we can get away with checking for the existence of the file beforehand, because it will catch the vast majority of cases.  However, in terms of documenting this API, we can't expect the caller to guarantee that the file should exist.

test/system/000-setup.bats
```diff
@@ -19,4 +19,6 @@ load 'libs/helpers'
         _pull_and_cache_distro_image fedora "$((system_version-1))" || false
         _pull_and_cache_distro_image fedora "$((system_version-2))" || false
     fi
+
+    _setup_docker_registry
```
debarshiray, line 23: Are we setting up our local Docker registry and tearing it down for every test case?  Even the ones that don't need it?

If so, then we should probably not do that because setting up the registry is quite expensive.  It involves generating generating cryptographic certificates, which is CPU intensive.  We are downloading the `quay.io/toolbox_tests/registry` image, and it doesn't look like we are caching that one, but even if it did the local disk I/O would still be measurable.  Then actualling running the container for the registry and adding an image to it, which both involve disk I/O.

test/system/101-create.bats
```diff
@@ -79,3 +79,35 @@ teardown() {
...
+  assert_success
+
+  run $TOOLBOX -y create -i "$DOCKER_REG_URI/$image"
+
+  assert_failure
+  assert_line --index 0 "Error: failed to pull image $DOCKER_REG_URI/$image"
+  assert_line --index 1 "If it was a private image, log in with: podman login $DOCKER_REG_URI"
+  assert_line --index 2 "Use 'toolbox --verbose ...' for further details."
+
+  run $TOOLBOX -y create -i "$DOCKER_REG_URI/$image" --authfile "$authfile"
```
debarshiray, line 106: Better not to mix the long and short variants of the command-line options.

The short ones are easier to type, but the long ones are better for readability.

test/system/libs/helpers.bash
```diff
@@ -119,6 +127,86 @@ function _clean_cached_images() {
...
+    -v "${DOCKER_REG_AUTH_DIR}:/auth" \
+    -e REGISTRY_AUTH=htpasswd \
+    -e REGISTRY_AUTH_HTPASSWD_REALM="Registry Realm" \
+    -e REGISTRY_AUTH_HTPASSWD_PATH="/auth/htpasswd" \
+    -v "${DOCKER_REG_CERTS_DIR}":/certs \
+    -e REGISTRY_HTTP_ADDR=0.0.0.0:443 \
+    -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
+    -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
+    -p 50000:443 \
+    IMAGES[docker-reg]
```
debarshiray, line 188: This is really impressive!  Well done.

debarshiray (changes requested):
Thanks for updating the patches!

debarshiray:
I took the liberty to tweak the branch as above.  Let's see how the tests react.

debarshiray:
This failure is odd:
```
not ok 1 test suite: Setup in 52570ms
# (from function `_setup_docker_registry' in file test/system/libs/helpers.bash, line 141,
#  in test file test/system/000-setup.bats, line 23)
#   `_setup_docker_registry' failed
# /home/zuul-worker/src/github.com/containers/toolbox/test/system/libs/helpers.bash: line 141: assert_success: command not found
```

I don't understand why the assertion at `test/system/libs/helpers.bash:141` would fail.

debarshiray:
@travier does the code here work for you?

This only adds a command line option.  Next step would be to add a configuration file knob.

HarryMichal:
> I don't understand why the assertion at `test/system/libs/helpers.bash:141` would fail.

Forgot to add an include.

debarshiray:
recheck

travier (approved):
Generally LGTM. Will have to test on a real system.

doc/toolbox-create.1.md
```diff
@@ -79,6 +80,14 @@ confusion.
 
 ## OPTIONS ##
 
+**--authfile** AUTHFILE
+
+Path to a file used for registry authentication. Exposes option `--authfile` of
```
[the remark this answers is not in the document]
HarryMichal: Fixed.

doc/toolbox-create.1.md
```diff
@@ -79,6 +80,14 @@ confusion.
...
+command `podman pull` that is used internally.
```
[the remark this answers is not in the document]
HarryMichal: The wording has been updated. The mention of `--authfile` is gone and only mention of `podman pull` is still present. I'd personally that one mention.

test/system/000-setup.bats
HarryMichal: No, `000-setup.bats` is a test file that is called before all other (considering the whole directory is used for testing or the file is hand-picked). The registry is set up only once for the whole duration of the system tests.

test/system/101-create.bats
HarryMichal: Makes sense.

test/system/libs/helpers.bash
HarryMichal: \o/

HarryMichal:
recheck

HarryMichal:
Finally managed to finalize the tests. Took much longer than I thought it would.

doc/toolbox-create.1.md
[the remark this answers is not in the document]
debarshiray: Oops!  I left this one dangling.

The current text in the `toolbox-create(1)` manual says:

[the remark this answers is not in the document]
debarshiray: Mentioned `containers-auth.json(5)` in https://github.com/containers/toolbox/pull/1108

doc/toolbox-create.1.md
```diff
@@ -79,6 +80,14 @@ confusion.
...
+
+The default is `${XDG_RUNTIME_DIR}/containers/auth.json`, which is set using
+`podman login`.
```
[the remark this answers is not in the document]
debarshiray: This overlaps with the conversation above.

test/system/libs/helpers.bash
```diff
@@ -10,6 +11,11 @@ readonly IMAGE_CACHE_DIR="${BATS_RUN_TMPDIR}/image-cache"
 readonly ROOTLESS_PODMAN_STORE_DIR="${TEMP_STORAGE_DIR}/storage"
 readonly ROOTLESS_PODMAN_RUNROOT_DIR="${TEMP_STORAGE_DIR}/runroot"
 readonly PODMAN_STORE_CONFIG_FILE="${TEMP_STORAGE_DIR}/storage.conf"
+readonly DOCKER_REG_ROOT="${TEMP_STORAGE_DIR}/docker-registry-root"
```
debarshiray, line 14: That brought me here, to the definition of `DOCKER_REG_ROOT`, and I saw that wiping out the `TEMP_STORAGE_DIR` solves the problem.

My question is, why are not putting `DOCKER_REG_ROOT` inside `BATS_RUN_TMPDIR` ?  Are we afraid of using tmpfs for `podman --root` ?

I think we have two options here.  Either we use a location that's unique to each run of the test suite (like `BATS_RUN_TMPDIR`), or we use `trap` to catch signals and clean everything up.  Ideally, we would do both.

test/system/libs/helpers.bash
```diff
@@ -130,6 +137,99 @@ function _clean_cached_images() {
...
+    -v "${DOCKER_REG_CERTS_DIR}":/certs \
+    -e REGISTRY_HTTP_ADDR=0.0.0.0:443 \
+    -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt \
+    -e REGISTRY_HTTP_TLS_KEY=/certs/domain.key \
+    -p 50000:443 \
+    "${IMAGES[docker-reg]}"
+  assert_success
+
+  run $PODMAN login \
+    --authfile ${TEMP_BASE_DIR}/authfile.json \
```
debarshiray, line 202: `TEMP_BASE_DIR` is `$XDG_CACHE_HOME`/toolbox.  So, we are putting this file in `~/.cache/toolbox`.  That's wrong for a test suite, even if we override `XDG_CACHE_HOME` to isolate the test suite from the running OS.  The Toolbx source code can be using `~/.cache/toolbox` for some other purpose and end up fighting with the test suite.

Why don't we put it in `BATS_RUN_TMPDIR`?

debarshiray (commented):
Recently, while hacking on Toolbx, I noticed that if I forcibly abort the tests with a `ctrl+c` then the next run of the test suite has a bunch of failures related to our Docker registry.

test/system/101-create.bats
```diff
@@ -117,3 +117,37 @@ teardown() {
...
+
+  assert_failure
+  assert_output "Error: file $file not found"
+}
+
+@test "create: Create a container based on an image from locked registry using an authentication file" {
+  local authfile="$BATS_RUN_TMPDIR/authfile"
+  local image="fedora-toolbox:32"
+
+  run $PODMAN login --authfile "$authfile" --username user --password user "$DOCKER_REG_URI"
```
debarshiray, line 134: It will be good to add another test that uses the default location used by `podman login` without making use of the `--authfile` option.  It will be give us added confidence that things are working as expected.

src/cmd/create.go
```diff
@@ -717,7 +729,7 @@ func pullImage(image, release string) (bool, error) {
 		defer s.Stop()
 	}
 
-	if err := podman.Pull(imageFull); err != nil {
+	if err := podman.Pull(imageFull, createFlags.authFile); err != nil {
```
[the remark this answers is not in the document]
debarshiray: Changed in https://github.com/containers/toolbox/pull/1240

PR #936, opened 2021-11-25
profile.d: Fix the PS1 on Z shell

debarshiray:
Currently, Z shell renders the PS1 like this:
```
\[\]⬢\[\][\u@\h \W]\$
```

Notice that it doesn't like that the terminal escape sequences for the
foreground colour are wrapped in `\[` and `\]` [1], and doesn't
understand the special characters like `\u` and `\h`.

This is fixed by making the PS1 specific to the shell. The prompt for
Z shell is based on the default prompt used on Fedora, just like the
one for Bash.

[1] Commit bc1a816ea3b80fc7
    https://github.com/debarshiray/toolbox/issues/190


HarryMichal:
recheck

PR #938, opened 2021-11-29
images/fedora: Remove mlocate

travier:
Remove mlocate from the base image. mlocate being installed by default
will trigger disk wide file listing scanning updates after each dnf
operation via the updatedb process.

This is probably not wanted in most cases and especially for ephemeral
containers usage where the scanning will happen every time and thus be
very IO intensive.

### DIFF
--- a/images/fedora/f34/extra-packages
+++ b/images/fedora/f34/extra-packages
@@ -20,7 +20,6 @@ less
 lsof
 man-db
 man-pages
-mlocate
 mtr
 nano-default-editor
 nss-mdns
--- a/images/fedora/f35/extra-packages
+++ b/images/fedora/f35/extra-packages
@@ -20,7 +20,6 @@ less
 lsof
 man-db
 man-pages
-mlocate
 mtr
 nano-default-editor
 nss-mdns
--- a/images/fedora/f36/extra-packages
+++ b/images/fedora/f36/extra-packages
@@ -20,7 +20,6 @@ less
 lsof
 man-db
 man-pages
-mlocate
 mtr
 nano-default-editor
 nss-mdns


debarshiray:
( Some past discussion at https://github.com/containers/toolbox/issues/391 )

> mlocate being installed by default
> will trigger disk wide file listing scanning updates after each dnf
> operation via the updatedb process.

Umm... do you know exactly how the disk wide file scanning is getting triggered?

For a Toolbox container, we should have more control over the `updatedb` process because the container's entry point runs it directly once a day.

Also, Fedora's [updatedb implementation](https://pagure.io/mlocate) is supposed to be able to incrementally update the database, but I haven't checked how effective it is.

> This is probably not wanted in most cases and especially for ephemeral
> containers usage where the scanning will happen every time and thus be
> very IO intensive.

Toolbox containers aren't meant to be completely ephemeral, though.  We expect many CLI users, such as developers, to live inside a toolbox container.

Maybe we can make `updatedb(8)` less agressive without completely removing it?

If `mlocate` is no longer considered desirable, maybe we should try to get it removed from the default Fedora Silverblue/Workstation installation?

travier:
> ( Some past discussion at #391 )

Thanks! Sorry I missed that.

> Umm... do you know exactly how the disk wide file scanning is getting triggered?
> 
> For a Toolbox container, we should have more control over the `updatedb` process because the container's entry point runs it directly once a day.

I don't know yet exactly what triggers this but it looks like it's either on a toolbox container startup (looks like https://github.com/containers/toolbox/blob/main/src/cmd/initContainer.go#L576) or right after a dnf operation/update.

> Also, Fedora's [updatedb implementation](https://pagure.io/mlocate) is supposed to be able to incrementally update the database, but I haven't checked how effective it is.
> 
> Toolbox containers aren't meant to be completely ephemeral, though. We expect many CLI users, such as developers, to live inside a toolbox container.

Yes, not all my toolbox containers are ephemeral but I have some that I re-create frequently and this re-triggers every-time. There only need to be one for that to be an issue unfortunately.

> Maybe we can make `updatedb(8)` less agressive without completely removing it?

We could run updatedb unprivileged (under the toolbox user instead of root inside the container) and store the database inside the home directory but that could potentially lead to issues with database format if shared between multiple versions of mlocate.

> If `mlocate` is no longer considered desirable, maybe we should try to get it removed from the default Fedora Silverblue/Workstation installation?

For the classic Workstation use case, it's unlikely that the db will get removed as frequently thus it's not really an issue to pay the IO cost once only.

For Silverblue, we might want to have mlocate in the base image and run both as root & as a user level service too to scan only the interesting content relevant for each case.

Another option would be to have a per toolbox option / property to disable updatedb launch from `toolbox init-container`.

cgwalters:
I think mlocate should be opt-in by default.  It also doesn't make much sense to have potentially multiple containers scanning the host filesystem.

debarshiray:
> I don't know yet exactly what triggers this but it looks like it's either
> on a toolbox container startup

Yes, the Toolbox container's entry point definitely invokes `updatedb(8)` at start-up.

travier:
After thinking about this for a while, I think it makes more sense to have mlocate as part of the Silverblue base image instead of having each instance of a container redundantly update its own database. In this setup, the container mlocate would have to look at the host db for file lookups. There is also the general question about what to index in a setup where the host and containers don't have the same view of the filesystem.

debarshiray:
Thinking about it some more ...

I suppose it's reasonable to wonder if `mlocate` should be opt-in or opt-out.  I don't have any strong opinion on it beyond *whatever people might have come to expect*.  I guess it makes sense to have it be part of the Silverblue image because it's not easy to get more RPMs added to it, but maybe it should be disabled by default?

Another option could be to delay the initial invocation of `updatedb(8)` so that it never gets run for short-lived containers.

Anyway, the current Toolbox code should already do the right thing if `updatedb(8)` is absent, so there's no problem there.

I wonder if it would be practical to bind mount the host's database into the container.  Does the database format differ across implementations.

cgwalters:
This behavior seems to have come from https://github.com/containers/toolbox/commit/e1635c06f3d2038097c875b7b1c293c7bc40bb1b

I personally don't use the default toolbox base images, so I don't have mlocate and never noticed this.

I know the initial focus of toolbx was the developer pet container case, but since we are also using it for the "root system debugging" case, I think we should do as little as possible in the container entrypoint by default and maximize reliability.  If I'm a system administrator and hop on a node to debug a performance issue, the last thing I want is to suddenly as a side effect start having `updatedb` run as a non-obvious side effect of starting my debugging container.

Even for the unpriv workstation scenario, I think it's better to instead encourage people to call out via e.g. `systemd-run mlocate` back to the host context.  While clearly a goal is to be able to live inside a container for *most* things, that will never be entirely true, and it's hard to avoid some level of having users understand the duality.

debarshiray:
Merged, after tweaking the commit message to make the problem clearer.

travier:
LGTM. Thanks!

PR #939, opened 2021-11-30
CONTRIBUTING: Content update & fixes

HarryMichal:


CONTRIBUTING.md
```diff
@@ -1,44 +1,26 @@
-![Toolbox logo](data/logo/toolbox-logo-landscape.svg)
+![Toolbox logo](data/gfx/powerup.gif)
```
debarshiray, line 1: Is @jimmac OK with using this animation here?

CONTRIBUTING.md
```diff
@@ -1,44 +1,26 @@
...
 - See our [documentation](https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/)
   if there are some steps that could help you solve your issue
 - Sometimes a bug is not reported in our bug tracker but instead people ask for
-  help somewhere else: IRC ([Freenode](https://freenode.net) - #silverblue,
-  #containers, #fedora, #fedora-devel,..), [Fedora forum](https://discussion.fedoraproject.org/tag/toolbox),
-  or somewhere else. In such cases we'd like you to still report the bug and
+  help somewhere else (e.g., chat channels). In such cases we'd like you to still report the bug and
   share with us any info that could be gathered from those places
 
 ## Writing a Bug Report
```
debarshiray, line 26: I'd suggest removing the *writing a bug report* section.  It should be covered by our GitHub issue template.

Same for the *making suggestions* section. Also, because the distinction between bugs and suggestions might not be clear, and GitHub calls all of them *issues* anyway.

CONTRIBUTING.md
```diff
@@ -1,44 +1,26 @@
...
   - If the issue is already reported and is marked as **OPEN**, comment on it
     and if possible and needed, share info about the issue just as if you were
-    submiting a new issue
+    submitting a new issue
   - If the issue is marked as **CLOSED**, check if your version of Toolbox is
     up-to-date or if there are some steps, described in the closed issue, that
     you should follow. If you are still experiencing the issue, please file a
     new issue
 - See our [documentation](https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/)
   if there are some steps that could help you solve your issue
```
debarshiray, line 21: We trying to consolidate our documentation around our [website](https://containertoolbx.org/), our manuals, and this code repository.  Linking to another documentation source goes against that.  I'd suggest linking to the [top-level manual](https://github.com/containers/toolbox/blob/main/doc/toolbox.1.md)

CONTRIBUTING.md
```diff
@@ -1,44 +1,26 @@
...
-- [First Contribution](#first-contribution)
-- [Pull Requests](#pull-requests)
-  - [Creating a Pull Request](#creating-a-pull-request)
-  - [After Creating a Pull Request](#after-creating-a-pull-request)
-- [Little Style Guide](#little-style-guide)
-
 # Reporting Bugs
 
-## Before Submiting a Bug Report
+## Before Submitting a Bug Report
```
debarshiray, line 10: I'd suggest using the word *issue* instead of *bug report* because that's the terminology GitHub uses.

debarshiray (commented):
Thanks for the updates!

CONTRIBUTING.md
HarryMichal: Didn't ask since the animation is already in the repo. @jimmac?

jimmac: I wouldn't overdo it. Maybe a static illustration for contributions.

CONTRIBUTING.md
HarryMichal: > I'd suggest removing the writing a bug report section. It should be covered by our GitHub issue template.

That's a valid point. I'll look at it going forward.

> Same for the making suggestions section. Also, because the distinction between bugs and suggestions might not be clear, and GitHub calls all of them issues anyway.

Yes, but the content is largely different. I could merge it with the issue templates, too. Also, GitHub has this Discussions feature that serves for this kind of purpose. It just has to be enabled in the repo.

CONTRIBUTING.md
HarryMichal: Good point. Adding to my backlog.

CONTRIBUTING.md
HarryMichal: Sure.

CONTRIBUTING.md
HarryMichal: I don't follow. Is it a problem to use the illustration or is it a problem that it is an animation?

jimmac: Something along this:
Toolbx_Contributors

HarryMichal: Oh, you want to create a new illustration? I mean... sure :). I'd be more than happy to reuse the work you've already done, though.

Btw, lefthanded is good :).

CONTRIBUTING.md
debarshiray: >> Same for the making suggestions section. Also, because the
>> distinction between bugs and suggestions might not be clear,
>> and GitHub calls all of them issues anyway.
>
> Yes, but the content is largely different. I could merge it with the
> issue templates, too.

I am mainly trying to keep the top-level `CONTRIBUTING.md` reasonably short.  Currently it's more than two vertical screen lengths, which seems to be a bit too long.  We can keep the length manageable by gradually leading the reader into more specific and detailed pages, instead of exposing everything at once.  In this case, that would be the issue template.

Also, the distinction between a bug, a feature request and a suggestion can be blurry.  We will never get it perfectly right.  So it's a trade-off -- too much text and people won't bother reading it, too little and it won't be sufficient.

> Also, GitHub has this Discussions feature
> that serves for this kind of purpose. It just has to be enabled in the repo.

Do we want even more communication channels? :)

HarryMichal: > Currently it's more than two vertical screen lengths, which seems to be a bit too long.

Agreed.

> In this case, that would be the issue template.

Agreed.

> Do we want even more communication channels? :)

If it makes the distinction between bugs, feature requests and suggestions less blurry, then yes.

debarshiray: >>    Do we want even more communication channels? :)
>
> If it makes the distinction between bugs, feature requests and
> suggestions less blurry, then yes.

It won't. :)

It will become yet another venue that we'd have to monitor, and it won't stop people from misfiling bugs as suggestions, suggestions as bugs, etc..  People already file pull requests to initiate discussions, which is confusing enough, and we already have too many communication channels at this point -- GitHub issues and PRs; Matrix; email (yes, I get bugs via email); Fedora's Discourse, mailing lists, IRC; etc..

I am afraid that I will have to start ignoring some if we keep accumulating more and more.

debarshiray: Oh, I forgot about Twitter!

HarryMichal: Fair enough!

Issue #885, opened 2021-09-25
Incorrectly created containers with some custom images


lopukhov:
**Describe the bug**

When testing different images to create containers with toolbox I have found that with some base images the start command of the container is not set correctly. When trying to enter to one of this containers the toolbox executable inside the container is not called.

I created an image based on `parrotsec/core`. When entering the toolbox the following error appears.

```
$ toolbox enter parrot
Error: invalid entry point PID of container parrot
```

When starting directly from podman, the output is the bash help output.

```
$ podman start --attach parrot
bash: --log-level: invalid option
Usage:	bash [GNU long option] [option] ...
	bash [GNU long option] [option] script-file ...
GNU long options:
	--debug
	--debugger
	--dump-po-strings
	--dump-strings
	--help
	--init-file
	--login
	--noediting
	--noprofile
	--norc
	--posix
	--pretty-print
	--rcfile
	--restricted
	--verbose
	--version
Shell options:
	-ilrsD or -c command or -O shopt_option		(invocation only)
	-abefhkmnptuvxBCHP or -o option
```

After inspecting the container, and comparing to another container with a working image, there is a difference in the start path.

```
$ podman --log-level debug inspect --format json --type container <container>
```

Not working container looks like:

```json
{
  "Path": "/bin/sh",
  "Args": [
      "-c",
      "bash $@",
      "toolbox",
      "--log-level",
      "debug",
      "init-container",
      "--gid",
      "1000",
      "--home",
      "/home/lopukhv",
      "--shell",
      "/bin/bash",
      "--uid",
      "1000",
      "--user",
      "lopukhv",
      "--monitor-host"
  ]
}
```

Working container look like:

```json
{
  "Path": "toolbox",
  "Args": [
      "--log-level",
      "debug",
      "init-container",
      "--gid",
      "1000",
      "--home",
      "/home/lopukhv",
      "--shell",
      "/bin/bash",
      "--uid",
      "1000",
      "--user",
      "lopukhv",
      "--monitor-host"
  ],
}
```

**Steps how to reproduce the behaviour**

1. Build local image with provided Containerfile (`podman build -t parrot .`)
2. Create a toolbox (`toolbox create -i localhost/parrot parrot`)
3. Enter toolbox (`toolbox enter parrot`)
4. Start container with podman (`podman start --attach parrot`)

**Expected behaviour**
Container should be created with the appropriate start command and should start normally.

**Actual behaviour**
[...]

PR #940, opened 2021-11-30
images, README.md: Remove misleading and redundant CMD and update the requirements around ENTRYPOINT

debarshiray:
https://github.com/containers/toolbox/issues/885


debarshiray:
recheck

PR #942, opened 2021-12-01
build: Drop the use of patchelf(1)

debarshiray:
Some downstream distributors like RHEL don't have patchelf(1). Relying
on patchelf(1) during the build will make it difficult for such
downstreams to distribute Toolbox.

Fortunately, the path of the dynamic linker (ie., PT_INTERP) is
hardcoded in the ABI specification of each architecture [1]. This means
that Toolbox's build system can keep it's own architecture to dynamic
linker mapping, and specify it during the build through the GNU ld
linker's --dynamic-linker flag, as opposed to using a tool like
patchelf(1) to change the path of the dynamic linker in the built
binary to the one inside /run/host. Currently, the list of
architectures covers the ones that Fedora builds for.

[1] https://sourceware.org/glibc/wiki/ABIList

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -3,7 +3,7 @@ project(
   'c',
   version: '0.0.99.2',
   license: 'ASL 2.0',
-  meson_version: '>= 0.53.0',
+  meson_version: '>= 0.56.0',
 )
 
 cc = meson.get_compiler('c')
@@ -21,13 +21,13 @@ profiledir = get_option('profile_dir')
 tmpfilesdir = get_option('tmpfiles_dir')
 if tmpfilesdir == ''
   systemd_dep = dependency('systemd')
-  tmpfilesdir = systemd_dep.get_pkgconfig_variable('tmpfilesdir')
+  tmpfilesdir = systemd_dep.get_variable(pkgconfig: 'tmpfilesdir')
 endif
 
 if bash_completion.found()
   install_data(
     'completion/bash/toolbox',
-    install_dir: bash_completion.get_pkgconfig_variable('completionsdir')
+    install_dir: bash_completion.get_variable(pkgconfig: 'completionsdir')
   )
 endif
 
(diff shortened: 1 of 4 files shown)


debarshiray:
We have a bootstrapping problem here.

The build fails because `go build -ldflags "-extldflags '-Wl,-dynamic-path,...'" specifies a dynamic linker under `/run/host`, which doesn't exist during the build because `/run/host` created by the build.

debarshiray:
> We have a bootstrapping problem here.
> 
> The build fails because `go build -ldflags "-extldflags '-Wl,-dynamic-path,...'"`
> specifies a dynamic linker under `/run/host`, which doesn't exist during the
> build because `/run/host` is created by the build.

Turns out that the linker flag is `-dynamic-linker`, not `-dynamic-path`.  With the correctly spelled flag, the toolchain doesn't bother checking if the path to the dynamic linker actually exists, and the build proceeds as expected. Phew!

We still need to double-check if the path to the dynamic linker is specified correctly for each architecture.

HarryMichal:
> > We have a bootstrapping problem here.
> > The build fails because `go build -ldflags "-extldflags '-Wl,-dynamic-path,...'"`
> > specifies a dynamic linker under `/run/host`, which doesn't exist during the
> > build because `/run/host` is created by the build.
> 
> Turns out that the linker flag is `-dynamic-linker`, not `-dynamic-path`. With the correctly spelled flag, the toolchain doesn't bother checking if the path to the dynamic linker actually exists, and the build proceeds as expected. Phew!
> 
> We still need to double-check if the path to the dynamic linker is specified correctly for each architecture.

But for the final binary to be operational the `/run/host` symlink to `/` has to exist, right? For that reason the build also has to adjusted as in #840 the created binary is used to generate the completion files.

debarshiray:
> But for the final binary to be operational the /run/host symlink to / has to
> exist, right?

Yes.

> For that reason the build also has to adjusted as in #840 the created
> binary is used to generate the completion files.

I haven't looked at https://github.com/containers/toolbox/pull/840 - we'll see when we get to that. Currently, we urgently need to roll a release, and we can't do that as long as we require `patchelf(1)` for the build.

PR #943, opened 2021-12-01
pkg/shell: Simplify code flow

cgwalters:
We only need to check against `err != nil` once, not twice.


ashcrow (approved):

HarryMichal:
Thanks for the PR, @cgwalters!

PR #944, opened 2021-12-01
shell: Add and use `RunGetOutput`

cgwalters:
This inherits https://github.com/containers/toolbox/pull/943
and is very similar to
https://github.com/openshift/machine-config-operator/pull/2587

Basically, it's useful at an API level to more clearly distinguish
between the "run a command for its side effect" versus
"run a command to capture stdout".

The *vast* majority of `Run()` cases pass `nil, nil, nil` which
is visual noise.  After this lands, we can do a pass which changes
its API signature to drop those.

The immediate motivation is fixing `Run()` to capture `stderr`
into the returned `error` buffer, which in turn would help me debug
why `usermod` is failing in this centos8 container.  Throwing
away the stderr from subprocesses makes debugging unnecessarily hard.


travier (approved):

HarryMichal:
I like the idea! Could you write some test cases for the added function in the `shell` package just for the sake of sanity :)? When that's done I'm happy to merge this. Thanks!

src/pkg/shell/shell.go
```diff
@@ -39,6 +40,20 @@ func Run(name string, stdin io.Reader, stdout, stderr io.Writer, arg ...string)
...
+// RunGetOutput spawns a command, capturing its stdout.
+func RunGetOutput(name string, arg ...string) ([]byte, error) {
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
+	exitCode, err := RunWithExitCode(name, nil, &stdout, &stderr, arg...)
+	if err != nil {
+		return nil, err
+	}
+	if exitCode != 0 {
+		return nil, fmt.Errorf("failed to invoke %s: %s", name, string(stderr.Bytes()))
```
HarryMichal, line 52: What about `err`? Does it hold exactly the same information like `stderr`?

HarryMichal: Mmm... nevermind, it is being handled in the previous if block.

Another concern could be multi-line output in `stderr`. Maybe print it with `logrus`?

cgwalters: > What about err? Does it hold exactly the same information like stderr?

No, `err` is not set here.  The command executed, but returned an error exit code.

> Another concern could be multi-line output in stderr. Maybe print it with logrus?

Ultimately, this API is for cases where we're just acting as a thin wrapper around the underlying CLI.  In the end, all we do with this API is dump the resulting text to stderr anyways, so multi-line output is fine.

debarshiray:
This one fell through the cracks.  :(

/me puts it on his list of PRs to review

Issue #941, opened 2021-12-01
Test the image definitions


debarshiray:
We should also run the tests on a set of images that were built on-the-fly with `podman build ...` from the in-tree image definitions, along with the images pulled from the registries.  This will assure us that are image definitions are correct.

We never know when some base image changes asynchronously behind our backs, or if some code change has some strange side-effect that nobody thought of, etc..

PR #947, opened 2021-12-03
playbooks: Install Go dependencies using dnf

HarryMichal:
Installing Go dependencies using the "Go approach" results in a slightly
different binary than when installing Fedora-hosted versions of the
packages since Fedora does not strictly follow the package versions
defined in the go.mod file. With this regressions in Fedora connected to
package updates should be caught.

### DIFF
--- a/playbooks/setup-env.yaml
+++ b/playbooks/setup-env.yaml
@@ -11,7 +11,19 @@
           - bats
           - flatpak-session-helper
           - golang
+          - golang-github-acobaugh-osrelease-devel
+          - golang-github-briandowns-spinner-devel
           - golang-github-cpuguy83-md2man
+          - golang-github-docker-units-devel
+          - golang-github-fsnotify-devel
+          - golang-github-godbus-dbus-devel
+          - golang-github-harrymichal-version-devel
+          - golang-github-mattn-isatty-devel
+          - golang-github-sirupsen-logrus-devel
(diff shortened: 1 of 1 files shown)


HarryMichal:
recheck

HarryMichal:
Hmm.. `go build` doesn't seem to pick up the installed packages. Needs work.

debarshiray (commented):
Umm... I am not sure about this.  The build and the build dependencies will be different on every OS distribution that's out there.  We can't possibly cover all of them in the upstream CI.

I think this is something that should be covered by downstream integration tests.

HarryMichal:
> I think this is something that should be covered by downstream integration tests.

Fair enough. Closing.

debarshiray:
I was thinking about this again today in the context of https://github.com/containers/toolbox/issues/941

I think the differentiating factor is that we are directly in charge of the `toolbox(1)` binary and the images defined in this code repository.  So we need to make sure that they work together.  The images are derived from the base images which don't have a well-defined ABI, which makes it necessary to keep a constant eye on them. Plus, we are (ab)using the OCI stack for interactive pet containers, which is not what most people do.

On the other hand, our Go dependencies are well-defined APIs.  So, it's easier to spot breakages, define tests for them in their own repositories, and chances are that we won't be the only ones to notice them.

PR #948, opened 2021-12-04
zuul: Run tests only for relevant files

HarryMichal:
Not all file are equal when it comes to testing. Unit tests are related
strictly to the source code and documentation changes do not concern it.
System tests have a wider range of influence but documentation and some
other areas also do not concern them.

I'm unsure about the effect of this change on the periodic pipeline
execution.

Split out from https://github.com/containers/toolbox/pull/918

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -3,6 +3,7 @@
     name: unit-test
     description: Run Toolbox's unit tests declared in Meson
     timeout: 300
+    files: ['playbooks/*', 'src/*', 'meson.build', 'meson_options.txt', '.zuul.yaml']
     nodeset:
       nodes:
         - name: ci-node-34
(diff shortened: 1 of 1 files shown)


.zuul.yaml
```diff
@@ -14,6 +15,7 @@
     name: system-test-fedora-34
     description: Run Toolbox's system tests in Fedora 34
     timeout: 1200
+    files: &system_test_files ['data/*', 'playbooks/*', 'profile.d/*', 'src/*', 'meson.build', 'meson_options.txt', '.zuul.yaml']
```
debarshiray, line 18: Please forgive my ignorance, but what does `&system_test_files` mean? Is this the syntax for defining it as a variable? In that case, wouldn't it be good to define it as part of the `system-test-fedora-rawhide` or globally, so that it doesn't get lost by mistake when Fedora 34 goes EOL?

debarshiray (commented):
Thanks @HarryMichal

.zuul.yaml
HarryMichal: This is so-called YAML anchor. I was advised by Zuul developers to use it. I don't think a global "variable" can be defined in YAML. This is a kind of a middle-ground.

debarshiray: Ok, I see. :)

What do you think about defining the anchor as part of the Rawhide tests?

HarryMichal: Sounds sensible.

debarshiray: Okay! I am trying that in https://github.com/containers/toolbox/pull/971

PR #949, opened 2021-12-08
README.md: Shorten it by linking to the website wherever possible

debarshiray:


HarryMichal:
Do we want remove most of the content? I think at least build instructions should stay in the README.

debarshiray:
> Do we want remove most of the content? I think at least build instructions
> should stay in the README.

I am trying to free up space in the code repository's `README.md` for documentation content that's still not linked from it. eg., `CONTRIBUTION.md`, `GOALS.md` and `SECURITY.md`. Then there are the manuals.

Generally, it seems to me that it's good to keep the code repository's `README.md` short. eg., see [Flatpak](https://github.com/flatpak/flatpak) and [systemd](https://github.com/systemd/systemd/). Readers already have to scroll past the contents of the Git repository to get to it. Longer the `README.md` gets, the less likely they would be to keep scrolling to read the whole of it; and things like build instructions do take up space.

Now that we have a website, we need to decide which part of the content goes where, and keep duplication to a minimum. Note that there's some content (eg., `CONTRIBUTING.md` and `SECURITY.md`) that does need to stay in the code repository to satisfy GitHub's requirements.

So, I see two options for the build instructions. We either link to a separate file in the code repository or one on the website. I prefer the website because:

* If someone is looking at the code, then chances are that they will soon notice the `meson.build` and know how to proceed with the build. That's the advantage of using a well-known build system, and Meson is a de-facto standard when it comes to building low-level components of a Linux OS -- everybody knows how to use it.

If for some strange reason, they can't figure it out, then they will end up reading the `README.md`, and it doesn't matter that much whether they get sent to another page in the code repository or to the website.

* The website is a more polished way to present the content. It's aimed at those who might not know or have time to read through the `meson.build`. eg., downstream distributors, commercial customers, journalists, etc..

* It's easier to delegate edits to the website to non-technical folks.

PR #951, opened 2021-12-09
Make the migration path for coreos/toolbox users optional

debarshiray:
Commit 6c86cabbe5da6e542b changed the command line interface to behave
a lot similar to that of github.com/coreos/toolbox, which makes things
easier for those switching over from it.

However, it makes things confusing for the vast majority of users who
have never used coreos/toolbox. The Toolbox CLI aims to be friendly to
new users by being self-documenting and offering a smooth onboarding
experience. It's jarring to new users when 'toolbox', without any
commands specified, suggests that it needs to perform a big download.

It's difficult to document two different sets of CLIs, and if the
manuals don't mention the second behaviour, then it just leaves the
users even more confused.

Hence, it will be good to keep the migration path for coreos/toolbox
behind a build-time option, so that only those OS distributors who
truly need it may enable it without impacting others. Fortunately,
coreos/toolbox doesn't have any manuals, which means that there's no
need to conditionalize the documentation.


debarshiray:
I don't know what to do with the tests for `coreos/toolbox` compatibility in `test/system/100-root.bats`.  Ideally we would run the test suite twice with the migration path turned off and on.  However that would require a more intrusive surgery of the test suite, and might not be that useful since the migration path is meant for a small subset of users.

@HarryMichal any thoughts?

debarshiray:
> I don't know what to do with the tests for `coreos/toolbox` compatibility
> in `test/system/100-root.bats`. Ideally we would run the test suite twice
> with the migration path turned off and on. However that would require a
> more intrusive surgery of the test suite, and might not be that useful
> since the migration path is meant for a small subset of users.

In the end, I decided to (temporarily?) remove `test/system/100-root.bats`.

It's important to test the default build configuration that the vast majority of users would be using, and it doesn't seem worth the hassle to block the 0.0.99.3 release while we figure out how to test the non-default build.

PR #958, opened 2021-12-15
CONTRIBUTING: custom graphic

jimmac:
- can't have right aligned images or inline css,
  better to just have them on the left.


HarryMichal:
Thanks @jimmac!

debarshiray:
Like I was saying in #toolbx:matrix.org ...

I didn't quite understand what you meant by *can't have right aligned images or inline CSS* since the previous animations seemed to be working fine, but specifically the left-aligned `README.gif` looks a bit odd because of the way the character is facing.  Wouldn't a laterally inverted version look better?

debarshiray:
> Wouldn't a laterally inverted version look better?

Inverted in https://github.com/containers/toolbox/pull/985

Issue #955, opened 2021-12-13
v0.0.99.3 builds but fails to execute on Ubuntu 20.04


antonc42:
**Describe the bug**
On Ubuntu 20.04, toolbox version 0.0.99.3 builds and installs but fails to execute with:
```
Failed to execute process '/usr/local/bin/toolbox'. Reason:
The file '/usr/local/bin/toolbox' does not exist or could not be executed.
```
File exists:
```
ll /usr/local/bin/toolbox
-rwxr-xr-x 1 root root 8.3M Dec 13 10:15 /usr/local/bin/toolbox
```
The interpreter for the binary seems to be incorrect
```
file /usr/local/bin/toolbox
/usr/local/bin/toolbox: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /run/host/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2, BuildID[sha1]=51a9844e89fe0d1aea45dd776804f298516fad68, for GNU/Linux 3.2.0, not stripped
```
The interpreter file doesn't exist on my system:
```
ll /run/host/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
ls: cannot access '/run/host/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2': No such file or directory
```

**Steps how to reproduce the behaviour**
1. Download https://github.com/containers/toolbox/releases/download/0.0.99.3/toolbox-0.0.99.3.tar.xz
2. Extract
3. Enter into extracted directory
4. Activate virtualenv
5. `meson setup -Dprofile_dir=/etc/profile.d builddir` [meson-setup.log](https://github.com/containers/toolbox/files/7705058/meson-setup.log)
6. `meson compile -C builddir` [meson-compile.log](https://github.com/containers/toolbox/files/7705057/meson-compile.log)
7. `sudo meson install -C builddir` [meson-install.log](https://github.com/containers/toolbox/files/7705056/meson-install.log)
8. `toolbox --help`

**Expected behaviour**
Toolbox builds, installs, and runs correctly.

**Actual behaviour**
Toolbox builds and installs correctly but does not run.
```
toolbox --help
Failed to execute process '/usr/local/bin/toolbox'. Reason:
The file '/usr/local/bin/toolbox' does not exist or could not be executed.
```

**Output of `toolbox --version` (v0.0.90+)**
```
toolbox --version
[...]

PR #965, opened 2021-12-17
build: Call 'systemd-tmpfiles --create' when installing

debarshiray:
It's only necessary to call 'systemd-tmpfiles --create' when building
and installing from source, and not when using a downstream package,
because:

  * When 'meson install' is called as part of building the package,
    that's not when the temporary files need to be created. They need
    to be created when the package is later downloaded and installed
    by the user.

  * Downstream tools can sometimes handle it automatically. eg., on
    Fedora, the systemd RPM installs a trigger that tells RPM to call
    'systemd-tmpfiles --create' automatically when a tmpfiles.d snippet
    is installed.

https://github.com/containers/toolbox/issues/955

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -6,6 +6,8 @@ project(
   meson_version: '>= 0.58.0',
 )
 
+fs = import('fs')
+
 cc = meson.get_compiler('c')
 if not cc.has_argument('-print-file-name=libc.so')
   error('C compiler does not support the -print-file-name argument.')
@@ -22,9 +24,12 @@ migration_path_for_coreos_toolbox = get_option('migration_path_for_coreos_toolbo
 profiledir = get_option('profile_dir')
 
 tmpfilesdir = get_option('tmpfiles_dir')
-if tmpfilesdir == ''
+if tmpfilesdir == '' or not fs.exists('/run/.containerenv')
   systemd_dep = dependency('systemd')
-  tmpfilesdir = systemd_dep.get_variable(pkgconfig: 'tmpfilesdir')
+
+  if tmpfilesdir == ''
+    tmpfilesdir = systemd_dep.get_variable(pkgconfig: 'tmpfilesdir')
+  endif
 endif
 
 if bash_completion.found()
@@ -65,3 +70,5 @@ subdir('data')
 subdir('doc')
 subdir('profile.d')
 subdir('src')
+
+meson.add_install_script('meson_post_install.py')
--- a/meson_post_install.py
+++ b/meson_post_install.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+import sys
+
+destdir = os.environ.get('DESTDIR', '')
+
+if not destdir and not os.path.exists('/run/.containerenv'):
+    print('Calling systemd-tmpfiles --create ...')
+
+    try:
+        subprocess.run(['systemd-tmpfiles', '--create'], check=True)
+    except subprocess.CalledProcessError as e:
+        print('Returned non-zero exit status', e.returncode)
+        sys.exit(e.returncode)
+
+sys.exit(0)
--- a/playbooks/setup-env.yaml
+++ b/playbooks/setup-env.yaml
@@ -52,10 +52,3 @@
       args:
         chdir: '{{ zuul.project.src_dir }}'
         creates: /usr/local/bin/toolbox
-
-    - name: Setup environment
-      become: yes
-      command:
-        cmd: systemd-tmpfiles --create
-        creates: /run/media
-        creates: /run/host


meson.build
```diff
@@ -65,3 +65,5 @@ subdir('data')
 subdir('doc')
 subdir('profile.d')
 subdir('src')
+
+meson.add_install_script('meson_post_install.py')
```
HarryMichal, line 69: `meson.add_install_script()` already has `DESTDIR` argument.

debarshiray: You mean [skip_if_destdir](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonadd_install_script) ?  I thought about using it, but then thought that if the script evolves more, then we might not be able to use it anymore.

Up to you.

debarshiray:
recheck

debarshiray:
recheck

HarryMichal:
Did you try running this locally? On my machine (when ran inside of a toolbx) this often results in:

```
Running custom install script '/usr/bin/systemd-tmpfiles --create'
--- stdout ---

--- stderr ---
fchownat() of /var/lib/systemd/coredump failed: Read-only file system
fchownat() of /tmp failed: Operation not permitted
fchownat() of /run/media failed: Operation not permitted
Setting access ACL "u::rwx,g::r-x,g:adm:r-x,g:wheel:r-x,g:4294967295:r-x,g:4294967295:r-x,m::r-x,o::r-x" on /var/log/journal failed: Read-only file system
Failed to re-open '/var/log/journal': Operation not permitted
fchownat() of /var/log/journal failed: Read-only file system
Setting access ACL "u::rwx,g::r-x,g:adm:r-x,g:wheel:r-x,g:4294967295:r-x,g:4294967295:r-x,m::r-x,o::r-x" on /var/log/journal/64ea8f1e4b534013843813476c178ab7 failed: Read-only file system
Failed to re-open '/var/log/journal/64ea8f1e4b534013843813476c178ab7': Operation not permitted
fchownat() of /var/log/journal/64ea8f1e4b534013843813476c178ab7 failed: Read-only file system
fchownat() of /dev/snd/seq failed: Operation not permitted
fchownat() of /dev/snd/timer failed: Operation not permitted
fchownat() of /dev/loop-control failed: Operation not permitted
fchownat() of /dev/kvm failed: Operation not permitted
fchownat() of /dev/vhost-net failed: Operation not permitted
fchownat() of /dev/vhost-vsock failed: Operation not permitted
Setting access ACL "u::rw-,g::r-x,g:adm:r--,g:wheel:r--,g:4294967295:r--,g:4294967295:r--,m::r--,o::---" on /var/log/journal/64ea8f1e4b534013843813476c178ab7/system.journal failed: Read-only file system
fchownat() of /var/log/journal/64ea8f1e4b534013843813476c178ab7/system.journal failed: Read-only file system

FAILED: install script '/usr/bin/systemd-tmpfiles --create' exit code 73, stopped
FAILED: meson-install
/usr/bin/meson install --no-rebuild
ninja: build stopped: subcommand failed.
```

The goal of this step is to create `/run/host` and not to create other tmpfiles. A way to solve this is to create only this file. I did it with this snipper:

```
meson.add_install_script(
  systemd_tmpfiles,
  '--create', '--prefix=/run/host', tmpfiles_toolbox_conf,
  skip_if_destdir: true,
)
```

meson.build
eli-schwartz: When using skip_if_destdir, meson internally detects that it doesn't need to fork and run the install script.

So you may still be interested in hoisting that check out of the script and into meson's installdata.

Either way it is possible to later on replace `meson.add_install_script('systemd-tmpfiles', skip_if_destdir: true)` with `meson.add_install_script('meson_post_install.py', skip_if_destdir: true)`

debarshiray:
> Did you try running this locally? On my machine (when ran inside of a toolbx)
> this often results in:

I only tried the build with `DESTDIR` locally on a Fedora Workstation host. I didn't try it without `DESTDIR` inside a Toolbox container. Thanks for testing it out.

By the way, this makes me think that we need a test that tries to build Toolbox from source inside a Toolbox container.

HarryMichal:
> By the way, this makes me think that we need a test that tries to build Toolbox from source inside a Toolbox container.

Do we? I'm wondering why is that required. What does it test? That Toolbx is buildable? That a software can be built?

If we were to add this test, it should not exist in the system tests but in a different group of tests.

debarshiray:
> > By the way, this makes me think that we need a test that tries to build
> > Toolbox from source inside a Toolbox container.
> 
> Do we? I'm wondering why is that required.

It's required because our test suite clearly failed to catch that this pull request breaks the build inside a Toolbx container. :)

> What does it test?

It should test that Toolbx can be built from source inside a Toolbx container.  Toolbx is necessary for hacking on Toolbx, so we shouldn't be pushing changes that break the build.

HarryMichal:
> It's required because our test suite clearly failed to catch that this pull request breaks the build inside a Toolbx container. :)

Technically speaking Toolbx can be built even without this fix being applied. It can not be run though due to missing `/run/host` symlink. I'm playing with words a bit but the problem does not lie in the build but the executable environment and that one is usually set up during installation (also this PR fixes the problem on installation). The fact we can run Toolbx without installing it is coincidental convenience.

> It should test that Toolbx can be built from source inside a Toolbx container.

Despite what I've written above, I agree the test can be written but I still believe such a test should exist outside of the current system test test suite.

> Toolbx is necessary for hacking on Toolbx

This generally applies only to Fedora Silverblue.

debarshiray:
> The goal of this step is to create `/run/host` and not to create other
> tmpfiles.

In the past, `toolbox create` [wouldn't work without /run/media](https://github.com/containers/toolbox/commit/0c848abd1893764f2e0bb62334972d67f95dafd1).  All those paths in `data/tmpfiles.d/toolbox.conf` are important for one reason or another.

> A way to solve this is to create only this file. I did it with this
> snipper:
> 
> ```
> meson.add_install_script(
>   systemd_tmpfiles,
>   '--create', '--prefix=/run/host', tmpfiles_toolbox_conf,
>   skip_if_destdir: true,
> )
> ```

Does this snippet actually work?  Or is it meant to be pseudo-code?  :)

The build failed inside a Toolbx container because `systemd` doesn't quite work inside the container.  I didn't check exactly which part of `systemd-tmpfiles` runs into problems, but the files we want to create are meant to live on the host, so it doesn't really matter.

debarshiray:
> > It's required because our test suite clearly failed to catch that this
> pull request breaks the build inside a Toolbx container. :)
> 
> Technically speaking Toolbx can be built even without this fix being
> applied. It can not be run though due to missing `/run/host` symlink.
> I'm playing with words a bit but the problem does not lie in the build
> but the executable environment and that one is usually set up during
> installation (also this PR fixes the problem on installation).

Eh?  :)

This pull request is trying to fix https://github.com/containers/toolbox/issues/955 which is about the built binary `toolbox` failing to execute.  Generally speaking, there's no point in building a binary that doesn't work.  This is also why when projects have tests, failing tests are equated to broken builds.

Since we are trying to fix https://github.com/containers/toolbox/issues/955, we might as well ensure that we do so without breaking the build.  :)

> > It should test that Toolbx can be built from source inside a Toolbx container.
> 
> Despite what I've written above, I agree the test can be written but I still
> believe such a test should exist outside of the current system test test suite.

It doesn't matter so much exactly what we call the test, as long as it tests the right thing and it is called from `ninja test`.  :)

> > Toolbx is necessary for hacking on Toolbx
> 
> This generally applies only to Fedora Silverblue.

... and Fedora Silverblue is the primary reason for Toolbx to exist.

It will be a major problem if it becomes impossible to hack on Toolbx on a Silverblue system.

HarryMichal:
> In the past, `toolbox create` [wouldn't work without /run/media](https://github.com/containers/toolbox/commit/0c848abd1893764f2e0bb62334972d67f95dafd1). All those paths in `data/tmpfiles.d/toolbox.conf` are important for one reason or another.

This is true but `/run/media` is needed in a later part of the execution process of Toolbx (exactly when creating + starting Toolbx containers). `/run/host` is needed to run the tool at all.

The reason for the suggestion is not to have a failing statement as part of Toolbx build. That way we'll never get a successful build.

> Does this snippet actually work? Or is it meant to be pseudo-code? :)

Yes, it works :).

Comparison of the `systemd-tmpfiles` invocation inside of a Toolbx container:

```sh
; systemd-tmpfiles --create
Failed to open directory 'cryptsetup': Permission denied
Failed to open directory 'portables': Permission denied
Failed to open directory 'sudo': Permission denied
Failed to open directory 'ts': Permission denied
Failed to open directory 'svnserve': Permission denied
fchownat() of /var/lib/systemd/coredump failed: Read-only file system
Failed to open directory 'private': Permission denied
Failed to open directory 'private': Permission denied
Failed to open directory 'private': Permission denied
fchownat() of /tmp failed: Operation not permitted
Setting access ACL "u::rwx,g::r-x,g:adm:r-x,g:wheel:r-x,g:4294967295:r-x,g:4294967295:r-x,m::r-x,o::r-x" on /var/log/journal failed: Read-only file system
Failed to re-open '/var/log/journal': Operation not permitted
fchownat() of /var/log/journal failed: Read-only file system
Setting access ACL "u::rwx,g::r-x,g:adm:r-x,g:wheel:r-x,g:4294967295:r-x,g:4294967295:r-x,m::r-x,o::r-x" on /var/log/journal/64ea8f1e4b534013843813476c178ab7 failed: Read-only file system
Failed to re-open '/var/log/journal/64ea8f1e4b534013843813476c178ab7': Operation not permitted
fchownat() of /var/log/journal/64ea8f1e4b534013843813476c178ab7 failed: Read-only file system
fchownat() of /dev/snd/seq failed: Operation not permitted
fchownat() of /dev/snd/timer failed: Operation not permitted
fchownat() of /dev/loop-control failed: Operation not permitted
fchownat() of /dev/kvm failed: Operation not permitted
fchownat() of /dev/vhost-net failed: Operation not permitted
fchownat() of /dev/vhost-vsock failed: Operation not permitted
Setting access ACL "u::rw-,g::r-x,g:adm:r--,g:wheel:r--,g:4294967295:r--,g:4294967295:r--,m::r--,o::---" on /var/log/journal/64ea8f1e4b534013843813476c178ab7/system.journal failed: Read-only file system
fchownat() of /var/log/journal/64ea8f1e4b534013843813476c178ab7/system.journal failed: Read-only file system
Setting default ACL "u::rwx,g::rwx,g:tss:rwx,m::rwx,o::r-x" on /var/lib/tpm2-tss/system/keystore failed: Operation not permitted
Setting default ACL "u::rwx,g::rwx,g:tss:rwx,m::rwx,o::r-x" on /run/tpm2-tss/eventlog failed: Operation not permitted

; echo $status
73

; systemd-tmpfiles --create --prefix=/run/host /var/home/marty/Documents/projects/toolbox/data/tmpfiles.d/toolbox.conf

; echo $status
0
```
 
> The build failed inside a Toolbx container because `systemd` doesn't quite work inside the container.

`systemd-tmpfiles` should work despite `systemd` being the active init, right?

> I didn't check exactly which part of `systemd-tmpfiles` runs into problems, but the files we want to create are meant to live on the host, so it doesn't really matter.

`/run/host` needs to be everywhere. Toolbx takes care of creating it in its containers.

debarshiray:
> > In the past, `toolbox create`
> > [wouldn't work without /run/media](https://github.com/containers/toolbox/commit/0c848abd1893764f2e0bb62334972d67f95dafd1).
> > All those paths in `data/tmpfiles.d/toolbox.conf` are important for one
> > reason or another.
> 
> This is true but `/run/media` is needed in a later part of the execution process
> of Toolbx (exactly when creating + starting Toolbx containers). `/run/host` is
> needed to run the tool at all.

Ideally, we don't want to differentiate between breakages.  We want our build to generate a fully working binary, as far as reasonably and practically possible.

Also, replicating a different subset of the files in `data/tmpfiles.d/toolbox.conf` in different parts of the build will be a maintenance nightmare.  We should try to keep things consolidated in one place as far as possible.

> The reason for the suggestion is not to have a failing statement as part of
> Toolbx build. That way we'll never get a successful build.

Umm... I didn't understand.

> > Does this snippet actually work? Or is it meant to be pseudo-code? :)
> 
> Yes, it works :).

I asked because this snippet didn't look valid:
```
meson.add_install_script(
  systemd_tmpfiles,
  '--create', '--prefix=/run/host', tmpfiles_toolbox_conf,
  skip_if_destdir: true,
)
```

... and the Meson logs didn't exactly match what I was seeing. eg.,
```
Running custom install script '/usr/bin/systemd-tmpfiles --create'
```

I suppose you had defined `systemd_tmpfiles` elsewhere.

> > The build failed inside a Toolbx container because `systemd` doesn't quite
> > work inside the container.
> 
> `systemd-tmpfiles` should work despite `systemd` being the active init, right?

`systemd-tmpfiles` is part of `systemd`.  We don't promise working `systemd` inside Toolbx containers, so it's not surprising that `systemd-tmpfiles` doesn't work either.  In theory, we could probably make it work, but it might also not be accepted by `systemd` upstream.

> > I didn't check exactly which part of `systemd-tmpfiles` runs into problems, but
> the files we want to create are meant to live on the host, so it doesn't really matter.
> 
> `/run/host` needs to be everywhere. Toolbx takes care of creating it in its containers.

We are talking about `systemd-tmpfiles --create` here.  Those files are host-specific.

HarryMichal:
> Ideally, we don't want to differentiate between breakages. We want our build to generate a fully working binary, as far as reasonably and practically possible.

Fair enough.
 
> Also, replicating a different subset of the files in `data/tmpfiles.d/toolbox.conf` in different parts of the build will be a maintenance nightmare. We should try to keep things consolidated in one place as far as possible.

Fair enough.

> Umm... I didn't understand.

Reason why I'm suggesting to run `systemd-tmpfiles` on a subset instead of running it with default settings.
 
> I asked because this snippet didn't look valid:
> 
> ```
> meson.add_install_script(
>   systemd_tmpfiles,
>   '--create', '--prefix=/run/host', tmpfiles_toolbox_conf,
>   skip_if_destdir: true,
> )
> ```

> I suppose you had defined `systemd_tmpfiles` elsewhere.

Of course `systemd_tmpfiles` has to be declared first. This snippet should do the trick:

```
systemd_tmpfiles = find_program('systemd-tmpfiles')
```

And `tmpfiles_toolbox_conf` has to also exist, first. Ideally this call is done in `data/tmpfiles.d/meson.build` where the variable holds a list of files with the config file present inside.

> `systemd-tmpfiles` is part of `systemd`. We don't promise that all `systemd` features will work inside Toolbx containers, so it's not surprising that `systemd-tmpfiles` doesn't work either. In theory, we could probably make it work, but it might also not be accepted by `systemd` upstream.

I don't understand what do you mean by "`systemd-tmpfiles` doesn't work". It works but in the container exist files/directories that the tool can not have privileges to work with because they've been bind-mounted from the host. There's also a difference between running the tool in system or user mode (`--user` option).

> > `/run/host` needs to be everywhere. Toolbx takes care of creating it in its containers.
> 
> We are talking about `systemd-tmpfiles --create` here. Those files are host-specific.

Of course. Still, the point about `/run/host` stands.

debarshiray:
> > Umm... I didn't understand.
> 
> Reason why I'm suggesting to run `systemd-tmpfiles` on a subset instead
> of running it with default settings.

I still don't understand.  Why would we run a subset of the `tmpfiles.d` snippet?

It already works without a subset.  We detect that we are inside a container and skip the `systemd-tmpfiles --create` call.

> > `systemd-tmpfiles` is part of `systemd`. We don't promise that all `systemd`
> > features will work inside Toolbx containers, so it's not surprising that
> > `systemd-tmpfiles` doesn't work either. In theory, we could probably make it
> > work, but it might also not be accepted by `systemd` upstream.
> 
> I don't understand what do you mean by "`systemd-tmpfiles` doesn't work".
> It works but in the container exist files/directories that the tool can not have
> privileges to work with because they've been bind-mounted from the host.

Ah, you are right that it's failing because the locations are owned by `nobody:nobody`.

Regardless, why `systemd-tmpfiles --create` is failing inside containers doesn't matter that much, because we don't need to run it inside containers anyway.

meson.build
debarshiray: It's no longer sufficient to just check for `DESTDIR`.  Now we need to look for the absence of `/run/.containerenv` as well.

(This is what I meant by *if the script evolves more, then we might not be able to use it anymore*.)

eli-schwartz: `skip_if_destdir: true` can overlap and coexist with checking for `/run/.containerenv`.

```diff
diff --git a/meson.build b/meson.build
index cf32ae3..8c41fd6 100644
--- a/meson.build
+++ b/meson.build
@@ -71,4 +71,4 @@ subdir('doc')
 subdir('profile.d')
 subdir('src')
 
-meson.add_install_script('meson_post_install.py')
+meson.add_install_script('meson_post_install.py', skip_if_destdir: true)
diff --git a/meson_post_install.py b/meson_post_install.py
index 2e22d87..c845e2a 100644
--- a/meson_post_install.py
+++ b/meson_post_install.py
@@ -4,9 +4,7 @@ import os
 import subprocess
 import sys
 
-destdir = os.environ.get('DESTDIR', '')
-
-if not destdir and not os.path.exists('/run/.containerenv'):
+if not os.path.exists('/run/.containerenv'):
     print('Calling systemd-tmpfiles --create ...')
 
     try:
```

debarshiray: I was considering using `skip_if_destdir` when the post-install script was essentially a one-liner, because that was the [original motivation](https://github.com/mesonbuild/meson/issues/9) behind adding that to Meson as an exception to the general rule about environment variables.

However, it's not a one-liner anymore because of the second condition.  So we might as well keep everything in one place.

eli-schwartz: Your link doesn't defend your claim that skip_if_destdir is an exception to "don't use environment variables.

Meson avoids environment variables during `mesin setup` because e.g. they result in different ninja build files if the environment variable changes on the next reconfigure.

The DESTDIR variable is only read during install, and meson already reads it to set the default `--destdir=...` value.

The skip_if_destdir kwarg has three practical effects here:
- it is processed in a fraction of a second if DESTDIR is set, and simply notes that it is skipped, rather than running in several fractions of a second by forking to python and doing nothing
- it helpfully logs "skipping custom install script because DESTDIR is set" (your script instead logs "running custom install script" and does nothing)
- it doesn't set the "did_install_something" bit, which may actually matter if no other files are installed, for example due to the use of install tags, in which case meson will log "nothing to install".

Nothing really serious, but my point is that on general principle there are reasons why for consistency, scripts should be using the skip_if_destdir kwarg.

debarshiray: > Your link doesn't defend your claim that skip_if_destdir is an exception to
> "don't use environment variables.

`skip_if_destdir` was added in https://github.com/mesonbuild/meson/pull/8269  It says:
```
It is common, at least in GNOME projects, to have scripts that must be
run only in the final destination, to update system icon cache, etc.
```

That's exactly what ptomato was saying [earlier](https://github.com/mesonbuild/meson/issues/9#issuecomment-385849310).

As a GNOME developer, I have dealt with my share of Meson post-install scripts to get the connection. ;)

> The skip_if_destdir kwarg has three practical effects here:

I don't see any of those reasons to be important enough to sacrifice the readability of the build scripts.  Without getting into too many details why they don't matter, I will just say that if `skip_if_destdir` was that important, then it wouldn't be (finally) added to Meson in January 2021.  ;)

eli-schwartz: Okay? That is not providing access to environment variables, and ptomato's mentioning DESTDIR in that issue was technically offtopic. DESTDIR is handled via minstall, always has been and always will be, and meson.build will still not read environment variables at configure time. (If it did then you'd have install scripts that ran during a DESTDIR install because DESTDIR wasn't set during `meson setup`...)

> I don't see any of those reasons to be important enough to sacrifice the readability of the build scripts.

That's a reasonable argument to make :) yes.

PR #970, opened 2022-01-04
.zuul: Try to prevent the CI from timing out running the unit tests

debarshiray:
Currently, the CI has been frequently timing out when running the unit
tests. It's possible that the current 5 minute timeout isn't enough,
because it's significantly lower than the 20 minute timeout on stable
Fedoras for the system tests.

Increase the timeout to 10 minutes to see if that makes the CI more
stable.


TristanCacqueray (approved):

HarryMichal:
Thanks!

PR #971, opened 2022-01-05
.zuul: Shuffle some code around

debarshiray:
Defining the YAML anchor as part of the Rawhide tests, instead of the
Fedora 34 test, will prevent it from getting lost by mistake when
Fedora 34 reaches its End of Life.


HarryMichal:
Thanks for this!

Issue #937, opened 2021-11-29
toolbox create argument 'distro' and 'release' invalid values handling


oturpe:
**Describe the bug**
`toolbox create` command accepts parameters `--distro` and `--release`.
It is not clear what values should be added to these,
and passing invalid values behaves strangely:

1. `--distro` seems to interpret any invalid value to mean Fedora,
which does not make sense.
If somebody decided to pass that argument,
for sure they were expecting to get what they passed in,
not any fallback because what they wanted is not available
or because they made a typo.
2. Except if `--distro` is used, but `--release` is not,
the resulting error message implies different behavior:
`release not found for non-default distribution <distro>`.
3. `--release` seems to use some kind of pattern matching depending on value of distro,
and prints `invalid argument for --release` if the pattern is not matched,
but does not say what pattern should be matched.

**Steps how to reproduce the behaviour**

`--distro` falling back to `fedora`:

```
$ toolbox create --distro alpine --release 35
Image required to create toolbox container.
Download registry.fedoraproject.org/fedora-toolbox:13 (500MB)? [y/N]:
```

`--distro` without `--release` implying that it is not going to fall back to Fedora:

```
$ toolbox create --distro alpine
Error: release not found for non-default distribution alpine
```

`--release` not helping with determining the correct pattern:

```
$ toolbox create --distro rhel --release 8
Error: invalid argument for '--release'
Run 'toolbox --help' for usage.
```

**Expected behaviour**
Fedora (nor anything else) is not used as a fallback for invalid `--distro` value.

Error messages help to fix problems:
Explain the expected pattern for  `--release`
and list supported values for `--distro`.

**Actual behaviour**
See above.

**Screenshots**
n/a

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.92.2`

**Toolbox package info (`rpm -q toolbox`)**
`toolbox-0.0.99.2^3.git075b9a8d2779-4.fc35.x86_64`

**Output of `podman version`**
```
Version:      3.4.2
API Version:  3.4.2
Go Version:   go1.16.8
Built:        Fri Nov 12 22:25:37 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-3.4.2-1.fc35.x86_64`

**Info about your OS**
Fedora Linux 35

**Additional context**
n/a

oturpe:
Related, it would be good if `toolbox create --help` would list
supported values for `--distro`
and related patterns for `--release`.

[...]

PR #972, opened 2022-01-05
man: toolbox-create distro broken

laolux:
Added a note that the --distro option in toolbox create is currently not working as expected.

This is being worked on in #937, but I thought it would be useful to warn users until this is resolved.
Certainly confused me and made me read the man page 3 times to see if I did something wrong.
Remove this note when #937 is solved.


HarryMichal:
Hi @laolux! Thank you for the contribution! The situation around #937 is definitely not ideal but instead of writing in the documentation that something does not work we should try to fix it for the next release. I already did some work towards fixing the issue, so hopefully you'll get a fix soon. Thanks for stopping by!

PR #974, opened 2022-01-07
ci: Add golangci linter

HarryMichal:
Use machines to do (at least some) of the work humans had to do!

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -1,0 +1,25 @@
+name: Golangci Lint
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  golangci:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


HarryMichal:
I don't know how to test this without pushing to the main branch. If there's an issue, I'll fix it in a follow up PR.

debarshiray (commented):
I wonder if we can run this as part of 'ninja test'.

Generally speaking, I want to consolidate our entire test suite to be run from 'ninja test`, so that I can run all the tests when doing a release, playing with the code locally, etc..  Currently our unit tests are run differently than our system tests, and now this adds a third variant.

Secondly, I don't want to have a hard dependency on some infrastructure, unless we really have to, so that we can continue working even if the infrastructure fails.  Also, this brings in yet another YAML file, which we need to keep up with.

Does `golangci-lint` already cover what `go fmt` does?

debarshiray:
> Does `golangci-lint` already cover what `go fmt` does?

https://github.com/containers/toolbox/pull/983 adds `gofmt` to `ninja test`.

Issue #961, opened 2021-12-17
Use newly published UBI toolbox image for RHEL


HarryMichal:
**Is your feature request related to a problem? Please describe.**

Currently Toolbox uses [`registry.access.redhat.com/ubi8/ubi`](https://catalog.redhat.com/software/containers/ubi8/ubi/5c359854d70cc534b3a3784e) and now there is [`registry.access.redhat.com/ubi8/toolbox`](https://catalog.redhat.com/software/containers/ubi8/toolbox/611bd665bd674341b5c5ed46) that has been created for Toolbx.

**Describe the solution you'd like**

Use `registry.access.redhat.com/ubi8/toolbox` for RHEL (`rhel`).

PR #976, opened 2022-01-08
pkg/utils: Use new UBI toolbox image

HarryMichal:
Red Hat has published a new UBI image made specificaly for Toolbx.
Make use of it from now on.

Fixes: https://github.com/containers/toolbox/issues/961


debarshiray:
Thanks, @HarryMichal !

PR #977, opened 2022-01-09
More robustly handle values passed to --distro and --release

HarryMichal:
This change is rather big despite my efforts to keep it as small as possible. The handling needed some separation and along the way I noticed some possible problems or duplication.

Read commits for detailed information.

Fixes https://github.com/containers/toolbox/issues/937

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -138,17 +138,23 @@ func create(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	var release string
-	if createFlags.release != "" {
+	distro, err := utils.ResolveDistro(createFlags.distro)
+	if err != nil {
+		err := createErrorInvalidDistro()
+		return err
+	}
+
+	release := createFlags.release
+	if release != "" {
 		var err error
-		release, err = utils.ParseRelease(createFlags.distro, createFlags.release)
+		release, err = utils.ParseRelease(distro, release)
 		if err != nil {
-			err := createErrorInvalidRelease()
+			err := createErrorInvalidRelease(distro)
 			return err
 		}
 	}
 
-	image, release, err := utils.ResolveImageName(createFlags.distro, createFlags.image, release)
+	image, release, err := utils.ResolveImageName(distro, createFlags.image, release)
 	if err != nil {
 		return err
 	}
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -249,7 +249,7 @@ func GetContainerNamePrefixForImage(image string) (string, error) {
 }
 
 func getDefaultImageForDistro(distro, release string) string {
-	if _, supportedDistro := supportedDistros[distro]; !supportedDistro {
+	if !IsDistroSupported(distro) {
 		distro = "fedora"
 	}
 
@@ -643,13 +643,10 @@ func ShortID(id string) string {
 // 'general.distro' key in a config file or 'fedora') is assumed.
 func ParseRelease(distro, release string) (string, error) {
 	if distro == "" {
-		distro = distroDefault
-		if viper.IsSet("general.distro") {
-			distro = viper.GetString("general.distro")
-		}
+		distro, _ = ResolveDistro(distro)
 	}
 
-	if _, supportedDistro := supportedDistros[distro]; !supportedDistro {
+	if !IsDistroSupported(distro) {
 		distro = "fedora"
 	}
 
@@ -796,10 +793,7 @@ func ResolveImageName(distroCLI, imageCLI, releaseCLI string) (string, string, e
 	distro, image, release := distroCLI, imageCLI, releaseCLI
 
 	if distroCLI == "" {
-		distro = distroDefault
-		if viper.IsSet("general.distro") {
-			distro = viper.GetString("general.distro")
-		}
+		distro, _ = ResolveDistro(distroCLI)
 	}
 
 	if distro != distroDefault && releaseCLI == "" && !viper.IsSet("general.release") {
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -250,7 +250,7 @@ func GetContainerNamePrefixForImage(image string) (string, error) {
 
 func getDefaultImageForDistro(distro, release string) string {
 	if !IsDistroSupported(distro) {
-		distro = "fedora"
+		distro = distroDefault
 	}
 
 	distroObj, supportedDistro := supportedDistros[distro]
@@ -647,7 +647,7 @@ func ParseRelease(distro, release string) (string, error) {
 	}
 
 	if !IsDistroSupported(distro) {
-		distro = "fedora"
+		distro = distroDefault
 	}
 
 	distroObj, supportedDistro := supportedDistros[distro]
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -44,6 +44,7 @@ type Distro struct {
 	ContainerNamePrefix    string
 	ImageBasename          string
 	ParseRelease           ParseReleaseFunc
+	ReleaseFormat          string
 	Registry               string
 	Repository             string
 	RepositoryNeedsRelease bool
@@ -58,6 +59,10 @@ const (
 	// Based on the nameRegex value in:
 	// https://github.com/containers/libpod/blob/master/libpod/options.go
 	ContainerNameRegexp = "[a-zA-Z0-9][a-zA-Z0-9_.-]*"
+)
+
+var (
+	ErrUnsupportedDistro = errors.New("linux distribution is not supported")
 )
 
 var (
@@ -98,6 +103,7 @@ var (
 			"fedora-toolbox",
 			"fedora-toolbox",
 			parseReleaseFedora,
+			"<release>/f<release>",
 			"registry.fedoraproject.org",
 			"",
 			false,
@@ -106,6 +112,7 @@ var (
 			"rhel-toolbox",
 			"toolbox",
 			parseReleaseRHEL,
+			"<major.minor>",
 			"registry.access.redhat.com",
 			"ubi8",
 			false,
@@ -407,6 +414,20 @@ func GetMountOptions(target string) (string, error) {
 	return mountOptions, nil
 }
 
+// GetReleaseFormat returns the format string signifying supported release
+// version formats.
+//
+// distro should be value found under ID in os-release.
+//
+// If distro is unsupported an empty string is returned.
+func GetReleaseFormat(distro string) string {
+	if !IsDistroSupported(distro) {
+		return ""
+	}
+
+	return supportedDistros[distro].ReleaseFormat
+}
+
 func GetRuntimeDirectory(targetUser *user.User) (string, error) {
 	gid, err := strconv.Atoi(targetUser.Gid)
 	if err != nil {
@@ -446,12 +467,29 @@ func GetRuntimeDirectory(targetUser *user.User) (string, error) {
 	return toolboxRuntimeDirectory, nil
 }
 
+// GetSupportedDistros returns a list of supported distributions
+func GetSupportedDistros() []string {
+	var distros []string
+	for d := range supportedDistros {
+		distros = append(distros, d)
+	}
+	return distros
+}
+
 // HumanDuration accepts a Unix time value and converts it into a human readable
 // string.
 //
 // Examples: "5 minutes ago", "2 hours ago", "3 days ago"
 func HumanDuration(duration int64) string {
 	return units.HumanDuration(time.Since(time.Unix(duration, 0))) + " ago"
+}
+
+// IsDistroSupported signifies if a distribution has a toolbx image for it.
+//
+// distro should be value found under ID in os-release.
+func IsDistroSupported(distro string) bool {
+	_, ok := supportedDistros[distro]
+	return ok
 }
 
 // ImageReferenceCanBeID checks if 'image' might be the ID of an image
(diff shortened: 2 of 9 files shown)


doc/toolbox-create.1.md
```diff
@@ -82,7 +82,8 @@ confusion.
 **--distro** DISTRO, **-d** DISTRO
 
 Create a toolbox container for a different operating system DISTRO than the
-host. Cannot be used with `--image`.
+host. Cannot be used with `--image`. Has to be coupled with `--release` unless
+the selected system matches the host system.
```
debarshiray, line 86: Nitpick: just "...matches the host" seems enough, since we spelled out "operating system DISTRO ... host" before.

debarshiray (commented):
Cool! Makes sense.

doc/toolbox.1.md
```diff
@@ -40,6 +40,20 @@ access to the user's home directory, the Wayland and X11 sockets, networking
 (including Avahi), removable devices (like USB sticks), systemd journal, SSH
 agent, D-Bus, ulimits, /dev and the udev database, etc..
 
+## Supported distribution images
```
debarshiray, line 43: We need to ensure that this section fits with the introduction above.

The introduction ends by mentioning the fedora-toolbox image, but it doesn't say anything about how the image relates to the host, if the host is used to select the image, or what happens when the host is not known.  So we should cover those points here.  Otherwise it feels a bit abrupt.

*Historically, we have been pretty brief or hand wavy about these things, because Fedora was the only supported distribution.  Now that we have added RHEL, and we are trying to spell things out more clearly, it's a good opportunity to make the documentation a bit more complete.*

So, it will be good to also include hosts here:
```
## Supported operating system distributions
```

doc/toolbox.1.md
```diff
@@ -40,6 +40,20 @@ access to the user's home directory, the Wayland and X11 sockets, networking
...
+
+Toolbox maintains a list of supported distribution images. These images are
+maintained by the team to be compatible with Toolbox. Some commands accept the
+`--distro` and `--release` flags. `--distro` accepts `ID` from `os-release`
+(e.g., on Fedora it is `fedora`). `--release` accepts version number of the
+selected distribution. Each distribution uses sligthly different versioning
+schema and some don't have any at all. Supported distributions are:
```
debarshiray, line 50: Then:
```
By default, Toolbox tries to create containers for the same operating system
distribution as the host.  If the host is not supported, then it creates a Fedora
container.

However, it's possible to create containers for a different distribution through
the use of the `--distro` and `--release` options that are accepted by the
relevant commands, or their counterparts in the configuration file.  The `--distro`
flag specifies the name of the distribution, and `--release` specifies its release.

Supported host operating systems:
...
...
```

We offer a configuration file now, so we might as well mention it.  :)

doc/toolbox.1.md
```diff
@@ -40,6 +40,20 @@ access to the user's home directory, the Wayland and X11 sockets, networking
...
+
+Name  |os-release|Format                     |URI
+------|----------|---------------------------|-----------------------------------------
+Fedora|fedora    |\<release\> or f\<release\>|registry.fedoraproject.org/fedora-toolbox
+RHEL  |rhel      |\<major\>.\<minor\>        |registry.access.redhat.com/ubi8/ubi
+
```
debarshiray, line 56: The column headers should be `distro` and `release`, and we should add one or two examples.

The manuals are user-facing documentation.  Their main goal is to document the user-visible interface.  The fact that we use `os-release(5)` is secondary because it's an implementation detail.  Since this is the top-level manual, it's better to stick to the higher-level big picture items.

I wouldn't document the exact URLs of the images.  Generally speaking, anything that we document will be considered API.  It will be difficult to change those things, and we have changed URLs in the past.  Moreover, the URLs are long enough that they don't render nicely in a terminal.  Especially not in the default width of 80 columns.

debarshiray (changes requested):

debarshiray (commented):
This reminds me of [commit fd756089efb89f05](https://github.com/containers/toolbox/commit/fd756089efb89f05cd8c163606659f21c7c45cd8).

I have been having doubts about it, because splitting `ResolveContainerAndImageNames` into two public functions pushes the burden on the callers.  The callers now need to carefully call the split functions in the right order, which opens the door for mistakes.

I think a good compromise would be to restore `ResolveContainerAndImageNames` as the single public API, but internally split it into smaller private functions.  It would keep things simple for both the callers and the implementation.

Ultimately, the `distro`, `release`, `image` and `container` values are very tightly related here.  So, we need a public API to which we feed the input values, and get back the resolved values, without worrying about the intricacies involved.

doc/toolbox.1.md
HarryMichal: > The column headers should be `distro` and `release`, and we should add one or two examples.

Ok.
 
> The manuals are user-facing documentation. Their main goal is to document the user-visible interface. The fact that we use `os-release(5)` is secondary because it's an implementation detail. Since this is the top-level manual, it's better to stick to the higher-level big picture items.

I agree with being cautious of getting too technical but the fact we use `ID` from `os-release(5)` is very important as we require an exact value. A user needs to know what value exactly to provide.

> I wouldn't document the exact URLs of the images. Generally speaking, anything that we document will be considered API. It will be difficult to change those things, and we have changed URLs in the past. Moreover, the URLs are long enough that they don't render nicely in a terminal. Especially not in the default width of 80 columns.

Ok.

doc/toolbox.1.md
HarryMichal: Reworded the section using suggestions.

doc/toolbox.1.md
HarryMichal: For now went the way of leaving out the technical details.

debarshiray (commented):
Thanks for the updates, @HarryMichal

doc/toolbox-create.1.md
```diff
@@ -82,7 +82,8 @@ confusion.
 **--distro** DISTRO, **-d** DISTRO
 
 Create a toolbox container for a different operating system DISTRO than the
-host. Cannot be used with `--image`.
+host. Cannot be used with `--image`. Has to be coupled with `--release` unless
+the selected system matches the host.
```
debarshiray, line 86: Nitpick: your commit message reminded me that *selected DISTRO* would be better, because it's a good opportunity to refer to the argument as we call it elsewhere.

debarshiray:
I merged the updates to the manuals in https://github.com/containers/toolbox/pull/986 after applying the above tweaks.

doc/toolbox.1.md
```diff
@@ -40,6 +40,28 @@ access to the user's home directory, the Wayland and X11 sockets, networking
 (including Avahi), removable devices (like USB sticks), systemd journal, SSH
 agent, D-Bus, ulimits, /dev and the udev database, etc..
 
+## Supported operating system distributions
+
+By default, Toolbox tries to use an image matching the host operating system
+distribution for creating containers. If the host is not supported, then it
+fallbacks to a Fedora image.
```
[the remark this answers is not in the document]
HarryMichal: > Nitpick: _falls back_, because fallback is the noun.

Oh, right. Thanks for catching this.

doc/toolbox.1.md
```diff
@@ -40,6 +40,28 @@ access to the user's home directory, the Wayland and X11 sockets, networking
...
+
+Some commands accept the `--distro` and `--release` flags allowing the user to
+use non-default distribtions and their versions. The `--distro` flag specifies
+the name of a distribution, and `--release` specifes its release version. Each
```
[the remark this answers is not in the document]
HarryMichal: It felt more natural to mention it separately. Both versions work for me.

doc/toolbox.1.md
```diff
@@ -40,6 +40,28 @@ access to the user's home directory, the Wayland and X11 sockets, networking
...
+distribution uses sligthly different versioning schema and some may not have
+any at all.
```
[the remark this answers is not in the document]
HarryMichal: > I suppose you were thinking of rolling distributions...

Yes.

> ...but we currently don't support any distribution that doesn't have a version number. So this part can be confusing.

Fair enough.

> And since we are going to list out the exact combinations that are supported, it would be clarified to the reader anyway.

:+1:

HarryMichal:
> I merged the updates to the manuals in #986 after applying the above tweaks.

Thanks!

doc/toolbox-create.1.md
debarshiray: This was fixed, and merged through https://github.com/containers/toolbox/pull/986

src/pkg/utils/utils.go
```diff
@@ -250,7 +250,7 @@ func GetContainerNamePrefixForImage(image string) (string, error) {
 
 func getDefaultImageForDistro(distro, release string) string {
 	if !IsDistroSupported(distro) {
-		distro = "fedora"
+		distro = distroDefault
```
debarshiray, line 253: As far as I can make out, this change seems functionally correct (ie., it won't break anything for the user), but it can cause confusion for the person reading the code.

The problem is that we have two similar, but subtly different, concepts that are important to way containers, distros, images and releases are handled; and we haven't been careful about clearly naming and separating the variables.

The *default* values are those that are used when the `create`, `enter` and `run` commands were used without any option.  These values are picked to match the host operating system.

However, if there's no supported Toolbox image matching the host operating system, and no options were provided to the `create`, `enter` and `run` commands, then the *fallback* values are used as a last resort.  Currently, the *fallback* values are set to Fedora.

Here, the string literal `"fedora"` was meant as a *fallback*, not *default*, value, because we never fall back to the *default* under any circumstances.  We should fall back to the *fallback* values if the *default* is not supported.

However, since the variables are neither clearly named nor separated, replacing the string literal with `distroDefault` doesn't change the behaviour of the code, but does make things even more confusing for the reader.

src/pkg/utils/utils.go
```diff
@@ -44,6 +44,7 @@ type Distro struct {
 	ContainerNamePrefix    string
 	ImageBasename          string
 	ParseRelease           ParseReleaseFunc
+	ReleaseFormat          string
```
debarshiray, line 47: I think it will be better to use the errors thrown by a `ParseReleaseFunc` to describe what was wrong with the `release` option, than to have a one-size-fits-all-errors string.  The `ParseReleaseFunc` knows exactly what went wrong while parsing, so it can clearly present the problem to the user.

A custom `ParseReleaseError` with a `hint` string to describe the problem in detail might work very well here.  See my comment about sentinels and custom `error` implementations.

The `toolbox(1)` manual already has a list of accepted `release` formats for each `distro` with examples.  It seems like the `ReleaseFormat` string is a more terse representation of that.  This means that accepted `release` formats are described in three different places:
* in the manual
* in the `ReleaseFormat` string
* in the `ParseReleaseFunc`

It will be good if we could reduce that to two.  :)

debarshiray (commented):
Time has only made me lean further towards my [earlier comment](https://github.com/containers/toolbox/pull/977#pullrequestreview-849758592).  We should re-unify the resolution of `container`, `distro`, `image` and `release`, and not split it even further.

src/pkg/utils/utils.go
```diff
@@ -643,13 +643,10 @@ func ShortID(id string) string {
 // 'general.distro' key in a config file or 'fedora') is assumed.
 func ParseRelease(distro, release string) (string, error) {
 	if distro == "" {
-		distro = distroDefault
-		if viper.IsSet("general.distro") {
-			distro = viper.GetString("general.distro")
-		}
+		distro, _ = ResolveDistro(distro)
```
debarshiray, line 646: To continue my [comment](https://github.com/containers/toolbox/pull/977#discussion_r960631867) from the previous commit ...

The `ResolveDistro` call is not just repeated across the front-end as part of the overall resolution code.  It's now spreading across the backend, and, since the function is fallible it makes me wonder if different invocations of it can yield different results.  It will be better if we could just resolve the `distro` only once and keep re-using the value.

src/cmd/create.go
```diff
@@ -138,17 +138,23 @@ func create(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	var release string
-	if createFlags.release != "" {
+	distro, err := utils.ResolveDistro(createFlags.distro)
```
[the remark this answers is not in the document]
debarshiray: Container and image name resolution is re-unified in https://github.com/containers/toolbox/pull/1080

src/pkg/utils/utils.go
debarshiray: Container and image name resolution is re-unified in https://github.com/containers/toolbox/pull/1080

src/pkg/utils/utils.go
debarshiray: https://github.com/containers/toolbox/pull/1100 uses the errors thrown by the `ParseReleaseFunc` implementations.

debarshiray: https://github.com/containers/toolbox/pull/1101 introduces a `ParseReleaseError`.

src/pkg/utils/utils.go
```diff
@@ -60,6 +61,10 @@ const (
 	ContainerNameRegexp = "[a-zA-Z0-9][a-zA-Z0-9_.-]*"
 )
 
+var (
+	ErrUnsupportedDistro = errors.New("linux distribution is not supported")
```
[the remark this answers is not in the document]
debarshiray: See https://github.com/containers/toolbox/pull/1101 for another step in this direction.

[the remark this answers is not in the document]
debarshiray: https://github.com/containers/toolbox/pull/1103 , https://github.com/containers/toolbox/pull/1104 and https://github.com/containers/toolbox/pull/1106 continue in this direction.

PR #978, opened 2022-01-09
ci: Remove only-new-issues option from linter run

HarryMichal:
The option is mutually exclusive with the working-directory option that
we need to use due to Toolbx code residing in a subdirectory. The tool
itself does not work recursively.

Fallout from https://github.com/containers/toolbox/pull/974

https://github.com/containers/toolbox/pull/978

### DIFF
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -22,4 +22,3 @@ jobs:
         with:
           version: latest
           working-directory: ./src
-          only-new-issues: true


debarshiray:
I wish there was some documentation. :)

eg., what does this mean:
```
The tool itself does not work recursively.
```

HarryMichal:
The tool uses as its working context either current working directory or one passed to it through a parameter. This parameter is exposed through the GitHub Action, too. The option is needed in case the project source code resides in a different directory than the root of a repository. The use of the option is mutually exclusive with the use of the `only-new-issues` option due to implementation detail of the tool. Seems like to achieve the effect of catching only newly introduced problems the tool compares the state on the main branch with the state in the introduced change. And the use of a different working context seems to be unsupported for this "specialized" workflow.

PR #982, opened 2022-01-11
build: Mark golang.org/x/term as a direct dependency

debarshiray:
Every invocation of 'go build' insists on updating src/go.mod to drop
the 'indirect' marker from golang.org/x/term.

Fallout from d323143c468e5e59dba4d209a4e4767d88dfc19e


HarryMichal:
Thanks for this!

PR #983, opened 2022-01-11
build: Enforce gofmt on all Go sources

debarshiray:
Gofmt returns with an exit code of 0 even when it's not happy with the
formatting of the source files. The recommendation is to check the
emptyness of the output [1].

[1] https://github.com/golang/go/issues/24230

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,6 +1,8 @@
 go_build_wrapper_file = files('go-build-wrapper')
 go_build_wrapper_program = find_program('go-build-wrapper')
 
+meson_go_fmt_program = find_program('meson_go_fmt.py')
+
 sources = files(
   'toolbox.go',
   'cmd/create.go',
@@ -66,4 +68,5 @@ if shellcheck.found()
   test('shellcheck go-build-wrapper', shellcheck, args: [go_build_wrapper_file])
 endif
 
+test('go fmt', meson_go_fmt_program, args: [meson.current_source_dir()])
 test('toolbox go unit tests', go, args: ['test', '-v', './...'], workdir: meson.current_source_dir())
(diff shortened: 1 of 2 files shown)


debarshiray:
recheck

debarshiray:
> Do we need a wrapper script for this? A `go fmt ./...` call seems simple enough to me.

Yes.  As the commit message mentions, `gofmt` doesn't offer any non-zero exit code for problems.  The recommendation is to check the standard output.  So we need a wrapper to turn that into a non-zero exit code for Meson to detect the failure.

Also, it seems easier and better to directly use `gofmt` instead of `go fmt`.  The latter is only a wrapper that invokes `gofmt -l -w`, which isn't ideal for a test suite.  Getting to see the actual problem through a diff can be quite helpful.

debarshiray:
recheck

debarshiray:
We should actually use Python for our wrapper scripts, instead of shell, because that's the lingua franca of the Meson world.

debarshiray:
> > We should actually use Python for our wrapper scripts, instead of
> > shell, because that's the lingua franca of the Meson world.
> 
> Actually, the rationale is explicitly a portability one.

I know.  I am not sure what you are trying to argue.

> With my Mesonbuild core dev hat on...
> 
> For linux-only projects I see no reason not to use shell wrapper scripts.
> And they will be a bit faster than loading a python interpreter.

Not everything is about speed.  Readability and ease of hacking are factors, and shell scripts are terrible in those.

Given that we spent considerable effort re-implementing Toolbx in a language other than shell, I'd really like to avoid using it as much as possible.  Otherwise, we might as well have kept using the shell implementation.

Meson already implies Python, so using shell instead of Python doesn't give us any meaningful benefit.  It's not like we are going to fork and exec a Python interpreter hundreds of times during a build.

For the sake of completeness, Toolbx itself isn't written in Python because Fedora CoreOS tries to avoid Python in the core OS as much as possible, but that's not a factor for the build scripts.

PR #997, opened 2022-01-24
Add two bind mounts needed by display servers

jadahl:
These two commits adds two new bind mounts to newly created toolboxes.

The first one binds `/run/systemd/sessions`, so that systemd calls such as `sd_session_get_seat()` works.

The second one binds `/run/udev/tags`. This makes it possible to run device enumerators that matches against some tag. Here is an example that would now work:

```c
/* gcc -o udev-enumerate `pkg-config --cflags libudev --libs libudev` udev-enumerate.c */

#include <libudev.h>
#include <stdio.h>

int
main(int argc, char **argv)
{
	struct udev *udev;
	struct udev_enumerate *e;
	struct udev_list_entry *entry;

	udev = udev_new();
	e = udev_enumerate_new(udev);
	udev_enumerate_add_match_subsystem(e, "drm");
	udev_enumerate_add_match_tag(e, "seat");
	udev_enumerate_scan_devices(e);
	udev_list_entry_foreach(entry, udev_enumerate_get_list_entry(e)) {
		const char *path;

		path = udev_list_entry_get_name(entry);
		printf("path: %s\n", path);
	}
	return 0;
}
```

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -56,6 +56,7 @@ var (
 		{"/run/libvirt", "/run/host/run/libvirt", ""},
 		{"/run/systemd/journal", "/run/host/run/systemd/journal", ""},
 		{"/run/systemd/resolve", "/run/host/run/systemd/resolve", ""},
+		{"/run/systemd/sessions", "/run/host/run/systemd/sessions", ""},
 		{"/run/udev/data", "/run/host/run/udev/data", ""},
 		{"/tmp", "/run/host/tmp", "rslave"},
 		{"/var/lib/flatpak", "/run/host/var/lib/flatpak", "ro"},
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -58,6 +58,7 @@ var (
 		{"/run/systemd/resolve", "/run/host/run/systemd/resolve", ""},
 		{"/run/systemd/sessions", "/run/host/run/systemd/sessions", ""},
 		{"/run/udev/data", "/run/host/run/udev/data", ""},
+		{"/run/udev/tags", "/run/host/run/udev/tags", ""},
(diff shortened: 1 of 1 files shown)


debarshiray:
recheck

jadahl:
Thanks for landing!

debarshiray:
> Thanks for landing!

Yeah, I was only more than a year late, but better late than never.  I tweaked the commit message a bit and merged.

debarshiray:
> This makes it possible to run device enumerators that matches against some tag. Here is an example that would now work:

Thanks for cooking up this example use-case.  I will try to turn it into a test case.

debarshiray:
> These two commits adds two new bind mounts to newly created toolboxes.
> 
> The first one binds `/run/systemd/sessions`, so that systemd calls such as `sd_session_get_seat()` works.

Don't you also want `/run/systemd/users`?  I see that `/run/systemd/users/$UID` also has:
```
SEATS=seat0
ACTIVE_SEATS=seat0
ONLINE_SEATS=seat0
```

Aren't those files used by some code path for similar purposes?  Isn't it possible that the implementation details change and various bits of code start reading `/run/systemd/users` instead of `/run/systemd/sessions`?

jadahl:
> Aren't those files used by some code path for similar purposes? Isn't it possible that the implementation details change and various bits of code start reading /run/systemd/users instead of /run/systemd/sessions?

Not impossible, I suppose, I haven't checked.

debarshiray:
> > Aren't those files used by some code path for similar purposes? Isn't it possible that the implementation details change and various bits of code start reading /run/systemd/users instead of /run/systemd/sessions?
> 
> Not impossible, I suppose, I haven't checked.

Okay!  I tried this in https://github.com/containers/toolbox/pull/1234

Issue #992, opened 2022-01-20
Support running display servers on a VT


jadahl:
Currently it's not possible to run a display server that relies using common technologies like logind and udev in a toolbox from a VT. This issue is meant to track the things than is needed to make this possible. The reason for wanting to do this is typically Wayland compositor development on a Silverblue system.

1. `sd_pid_get_session()` doesn't find the right session.

This function tries to find the session by looking at the cgroup on `getpid()`. This fails due to the cgroup not being what it expects. A typical cgroup it'd expect looks like `0::/user.slice/user-1000.slice/session-35.scope`, where `35` is the session, but instead it'll contain something like `0::/user.slice/user-1000.slice/user@1000.service/user.slice/libpod-67cfc51a5110905b4c7fb6b49607fcf4aba9a9b88089d0fe57fc95cc87ce2755.scope/container`.

It might be possible to work around this issue in compositors by finding the session some other way. However, for this to be possible using the C API, some things seems to be needed:
  * `/run/systemd/users/$UID` needs to be exposed for systemd to find the sessions
  * `/run/systemd/sessions/` needs to be exposed

Some help from toolbox might be needed as well for allowing the display server to figure out what tty `toolbox enter` came from.

2. udev device enumeration doesn't work when matching seat.

Running e.g `udevadm trigger --dry-run --subsystem-match=drm --verbose -g seat` in a toolbox doesn't enumerate the available devices as it does when run outside of toolbox. Not passing `-g seat` makes it appear. Compositors use this to only attempt to open drm devices assigned to its seat. Have yet to investigate further what is needed.

If one works around these two issues (e.g. hard coding them), one can sucessfully run a display server running as drm master, managing mode setting devices and input devices.

Cc: @debarshiray, @swick, @HarryMichal

jadahl:
Some more findings:

> sd_pid_get_session() doesn't find the right session.

`$XDG_SESSION_ID` is already exposed inside the toolbox; the display server should use use this instead of `sd_pid_get_session()`.

> udev device enumeration doesn't work when matching seat.

This needed exposing `/run/udev/tags` from the host to work.

jadahl:
For reference, https://github.com/containers/toolbox/pull/997 makes things work better. Together with using `$XDG_SESSION_ID` things appears to launch properly with toolboxes created with that PR applied.

debarshiray:
> Some more findings:
> 
> > sd_pid_get_session() doesn't find the right session.
> 
[...]

PR #1009, opened 2022-02-14
cmd/create: use the host cgroup namespace

swick:
Podman creates a cgroup namespace for cgroups v2 by default. The host
cgroupfs is mounted at /sys/fs/cgroup giving an inconsistent view of the
cgroups. There is no need for a cgroup namespace with toolbox so just
use the host namespace.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -403,6 +403,7 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 	createArgs := []string{
 		"--log-level", logLevelString,
 		"create",
+		"--cgroupns", "host",
 		"--dns", "none",
 		"--env", toolboxPathEnvArg,
 	}


debarshiray (changes requested):
Thanks @swick and my apologies for the delay!

This seems sane to me, but I wonder what the user-facing effect or advantage of this is.  Does this help with https://github.com/containers/toolbox/issues/992 ?  I am asking so that some time in the future we can easily answer the question: *why does Toolbx use --cgroupns=host?*

debarshiray:
> That's a really good question. I know this was the result of working on #992 but everything seems to be working just fine without this change.
> 
> This could very well be from experimenting with systemd user sessions in the toolbox to run a whole desktop session. Same for #1010.

I see.  Even if things are working without it, it doesn't seem obviously harmful for Toolbx containers to use the host's cgroup namespace.

In fact, since this requires a change in the `podman create` arguments, it's not going to have an effect on existing containers, and re-creating containers is annoying for users.  So it might be better to get ahead of the curve and do it preemptively.

src/cmd/create.go
```diff
@@ -389,6 +389,10 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 		"--env", toolboxPathEnvArg,
 	}
 
+	createArgs = append(createArgs, []string{
+		"--cgroupns=host",
```
debarshiray, line 393: Nitpick: this could just be folded into the array above, no?

PR #1010, opened 2022-02-14
cmd/initContainer: Bind mount /run/systemd/system

swick:
Required for systemd --user to realize the system was booted with
systemd.


akdev1l (approved):
lgtm

debarshiray:
> This seems sane to me, but I wonder what the user-facing effect or advantage of this is. Does this help with #992 ? I am asking so that some time in the future we can easily answer the question: _why is it necessary for systemd --user to realize the system was booted with systemd?_

Answered in https://github.com/containers/toolbox/pull/1009

Issue #1011, opened 2022-02-16
Create a Fedora 37 toolbox image


travier:
**Is your feature request related to a problem? Please describe.**

Create a Fedora 37 toolbox. We need this to test toolbox in Fedora CoreOS. See https://github.com/coreos/fedora-coreos-tracker/issues/1103

travier:
There are probably other steps that I don't know about on the infra side of things.

HarryMichal:
There seems to be an upstream issue with building images -> https://pagure.io/releng/issue/10658

juhp:
Actually the first step is to have an image tagged with `fedora:37` which is currently missing.

juhp:
fedora:37 is fixed, but now the fedora-toolbox:37 build fails in OBS in the same way as fedora-toolbox:36.

juhp:
I have built the container images now - see the releng if you want to know the details..

This ticket can be closed now

PR #1012, opened 2022-02-16
images: Create F37 toolbox image from F36

travier:
Related to https://github.com/containers/toolbox/issues/1011


juhp:
Thank you! 
But note the actual container definition lives in
https://src.fedoraproject.org/container/fedora-toolbox/tree/rawhide
for now - the files are just a reference copy of that basically.

HarryMichal:
> Thank you! 
> But note the actual fedora-toolbox container definition lives in
> https://src.fedoraproject.org/container/fedora-toolbox/tree/rawhide

I'll add that piece of info into the directory. Without it is confusing.

travier:
Hum, OK. I'll update that with the content from the Fedora repo then.

HarryMichal:
Thank you, @travier!

Issue #848, opened 2021-07-21
Using `toolbox run` in a pipeline sometimes leaves tty in wrong state


GregoryHinton:
**Description**
Using `toolbox run` in a pipeline sometimes leaves the tty with echo and other settings disabled.
The `stty sane` command must be issued to return the tty to normal.
For example, the command
```
    $ toolbox run echo hello | awk ''
```
will cause this to occur.

**Steps to reproduce**
```
1. toolbox create
2. stty -a >stty.before
3. toolbox run echo hello | awk ''; stty -a >stty.after; stty sane
4. diff stty.before stty.after
```

**Expected behaviour**
The `diff` command from step 4 above should have no output, i.e. the tty should be in
the same state as before the `toolbox run` command.

**Actual behaviour**
The following `stty -a` settings are changed:
```
icrnl --> -icrnl
opost --> -opost
isig --> -isig
icanon --> -icanon
iexten --> -iexten
echo --> -echo
```

**Output of `toolbox --version` (v0.0.90+)**
```
toolbox version 0.0.99.2
```

**Toolbox package info (`rpm -q toolbox`)**
```
toolbox-0.0.99.2-1.fc34.x86_64
```

**Output of `podman version`**
```
Version:      3.2.2
API Version:  3.2.2
Go Version:   go1.16.4
Built:        Fri Jun 25 16:49:19 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
```
podman-3.2.2-1.fc34.x86_64
```

**Info about your OS**
```
$ head -2 /etc/os-release 
NAME=Fedora
VERSION="34.20210721.0 (Silverblue)"
```

**Additional context**
This may be related to [#157](https://github.com/containers/toolbox/issues/157).

I have not seen this bug occur when `sed` is used as the second stage in the pipeline.
For some reason, `awk` is particularly vulnerable.

I have confirmed that each line of the `toolbox run` output is terminated with CRLF 
instead of just LF. I thought this might be confusing `awk`; but when a hand-built file 
terminated with CRLF is piped it into `awk`, it works just fine.

I have seen no problems when I use `podman exec` for the first stage of the pipeline 
(with or without the `-it` flags). Doesn't `toolbox run` just do some initial checking and then 
run `podman exec`?

I also reported this bug at [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1983378).

debarshiray:
[...]

Issue #157, opened 2019-05-13
toolbox run does not work when no tty is attached


tosmi:
the current implementation of run() always uses the flags `--interactive` and `--tty` for podman exec. i'm using toolbox run mostly from `dmenu`, so there is no tty attached. Maybe `--tty` and `--interactive` should only be used when enterning a container. 

will create a pull request asap.

thanks
toni

debarshiray:
You need `--interactive --tty` in `toolbox run` if you want to run a terminal text editor, for example. Maybe `toolbox run` should grow a flag for this?

tosmi:
yep maybe. i'll try to come up with something next week. didn't find the time for anything the last week...
i mostly use `toolbox run` to start my containerized apps within X11 (e.g. vlc, firefox, emacs...). 

i suppose the default should be to use a tty and optionally disable allocating a tty, so `--no-tty`?

thanks
toni

tosmi:
so #179 should fix this and works for me. 

thanks
toni

ibotty:
Any reason why this is not acted upon?  Should it be determined whether there actually is a tty and automatically disable interactive and tty forward?

tosmi:
it would be great if this could be merged

thanks

HarryMichal:
Hey @tosmi! I just added your PR (#179) in the rewritten version of Toolbox (tracked here: #318). I just adjusted it to the Go code. Thank you for your contribution and I apologize for not merging this sooner!

jmou:
Hi! I came looking for a similar issue but I think the problem is a little more general. In particular I think there are broadly 3 use cases for `toolbox run`:

- Running GUI (non-terminal) applications, possibly without a controlling tty. This is @tosmi 's use case. We SHOULD NOT allocate a tty in this case.
- Running interactive terminal applications that need a tty (bash, screen, vim, anything curses). We SHOULD allocate a tty in this case.
- Running non-interactive terminal applications. These may be used with stdin often attached to a pipe (netcat), which does not work with a tty. We SHOULD NOT allocate a tty in this case.

An app like ssh may be either the second (if invoked without a command, it allocates a tty) or the third (does not allocate a tty if invoked with a command).

[...]

PR #1013, opened 2022-02-21
cmd/run, root: Exit with exit code of invoked command

HarryMichal:
When a command is executed with toolbox run and it returns a non-zero
exit code, it is just ignored if that exit code is not handled. This
prevents users to identify errors when executing commands in toolbox.

With this fix, the exit codes of the invoked command are propagated
and returned by 'toolbox run'. This includes even exit codes returned
by Podman on error.

Supersedes https://github.com/containers/toolbox/pull/867

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -325,7 +325,7 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 			logrus.Debugf("%s", arg)
 		}
 
-		exitCode, err := shell.RunWithExitCode("podman", os.Stdin, os.Stdout, nil, execArgs...)
+		exitCode, err := shell.RunWithExitCode("podman", os.Stdin, os.Stdout, os.Stderr, execArgs...)
 
 		if emitEscapeSequence {
 			fmt.Printf("\033]777;container;pop;;;%s\033\\", currentUser.Uid)
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -159,6 +159,15 @@ func run(cmd *cobra.Command, args []string) error {
 		false,
 		false,
 		true); err != nil {
+		// runCommand returns exitError for the executed commands to properly
+		// propagate return codes. Cobra prints all non-nil errors which in
+		// that case is not desirable. In that scenario silence the errors and
+		// leave the error handling to the root command.
+		var errExit *exitError
+		if errors.As(err, &errExit) {
+			cmd.SilenceErrors = true
+		}
+
 		return err
 	}
 
@@ -338,9 +347,9 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 			}
 			return nil
 		case 125:
-			return fmt.Errorf("failed to invoke 'podman exec' in container %s", container)
+			return &exitError{exitCode, fmt.Errorf("failed to invoke 'podman exec' in container %s", container)}
 		case 126:
-			return fmt.Errorf("failed to invoke command %s in container %s", command[0], container)
+			return &exitError{exitCode, fmt.Errorf("failed to invoke command %s in container %s", command[0], container)}
 		case 127:
 			if pathPresent, _ := isPathPresent(container, workDir); !pathPresent {
 				if runFallbackWorkDirsIndex < len(runFallbackWorkDirs) {
@@ -357,7 +366,7 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 					fmt.Fprintf(os.Stderr, "Using %s instead.\n", workDir)
 					runFallbackWorkDirsIndex++
 				} else {
-					return fmt.Errorf("directory %s not found in container %s", workDir, container)
+					return &exitError{exitCode, fmt.Errorf("directory %s not found in container %s", workDir, container)}
 				}
 			} else if _, err := isCommandPresent(container, command[0]); err != nil {
 				if fallbackToBash && runFallbackCommandsIndex < len(runFallbackCommands) {
@@ -371,13 +380,13 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 
 					runFallbackCommandsIndex++
 				} else {
-					return fmt.Errorf("command %s not found in container %s", command[0], container)
+					return &exitError{exitCode, fmt.Errorf("command %s not found in container %s", command[0], container)}
 				}
 			} else {
 				return nil
 			}
 		default:
-			return nil
+			return &exitError{exitCode, nil}
 		}
 	}
 }
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -26,6 +26,7 @@ import (
 	"github.com/containers/toolbox/pkg/podman"
 	"github.com/containers/toolbox/pkg/shell"
 	"github.com/containers/toolbox/pkg/utils"
+	"github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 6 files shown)


HarryMichal:
recheck

src/cmd/run.go
```diff
@@ -325,7 +325,7 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 			logrus.Debugf("%s", arg)
 		}
 
-		exitCode, err := shell.RunWithExitCode("podman", os.Stdin, os.Stdout, nil, execArgs...)
+		exitCode, err := shell.RunWithExitCode("podman", os.Stdin, os.Stdout, os.Stderr, execArgs...)
```
debarshiray, line 337: I am unsure about this one.  The standard error stream was being intentionally set to `nil`, but not doing that also helps in some cases.

First, the behaviour shown in the commit message is wrong, but don't be upset.  It took me 5 hours before I could make any sense of what's going on, and I am still unsure.   :)

The behaviour before this patch was:
```
$ output=$(toolbox run /etc)
Error: failed to invoke command /etc in container fedora-toolbox-36
$ echo $output

$
```

ie., `output` was no longer being polluted by the error messages from the inner shell.  This was caused by a combination of the previous [commit](https://github.com/containers/toolbox/commit/a22d7821cb8cea134791dcd4643de21ef565d645) that skips the `--tty` flag when the standard output isn't connected a terminal device, and the [fact](https://github.com/containers/podman/issues/16254) that `--tty` connects the standard error stream of the inner process to the standard output of the outer.

Anyway, the standard error stream was being intentionally set to `nil` to suppress unexpected or redundant error messages from Podman or the inner shells, unless `--verbose` was being used.  eg., in this case, the top two lines are a bit redundant and reveal ugly implementation details:
```
output=$(toolbox run /etc)
/bin/sh: line 1: /etc: Is a directory
/bin/sh: line 1: exec: /etc: cannot execute: Is a directory
Error: failed to invoke command /etc in container fedora-toolbox-36
```

Instead, a better error might be something like:
```
Error: failed to invoke command /etc in container fedora-toolbox-36
Is /etc an executable?  Do you have permission to execute it?
```

... but that doesn't require the standard error stream from Podman or the inner shells.

However, having the standard error stream can be useful in some other cases:
```
$ output=$(./builddir/src/toolbox run grep bar /foo)
grep: /foo: No such file or directory
```

... because the error comes directly from the command that the user actually tried to run.  We can't detect it through `podman exec` exit codes and such, and showing it doesn't expose any ugly implementation details.  In fact, suppressing these errors hurt usability.

debarshiray: I will think about this some more and try to see if I can figure out a way to have both.

debarshiray (commented):
Thanks, @HarryMichal - this does fix https://github.com/containers/toolbox/issues/848

However, I didn't bother setting up [dmenu](https://tools.suckless.org/dmenu/) (nor [dwm](https://dwm.suckless.org/)) as mentioned in https://github.com/containers/toolbox/issues/157 because ... X.  :)

src/cmd/run.go
debarshiray: One correction after reading Matt's [comment](https://github.com/containers/podman/issues/16254#issuecomment-1287778526) on https://github.com/containers/podman/issues/16254:

*Remember: `podman exec --tty` connects the standard error stream of the inner process to the standard output of the outer.*

> the standard error stream was being intentionally set to nil
> to suppress unexpected or redundant error messages from
> Podman or the inner shells, unless --verbose was being
> used. eg., in this case, the top two lines are a bit redundant
> and reveal ugly implementation details:

We weren't able to suppress the standard error stream from *the inner shells* by setting it to `nil`, when `--tty` was being used, because the use of `--tty` merges the standard error stream of the inner process to the standard output.  So, the behaviour in these cases didn't change:
```
$ toolbox run /etc
/bin/sh: line 1: /etc: Is a directory
/bin/sh: line 1: exec: /etc: cannot execute: Is a directory
Error: failed to invoke command /etc in container fedora-toolbox-36
$ toolbox run grep bar /foo
grep: /foo: No such file or directory
```

The previous [commit](https://github.com/containers/toolbox/commit/a22d7821cb8cea134791dcd4643de21ef565d645) fixed the behaviour when the standard output of `toolbox` wasn't connected to a terminal, by not using `--tty`.  This prevented the standard error streams from the inner shells from polluting the `output`, because, in the absence of `--tty`, the standard error stream of the inner process stayed separate from the standard output.  Unfortunately, since Toolbx wasn't reading the standard error stream, because it was set to `nil`, the error messages completely disappeared after the previous patch.  So, as seen before, the behaviour with the previous commit, but before this patch was:
```
$ output=$(toolbox run /etc)
Error: failed to invoke command /etc in container fedora-toolbox-36
$ echo $output

$ output=$(toolbox run grep bar /foo)
$ echo $output

$
```

This patch brings them back the errors from the inner shells by reading the standard error stream:
```
$ output=$(toolbox run /etc)
/bin/sh: line 1: /etc: Is a directory
/bin/sh: line 1: exec: /etc: cannot execute: Is a directory
Error: failed to invoke command /etc in container fedora-toolbox-36
$ echo $output

$ output=$(toolbox run grep bar /foo)
grep: /foo: No such file or directory
$ echo $output

$
```

Originally, what we were able to suppress was the standard error stream of Podman, unless `--verbose` was being used:
```
$ cd /usr/lib64/gnome-shell  # location only present on the host, not the container
$ toolbox run true
Error: directory /usr/lib64/gnome-shell not found in container fedora-toolbox-36
Using /home/rishi instead
```

Now with this patch, the errors from Podman show up even without `--verbose`:
```
$ cd /usr/lib64/gnome-shell  # location only present on the host, not the container
$ toolbox run true
Error: crun: chdir `/usr/lib64/gnome-shell`: No such file or directory: OCI runtime attempted to invoke a command that was not found
Error: directory /usr/lib64/gnome-shell not found in container fedora-toolbox-36
Using /home/rishi instead
```

debarshiray: Given that `podman exec --tty` [connects](https://github.com/containers/podman/issues/16254) the standard error stream of the inner process to the standard output of the outer, we won't be able to selectively suppress the error stream of the inner shells to hide ugly implementation details and only reveal what's relevant to the user.  At least not without some serious contortions like overriding the behaviour of `capsh`, `sh`, etc..

What we can suppress is Podman's standard error stream, and I think we should continue to do that as before (ie., unless `--verbose` is used).

src/cmd/run.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/containers/toolbox/pkg/podman"
 	"github.com/containers/toolbox/pkg/shell"
 	"github.com/containers/toolbox/pkg/utils"
+	"github.com/mattn/go-isatty"
```
debarshiray, line 29: Nitpick: ideally, we should be using `golang.org/x/term`.  :)

debarshiray: I attempted a fix in https://github.com/containers/toolbox/pull/1144

test/system/104-run.bats
```diff
@@ -121,3 +121,48 @@ teardown() {
   assert_success
   assert_output --partial "uid=0(root)"
 }
+
+@test "run: Run command exiting with zero code in the default container" {
+  create_default_container
+
+  run $TOOLBOX run /bin/sh -c 'exit 0'
```
debarshiray, line 128: Nitpick: `true` is likely going to be quicker.

debarshiray: Changed in https://github.com/containers/toolbox/pull/1145

PR #1029, opened 2022-03-20
.zuul: Use labels/nodes without the '-small' suffix

HarryMichal:


HarryMichal:
Seems like it helps a bit with CI execution times. Nice!

PR #1031, opened 2022-03-22
src: Port Go build wrapper from Shell to Python & support externaly defined build options

martymichal:


debarshiray (commented):
Yes, this will be good to have.  I once thought of it, but the shell script started giving me nightmares.

Maybe this is be a good time to add some tests to sanity check the built binary?

The `go-build-wrapper` script encapsulates a lot of subtle, but crucial, details and we don't want to regress anything.  eg., the path to the dynamic linker, the rpath, etc..  Ideally we would add the test first to test the old script, and then introduce the new script -- that way we can trivially verify (through `git bisect`, etc.) that everything is good.

src/go-build-wrapper.py
```diff
@@ -0,0 +1,99 @@
...
+                           [DYNAMIC LINKER]
+                           [MIGRATION PATH FORCOREOS/TOOLBOX]
+                           [BUILD FLAGS]'''.format(prog_name), file=sys.stderr)
+    sys.exit(1)
+
+source_dir = sys.argv[1]
+output_dir = sys.argv[2]
+version = sys.argv[3]
+cc = sys.argv[4]
+dyn_linker = sys.argv[5]
```
debarshiray, line 41: Would be better to use terms like *dynamic_linker*, *dynamic-linker* everywhere.  They are easier to `grep` for (eg., searching for `dynamic` will get you better results), and can defend against [typos](https://github.com/containers/toolbox/pull/942) by making them stand out.

test/binary/binary_test.go
```diff
@@ -0,0 +1,118 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package binary
+
+import (
+	"debug/elf"
```
debarshiray, line 20: How does Go's `debug/elf` compare with other implementations like GNU Binutils, in terms of architecture support and the ability to handle a diverse variety of ELFs?

HarryMichal: Overall I'd say the support is lower than of `readelf` from GNU Binutils. From what I understood though, `debug/elf` supports both [32-bit and 64-bit architectures](https://pkg.go.dev/debug/elf#Class) and recognizes many machine types and many OS ABIs. Though, am not sure how the last two parts translate into "architecture support". Most likely didn't answer your question in full.

Probably is noteworthy that the tool is for checking the Toolbx binary and not an arbitrary binary. My presumption is that whether Go is capable of producing a binary on a certain architecture, the created binary also has to be processable by the `debug/elf` package.

debarshiray: > Probably is noteworthy that the tool is for checking the Toolbx binary and not an arbitrary binary. My presumption is that whether Go is capable of producing a binary on a certain architecture, the created binary also has to be processable by the debug/elf package.

This is a good point!

completion/meson.build
```diff
@@ -1,14 +1,14 @@
 generate_completions_program = find_program('generate_completions.py')
 
 if bash_completion.found()
-  bash_comp_dir = bash_completion.get_pkgconfig_variable('completionsdir')
+  bash_comp_dir = bash_completion.get_variable(pkgconfig: 'completionsdir')
```
debarshiray, line 4: Oops!  This was done a few months after this pull request was submitted in commit [1567d207c6c6317c](https://github.com/containers/toolbox/commit/1567d207c6c6317c264ffcb056f564e114333f38) or https://github.com/containers/toolbox/pull/1096.  Sorry about that.

completion/meson.build
```diff
@@ -1,14 +1,14 @@
...
 else
   bash_comp_dir = get_option('datadir') / 'bash-completion' / 'completions'
   message('bash-completion not found: using', get_option('prefix') / bash_comp_dir, 'as a falback install directory')
 endif
 
 if fish.found()
-  fish_comp_dir = fish.get_pkgconfig_variable('completionsdir')
+  fish_comp_dir = fish.get_variable(pkgconfig: 'completionsdir')
```
debarshiray, line 11: Ditto.

src/meson.build
```diff
@@ -58,16 +57,18 @@ custom_target(
...
-  test('shellcheck go-build-wrapper', shellcheck, args: [go_build_wrapper_file])
-endif
-
 test('go fmt', meson_go_fmt_program, args: [meson.current_source_dir()])
-test('toolbox go unit tests', go, args: ['test', '-v', './...'], workdir: meson.current_source_dir())
+test('toolbox go unit tests',
+  go,
+  args: ['test', '-v', './...'],
+  depends: toolbox,
+  workdir: meson.current_source_dir()
```
debarshiray, line 73: Is it really necessary to make the Go unit tests depend on the `toolbox(1)` binary?  They seem to be independent, as far as I can tell.  If it's really needed then it will be good to add the reason to the Git commit message.

debarshiray (changes requested):

src/go-build-wrapper
```diff
@@ -68,12 +68,16 @@ if ! dynamic_linker_canonical_dirname=$(dirname "$dynamic_linker_canonical"); th
 fi
 
 dynamic_linker="/run/host$dynamic_linker_canonical_dirname/$dynamic_linker_basename"
+rpath="/run/host$libc_dir_canonical_dirname"
+
+# Write overriden binary values to a file for testing
+echo "${dynamic_linker} ${rpath}" > "$2/go-build-overrides"
```
debarshiray, line 74: I wonder if we can pass these values straight to the test through `go test ... -args ... <dynamic-linker> <rpath>` instead of inventing an intermediate file.  If nothing else ...

test/binary/binary_test.go
```diff
@@ -0,0 +1,118 @@
...
+		fmt.Fprintf(os.Stderr, "%s: failed to read file %s: %s\n", os.Args[0], buildOverridesPath, err)
+	}
+
+	buildOverrides := strings.Split(strings.TrimSpace(string(content)), " ")
+	if len(buildOverrides) != 2 {
+		fmt.Fprintf(os.Stderr, "%s: overrides file needs to have 2 space-separated values written inside", os.Args[0])
+		os.Exit(1)
+	}
+	dynamicLinker = buildOverrides[0]
+	runpath = buildOverrides[1]
```
debarshiray, line 70: ... we won't have to do all of the above to parse the file.

PR #1035, opened 2022-04-06
cmd/list: Add option to display container/image size

### LINKED ISSUE #1026: Show container size when running list command
## Is your feature request related to a problem? Please describe.
I want to be able to know the size of my installed containers. If you have many different toolboxes with different configurations like different packages, different distributions and different installed packages it would be good to be able to compare sizes between them.

## Describe the solution you'd like
A column with size when running `toolbox list`.

## Describe alternatives you've considered
A flag to pass to the `toolbox list` command to show sizes.

Parth576:
- Add an additional option 'size' for the list command
- Works for containers as well as images
- Fixes #1026


whs-dot-hk:
I think there is too many if else statements

PR #1047, opened 2022-05-09
Make toolbox list behave in the same way podman images does

### LINKED ISSUE #1043: toolbox list output contains duplicate image records
**Describe the bug**
this bug was initially created downstream, I am creating it after discussion with @debarshiray as we want to extend discussion to upstream 

```
$ podman pull <INTERNAL_URL>/toolbox-container:9.0.0-9  
Trying to pull <INTERNAL_URL>/toolbox-container:9.0.0-9  ...
Getting image source signatures
Copying blob b1641283b956 done  
Copying blob 26f1167feaf7 done  
Copying blob adffa6963146 done  
Copying config e793d00bc6 done  
Writing manifest to image destination
Storing signatures
e793d00bc6295609626f1eae4084d4a767a751f4f1d4c3c3fc5c75d1b1c6680c

[test@rhel-8 ~]$ skopeo copy <INTERNAL_URL>/toolbox-container:9.0.0-9  containers-storage:registry.access.redhat.com/ubi9/toolbox:latest
Getting image source signatures
Copying blob e6a13b8f88ad skipped: already exists  
Copying blob 352ba846236b skipped: already exists  
Copying blob 3ba8c926eef9 skipped: already exists  
Copying config e793d00bc6 done  
Writing manifest to image destination
Storing signatures
[test@fedora redhat-linux-build]$ podman images
REPOSITORY                                                       TAG         IMAGE ID      CREATED      SIZE
registry.access.redhat.com/ubi9/toolbox                          latest      c24104d7b237  13 days ago  596 MB
<INTERNAL_URL>/rh-osbs/toolbox-container  9.0.0-9     c24104d7b237  13 days ago  596 MB
[test@fedora redhat-linux-build]$ toolbox list 
IMAGE ID      IMAGE NAME                                      CREATED
c24104d7b237  registry.access.redhat.com/ubi9/toolbox:latest  13 days ago
c24104d7b237  registry.access.redhat.com/ubi9/toolbox:latest  13 days ago
```

As you can see, output of `toolbox list` show those images as the same image, however those are actually 2 images that have 2 human readable names:

$ podman images --format json":
```
Image 1:

"Id": "e9e5f02b635db0130339d8e6ba8c873030dfcc24b075f96a8c4cdb743a9b6eac",
...
 "Names": [
            " registry.access.redhat.com/ubi9/toolbox:latest ",
            " <INTERNAL_URL>/toolbox-container:9.0.0-9 "
        ],

Image 2:

"Id": "e9e5f02b635db0130339d8e6ba8c873030dfcc24b075f96a8c4cdb743a9b6eac",
...
"Names": [
            " registry.access.redhat.com/ubi9/toolbox:latest  ",
            " <INTERNAL_URL>/toolbox-container:9.0.0-9 "
        ],  
```
The field that could actually distinguish those images is a TAG but it's not a part of JSON output, to obtain the TAG you have to switch to Go Template:

```
$  podman images   --format "{{.ID}}  {{.Tag}}"
c24104d7b237  latest
c24104d7b237  9.0.0-9
```

**Expected behaviour**
I am not sure what should be the expected behaviour but I would add the tag so we can see that those are actually two images.

toolbox version 0.0.99.3

toolbox-0.0.99.3-4.fc36.x86_64

e.g.,
```
Client:       Podman Engine
Version:      4.0.3
API Version:  4.0.3
Go Version:   go1.18
Built:        Fri Apr  1 20:08:04 2022
OS/Arch:      linux/amd64
```

podman-4.0.3-1.fc37.x86_64

**Info about your OS**
Fedora release 37 (Rawhide)

mkrajnak:
This PR should fix #1043.

It makes `toolbox list` behave in the same way as `podman images`
```
[test@fedora build]$ podman images 
REPOSITORY                                                       TAG         IMAGE ID      CREATED      SIZE
registry.access.redhat.com/ubi9/toolbox                          latest      c24104d7b237  2 weeks ago  596 MB
<INTERNAL_URL>/toolbox-container                             9.0.0-9     c24104d7b237  2 weeks ago  596 MB
[test@fedora build]$ toolbox list 
IMAGE ID      IMAGE NAME                                                               CREATED
c24104d7b237  registry.access.redhat.com/ubi9/toolbox:latest                           2 weeks ago
c24104d7b237  <INTERNAL_URL>/toolbox-container:9.0.0-9                            2 weeks ago
```

As mentioned in the issue `podman images --format json` returns image twice with multiple `Names`, so my approach is to detect images with the same ID (this comes from the assumption that those image record are really identical), then print image twice with both names.

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"sort"
 	"text/tabwriter"
 
 	"github.com/containers/toolbox/pkg/podman"
@@ -105,7 +106,7 @@ func list(cmd *cobra.Command, args []string) error {
 	var err error
 
 	if lsImages {
-		images, err = getImages()
+		images, err = getImages(false)
 		if err != nil {
 			return err
 		}
@@ -180,26 +181,41 @@ func listHelp(cmd *cobra.Command, args []string) {
 	}
 }
 
-func getImages() ([]podman.Image, error) {
+func getImages(fillNameWithID bool) ([]podman.Image, error) {
 	logrus.Debug("Fetching all images")
-	args := []string{"--sort", "repository"}
+	var args []string
 	images, err := podman.GetImages(args...)
 	if err != nil {
 		logrus.Debugf("Fetching all images failed: %s", err)
 		return nil, errors.New("failed to get images")
 	}
 
+	processed := make(map[string]struct{})
 	var toolboxImages []podman.Image
 
 	for _, image := range images {
+		if _, ok := processed[image.ID]; ok {
+			continue
+		}
+
+		processed[image.ID] = struct{}{}
+		var isToolboxImage bool
+
 		for label := range toolboxLabels {
 			if _, ok := image.Labels[label]; ok {
-				toolboxImages = append(toolboxImages, image)
+				isToolboxImage = true
 				break
 			}
 		}
+
+		if isToolboxImage {
+			flattenedImages := image.FlattenNames(fillNameWithID)
+			toolboxImages = append(toolboxImages, flattenedImages...)
+		}
+
 	}
 
+	sort.Sort(podman.ImageSlice(toolboxImages))
 	return toolboxImages, nil
 }
 
@@ -209,14 +225,13 @@ func listOutput(images []podman.Image, containers []toolboxContainer) {
 		fmt.Fprintf(writer, "%s\t%s\t%s\n", "IMAGE ID", "IMAGE NAME", "CREATED")
 
 		for _, image := range images {
-			imageName := "<none>"
-			if len(image.Names) != 0 {
-				imageName = image.Names[0]
+			if len(image.Names) != 1 {
+				panic("cannot list unflattened Image")
 			}
 
 			fmt.Fprintf(writer, "%s\t%s\t%s\n",
 				utils.ShortID(image.ID),
-				imageName,
+				image.Names[0],
 				image.Created)
 		}
 
--- a/test/system/102-list.bats
+++ b/test/system/102-list.bats
@@ -76,6 +76,27 @@ teardown() {
   fi
 }
 
+@test "list: Image and its copy" {
+  local default_image
+  default_image="$(get_default_image)"
+
+  pull_default_image_and_copy
+
+  local num_of_images
+  num_of_images="$(list_images)"
+  assert_equal "$num_of_images" 2
+
+  run --keep-empty-lines --separate-stderr "$TOOLBOX" list
+
+  assert_success
+  assert_line --index 1 --partial "$default_image"
+  assert_line --index 2 --partial "$default_image-copy"
+  assert [ ${#lines[@]} -eq 4 ]
+  if check_bats_version 1.7.0; then
+    assert [ ${#stderr_lines[@]} -eq 0 ]
+  fi
+}
+
 @test "list: Try to list images and containers (no flag) with 3 containers and 2 images (the list should have 3 images and 2 containers)" {
   # Pull the two images
   pull_default_image
--- a/test/system/102-list.bats
+++ b/test/system/102-list.bats
@@ -90,8 +90,8 @@ teardown() {
   run --keep-empty-lines --separate-stderr $TOOLBOX list --images
 
   assert_success
-  assert_line --index 1 --partial "$(get_system_id)-toolbox:$(get_system_version)"
-  assert_line --index 2 --partial "fedora-toolbox:34"
+  assert_line --index 1 --partial "fedora-toolbox:34"
+  assert_line --index 2 --partial "$(get_system_id)-toolbox:$(get_system_version)"
   assert [ ${#lines[@]} -eq 4 ]
   if check_bats_version 1.7.0; then
     assert [ ${#stderr_lines[@]} -eq 0 ]
@@ -113,8 +113,8 @@ teardown() {
   run --keep-empty-lines --separate-stderr $TOOLBOX list
 
   assert_success
-  assert_line --index 1 --partial "$(get_system_id)-toolbox:$(get_system_version)"
-  assert_line --index 2 --partial "fedora-toolbox:34"
+  assert_line --index 1 --partial "fedora-toolbox:34"
+  assert_line --index 2 --partial "$(get_system_id)-toolbox:$(get_system_version)"
   assert_line --index 5 --partial "$(get_system_id)-toolbox-$(get_system_version)"
   assert_line --index 6 --partial "non-default-one"
   assert_line --index 7 --partial "non-default-two"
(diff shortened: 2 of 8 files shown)


src/cmd/list.go
```diff
@@ -206,6 +206,7 @@ func getImages() ([]toolboxImage, error) {
 	for _, image := range images {
 		var i toolboxImage
 		var isToolboxImage bool = false
+		var isDuplicatedImage bool = false
```
debarshiray, line 209: Nitpick: there's no need for the `= false`.  Just this much is enough in Go:
```
var isDuplicatedImage bool
```

I know the (wrong) pattern was already [there](https://github.com/containers/toolbox/pull/469) in the code, so you can use a separate commit to fix it up.

debarshiray:
Two other things that came to my mind:

* Would it be possible to extend our test suite to cover this case?  Maybe we can stick a `skopeo copy ...` somewhere to mimic the error condition?
* Could you please add a link to the [original issue](https://github.com/containers/toolbox/pull/1047) in the commit message, like we do in the other commits?  That makes it easier to follow the paper trail from a terminal when looking at old code.  Otherwise one has to open a browser tab, construct the URL to the commit, etc. which gets a bit tedious when one is confused and under pressure.  :)

src/cmd/list.go
```diff
@@ -226,7 +227,14 @@ func getImages() ([]toolboxImage, error) {
 			}
 		}
 
-		if isToolboxImage {
+		for _, image := range toolboxImages {
+			if image.ID == i.ID {
+				isDuplicatedImage = true
+				break
+			}
```
debarshiray, line 234: It will be better to use a map instead of a loop, because it's a more efficient way of checking membership in a collection.  Since `toolboxImages` isn't of constant size and can potentially be very long, this is essentially becoming a O(n^2) algorithm.  :)

src/cmd/list.go
```diff
@@ -240,15 +248,16 @@ func listOutput(images []toolboxImage, containers []toolboxContainer) {
 		fmt.Fprintf(writer, "%s\t%s\t%s\n", "IMAGE ID", "IMAGE NAME", "CREATED")
 
 		for _, image := range images {
-			imageName := "<none>"
-			if len(image.Names) != 0 {
-				imageName = image.Names[0]
+			if len(image.Names) == 0 {
+				image.Names = []string{"<none>"}
 			}
+			for _, imageName := range image.Names {
```
debarshiray, line 254: Looping over the multiple names of an image here will break the sorting.  Even though the code broke in [commit 2369da5d31830e5c](https://github.com/containers/toolbox/commit/2369da5d31830e5cd3e1a13857a686365875ae61), the intention is to sort the output by the names of the images.

So, we need to flatten the names earlier, possibly in `getImages()`, then sort them, and then list them.

test/system/102-list.bats
```diff
@@ -102,3 +102,19 @@ teardown() {
 
     rm -f "$BATS_TMPDIR"/Containerfile
 }
+
+@test "list: List should not show two human readble names twice" {
+  pull_and_rename_default_image fedora 32
+
+  run podman images
```
debarshiray, line 109: It's not necessary to `RUN podman images` to double check whether the image really showed up or not.  For example, we don't check whether `create_default_container` worked or not.

test/system/libs/helpers.bash
```diff
@@ -286,6 +286,36 @@ function pull_default_image() {
...
+# - distro - os-release field ID (e.g., fedora, rhel)
+# - version - os-release field VERSION_ID (e.g., 33, 34, 8.4)
+function pull_and_rename_default_image() {
+  local distro
+  local version
+  local image
+  
+  distro="$1"
+  version="$2"
+  image="${IMAGES[$distro]}:${version}"
```
debarshiray, line 304: This could be simpler if we just used the default image for the host where the tests are running.

test/system/libs/helpers.bash
```diff
@@ -286,6 +286,36 @@ function pull_default_image() {
...
+  pull_distro_image $distro $version
+
+  # https://github.com/containers/skopeo/issues/547 for the options for containers-storage
+  run $SKOPEO copy "containers-storage:[overlay@$ROOTLESS_PODMAN_STORE_DIR+$ROOTLESS_PODMAN_STORE_DIR]${image}" "containers-storage:[overlay@$ROOTLESS_PODMAN_STORE_DIR+$ROOTLESS_PODMAN_STORE_DIR]${image}-copy"
+  if [ "$status" -ne 0 ]; then
+    echo "Failed to load image ${image} from cache ${IMAGE_CACHE_DIR}/${image_archive}"
+    assert_success
+  fi
+
+  $PODMAN images
```
debarshiray, line 315: This should be removed.  See: https://github.com/containers/toolbox/pull/1193

debarshiray (changes requested):

debarshiray:
I took the liberty to fix up the branch, along with various other improvements to the 'list' tests in https://github.com/containers/toolbox/pull/1192.

test/system/102-list.bats
debarshiray: Done.

src/cmd/list.go
debarshiray: Done.  However, this has an impact on the shell completions, so those need to be tested.

src/cmd/list.go
debarshiray: Done.

debarshiray:
> Build failed.
> 
> heavy_check_mark [unit-test ](https://softwarefactory-project.io/zuul/t/local/build/4a8cf35f9f87459e9befdbab23846322) SUCCESS in 8m 33s heavy_check_mark [unit-test-migration-path-for-coreos-toolbox ](https://softwarefactory-project.io/zuul/t/local/build/5b8fd49186ff484e86c05001b2cbdcc1) SUCCESS in 8m 28s x [system-test-fedora-rawhide ](https://softwarefactory-project.io/zuul/t/local/build/4a6279af908d47e9b417dda950699876) **RETRY_LIMIT** in 1m 10s heavy_check_mark [system-test-fedora-36 ](https://softwarefactory-project.io/zuul/t/local/build/b48c01ecc0184b06bdc341ed89f79421) SUCCESS in 12m 32s heavy_check_mark [system-test-fedora-35 ](https://softwarefactory-project.io/zuul/t/local/build/47b25ebf68c94c78b9449f6d96878ea9) SUCCESS in 13m 34s

```
Failed to download packages: aardvark-dns-1.3.0-1.fc38.x86_64: Cannot download, all mirrors were already tried without success
```

debarshiray:
recheck

test/system/102-list.bats
debarshiray: On second thoughts, there's this `list_images` function that we use to count the number of images.  We can use it here.

src/cmd/list.go
debarshiray: I tested the shell completions manually and they seem to work.

Since `getImages` is also used by the `rmi` command, I spent some time beefing up the tests for it.  See https://github.com/containers/toolbox/pull/1195

debarshiray:
> > Two other things that came to my mind:
> > ```
> > * Would it be possible to extend our test suite to cover this case?  Maybe we can stick a `skopeo copy ...` somewhere to mimic the error condition?
> > ```
> 
> We already have the busybox image available and it seems to
> me (though I didn't read the original issue, yet) that this is not
> tied to an exact image but to a naming scheme combined.

The busybox image isn't labeled with `com.github.containers.toolbox:true`, and hence isn't counted as a Toolbx image.  We need a Toolbx image because this is a bug in `list` when showing Toolbx images.

> Commands invoking `podman build` in a test case could be
> enough to test this, no?

The bug is about images that have been copied using `skopeo copy ...` or `podman tag ...`.  So, we need to do one of those on a Toolbx image to test the fix.

debarshiray:
Thanks for your work, @mkrajnak !

PR #1049, opened 2022-05-12
Extend toolbox command to accept custom hostnames.

absenth:
The Default behavior of toolbox does not support the dynamic setting of hostnames. 
The current behavior automatically sets the hostname of all podman containers to the hostname "toolbox"

Ideally like podman an operator might want the container name to be set at create time, so the prompt says something like....  `$username@testing` if the container was named `testing`.  This PR should continue the default behavior, but also allow the default hostname to be overwritten with the argument`--hostname foo` at toolbox create time.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -45,6 +45,7 @@ var (
 		authFile  string
 		container string
 		distro    string
+		hostname  string
 		image     string
 		release   string
 	}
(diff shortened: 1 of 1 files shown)


absenth:
This PR was premature.  I will re-open it once I've fixed the things I've done wrong.

Issue #959, opened 2021-12-16
Test suite doesn't offer a way to override the default image


debarshiray:
We currently offer [a configuration file](https://github.com/containers/toolbox/pull/828) to override the default image that's used for things like the `create` command.  Some downstream distributors use this to significantly change the default image to the point that the test suite can't be run unless it uses the same default image that's specified in the configuration file.

As an aside, I am worried that the test suite seems to be growing it's own logic to resolve container and image names parallel to the actual Toolbox code itself. We should try to fix this, because the logic isn't trivial.

When running the test suite on RHEL, @olivergs was using a downstream patch to introduce some environment variables to override the defaults, but I don't see how it can possibly work with the state of the test suite in Toolbox 0.0.99.3.

I think that we need a proper upstream solution for this.

debarshiray:
For reference, here's the downstream patch from RHEL:
```
From ae6de45466f93cf441408421f1804033c1525c0b Mon Sep 17 00:00:00 2001
From: Oliver Gutierrez <ogutsua@gmail.com>
Date: Tue, 14 Sep 2021 13:41:17 +0100
Subject: [PATCH] Tests fixes for gating

---
 test/system/102-list.bats      |  8 +++----
 test/system/103-container.bats |  2 +-
 test/system/libs/helpers.bash  | 38 +++++++++++++++++++++++++++-------
 3 files changed, 36 insertions(+), 12 deletions(-)

diff --git a/test/system/102-list.bats b/test/system/102-list.bats
index 1989409b1207..c6ddf15c34d1 100644
--- a/test/system/102-list.bats
+++ b/test/system/102-list.bats
@@ -62,14 +62,14 @@ teardown() {
   run $TOOLBOX list --images
 
   assert_success
-  assert_output --partial "$(get_system_id)-toolbox:$(get_system_version)"
+  assert_output --partial "$(get_default_image_name)"
   assert_output --partial "fedora-toolbox:32"
 
   # Check containers
   run $TOOLBOX list --containers
 
   assert_success
-  assert_output --partial "$(get_system_id)-toolbox-$(get_system_version)"
+  assert_output --partial "$(get_default_container_name)"
   assert_output --partial "non-default-one"
   assert_output --partial "non-default-two"
 
@@ -77,9 +77,9 @@ teardown() {
   run $TOOLBOX list
 
   assert_success
-  assert_output --partial "$(get_system_id)-toolbox:$(get_system_version)"
[...]

PR #1050, opened 2022-05-12
src,test/system: Respect configuration file in system tests

HarryMichal:
Commit 20f4f68 added support for configuration files. If keys in the
configuration file were set the system test suite would not honor them.
This puts additional strain on distributors overriding the default
settings.

This solves the problem by creating a new hidden command in Toolbx that
provides interface for getting some information Toolbx uses
under-the-hood (e.g., the default container name, image).

This commit does not change the way images are cached in the system test
suite. In case a different image is needed, the tests still need to be
patched.

Fixes https://github.com/containers/toolbox/issues/959


HarryMichal:
Just realised this does not handle the potential failures of calls to `toolbox __test` failing. Will have to address that before this can be merged.

HarryMichal:
> Just realised this does not handle the potential failures of calls to `toolbox __test` failing. Will have to address that before this can be merged.

On second thought this might not be as problematic. While all calls to `toolbox __test` can fail, the returned (or rather not returned) value will most likely cause the tests to fail, causing us to look into the failure. And only the `config-key` request type accepts an argument at the moment and an unknown key will yield nothing on `stdout`. @debarshiray, thoughts?

PR #1051, opened 2022-05-12
cmd/completion: Add prefix to command to hide it better

HarryMichal:


debarshiray:
This was reverted in https://github.com/containers/toolbox/pull/1055 (and https://github.com/containers/toolbox/pull/1121).

Issue #957, opened 2021-12-15
`toolbox create` command (ran from within toolbx) is not failing when container already exists.


JayDoubleu:
**Describe the bug**
`toolbox create` command is not failing when container already exists while running from within toolbox

**Steps how to reproduce the behaviour**
```
[user@hostname ~]$ toolbox create --distro fedora --release f35
Error: container fedora-toolbox-35 already exists
Enter with: toolbox enter
Run 'toolbox --help' for usage.
[user@hostname ~]$ echo $?
1

[user@hostname ~]$ toolbox run toolbox create --distro fedora --release f35
[user@hostname ~]$ echo $?
0
[user@hostname ~]$ 
```

**Expected behaviour**
Should return `Error: container fedora-toolbox-35 already exists` and return code 1

**Actual behaviour**
No output and return code 0

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.99.2`

**Toolbox package info (`rpm -q toolbox`)**
e.g., `toolbox-0.0.18-2.fc32.noarch`

**Output of `podman version`**
```
Version:      3.4.2
API Version:  3.4.2
Go Version:   go1.16.8
Built:        Fri Nov 12 20:25:37 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-3.4.2-1.fc35.x86_64`

**Info about your OS**
e.g., Fedora Silverblue 35

**Other information**

Same result when running from interactive shell inside toolbox as well as with `toolbox run`

JayDoubleu:
In fact, seems like `toolbox create` just returns 0 when ran in general 
```
toolbox create --distro fedora --release f34
echo $?
0
```

debarshiray:
Good catch! It seems that while tunneling things back and forth over `flatpak-spawn`, the standard error output and the exit codes are getting lost.

JayDoubleu:
There is still few bugs here and there with flatpak-spawn it seems (https://github.com/flatpak/flatpak/issues/3697). I'm not sure if it was designed to perform things toolbx requires. 

Out of curiosity, would it be a difficult task for this logic to be embedded into a toolbx itself in case flatpak-spawn is trying to solve different problems and has its own roadmap?

debarshiray:
@HarryMichal we might have to take this off the milestone for 0.1.0 if this ends up requiring some major surgery.

debarshiray:
[...]

PR #1052, opened 2022-05-13
Fixes for Toolbx forwarding

HarryMichal:
See commit messages for detailed information.

Fixes https://github.com/containers/toolbox/issues/957

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -115,11 +115,9 @@ func create(cmd *cobra.Command, args []string) error {
 			return errors.New("this is not a toolbox container")
 		}
 
-		if _, err := utils.ForwardToHost(); err != nil {
-			return err
-		}
-
-		return nil
+		cmd.SilenceErrors = true
+		exitCode, err := utils.ForwardToHost()
+		return &exitError{exitCode, err}
 	}
 
 	if cmd.Flag("distro").Changed && cmd.Flag("image").Changed {
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -204,7 +204,7 @@ func ForwardToHost() (int, error) {
 		logrus.Debugf("%s", arg)
 	}
 
-	exitCode, err := shell.RunWithExitCode("flatpak-spawn", os.Stdin, os.Stdout, nil, flatpakSpawnArgs...)
+	exitCode, err := shell.RunWithExitCode("flatpak-spawn", os.Stdin, os.Stdout, os.Stderr, flatpakSpawnArgs...)
 	if err != nil {
 		return exitCode, err
 	}
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -166,3 +166,14 @@ teardown() {
   assert_line --index 2 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
   assert [ ${#lines[@]} -eq 3 ]
 }
+
+@test "run: Try to run a Toolbx command from a container (check error message propagation)" {
+  create_default_container
+
+  run $TOOLBOX run toolbox foo
+
+  assert_failure
+  assert [ $status -eq 1 ]
+  assert_line --index 0 "Error: unknown command \"foo\" for \"toolbox\""
+  assert_line --index 1 "Run 'toolbox --help' for usage."
+}
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -177,3 +177,12 @@ teardown() {
   assert_line --index 0 "Error: unknown command \"foo\" for \"toolbox\""
   assert_line --index 1 "Run 'toolbox --help' for usage."
 }
+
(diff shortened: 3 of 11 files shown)


debarshiray:
recheck

src/pkg/utils/utils.go
```diff
@@ -204,7 +204,7 @@ func ForwardToHost() (int, error) {
 		logrus.Debugf("%s", arg)
 	}
 
-	exitCode, err := shell.RunWithExitCode("flatpak-spawn", os.Stdin, os.Stdout, nil, flatpakSpawnArgs...)
+	exitCode, err := shell.RunWithExitCode("flatpak-spawn", os.Stdin, os.Stdout, os.Stderr, flatpakSpawnArgs...)
```
debarshiray, line 207: At least at first glance, this change looks correct to me, but ...

test/system/104-run.bats
```diff
@@ -166,3 +166,14 @@ teardown() {
   assert_line --index 2 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
   assert [ ${#lines[@]} -eq 3 ]
 }
+
+@test "run: Try to run a Toolbx command from a container (check error message propagation)" {
+  create_default_container
+
+  run $TOOLBOX run toolbox foo
```
debarshiray, line 173: ... not this test case, because it doesn't demonstrate the problem that the above code change tries to address.

When trying to run an invalid command line, such as with an unknown command, the error is returned very early in the lifetime of the `toolbox(1)`  binary before anything has been forwarded over `flatpak-spawn(1)`.  Specifically, within the `rootCmd.Execute()` call.

Instead, we need something that happens after the CLI invocation has been forwarded over `flatpak-spawn(1)`.  Such as, within a command's `RunE` function.  I think `toolbox create --distro unknown --release invalid` can work.

I would still keep this test case, because it still tests *some* thing, just not *this* thing.  :)

debarshiray (changes requested):
I managed to finally fish this out from my backlog.  Sorry about that.

test/system/104-run.bats
debarshiray: > I would still keep this test case, because it still tests some thing, just not this thing. :)

I tried to resurrect just the test to prove my theory in https://github.com/containers/toolbox/pull/1525

src/cmd/create.go
```diff
@@ -115,11 +115,9 @@ func create(cmd *cobra.Command, args []string) error {
 			return errors.New("this is not a toolbox container")
 		}
 
-		if _, err := utils.ForwardToHost(); err != nil {
-			return err
-		}
-
-		return nil
+		cmd.SilenceErrors = true
```
debarshiray, line 118: I wonder if we can set `SilenceErrors` on the root command and take more responsibility for printing the errors.  That way we wouldn't need this boilerplate in each command.

debarshiray: I have submitted https://github.com/containers/toolbox/pull/1533 to verify the idea.  Let's see if it works out.

debarshiray (commented):
Don't we also need to handle the help functions in the same way?

debarshiray:
> Don't we also need to handle the help functions in the same way?

To answer my own question:

First of all, if `man(1)` is missing from the host, we show a hard-coded very brief summary.  So, we don't need to propagate any errors or exit codes because of that.

Secondly, the help functions deal with the `--help` flag.  By that point, the command line parsing has ensured that the flag is being used against a legitimate command.  So, we don't need to worry about missing manual pages for bogus commands.

Things like `toolbox help foo` are handled by the `help` command's main `help()` function, which already propagates errors and exit codes like the other commands.

This leaves us with the scenario where `flatpak-spawn(1)` is missing from the container.  Unfortunately, Cobra's [help functions](https://pkg.go.dev/github.com/spf13/cobra#Command.SetHelpFunc) cannot return an `error`, and currently we have this behaviour:
```
⬢$ toolbox create --help
Error: flatpak-spawn(1) not found
⬢$ echo $?
0
```

Ideally, we would have returned a non-zero exit code, but it's probably not a big enough problem and we can set it aside for another day.  Ultimately, Toolbx containers are expected to have `flatpak-spawn(1)`.

src/cmd/create.go
debarshiray: It seems to work.

debarshiray: So, I resurrected this change based on setting `SilenceErrors` on the root command in https://github.com/containers/toolbox/pull/1534

debarshiray (commented):
I believe we have now resurrected and merged all the commits to fix https://github.com/containers/toolbox/issues/957 through:
* https://github.com/containers/toolbox/pull/1525
* https://github.com/containers/toolbox/pull/1532
* https://github.com/containers/toolbox/pull/1533
* https://github.com/containers/toolbox/pull/1534
* https://github.com/containers/toolbox/pull/1538

debarshiray:
Closing.  My apologies that it took so long to get to it.

PR #1055, opened 2022-05-20
Post-merge fixes for https://github.com/containers/toolbox/pull/840

HarryMichal:


HarryMichal:
This fail is truly confusing to me. My local machine has no problem with it. If I don't find a solution soon, I'll just drop most of https://github.com/containers/toolbox/pull/1055/commits/a30f76cc5538f417fc89b7b0a4174e078ac76a6e.

HarryMichal:
Ah, ok. This is the same as https://github.com/containers/toolbox/pull/840#issuecomment-989827448. Will fall back to the previous approach of invoking `go run`.

debarshiray (commented):
Thanks for these follow up fixes, Ondřej!

I see that this branch needs to be rebased against `main`.  While I try to work through the conflicts, I have split this patch out into https://github.com/containers/toolbox/pull/1112

debarshiray (commented):
I split this commit to remove the unneeded documentation into https://github.com/containers/toolbox/pull/1113

debarshiray (commented):
Is this revert correct?

I don't see the name of the command (ie., the `Use` field of `cobra.Command`) being updated in `src/cmd/completion.go`.

src/cmd/completion.go
```diff
@@ -43,13 +42,15 @@ func completionEmpty(_ *cobra.Command, _ []string, _ string) ([]string, cobra.Sh
 }
 
 func completionCommands(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
-	commandNames := []string{}
+	// init-container is hidden but documented command that is meant to be
+	// accessed by users at least via a manual page
```
debarshiray, line 46: This comment is a bit confusing because one wonders what *hidden but documented* really is?  If it's hidden then why is it documented, and if it's documented then why is it hidden?

The main difference between `completion` and `init-container` is that the former is purely an implementation detail of our build that's of no interest to the user, while the latter does show up in things like `podman inspect`.

A user who is trying to troubleshoot or poke around at their Toolbx containers can be left wondering about `toolbox init-container ...`, which is a crucial detail of how Toolbx containers work.  Plus, the command does have some expectation of stability.  Otherwise existing Toolbx containers would break.  This is why it's worth documenting it.

However, `toolbox completion` is different.  It doesn't seem to show up in the generated shell completion files, and even if it did, it's not that fundamental to how Toolbx works.  Plus, there's really no need or expectation of stability.  It can change if Cobra changes.  Therefore, there's not much value in having a manual for it.

https://github.com/containers/podman/issues/14110 has a similar discussion but in the context of Podman, which does have a `podman-completion(1)` manual.

src/cmd/completion.go
```diff
@@ -43,13 +42,15 @@ func completionEmpty(_ *cobra.Command, _ []string, _ string) ([]string, cobra.Sh
...
+	commandNames := []string{"init-container"}
```
debarshiray, line 47: Anyway, when it comes to shell completion, I think that we should skip both `completion` and `init-container`.  We don't expect users to directly invoke either of them.  eg., `toolbox-init-container(1)` says:
```
It is not expected to be directly invoked by humans
```

Having it show up in the completion will only encourage enterprising souls to mess with it, leading to unforeseen consequences.

I see that it [used to be](https://github.com/containers/toolbox/commit/8b84b5e4604921fad818ede5267591de7f0993e6) part of the older hand-written competion file.  So, dropping it would be a change in behaviour, but I think it's right direction.

debarshiray (changes requested):

debarshiray (commented):
I split this commit to make the completion depend on the Toolbx binary to https://github.com/containers/toolbox/pull/1115

debarshiray (commented):
I split the typo fix for ZSH completions to https://github.com/containers/toolbox/pull/1118

debarshiray (commented):
I split the switch to Cobra's Bash completion v2 into https://github.com/containers/toolbox/pull/1119

debarshiray:
> Is this revert correct?
> 
> I don't see the name of the command (ie., the `Use` field of `cobra.Command`) being updated in `src/cmd/completion.go`.

I took the liberty to also include the `Use` field of `cobra.Command` in `src/cmd/completion.go` in https://github.com/containers/toolbox/pull/1121

debarshiray (changes requested):
Instead of hard-coding the locations for the shell completions, we should have build options to override them.  See https://github.com/containers/toolbox/pull/840#discussion_r866196849

I have tried this in https://github.com/containers/toolbox/pull/1123

debarshiray:
I have merged all the commits in this pull request, except this one:
```
[cmd/completion: Only suggest whitelisted and visible command for help](https://github.com/containers/toolbox/pull/1055/commits/3c0e092d3f068c4de0508385eb93d45285d8c36f)
```

... which is related to the two unresolved conversations above.  I don't feel qualified to fix it up, and since I am about to go on vacation, I don't have time anyway.  So, I will leave it to you.

I found two new problems with https://github.com/containers/toolbox/pull/840  Feel free to address them in this pull request, if you want to.

src/cmd/completion.go
HarryMichal: I second the opinion to not show `init-container` in the completion suggestions but due to the completion of the `help` command being related more towards the immediate help of the user.

We document, should keep on documenting and should propagate the way we bootstrap toolbxes. It is a vital part of the project and deserves some explanation into its inner workings.

As a result the last commit can be dropped from the PR.

HarryMichal:
> ... which is related to the two unresolved conversations above. I don't feel qualified to fix it up, and since I am about to go on vacation, I don't have time anyway. So, I will leave it to you.

There's not much to fix, only to drop :).

> I found two new problems with #840 Feel free to address them in this pull request, if you want to.

Okay :+1:.

src/cmd/completion.go
debarshiray: I forgot what the final conclusion for this discussion was.

As I mentioned [before](https://github.com/containers/toolbox/pull/1055#discussion_r966183914), `init-container` can be important for those who are trying to troubleshoot or poke around at their Toolbxes.  It shows up in `podman inspect` and does have some expectation of stability.  That's why we want to document it.

However, we also don't want casual users start messing with `init-container`.  That's why we shouldn't expose it in the shell completion for `toolbox init-co<TAB>`, but I suppose it's fine to expose it for `toolbox help init-co<TAB>`.  However, I don't know if it's feasible to make that distinction through Cobra.

src/cmd/completion.go
```diff
@@ -32,21 +31,16 @@ var completionCmd = &cobra.Command{
...
-				fmt.Fprintf(os.Stderr, "Error: %v", err)
-			}
+			return cmd.Root().GenZshCompletion(os.Stdout)
 		case "fish":
-			if err := cmd.Root().GenFishCompletion(os.Stdout, true); err != nil {
-				fmt.Fprintf(os.Stderr, "Error: %v", err)
-			}
+			return cmd.Root().GenFishCompletion(os.Stdout, true)
 		}
+		return nil
```
debarshiray, line 43: There should be a `panic("code should not be reached")` here, because it really shouldn't happen.  :)

debarshiray: I took the liberty to update the branch.

debarshiray (approved):

debarshiray:
Thanks for all the fixes, @HarryMichal !

PR #1057, opened 2022-05-20
test/system: Update error handling in system tests

martymichal:
~Depends on https://github.com/containers/toolbox/pull/1050~

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -406,7 +406,7 @@ function find_os_release() {
   elif [[ -f "/usr/lib/os-release" ]]; then
     echo "/usr/lib/os-release"
   else
-    echo ""
+    fail "os-release was not found"
   fi
 }
 
@@ -418,8 +418,7 @@ function get_system_id() {
   os_release="$(find_os_release)"
 
   if [[ -z "$os_release" ]]; then
-    echo ""
-    return
+    return 1
   fi
 
   echo $(awk -F= '/ID/ {print $2}' $os_release | head -n 1)
@@ -433,8 +432,7 @@ function get_system_version() {
   os_release="$(find_os_release)"
 
   if [[ -z "$os_release" ]]; then
-    echo ""
-    return
+    return 1
   fi
 
   echo $(awk -F= '/VERSION_ID/ {print $2}' $os_release | head -n 1)
--- a/test/system/000-setup.bats
+++ b/test/system/000-setup.bats
@@ -1,11 +1,17 @@
 #!/usr/bin/env bats
 
+load 'libs/bats-support/load'
+load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 @test "test suite: Setup" {
     local os_release="$(find_os_release)"
     local system_id="$(get_system_id)"
     local system_version="$(get_system_version)"
+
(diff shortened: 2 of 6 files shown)


test/system/libs/helpers.bash
```diff
@@ -418,8 +418,7 @@ function get_system_id() {
   os_release="$(find_os_release)"
 
   if [[ -z "$os_release" ]]; then
-    echo ""
-    return
+    return 1
   fi
 
   echo $(awk -F= '/ID/ {print $2}' $os_release | head -n 1)
```
debarshiray, line 424: This isn't about this specific change, but a more general comment.

This has always looked needlessly complicated to me.  eg., in the POSIX shell implementation, we simply sourced `os-release` file in a sub-shell and echoed the specific variable we want:
```
get_host_id()
(
    # shellcheck disable=SC1091                                                 
    . /usr/lib/os-release
    echo "$ID"
)

get_host_variant_id()
(
    # shellcheck disable=SC1091                                                 
    . /usr/lib/os-release
    echo "$VARIANT_ID"
)

get_host_version_id()
(
    # shellcheck disable=SC1091                                                 
    . /usr/lib/os-release
    echo "$VERSION_ID"
)
```

debarshiray (changes requested):

test/system/libs/helpers.bash
```diff
@@ -258,14 +257,13 @@ function pull_distro_image() {
   # No need to copy if the image is already available in Podman
   run $PODMAN image exists ${image}
   if [[ "$status" -eq 0 ]]; then
-    return
+    return 0
```
debarshiray, line 260: This isn't about this specific change, but a more general comment.

Why are we running `$PODMAN image exists` under Bats' `run`?  It's meant for cases where we are want to *test* a command.  Here, we are only doing a boolean check, so we don't need all the overhead of `run`.  Can we do something like this:
```
if "$PODMAN" image exists "$image" >/dev/null 2>&1; then
    return 0
fi
```

test/system/libs/helpers.bash
```diff
@@ -258,14 +257,13 @@ function pull_distro_image() {
...
   fi
 
   # https://github.com/containers/skopeo/issues/547 for the options for containers-storage
   run $SKOPEO copy "dir:${IMAGE_CACHE_DIR}/${image_archive}" "containers-storage:[overlay@$ROOTLESS_PODMAN_STORE_DIR+$ROOTLESS_PODMAN_STORE_DIR]${image}"
   if [ "$status" -ne 0 ]; then
-    echo "Failed to load image ${image} from cache ${IMAGE_CACHE_DIR}/${image_archive}"
-    assert_success
+    fail "Failed to load image ${image} from cache ${IMAGE_CACHE_DIR}/${image_archive}"
```
debarshiray, line 266: Ditto.

debarshiray (commented):
The change seems OK, but I am also unsure why it's needed.

It says that it's a fallout from commit ae8bd1a9c99c27e5fb4d1707d4b55ca9d7c3be04 which in turn is a fallout from commit da6b6a7c5a6e5e284e6642c29159a872ceec94e2 but that one doesn't exist in our Git history.  So, I have no idea what makes the code *proper* now.  :)

PR #1063, opened 2022-05-30
src: Run 'go test' unit tests with LC_ALL=C

HarryMichal:
The system language interferes with some of the tests. And while
language specific code-paths may need to be tested, we don't have any at
the moment.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -70,4 +70,10 @@ if shellcheck.found()
 endif
 
 test('go fmt', meson_go_fmt_program, args: [meson.current_source_dir()])
-test('toolbox go unit tests', go, args: ['test', '-v', './...'], workdir: meson.current_source_dir())
+test(
+  'toolbox go unit tests',
+  go,
+  args: ['test', '-v', './...'],
+  env: 'LC_ALL=C',
+  workdir: meson.current_source_dir()
+)


src/meson.build
```diff
@@ -70,4 +70,10 @@ if shellcheck.found()
 endif
 
 test('go fmt', meson_go_fmt_program, args: [meson.current_source_dir()])
-test('toolbox go unit tests', go, args: ['test', '-v', './...'], workdir: meson.current_source_dir())
+test(
+  'toolbox go unit tests',
+  go,
+  args: ['test', '-v', './...'],
+  env: 'LC_ALL=C',
```
debarshiray, line 77: Does it really have to be the `C` locale?  Can't we use an UTF-8 locale like `en_US.UTF-8`?

I am saying this because for quite a long time, GNOME Terminal [doesn't even start](https://wiki.gnome.org/Apps/Terminal/FAQ#Exit_status_8) without a UTF-8 locale.  So, a non-UTF-8 locale isn't very representative of the environment in which Toolbx is likely to be used.

debarshiray (commented):
Sorry, I didn't notice this one before.  Do you have an example of the breakage?  I am curious.

src/meson.build
matthiasclasen: I agree that C shouldn't be used

debarshiray:
> Do you have an example of the breakage? I am curious.

@HarryMichal had once mentioned that he had his computer set to Finnish when this happened.  So, I tried:
```
$ LANG=fi_FI.UTF-8 meson setup ...
$ LANG=fi_FI.UTF-8 meson compile ...
$ LANG=fi_FI.UTF-8 meson test ...
```

... and:
```
$ LC_ALL=fi_FI.UTF-8 meson setup ...
$ LC_ALL=fi_FI.UTF-8 meson compile ...
$ LC_ALL=fi_FI.UTF-8 meson test ...
```

... but couldn't trigger any failures.

debarshiray:
I am going to close this one, because I don't want to fiddle with locales without any sort of reproducer.  We can revisit this if the problem comes up again.

PR #1067, opened 2022-06-16
cmd/run: add `--preserve-fds` command-line argument

### LINKED ISSUE #1066: Add `--forward-fd` to `toolbox run`
Basically, similar in spirit to `flatpak-spawn --forward-fd=` and `podman exec --preserve-fds=` it would be lovely to be able to do the same with Toolbox.

An alternative to this would be a "dry-run" mode for `toolbox run` that shows me which `podman` commandline would be used, allowing me to hack it for myself.

As a workaround, for the time being, I'm running `toolbox run env` to start the toolbox and get a list of the environment variables and then constructing an approximate `podman exec` invocation using the output.

allisonkarlitskaya:
This mirrors the `--preserve-fds` argument of podman.

Fixes #1066

Welcome to "I've never written anything in Go before" hour :)

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -34,9 +34,10 @@ import (
 
 var (
 	runFlags struct {
-		container string
-		distro    string
-		release   string
+		container   string
+		distro      string
+		preserveFDs uint
+		release     string
 	}
 
 	runFallbackCommands = [][]string{{"/bin/bash", "-l"}}
@@ -65,6 +66,11 @@ func init() {
 		"d",
 		"",
 		"Run command inside a toolbox container for a different operating system distribution than the host")
+
+	flags.UintVar(&runFlags.preserveFDs,
+		"preserve-fds",
+		0,
+		"Pass down to command N additional file descriptors (in addition to 0, 1, 2)")
 
 	flags.StringVarP(&runFlags.release,
 		"release",
@@ -134,6 +140,7 @@ func run(cmd *cobra.Command, args []string) error {
 		defaultContainer,
 		image,
 		release,
+		runFlags.preserveFDs,
 		command,
 		false,
 		false,
@@ -156,6 +163,7 @@ func run(cmd *cobra.Command, args []string) error {
 func runCommand(container string,
 	defaultContainer bool,
 	image, release string,
+	preserveFDs uint,
 	command []string,
 	emitEscapeSequence, fallbackToBash, pedantic bool) error {
 	if !pedantic {
@@ -277,14 +285,21 @@ func runCommand(container string,
 
 	logrus.Debugf("Container %s is initialized", container)
 
-	if err := runCommandWithFallbacks(container, command, emitEscapeSequence, fallbackToBash); err != nil {
+	if err := runCommandWithFallbacks(container,
+		preserveFDs,
+		command,
+		emitEscapeSequence,
+		fallbackToBash); err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func runCommandWithFallbacks(container string, command []string, emitEscapeSequence, fallbackToBash bool) error {
+func runCommandWithFallbacks(container string,
+	preserveFDs uint,
+	command []string,
+	emitEscapeSequence, fallbackToBash bool) error {
 	logrus.Debug("Checking if 'podman exec' supports disabling the detach keys")
 
 	var detachKeysSupported bool
@@ -295,6 +310,7 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 	}
 
 	envOptions := utils.GetEnvOptionsForPreservedVariables()
+	preserveFDsString := fmt.Sprint(preserveFDs)
 
 	var stderr io.Writer
 	var ttyNeeded bool
@@ -320,6 +336,7 @@ func runCommandWithFallbacks(container string, command []string, emitEscapeSeque
 
 	for {
 		execArgs := constructExecArgs(container,
+			preserveFDsString,
 			command,
 			detachKeysSupported,
 			envOptions,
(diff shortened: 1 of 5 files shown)


allisonkarlitskaya:
@debarshiray ping?

src/cmd/run.go
```diff
@@ -33,9 +34,10 @@ import (
 
 var (
 	runFlags struct {
-		container string
-		distro    string
-		release   string
+		container    string
+		distro       string
+		release      string
+		preserve_fds int
```
debarshiray, line 40: Nitpick: camel case is the norm in Go, so something like `preserveFDs` will be better.

src/cmd/run.go
```diff
@@ -475,6 +483,10 @@ func constructExecArgs(container string,
 
 	execArgs = append(execArgs, envOptions...)
 
+	if runFlags.preserve_fds != 0 {
+		execArgs = append(execArgs, "--preserve-fds="+strconv.Itoa(runFlags.preserve_fds))
```
debarshiray, line 487: This part will require rebasing against `main`.

debarshiray (commented):
Thanks for working on this @allisonkarlitskaya !  My apologies - it fell off my radar.

It looks mostly good to me.  It will need some rebasing, but I will take care of it, because right now I am staring at the need for `podman exec --preserve-fds ...` myself for the internal use of Toolbx.  :)

src/cmd/run.go
```diff
@@ -71,6 +73,12 @@ func init() {
 		"",
 		"Run command inside a toolbox container for a different operating system release than the host")
 
+	flags.IntVarP(&runFlags.preserve_fds,
```
debarshiray, line 76: Nitpick: with [flags.IntVar](https://pkg.go.dev/github.com/spf13/pflag#IntVar) (ie., without the trailing `P`) ...

src/cmd/run.go
```diff
@@ -71,6 +73,12 @@ func init() {
...
+		"preserve-fds",
+		"",
```
debarshiray, line 78: ... this empty parameter can be dropped.

src/cmd/run.go
allisonkarlitskaya: Done.

src/cmd/run.go
allisonkarlitskaya: git seems to have done a good enough job here, on its own.

src/cmd/run.go
allisonkarlitskaya: Done.

src/cmd/run.go
allisonkarlitskaya: Done.

src/cmd/run.go
```diff
@@ -481,6 +488,10 @@ func constructExecArgs(container string,
 
 	execArgs = append(execArgs, envOptions...)
 
+	if runFlags.preserveFDs != 0 {
```
debarshiray, line 491: I suspect that this should be `> 0` because that's also what [Podman](https://github.com/containers/podman/commit/0b34327ad40e04861dac7f73870d87633a5c637e) uses.  Otherwise, we will only error out later from `podman exec ...` itself:
```
$ toolbox run --preserve-fds -12 true
Error: failed to invoke 'podman exec' in container fedora-toolbox-36
$ echo $?
125
```

This makes me wonder if we should be using an unsigned integer flag.  :)

debarshiray, line 491: Another thing that I just realized ...

It might be better if `runFlags.preserveFDs` is passed down to `runCommand()` as a parameter, because the code from `runCommand()` downwards is shared between the `enter` and `run` commands.

I am still getting used to Go where nothing is `static` to one file.  Everything is either global within a package or exported from the package.  So, far, it seems like people tend to restrict the parsed flag variables to the topmost function.

debarshiray: I see that newer versions of Podman [do use](https://github.com/containers/podman/commit/4d895dcb5472da19b886ca1662182556242fe5d4) an unsigned integer flag.

*It's not obvious from the commit message, but that's where it changed.*

debarshiray (commented):
Thanks for the updates!

src/cmd/run.go
debarshiray: Damn!  Converting an *unsigned* integer to string is surprisingly annoying.  :)

[strconv.Itoa](https://pkg.go.dev/strconv#Itoa) takes a *signed* `int`, which would require a cast, and there's no unsigned counterpart.  There's [strconv.FormatUint](https://pkg.go.dev/strconv#FormatUint) which takes an unsigned `uint64`, which is better, but would still require a cast.

So, [fmt.Sprint](https://pkg.go.dev/fmt#Sprint) it is, if we want to avoid the casting.  It's a bit more expensive than the other two functions, but we don't have to worry about it unless it's proven to be a performance bottle neck.

debarshiray: I took the liberty to change this.

debarshiray:
> As per our conversation on IRC, I'll step back and let @debarshiray address the outstanding issues.

I managed to add tests.  It took longer than I expected because the [test harness](https://github.com/bats-core/bats-core) reserves two extra file descriptors.

src/cmd/run.go
debarshiray: Yes, indeed.  :)

src/cmd/run.go
debarshiray: Thanks.

src/cmd/run.go
debarshiray: Thanks.

debarshiray:
recheck

debarshiray:
Damn:
```
#   Trying to pull quay.io/toolbox_tests/registry:latest...
#   Getting image source signatures
...
#   Error: writing blob: storing blob to file "/var/tmp/storage1659742604/3": happened during read: unexpected EOF
```

debarshiray:
recheck

debarshiray:
I suspect that `unit-test-migration-path-for-coreos-toolbox` was failing because I forgot to update the `runCommand` call in `src/rootMigration.go` and the code was failing to compile.

debarshiray:
Thanks for working on this, @allisonkarlitskaya !

PR #1068, opened 2022-06-17
images: add `findutils` and `sudo` to missing docs for f36+

jlebon:
Noticed today that `man xargs` was returning the POSIX manpage instead
of the one shipped by `findutils`.

### DIFF
--- a/images/fedora/f36/missing-docs
+++ b/images/fedora/f36/missing-docs
@@ -1,6 +1,7 @@
 acl
 bash
 curl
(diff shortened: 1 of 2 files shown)


jlebon:
Updated this now with sign-off!

jlebon:
Can we get this in?

nievesmontero:
This pull request is solved in https://github.com/containers/toolbox/pull/1133

PR #1007, opened 2022-01-28
Allow toolbox containers to be created with a custom hostname

buck-ross:
This feature consists of a new optional flag `--hostname HOSTNAME`, which can be used when invoking `toolbox create` to create a new container with a custom hostname.

This is essentially a re-implementation of #210 from scratch, written in golang.

This may also address many of the concerns brought up in the comments of #771, and could potentially assist in the adoption of a new default hostname, since it allows for more flexibility in that regard.

This PR includes:
- [x] A new feature, which is backwards compatible with all existing use-cases of `toolbox create`
- [x] Associated system tests for the new `--hostname` flag
- [x] Associated updates to the `toolbox create` documentation in the manual.

Issue #969, opened 2021-12-29
Use something other than hostname to indicate container name in command line prompt


castedo:
Below is more or less the same issue I describe in https://github.com/89luca89/distrobox/issues/62

**Describe the bug**
The container created by `toolbx` is setting the hostname in the container to `toolbox`. For example /etc/hostname is a value that is not actually the DNS name of any host on any network. See `man hostname` for all the assumptions made about and effects of setting /etc/hostname.

**Expected behaviour**
The hostname in the container should be the same hostname as the host because they are sharing the network.

On Debian there is a precedence for indicating the chroot in the command line prompt (chroot being a precursor to containers). To my knowledge there isn't an equivalent on Fedora based distros. Not knowing of an alternative I started using `osvirtalias` in [cnest](https://github.com/castedo/cnest).

I describe the situation in more detail here:
https://github.com/castedo/cnest/tree/master/prompt

So I can suggest doing similarly and set `osvirtalias` and/or `debian_chroot` to be the name of the container, and let the distro and container configuration determine the command line prompts however they are supposed to. And do this without incorrectly repurposing the hostname for a purpose other than what hostname is supposed to mean.

If you come across an alternative to `osvirtalias` please let me know. I don't want to start a precedent that is non-standard, but I do want to start a precedent other than `debian_chroot`. If it catches on, Debian distros probably want to use a name without chroot in it anyway.

castedo:
Here's an example of how any application that actually wants to programmatically get the real hostname or [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) will not actually get the real hostname or FQDN when run inside a toolbox container. Python example:
```
import socket
socket.gethostname()
socket.getfqdn()
```
It's not hard to imagine applications where this is not the desired behavior. For instance, a web application that reports to the users how to connect to the machine via browser URL etc...

castedo:
FYI update on what Luca (distrobox) and I (cnest) decided makes sense in 89luca89/distrobox#62

[...]

PR #1086, opened 2022-08-17
fix: don't use `toolbox` as constant for FQDN

akdev1l:
## Issue: https://github.com/containers/toolbox/issues/969

Setting the hostname to `toolbox` causes timeouts whenever anything tries
to resolve the name of the machine - for example `sudo` does this.

This change makes it so the FQDN is set to
`${container_name}.${hostname}` as recommended in the linked issue.

After this change commands can properly resolve the local FQDN.

I removed the symlink to `/run/host/etc/hosts` because podman already copies that information in and then we can use `--add-host` to add a mapping to localhost for the container - this way calling `ping $(hostname)` does what is expected.

Pending PRs:

1. https://github.com/containers/toolbox/pull/1007 - this adds a new flag to allow setting the hostname of the toolbox, I think this should just follow the `container-name.hostname` convention otherwise it seems confusing
2. https://github.com/containers/toolbox/pull/771/files - similar but sets the hostname of the container to be equal to the container name - the name is still unresolvable however.
3. https://github.com/containers/toolbox/pull/383/files - same as above but tries to sanitize the container name
4. https://github.com/containers/toolbox/pull/573/files - Obsolete, PR against bash toolbox

None of these PRs address my issue with delays due to unresolvable hostnames. So this one tries to do that.


akdev1l:
this seems to trigger a minor bug, when exiting the toolbox if I pressed `ctrl+c` at the prompt then toolbox falsely prints out an empty `Error: ` message

I'll have to dig into that

akdev1l:
sorry for the noise the bug I hit seems to be because toolbox exits with return code 130 (SIGTERM) whenever `ctrl+c` is pressed - seriously wondering how my changes triggered this

akdev1l:
```
capsh --caps= -- -c exec "$@" /bin/sh /bin/bash -l
```

 why is so convoluted ...  it would seems this is equivalent
 
 ```
 capsh --caps= -- -c 'exec /bin/bash -l'
 ```
 
 mm from further experimentation it seems that bash just does this and `toolbox` should just ignore the `130` error code
 
 
 toolbox without modification does this:
 
 ```
 [akdev@canzuk toolbox]$ toolbox enter test10
⬢[akdev@test10 toolbox]$ ^C
⬢[akdev@test10 toolbox]$ ^C
⬢[akdev@test10 toolbox]$ ^C
⬢[akdev@test10 toolbox]$
logout
[akdev@canzuk toolbox]$ echo $?
0
[akdev@canzuk toolbox]$ bash
[akdev@canzuk toolbox]$ ^C
[akdev@canzuk toolbox]$ ^C
[akdev@canzuk toolbox]$
exit
[akdev@canzuk toolbox]$ echo $?
130
```

imo this should keep the error code of the last command execute so it should be `130` at the end to match the behaviour of bash - we could probably simplify that `capsh` shenanigans, looks like there's at least an unnecessary fork there

**Comment by akdev1l** [comment]
I have continued development of toolbox on my own as I'm not really having this merged on a reasonable time frame.

my fork is at: https://github.com/akdev1l/toolbox/tree/akdev

I have fixed some long standing issues and added some features (particularly I have enabled a static build and containerized `toolbox` itself, resolved the DNS issues and added basic `export` support) - feel free to ping me if there's any interest on that. 

Otherwise I will leave this PR to die - feel free to close.

File: src/cmd/run.go [hunk 1]
```diff
@@ -477,10 +479,10 @@ func constructExecArgs(container string,
 
 	execArgs = append(execArgs, []string{
 		container,
-		"capsh", "--caps=", "--", "-c", "exec \"$@\"", "/bin/sh",
+		"capsh", "--caps=", "--", "-c",
```
debarshiray, line 482: This doesn't seem to have anything to do with what the commit message describes.  :)

src/cmd/run.go
```diff
@@ -477,10 +479,10 @@ func constructExecArgs(container string,
...
 	}...)
 
-	execArgs = append(execArgs, command...)
+	execArgs = append(execArgs, fmt.Sprintf("exec %s", strings.Join(command, " ")))
```
debarshiray, line 485: Ditto.  Even more importantly, how is this objectively better than the existing code?  It also seems wrong.

If some of the elements in `command` themselves contain spaces, then `strings.Join(command, " ")` will split them into separate arguments for the `exec`.  That would be the equivalent of `$*` or `$@` without the quotes.  Instead they need to be passed as is, without further splitting.  Hence the use of ["$@" with the quotes](https://stackoverflow.com/questions/21071943/difference-between-and-in-bash-script).

Let's not deal with the escaping of whitespaces and quoting, and leave it to the shell.

debarshiray (changes requested):
It will be good if the commit message had some explanation around exit code 130 and `SIGTERM`.  Particularly the user facing behaviour.

debarshiray:
> sorry for the noise the bug I hit seems to be because toolbox
> exits with return code 130 (SIGTERM) whenever `ctrl+c` is
> pressed - seriously wondering how my changes triggered this

I don't think your changes introduced that.  :)

A few months ago, Toolbx [started](https://github.com/containers/toolbox/commit/f8e21a31b317a65f57773c280926aa471ae060ff) propagating the exit code of the command, which triggered this.  You wouldn't have encountered this behaviour before that.

From the `bash(1)` [manual](https://www.man7.org/linux/man-pages/man1/bash.1.html):
```
When a command terminates on a fatal signal N,
bash uses the value of 128+N as the exit status.
```

`Ctrl+c` is `SIGINT` (not `SIGTERM`) and it's numerical value is 2:
```
$ kill -L
 1) SIGHUP	 2) SIGINT ...
```

images/fedora/f32/README.md
```diff
@@ -54,7 +54,7 @@ This will create a container called `fedora-toolbox-<version-id>`.
 ### Enter the toolbox:
 ```console
 [user@hostname ~]$ toolbox enter
-⬢[user@toolbox ~]$
+⬢[user@fedora-toolbox-<version-id> ~]$
```
debarshiray, line 57: This shows the problem with using the name of the container in the hostname. :)

As I mentioned [elsewhere](https://github.com/containers/toolbox/issues/98#issuecomment-1318592340), it makes shell prompts with `\h`, which is often the case, too long.  Users who are fine with longer prompts can customize it on their own using [metadata](https://github.com/containers/podman/issues/6192) from `/run/.containerenv`, but we shouldn't do that through the hostname because of how common it is for default prompts to have `\h` in them.

*For what it's worth, https://github.com/containers/toolbox/issues/98 is the issue about making it easier to distinguish one Toolbx container from another.*

src/cmd/create.go
```diff
@@ -418,8 +418,12 @@ func createContainer(container, image, release string, showCommandToEnter bool)
 
 	createArgs = append(createArgs, xdgRuntimeDirEnv...)
 
+	hostname, err := os.Hostname()
+	containerHostname := container + "." + hostname
```
debarshiray, line 422: For the purpose of implementing https://github.com/containers/toolbox/issues/969 let's use `toolbox.<host's hostname>` as the container's hostname.  We might still have to revisit the `toolbox` prefix, but it will still be an improvement.

The [issue](https://github.com/containers/toolbox/issues/98) about making it easier to distinguish one Toolbx container from another is different from https://github.com/containers/toolbox/issues/969  It's already possible today to use `/run/.containerenv` and `/run/.toolboxenv` to come up with one's own shell prompt without making any changes to the Toolbx code.  Having said that, it will definitely be good to improve the default prompt that Toolbx provides, but let's not worry about that in this pull request.

Baby steps.  :)

debarshiray:
> Setting the hostname to toolbox causes timeouts whenever
> anything tries to resolve the name of the machine - for example
> sudo does this.

I am curious.  I have never experienced delays when using `sudo(8)` inside a Toolbx container, but maybe you are doing something that I never do.  Could you please describe this in a bit more detail?

debarshiray:
> I have enabled a static build

The way things stand today, we are unlikely to build statically or disable CGO.  See: https://github.com/containers/toolbox/issues/832

Of course, this may change if our realities change or if new information comes to light.

images/fedora/f32/README.md
akdev1l: I am okay with  leaving this as `toolbox` for this change - will pick this up again and make this change

however my shell prompt with `\u @ \h` crops the FQDN at the first component so it doesn't get long, example:

```
[akdev@vancouver ~]$ export PS1='\u @ \h > '
akdev @ vancouver > hostname
vancouver.hq.akdev.xyz
```

with FQDN = `vancouver.hq.akdev.xyz` the prompt shows `vancouver`

src/cmd/run.go
akdev1l: yeah I think my change was incorrect here

akdev1l:
@debarshiray hi! thanks for your comments - I'll clean this up and just deal with the fqdn change

you gave me some historical background on toolbx and I appreciate that.

> I am curious. I have never experienced delays when using sudo(8) inside a Toolbx container, but maybe you are doing something that I never do. Could you please describe this in a bit more detail?

this happened to me because I build/distribute my own toolbx images (https://github.com/akdev1l/toolbox-images) - there is a requirement that isn't specified in the documentation which is having `nss-myhostname` installed and enabled in `/etc/nsswitch.conf` in the toolbox image. Without that `sudo` tries to resolve the container hostname and fails. (using the fqdn as originally intended in this PR solves this issue too as the name is resolvable, hence that was original drive for this change)

debarshiray:
> this happened to me because I build/distribute my own toolbx
> images (https://github.com/akdev1l/toolbox-images) - there is
> a requirement that isn't specified in the documentation which
> is having `nss-myhostname` installed and enabled in
> `/etc/nsswitch.conf` in the toolbox image. Without that `sudo`
> tries to resolve the container hostname and fails. (using the
> fqdn as originally intended in this PR solves this issue too as the
> name is resolvable, hence that was original drive for this change)

*nod*

@HarryMichal explained that to me later in real life.  My apologies, I forgot to mention that here.

akdev1l:
@debarshiray minimal cleaned this up, this should be the minimal change required for a proper FQDN - looks like it passes the tests

I still think we should change the prefix from "toolbox" to the name of the toolbox, bash prompts seem to show the first component of the FQDN (with this change there won't be any user visible prompt changes as we have hardcoded it the first component to `toolbox` for now)

```
[akdev@toronto toolbox]$ echo $PS1
[\u@\h \W]\$
```

so this would solve the issue of distinguishing the containers very elegantly imo (no changes required on the images, no big changes required on toolbox, no custom scripts nor custom prompts for bash or other shells)

jkemp814:
https://github.com/containers/toolbox/issues/98#issuecomment-1422038509

nievesmontero:
Hey @akdev1l, sorry for the delay on this PR. I am reviewing everything to  merge it as soon as possible. 

I just wanted to ask a quick question, at the beginnning of the issue, in the initial explanation, you say that the expected hostname is the following `${container_name}.${hostname}`. However, after reviewing the code I realised that it is actually `toolbox.${hostname}`, am I right?

runiq:
> At the beginnning of the issue, in the initial explanation, you say that the expected container hostname is the following `${container_name}.${hostname}`. However, after reviewing the code I realised that it is actually `toolbox.${hostname}`, am I right?

As someone who uses multiple toolboxes (one for regular development work, and a "grab-bag" one for other stuff), I would prefer the container name. (But it's not a big deal, honestly, since I can always change the prompt from within the toolbox itself.)

nievesmontero:
I also wanted to reproduce those timeouts you were talking about. Could you please let me know which commands led you to those timeouts so that I can reproduce them on my machine?

jnohlgard:
> I also wanted to reproduce those timeouts you were talking about. Could you please let me know which commands led you to those timeouts so that I can reproduce them on my machine?

Personally I encountered them mostly when using Jetbrains IDEs, e.g. IntelliJ, CLion, from within toolbox on a Fedora Kinoite (KDE plasma) system, but it should be possible to reproduce with other GUI apps on a KDE system because the problems seem to be with kwin trying to do some DNS resolution in relation to putting the host name in the title bar of the windows ("`xterm <@toolbox>`". Sorry I don't have any more exact recipe for this though.

juhp:
+1 for using the container name for the hostname.
Distrobox already does this I believe.
With multiple toolboxes it makes much more sense.

runiq:
I just realized something: With the Foot terminal (and others), I can [emit OSC 7 on directory change to have new terminals spawn in that directory](https://codeberg.org/dnkl/foot/wiki#user-content-how-to-configure-my-shell-to-emit-the-osc-7-escape-sequence). However, if you look at the linked code snippet:

> ```fish
> printf \e\]7\;file://%s%s\e\\ $hostname (string escape --style=url $PWD)
> ```

You can see that this depends on the correct hostname (so that it doesn't erroneously catch directory changes over SSH connections, for example).

So, the toolbox not having the host's hostname breaks at least that bit of functionality.

castedo:
> +1 for using the container name for the hostname. Distrobox already does this I believe. With multiple toolboxes it makes much more sense.

@juhp Distrobox has now switched to NOT set the hostname to container name because messing with the hostname caused too many problems. See
https://github.com/89luca89/distrobox/issues/62

lawhiteley:
Hello :wave:

Apologies to resurrect an old issue but this is prohibitive for those of us who wish to run certain graphical IDEs inside Toolbox. Jetbrains' products often make calls for the hostname, resulting in system-wide hangs fairly frequently ([this user](https://www.reddit.com/r/Fedora/comments/xvj3m9/what_makes_my_ide_slower_when_run_from_the_toolbox/jg2ffns/) seems to have encountered the same problem).

FWIW I'm running Kinoite, and I understand that there doesn't seem to be a straightforward fix, but switching to Distrobox with `--hostname localhost` has completely resolved the issue for me -- it'd be nice to have the same option here.

buonhobo:
So just to recap the situation with this long standing issue:

As of today, users have to manually add a "toolbox" entry to /etc/hosts in order to fix the issue?
Are there better workarounds?

diamondburned:
> Distrobox has now switched to NOT set the hostname to container name because messing with the hostname caused too many problems. See
> [89luca89/distrobox#62](https://github.com/89luca89/distrobox/issues/62)

Clarification: their fix was to use the same container hostname as the host's. The definite fix to include this would have to be something like distrobox's `--hostname` flag rather than a container name-derived hostname.

Issue #151, opened 2019-05-07
Links in welcome messages are a little confusing


allanday:
The links that are shown in the welcome messages are cool, but I think that they might confuse some people:

![image](https://user-images.githubusercontent.com/213068/57293860-d0757800-70bd-11e9-978d-57a0a3b8beed.png)

Links of this sort aren't typical in the terminal, and I'm not sure that the dotted underline will always be understood to mean that it's a link.

The other issue is that, if it is a link, clicking it should open something, and it doesn't.

It might be better to change the link text to be the address, so people are able to easily identify it as a link.

aniruddha2000:
Hello! I am new to this project and would like to contribute :D 
Seems it's a documentation issue so I would like to work on this.

debarshiray:
I mean, `hostnamectl` and `systemctl status ...` are vastly more exposed to users than the one-time Toolbx banner.  So it's worth trying to see what the consensus is, and if it can be improved, before doing a one-off in Toolbx.

PR #1090, opened 2022-08-25
Fix Issue 151: updates welcome message to remove formatted link

tonytheleg:
- Fixes [Issue 151](https://github.com/containers/toolbox/issues/151) which has a PR pending but looks to have failed tests and is from 2020. Not sure if it wasn't merged for that reason.
- Tested locally with Fedora 35, 36, and 37 releases in toolbox

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -27,10 +27,7 @@ if [ -f /run/ostree-booted ] \
     echo "host system. You may want to try out the Toolbox for a directly"
     echo "mutable environment that allows package installation with DNF."
     echo ""
-    printf "For more information, see the "
-    # shellcheck disable=SC1003
-    printf '\033]8;;https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/\033\\documentation\033]8;;\033\\'
-    printf ".\n"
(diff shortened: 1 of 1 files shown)


matthiasclasen:
I'm not convinced that is an improvement, tbh.

Reading the original issue, I would say the real problem is that clicking did not open the link

tonytheleg:
> I'm not convinced that is an improvement, tbh.
> 
> Reading the original issue, I would say the real problem is that clicking did not open the link

Yeah I can't disagree there, the link worked fine in Gnome Terminal  and personally I think looks more visually appealing. I was already looking at another issue and saw this one and figured I'd take it since it was an easy adjustment

debarshiray:
> I'm not convinced that is an improvement, tbh.
> 
> Reading the original issue, I would say the real problem is that clicking did not open the link

Yes, I agree.  As I mentioned in the original [issue](https://github.com/containers/toolbox/issues/151), the exact handling of hyperlinks is something that's worth discussing with the terminal emulators, before removing it from Toolbx.

debarshiray:
Closing.

@tonytheleg Thanks for your interest in Toolbx, though!

PR #1101, opened 2022-09-01
Increase the validation coverage for the container & release options

debarshiray:
Currently, the container name and release are only validated if they
were specified as command line options.  Neither the value of release
in the configuration file nor the container name generated from an
image are validated.

There's also a lot of repeated code in the command front-ends to
validate the container name and release.  This opens the door for
mistakes.  Any adjustment to the code must be repeated elsewhere, and
there are subtle interactions and overlaps between the validation code
and the code to resolve container and image names.

It's worth noting that the container and image name resolution happens
for both the command line and configuration file options, and generates
the container name from the image when necessary.

Therefore, validating everything while resolving cleans up the command
front-ends and increases the coverage of the validation.

This introduces the use of sentinel error values and custom error
implementations to identify the different errors that can occur while
resolving the container and images, so that they can be appropriately
shown to the user.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -147,28 +147,12 @@ func create(cmd *cobra.Command, args []string) error {
 		containerArg = "--container"
 	}
 
-	if container != "" {
-		if !utils.IsContainerNameValid(container) {
-			err := createErrorInvalidContainer(containerArg)
-			return err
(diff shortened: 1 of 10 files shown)


debarshiray:
recheck

PR #1117, opened 2022-09-08
build: Wire go-build-wrapper's output to Meson's

debarshiray:
Note that Meson's '@OUTPUT@' is not just the basename of the output, but
includes the relative path under the project's root directory.

### DIFF
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright © 2020 – 2021 Red Hat Inc.
+# Copyright © 2020 – 2022 Red Hat Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


debarshiray:
recheck

debarshiray:
recheck

eli-schwartz:
> Note that Meson's '@output@' is not just the basename of the output, but  includes the relative path under the project's root directory.

Hmm, then wouldn't it make sense to use the basename of the `@OUTPUT@` argument? Since you're manually constructing the output path yourself with current_build_dir and all.

Alternatively, you could assign `outfile=$(readlink --canonicalize "$2")` ***before*** performing `cd`. Then simply don't bother passing current_build_dir as an argument from within meson.

debarshiray:
> > Note that Meson's '@output@' is not just the basename of the output, but
> > includes the relative path under the project's root directory.
> 
> Hmm, then wouldn't it make sense to use the basename of the `@OUTPUT@`
> argument? Since you're manually constructing the output path yourself with
> current_build_dir and all.

Umm... I was trying to minimize the amount of shell scripting and use as much from Meson as possible.  Or, maybe I misunderstood something?

> Alternatively, you could assign `outfile=$(readlink --canonicalize "$2")` _**before**_
> performing `cd`. Then simply don't bother passing current_build_dir as an argument
> from within meson.

Would it be possible for you submit a pull request illustrating this?  I don't think I fully understand, and I don't have much time because I am about to leave on vacation.

PR #1123, opened 2022-09-12
build: Enable changing the completion paths & drop install_completions

debarshiray:
The `bash-completion` and `fish` dependencies were already optional - the shell completions for Bash and fish won't be generated and installed if they are absent; and there's no dependency required for Z shell.  So the `install_completions` build option wasn't reducing the dependency burden.

The build option was a way to disable the generation and installation of the shell completions, regardless of whether the necessary dependencies are present or not.  The only use-case for this is when installing to a non-system-wide prefix while hacking on Toolbox as a non-root user, because the locations for the completions advertised by the shells' APIs might not be accessible.  Being able to disable the completions prevents the installation from failing.

A different way of ensuring a smooth developer experience for a Toolbx hacker is to offer a way to change the locations where the shell completions are installed, which is necessary and beneficial for other use-cases.

Z shell, unlike Bash's `bash-completion.pc` and fish's `fish.pc`, doesn't offer an API to detect the location for the shell completions.  This means that Debian and Fedora use different locations [1, 2].  Namely, `/usr/share/zsh/vendor-completions` and `/usr/share/zsh/site-functions`.

An option to specify the locations for the shell completions can optimize the build, if there's an alternate API for the location that
doesn't involve using `bash-completion.pc` and `fish.pc` as build dependencies.  eg., Fedora provides the `_tmpfilesdir` RPM macro to specify the location for vendor-supplied `tmpfiles.d(5)` files, which makes it possible to avoid having `systemd.pc` as a build dependency [3].

Fallout from bafbbe81c9220cb3749a19a244e45a61477553a6

[1] Debian zsh commit bf0a44a8744469b5
    https://salsa.debian.org/debian/zsh/-/commit/bf0a44a8744469b5
    https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=620452

[2] https://src.fedoraproject.org/rpms/zsh/blob/f37/f/zsh.spec

[3] Fedora toolbox commit 9bebde5bb60f36e3
    https://src.fedoraproject.org/rpms/toolbox/c/9bebde5bb60f36e3

https://github.com/containers/toolbox/pull/840

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -18,8 +18,26 @@ go_md2man = find_program('go-md2man')
 shellcheck = find_program('shellcheck', required: false)
 skopeo = find_program('skopeo', required: false)
 
-bash_completion_dep = dependency('bash-completion', required: false)
-fish_dep = dependency('fish', required: false)
+bashcompletionsdir = get_option('bash_completions_dir')
+if bashcompletionsdir == ''
+  bash_completion_dep = dependency('bash-completion', required: false)
+  if bash_completion_dep.found()
+    bashcompletionsdir = bash_completion_dep.get_variable(pkgconfig: 'completionsdir')
+  endif
+endif
+
+fishcompletionsdir = get_option('fish_completions_dir')
+if fishcompletionsdir == ''
+  fish_completion_dep = dependency('fish', required: false)
+  if fish_completion_dep.found()
+    fishcompletionsdir = fish_completion_dep.get_variable(pkgconfig: 'completionsdir')
+  endif
+endif
+
+zshcompletionsdir = get_option('zsh_completions_dir')
+if zshcompletionsdir == ''
+  zshcompletionsdir = get_option('datadir') / 'zsh' / 'site-functions'
+endif
 
 migration_path_for_coreos_toolbox = get_option('migration_path_for_coreos_toolbox')
 profiledir = get_option('profile_dir')
@@ -70,8 +88,6 @@ subdir('data')
 subdir('doc')
 subdir('profile.d')
 subdir('src')
-if get_option('install_completions')
-  subdir('completion')
-endif
+subdir('completion')
 
 meson.add_install_script('meson_post_install.py')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,4 +1,22 @@
 option(
+  'bash_completions_dir',
+  description: 'Directory for Bash completion scripts',
+  type: 'string',
+)
+
+option(
+  'fish_completions_dir',
+  description: 'Directory for fish completion scripts',
+  type: 'string',
+)
+
+option(
+  'zsh_completions_dir',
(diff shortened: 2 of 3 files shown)


meson.build
```diff
@@ -33,6 +46,11 @@ if tmpfilesdir == '' or not fs.exists('/run/.containerenv')
   endif
 endif
 
+zshcompletionsdir = get_option('zsh_completions_dir')
+if zshcompletionsdir == ''
+  zshcompletionsdir = join_paths(get_option('datadir'), 'zsh', 'site-functions')
+endif
+
```
HarryMichal, line 53: Could you group this together with the other completion related lines?

meson_options.txt
```diff
@@ -19,8 +31,7 @@ option(
 )
 
 option(
-  'install_completions',
-  description: 'Install bash, zsh and fish command completions',
-  type: 'boolean',
-  value: true,
+  'zsh_completions_dir',
+  description: 'Directory for Z shell completion scripts (default=$datadir/zsh/site-functions)',
```
HarryMichal, line 35: Is it not possible to use `get_option('data_dir')` here so that the `value` option in `option()` can be used?

debarshiray: You mean, to use `get_option('data_dir')` in `meson_options.txt`, so that that the right value shows up in `meson configure`?  That I don't know at all.  I will try.

Or did you mean something else?

meson.build
debarshiray: I *think* I was keeping it in the same order as they are in `meson_options.txt`.  I can change the grouping in both places.

meson_options.txt
HarryMichal: Yes, I meant that.

debarshiray: This doesn't seem to work.

```
meson_options.txt:35:78: ERROR: Expecting rparen got id.
  description: 'Directory for Z shell completion scripts (default=get_option('datadir')/zsh/site-functions)',
                                                                              ^
For a block that started at 33,6
option(
      ^
```

```
meson_options.txt:35:77: ERROR: Expecting rparen got id.
  description: 'Directory for Z shell completion scripts default=get_option('datadir')/zsh/site-functions',
                                                                             ^
For a block that started at 33,6
option(
      ^
```

debarshiray: I wish I could somehow use `value` to represent the default value, so that we don't have the maintain default value in two places - one in `meson_options.txt` and then in `meson.build`.

However, it's a lot easier to compare against `''`, than check for a `$` prefix.  Even if it's a `$` prefix, we won't know if it's `$datadir` or some other variable.  It could also not be a variable but the beginning of a legitimate path, because this works:
```
$ touch $
```

So, I am inclined to give up.

meson.build
debarshiray: Done.

Issue #1110, opened 2022-09-08
Enable OpenGL and Vulkan inside the fedora-toolbox images


debarshiray:
Currently, a user needs to `dnf install` the relevant RPM packages inside a `fedora-toolbox` container to be able to use OpenGL and Vulkan applications.  Specifically:
* mesa-dri-drivers
* mesa-vulkan-drivers
* vulkan-loader

It will be good if the image already came with these packages pre-installed, just like Fedora Silverblue and Workstation.  Otherwise, if someone tries to use OpenGL or Vulkan inside Toolbx,  it's not obvious why things are failing and exactly what to do (or install) to solve it.

matthiasclasen:
We should probably figure out how to test that it actually works.

HarryMichal:
> We should probably figure out how to test that it actually works.
> 
> Maybe install glxgears or vkcube inside the toolbox and check that the run successfully.

To give some brain food on how to do this, I was thinking of creating a dedicated testing Toolbx image based on our existing images with all the testing tools installed in it. Those then could be used to test the capabilities in a reliable manner.

juhp:
I tried vkcube (by installing vulkan-tools) in current Fedora toolbox'es and it runs fine anyway.

Tested both with fedora-toolbox:36 using SB37 and fedora-toolbox:37 with SB36.

debarshiray:
> > We should probably figure out how to test that it actually works.
> > Maybe install glxgears or vkcube inside the toolbox and check that the run successfully.
> 
> To give some brain food on how to do this, I was thinking of creating a
> dedicated testing Toolbx image based on our existing images with all
> the testing tools installed in it. Those then could be used to test the
> capabilities in a reliable manner.

We could just `podman build ...` a custom image, if necessary, as part of our test suite?

It would be really good to avoid putting up another image on a registry somewhere.  We already use a few such auxiliary images.  They make the test suite flaky due to network issues and it makes things just a bit more opaque because it's yet another thing that's maintained elsewhere.

debarshiray:
I wonder if we could hack up a tiny bit of C or shell, with minimal dependencies, to poke at the right things to know that we got everything set up correctly at our end?

PR #1124, opened 2022-09-12
Added new packages

nievesmontero:
The following packages have been added to the
image to make OpenGL and Vulkan work:
mesa-dri-drivers
mesa-vulkan-drivers
vulkan-loader

https://github.com/containers/toolbox/issues/1110

### DIFF
--- a/images/fedora/f35/extra-packages
+++ b/images/fedora/f35/extra-packages
@@ -20,6 +20,8 @@ less
 lsof
 man-db
 man-pages
+mesa-dri-drivers
+mesa-vulkan-drivers
 mtr
 nano-default-editor
 nss-mdns
@@ -37,6 +39,7 @@ tree
 unzip
 util-linux
 vte-profile
+vulkan-loader
 wget
 which
 words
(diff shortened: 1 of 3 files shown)


nievesmontero:
The packages have been tested and they work on the image.

matthiasclasen (approved):
Looks all good to me.

debarshiray (commented):
Thanks for working on this, Nieves!

This looks good to me.  Could you please do the same thing for the Fedora 36 and 35 images? Those Fedoras are still supported, and since this isn't a big change, it will be good to make life easier for our users.

nievesmontero:
The new packages have been added to both, f36 and f35, they're pending to be tested.

matthiasclasen:
You will have to redo the first commit with a sign-off to make check go green

travier:
Hey folks. When you add packages to the Fedora toolbox image, could you also open an issue in the https://github.com/toolbx-images/images repo? That would it much easier for us to keep things in sync. Thanks!

Issue #60, opened 2019-02-23
Locale issue, need to install glibc-langpack-* package


mildred:
using `stack` to build a haskell program inside Fedora toolbox, I got a strange error that I could track down to a locale issue. There were other issues such as `less` not showing up UTF-8 byte sequences. Running `locale` showed up errors:

```
$ locale
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_ALL to default locale: No such file or directory
LANG=fr_FR.UTF-8
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL=
```

Installing the `glibc-langpack-fr` package (in my case) solved the issue. That should probably be performed when the toolbox is created the first time. Or it could be part of the documentation.

debarshiray:
Yes, I have heard of similar oddity reported by Czech users too, but since `locale` seemed to work fine for variants of `en` (not just `en_US`), I forgot about it.

Do you know how the `glibc-langpack-fr` package gets installed on your Fedora host?

debarshiray:
Thanks to @halfline I understand this a little better.

It's likely that replacing `glibc-minimal-langpack` with `glibc-all-langpacks` will solve the most pressing issues:
```
# dnf -y swap glibc-minimal-langpack glibc-all-langpacks
```

I think that a complete fix would also involve removing `/etc/rpm/macros.image-language-conf` and reinstalling all packages with missing translations. That would involve something similar to #55, which undoes the effects of `tsflags=nodocs`.

debarshiray:
It would also be helpful to have a concrete test case that I can use to validate my experiments.

juhp:
A simpler fix or workaround for now might be just to run podman with `LANG=C.utf8`.
That should fix most of the current problems I reckon.

[...]

PR #1130, opened 2022-09-29
Add package glibc-all-langpacks to f37 and f38

nievesmontero:
This new packet allows the user to set a locale inside the toolbox and make locale dependent commands work

https://github.com/containers/toolbox/issues/60

### DIFF
--- a/images/fedora/f37/extra-packages
+++ b/images/fedora/f37/extra-packages
@@ -7,6 +7,7 @@ findutils
 flatpak-spawn
 fpaste
 git
+glibc-all-langpacks
 gnupg
 gnupg2-smime
 gvfs-client
--- a/images/fedora/f38/extra-packages
+++ b/images/fedora/f38/extra-packages
@@ -7,6 +7,7 @@ findutils
 flatpak-spawn
 fpaste
 git
+glibc-all-langpacks
 gnupg
 gnupg2-smime
 gvfs-client


images/fedora/f37/extra-packages
```diff
@@ -7,6 +7,7 @@ findutils
 flatpak-spawn
 fpaste
 git
+glibc-all-langpacks
```
debarshiray, line 10: Just adding `glibc-all-langpacks` leaves us with both `glibc-minimal-langpack` and `glibc-all-langpacks` in the image.  It's probably harmless, but is a bit odd.  It will be better to use `dnf swap` as I mentioned before.

Secondly, the `/etc/rpm/macros.image-language-conf` file needs to be removed.  Otherwise, the translations get ripped out of the RPMs.  For example, try:
```
$ rpm --list -q sudo | grep locale
```

None of those files actually exist in the image.

We should reinstall all the packages in the base image whose translations went missing.  This is the same problem that we have with the missing documentation that we try to solve through `missing-docs`.

Ideally we would have some script to help us generate this list.  Otherwise it's too tedious and error-prone.  There's an initial attempt at `gen-docs-list`, which could probably be improved.

debarshiray (commented):
Did you see my [comment](https://github.com/containers/toolbox/issues/60#issuecomment-467174278) on the original [issue](https://github.com/containers/toolbox/issues/60#issuecomment-467174278) ?

debarshiray:
Since Fedora 36 is still supported, I went ahead and pulled these changes for F36 in https://github.com/containers/toolbox/pull/1225

images/fedora/f37/extra-packages
debarshiray: The first two items were fixed in https://github.com/containers/toolbox/pull/1139

However, we still need to figure out a way to restore all the missing translations that got ripped out from the base image.

PR #1133, opened 2022-10-13
Add findutils and sudo to missing docs for images f36, f37 and f38

nievesmontero:
This pull request is based on the following one https://github.com/containers/toolbox/pull/1068

### DIFF
--- a/images/fedora/f36/missing-docs
+++ b/images/fedora/f36/missing-docs
@@ -1,6 +1,7 @@
 acl
 bash
 curl
+findutils
 gawk
 grep
 gzip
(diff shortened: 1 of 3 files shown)


jlebon:
Hmm, it doesn't seem like the toolbox container in the Fedora registry has been rebuilt with this. In fact, it seems like it hasn't been updated in a while:

```
$ skopeo inspect docker://registry.fedoraproject.org/fedora-toolbox:36 | jq -r .Created
2022-08-01T05:04:56.72287Z
```

Is that by design or are we missing something to get these containers regularly rebuilt?

tpopela:
@jlebon see https://pagure.io/releng/issue/11092 / https://pagure.io/fedora-workstation/issue/337

Issue #1137, opened 2022-10-17
'duplicate' langpacks


matthiasclasen:
Debarshi pointed out that after we added glibc-all-langpacks to the extra-packages list, we now end up
with both glibc-all-langpacks and glibc-minimal-langpack in the container.

This is not causing any problems, but we could avoid this if we instead added a line

```
RUN dnf -y swap glibc-minimal-langpack glibc-all-langpacks
```

to the Containerfile

Issue #1136, opened 2022-10-17
More langpack work


matthiasclasen:
This is a followup to the recent change to add glibc-all-langpacks to the package list:

The Fedora base image contains a file that instructs rpm to strip out translations from installed packages: /etc/rpm/macros.image-language.conf

We should remove that file in our Containerfiles, so we get all translations for the packages we install.

The suggestion is to add an line like:

```
RUN rm /etc/rom/macros.image-language.conf
```

PR #1139, opened 2022-10-20
Add corrections to avoid duplicate packages

nievesmontero:
The following lines have been added to the fedora 37 and 38 images: RUN dnf -y swap glibc-minimal-langpack glibc-all-langpacks RUN rm /etc/rom/macros.image-language.conf
These two lines avoid redundant packages inside the images.

https://github.com/containers/toolbox/issues/1136
https://github.com/containers/toolbox/issues/1137

### DIFF
--- a/images/fedora/f37/Containerfile
+++ b/images/fedora/f37/Containerfile
@@ -13,6 +13,8 @@ COPY README.md /
 
 RUN sed -i '/tsflags=nodocs/d' /etc/dnf/dnf.conf
 RUN dnf -y swap coreutils-single coreutils-full
+RUN dnf -y swap glibc-minimal-langpack glibc-all-langpacks
+RUN rm /etc/rpm/macros.image-language.conf
 
 COPY missing-docs /
 RUN dnf -y reinstall $(<missing-docs)
--- a/images/fedora/f38/Containerfile
+++ b/images/fedora/f38/Containerfile
@@ -13,6 +13,8 @@ COPY README.md /
 
 RUN sed -i '/tsflags=nodocs/d' /etc/dnf/dnf.conf
 RUN dnf -y swap coreutils-single coreutils-full
+RUN dnf -y swap glibc-minimal-langpack glibc-all-langpacks
+RUN rm /etc/rpm/macros.image-language.conf
 
 COPY missing-docs /
 RUN dnf -y reinstall $(<missing-docs)
--- a/images/fedora/f37/extra-packages
+++ b/images/fedora/f37/extra-packages
@@ -7,7 +7,6 @@ findutils
 flatpak-spawn
 fpaste
 git
-glibc-all-langpacks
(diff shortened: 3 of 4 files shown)


images/fedora/f37/Containerfile
```diff
@@ -13,6 +13,8 @@ COPY README.md /
 
 RUN sed -i '/tsflags=nodocs/d' /etc/dnf/dnf.conf
 RUN dnf -y swap coreutils-single coreutils-full
+RUN dnf -y swap glibc-minimal-langpack glibc-all-langpacks
+RUN rm /etc/rpm/macros.image-language.conf
```
debarshiray, line 17: It will be better to make changes to the RPM configuration before running DNF.  Otherwise, those changes won't affect the DNF invocations.

I don't know if it actually makes a difference in this case, but it's better to be safe than sorry.  :)

images/fedora/f38/Containerfile
```diff
@@ -13,6 +13,8 @@ COPY README.md /
 
 RUN sed -i '/tsflags=nodocs/d' /etc/dnf/dnf.conf
 RUN dnf -y swap coreutils-single coreutils-full
+RUN dnf -y swap glibc-minimal-langpack glibc-all-langpacks
+RUN rm /etc/rpm/macros.image-language.conf
```
debarshiray, line 17: Ditto.

images/fedora/f37/Containerfile
debarshiray: I took the liberty to change this in https://github.com/containers/toolbox/pull/1225

images/fedora/f38/Containerfile
debarshiray: Changed in https://github.com/containers/toolbox/pull/1225

debarshiray:
Since Fedora 36 is still supported, I went ahead and pulled these changes for F36 in https://github.com/containers/toolbox/pull/1225

images/fedora/f37/Containerfile
debarshiray, line 17: Typo alert!  Did you try a `podman build ...` with these changes?  :)

The file is `/etc/rpm/macros.image-language-conf`, not `/etc/rpm/macros.image-language.conf`.

images/fedora/f38/Containerfile
debarshiray, line 17: Ditto.

images/fedora/f37/Containerfile
debarshiray: Fixed in https://github.com/containers/toolbox/pull/1225

images/fedora/f38/Containerfile
debarshiray: Fixed in https://github.com/containers/toolbox/pull/1225

PR #1144, opened 2022-10-23
cmd/list, cmd/run: Consolidate the dependencies for the IsTerminal() API

debarshiray:
It was decided in commit 950f510872404da8 that golang.org/x/* would be used for the IsTerminal() API, not github.com/mattn/go-isatty.  However, github.com/mattn/go-isatty had crept in through commits f49df914f4b64fed and a22d7821cb8cea13.

The size savings seem to have been lost, because with Go 1.18.6, the binary size actually grew from 9410616 bytes to 9410912.  However, it seems better to stick to packages from the golang.org domain, whenever possible.

### DIFF
--- a/src/cmd/list.go
+++ b/src/cmd/list.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 – 2021 Red Hat Inc.
+ * Copyright © 2019 – 2022 Red Hat Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


debarshiray:
@HarryMichal is there something we can do in our tests or CI to ensure that such things don't happen in the future?  Eg., we could have a block list of packages that we explicitly chose not to use, or something better.

I came across https://github.com/ryancurrah/gomodguard and [it seems like](https://github.com/ryancurrah/gomodguard/issues/12) it can be used through golangci-lint.  Any thoughts?

Issue #1076, opened 2022-07-11
`.local/bin` not on path for applications which are run with toolbox run


VarLad:
**Describe the bug**
If I open an editor as such `toolbox run hx /path/to/file.extension`, LSP doesn't work, since the language is installed in `.local/bin`. The only workaround is to do `toolbox enter` and `hx /path/to/file.extension`.
The same occurs for shells. Whether I use `SHELL=something toolbox enter` or `toolbox run shell`, I get the same, other than bash, no shell has `.local/bin` on path. I've tried it for nushell and fish

Essentially, anything opened with `toolbox run` doesn't have `.local/bin` on path

**Steps how to reproduce the behaviour**
Mentioned above

**Expected behaviour**
Everything opened with `toolbox enter` has `.local/bin` on the PATH

**Actual behaviour**
`.local/bin` is not on the path

**Output of `toolbox --version` (v0.0.90+)**
toolbox version 0.0.99.3

**Toolbox package info (`rpm -q toolbox`)**
e.g., `toolbox-0.0.18-2.fc32.noarch`

**Output of `podman version`**
e.g.,
```
Client:       Podman Engine
Version:      4.1.1
API Version:  4.1.1
Go Version:   go1.18.3
Built:        Wed Jun 15 20:01:58 2022
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
podman-4.1.1-1.fc36.x86_64

**Info about your OS**
Fedora Silverblue 36

**Additional context**
Being unable to use files installed on local path is an issue since a lot of applications are installed there. Thats a problem when using toolbox with editors or shells. Moreso on immutable distros where toolbox are usually used because its easier to install stuff locally

Fowindev:
Happened to me too, but "toolbox run bash -i -c command" seems to work

debarshiray:
Yes, you are right.  There's something odd going on.  Before going ahead and changing anything, I would like to understand exactly what Podman is doing.  See https://github.com/containers/podman/issues/16280

debarshiray:
This should be fixed by https://github.com/containers/toolbox/pull/1148

Please feel free to re-open or leave a comment if you have reason to believe otherwise.

debarshiray:
Thanks for your interest in Toolbx.  :)

PR #1148, opened 2022-10-25
cmd/run: Ensure that 'run' has the same container environment  as 'enter'

debarshiray:
Currently, commands invoked using `toolbox run` have a different environment than the interactive environment offered by `toolbox enter`. This is because 'toolbox run' was invoking the commands using something like this:
```
$ bash -c 'exec "$@"' bash [COMMAND]
```

... whereas, `toolbox enter` was using something like this:
```
$ bash -c 'exec "$@"' bash bash --login
```

In the first case, the helper Bash shell is a non-interactive non-login shell.  This means that it doesn't read any of the usual start-up files, and, hence, it doesn't pick up anything that's specified in them.  It runs with the default environment variables set up by Podman and the Toolbx image, plus the environment variables set by Toolbx itself.

In the second case, even though the helper Bash shell is still the same as the first, it eventually invokes a login shell, which runs the usual set of start-up files and picks up everything that's specified in them.

Therefore, to ensure parity, `toolbox run` should always have a login shell in the call chain inside the Toolbx container.

The easiest option is to always use a helper shell that's a login shell with `toolbox run`, but not `toolbox enter` so as to avoid reading the same start-up files twice, due to two login shells in the call chain.  It will still end up reading the same start-up files twice, if someone tried to invoke a login shell through `toolbox run`, which is fine.  It's very difficult to be sure that the user is invoking a login shell through `toolbox run`, and it's not what most users will be doing.

https://github.com/containers/toolbox/issues/1076

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -421,6 +421,13 @@ func callFlatpakSessionHelper(container string) error {
 	return nil
 }
 
+func constructCapShArgs(command []string) []string {
+	capShArgs := []string{"capsh", "--caps=", "--", "-c", "exec \"$@\"", "bash"}
+	capShArgs = append(capShArgs, command...)
+
+	return capShArgs
+}
+
 func constructExecArgs(container string,
 	command []string,
 	detachKeysSupported bool,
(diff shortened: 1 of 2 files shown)


debarshiray:
recheck

debarshiray:
Looks like something is wrong with softwarefactory-project.io.  It's not picking up builds for this pull request, and [this](https://softwarefactory-project.io/zuul/t/local/buildsets) says that there's an invalid start time.

debarshiray:
The tests pass locally on Fedora 36 on each of the two commits.  Merging.

PR #1152, opened 2022-11-08
test/system: Group the test cases somewhat logically

debarshiray:
It seems that as new test cases got developed they got appended towards the end of the file.  Now that there are a non-trivial number of test cases, it's difficult to look at the file and get a gist of all the scenarios being tested.

It will be better to have some logical grouping -- starting with the most basic functionality, then moving on to more advanced features, and then finally the errors.

This a step towards that.

### DIFF
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -13,6 +13,15 @@ teardown() {
   cleanup_containers
 }
 
+@test "run: Run command exiting with zero code in the default container" {
+  create_default_container
+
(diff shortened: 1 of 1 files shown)


debarshiray:
recheck

debarshiray:
> Other than the tests timing out, I don't know why `unit-test` and
> `unit-test-migration-path-for-coreos-toolbox` aren't even being run.

That's because of the optimizations we have in place.  Silly me.

PR #1162, opened 2022-11-16
test/system: Use long options, instead of their shorter aliases

debarshiray:
The long options are easier to `grep(1)` for in the sources than their shorter aliases.

### DIFF
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -241,18 +241,18 @@ teardown() {
   assert_output ""
 }
 
-@test "run: Try to run non-existent command in the default container" {
-  local cmd="non-existent-command"
-
+@test "run: Pass down 1 invalid file descriptor" {
+  local default_container_name="$(get_system_id)-toolbox-$(get_system_version)"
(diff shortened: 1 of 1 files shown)


debarshiray:
```
not ok 1 test suite: Setup in 10344ms
# (from function `assert_success' in file test/system/libs/bats-assert/src/assert.bash, line 114,
#  from function `_setup_docker_registry' in file test/system/libs/helpers.bash, line 176,
#  in test file test/system/000-setup.bats, line 23)
#   `_setup_docker_registry' failed
#
# -- command failed --
# status : 125
# output (13 lines):
#   Trying to pull quay.io/toolbox_tests/registry:latest...
#   Getting image source signatures
...
#   Copying blob sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4
#   Error: writing blob: storing blob to file "/var/tmp/storage4072376123/3": happened during read: unexpected EOF
```

debarshiray:
recheck

Issue #1017, opened 2022-03-01
/etc/profile.d/toolbox.sh:26: bad substitution on Fedora Kinoite


FilBot3:
# Describe the bug

> A clear and concise description of *what the bug is*. If possible, re-run the command(s) with `--log-level debug` and put the output here.

When I launch a new Zsh (oh-my-zsh) session, I get this output.

```bash
/etc/profile.d/toolbox.sh:26: bad substitution
```

## Steps how to reproduce the behaviour 

1. Open Konsole, see error.

## Expected behaviour

> A clear and concise description of what you *expected to happen*.

No output when starting the session.

## Actual behaviour

> A clear and concise description of what *actually happened*.

I see.

```bash
/etc/profile.d/toolbox.sh:26: bad substitution
```

## Screenshots 

> If applicable, add screenshots to help explain your problem.

![image](https://user-images.githubusercontent.com/3039509/156198255-3ce16860-af66-45fa-971f-c973fdac111e.png)

**Output of `toolbox --version` (v0.0.90+)**

```bash
❯ toolbox --version
toolbox version 0.0.99.3
```

**Toolbox package info (`rpm -q toolbox`)**

```bash
❯ rpm -q toolbox
toolbox-0.0.99.3-2.fc35.x86_64
```

**Output of `podman version`**

```bash
❯ podman version
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.16.8
Built:        Wed Dec  8 15:45:07 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**

```bash
❯ rpm -q podman
podman-3.4.4-1.fc35.x86_64
```

**Info about your OS**

```bash
❯ cat /etc/os-release 
NAME="Fedora Linux"
VERSION="35.20220228.0 (Kinoite)"
ID=fedora
VERSION_ID=35
VERSION_CODENAME=""
PLATFORM_ID="platform:f35"
PRETTY_NAME="Fedora Linux 35.20220228.0 (Kinoite)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:35"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-kinoite/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
[...]

PR #1177, opened 2022-11-24
profile.d: Restore compatibility with Z shell

debarshiray:
Otherwise, every zsh instance on Fedora Kinoite and Silverblue was running into:
```
/etc/profile.d/toolbox.sh:30: bad substitution
```

... because case modification with `"${VARIANT_ID^}"` is undefined in POSIX shell [1], and doesn't work with Z shell.

Fedora Silverblue got its own `PRETTY_NAME` (and `VARIANT` and `VARIANT_ID`) starting from Fedora 32 [2].  Therefore, it's better to use `PRETTY_NAME` and let the downstream distributor of the host operating system decide how it should be presented to the user, instead of coming up with a custom string.   eg., `PRETTY_NAME` isn't the same as `"Fedora $VARIANT"` on Fedora Silverblue.

One nice side-effect of this is that while `VARIANT` and `VARIANT_ID` are optional fields, `PRETTY_NAME` has a well-defined fallback value of 'Linux' [3].  This makes this a little less specific to Fedora Kinoite and Silverblue.

The rest of the welcome text was reformatted to prevent it from getting too wide depending on the contents of `PRETTY_NAME`.

Fallout from 3641a0032fff04ed50874de52416a75d1ab63597

[1] https://www.shellcheck.net/wiki/SC3059

[2] https://pagure.io/workstation-ostree-config/c/c18ef957d11862d32f362722931dbfdf1f5beb0d

[3] https://www.freedesktop.org/software/systemd/man/os-release.html

https://github.com/containers/toolbox/issues/1017

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -18,6 +18,7 @@ eval $(
           fi
 
           echo ID="$ID"
+          echo PRETTY_NAME="\"$PRETTY_NAME\""
           echo VARIANT_ID="$VARIANT_ID"
       )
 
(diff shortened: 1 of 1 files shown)


debarshiray:
Needs to be tested on an actual Fedora Kinoite or Silverblue host.

Issue #1074, opened 2022-07-05
Support subordinate user and group IDs provided by SSSD on enterprise FreeIPA set-ups


mhjacks:
**Describe the bug**

Toolbox currently seems to be hard-coded to look for subuid and subgid in /etc/sub{u,g}id respectively. These ids may now be provided by FreeIPA in a centralized way. Podman and buildah can both take advantage of this new mechanism; maybe toolbox should too

**Steps how to reproduce the behaviour**
1. Enable centralized subid support in FreeIPA/IdM (subid: sss in /etc/nsswitch.conf)
2. Remove /etc/sub{u,g}id entries for the user under test
3. `getsubids <user>` should still return a valid subid range
4. `toolbox enter` displays an error like this:

```
$ toolbox enter
Error: /etc/subgid and /etc/subuid don't have entries for user mjackson
See the podman(1), subgid(5), subuid(5) and usermod(8) manuals for more
information.
```

**Expected behaviour**
Toolbox should come up and be usable

**Actual behaviour**
Program reports error and refuses to proceed

**Output of `toolbox --version` (v0.0.90+)**
e.g., `toolbox version 0.0.90`

**Toolbox package info (`rpm -q toolbox`)**
toolbox-0.0.99.3-5.fc36.x86_64

**Output of `podman version`**
```
Client:       Podman Engine
Version:      4.1.1
API Version:  4.1.1
Go Version:   go1.18.3
Built:        Wed Jun 15 09:31:58 2022
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
podman-4.1.1-1.fc36.x86_64

**Info about your OS**
Fedora Workstation 36

**Additional context**
We recently troubleshot and fixed essentially the same issue in buildah: https://github.com/containers/buildah/issues/4063

debarshiray:
Thanks for pointing this out.

We [started checking](https://github.com/containers/toolbox/commit/0886fb466e7a73e7382255cf80c9459ffb1d84ac) `/etc/subgid` and `/etc/subuid` for better error handling on hosts where the current user might have been created long ago with an old version of `shadow-utils`.

I see that `getsubids <user>` works on my Fedora 36, even though I don't think I enabled centralized subid support in FreeIPA.  I don't see `subid: sss` in my `/etc/nsswitch.conf`.

So, I suppose we should first check `getsubids <user>`.  In case the `getsubids` binary is absent, we can fallback to the directly checking the files.  Did I get that right?

[...]

PR #1184, opened 2022-11-30
Enhance subid check by using library function

mhjacks:
Error correctly if no ranges found

Add subid-devel dependency for zuul

Refactor subid check to move cgo glue elsewhere

Signed-off-by: Martin Jackson <martjack@redhat.com>

### DIFF
--- a/playbooks/dependencies.yaml
+++ b/playbooks/dependencies.yaml
@@ -31,6 +31,7 @@
       - ninja-build
       - openssl
       - podman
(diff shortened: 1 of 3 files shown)


mhjacks:
This will fix https://github.com/containers/toolbox/issues/1074 and https://github.com/fedora-silverblue/issue-tracker/issues/263 and is the final form of #1180 where there is more discussion of what and why.

Issue #1070, opened 2022-06-30
Reduce `toolbox enter` start time


CMGeldenhuys:
**Is your feature request related to a problem? Please describe.**
I'm new to using `toolbox`, but `toolbox enter` seems to be a bit sluggish to launch(~1s). While still fast (I might just be nit-picking), my workflow is predominantly on the command line and so I'm constantly opening new terminal sessions, where the 1s is a significant pause before the prompt is visible. 
 
```
$ time toolbox run true

real	0m1.009s
user	0m0.465s
sys	0m0.174s

$ time podman exec fedora-toolbox-36 true

real	0m0.257s
user	0m0.103s
sys	0m0.039s
```
Comparing the run times between `toolbox run ...` and `podman exec ...` makes me think the problem is similar to that of #654. Might be related to `toolbox` evaluating expensive paths to ensure safe fallback (speculation).  

**Describe the solution you'd like**
Either a way to bypass the fallback sanity/safety checks with a flag, or perhaps narrow the gap between `podman exec` and `toolbox enter` run times.

**Describe alternatives you've considered**
Directly running `podman exec ...` instead of `toolbox enter`, on the launch of new terminal sessions.

Some reduction can be achieved by specifying the container:
```
$ time toolbox run --container fedora-toolbox-36 true

real	0m0.907s
user	0m0.469s
sys	0m0.183s

$ time toolbox run true

real	0m1.011s
user	0m0.519s
sys	0m0.173s
```

**Additional context**
I've tried to mark (`>>`) where the potential slow down might occur
```
$ toolbox run --verbose true |& ts '[%Y-%m-%d %H:%M:%.S]'
[2022-06-30 08:59:46.242384] level=debug msg="Running as real user ID 1000"
[2022-06-30 08:59:46.242466] level=debug msg="Resolved absolute path to the executable as /usr/bin/toolbox"
[2022-06-30 08:59:46.242519] level=debug msg="Running on a cgroups v2 host"
[2022-06-30 08:59:46.242543] level=debug msg="Checking if /etc/subgid and /etc/subuid have entries for user chrisg"
[2022-06-30 08:59:46.242564] level=debug msg="Validating sub-ID file /etc/subuid"
[2022-06-30 08:59:46.242582] level=debug msg="Validating sub-ID file /etc/subgid"
[2022-06-30 08:59:46.242599] level=debug msg="TOOLBOX_PATH is /usr/bin/toolbox"
[...]

PR #1187, opened 2022-12-05
test/system: Ensure that binaries are run against their build-time ABI

debarshiray:


akdev1l:
It seems like you are going to end up playing whack-a-mole with any foreseeable ABI incompatibility

I don't think the binary patching that is being done currently is supported by glibc and it seems likely to introduce all sorts of unintended consequences.

If toolbox is meant to run containerized and support multiple distributions then it should really not depend on the host ABI.

Depending on CLI and POSIX has turned out ok for the sister project distrobox - I don't see why we can't do the same here. 

just my 2¢

debarshiray:
> It seems like you are going to end up playing whack-a-mole with any
> foreseeable ABI incompatibility
> 
> I don't think the binary patching that is being done currently is
> supported by glibc and it seems likely to introduce all sorts of
> unintended consequences.

At least some of the GNU C Library maintainers are aware of this use-case.

> If toolbox is meant to run containerized and support multiple
> distributions then it should really not depend on the host ABI.

Have you seen [this](https://debarshiray.wordpress.com/2022/10/02/toolbx-running-the-same-binary-on-arch-linux-fedora-ubuntu-etc-containers/) blog post?

> Depending on CLI and POSIX has turned out ok for the sister project
> distrobox - I don't see why we can't do the same here.

Are you proposing that Toolbx go back to the POSIX shell implementation?

I am reluctant to do that for various reasons, but before we get into that, did you notice that Toolbx actually used to be implemented in POSIX shell?

mhjacks:
As far as subids go, we also have the option of not doing the check at all, and letting podman throw the error if it's going to.

debarshiray:
> > However, that would [prevent us](https://www.reddit.com/r/golang/comments/8m4xrh/do_linux_golang_binaries_depend_on_libc/) from using glibc’s [Name Service Switch](https://en.wikipedia.org/wiki/Name_Service_Switch) to look up usernames and groups, or to resolve host names.
> 
> Not sure why toolbx would need to resolve names by itself

Off the top of my head:
* To prompt the user about missing updates and containers that have reached End of Life.
* To provide better progress reporting about image downloads.  There's an inherent tension between keeping Toolbx's UI/UX clean and *just* being a Podman wrapper.  I won't be surprised if we end up pulling the images in-process at some point.

Some of these can be found under the [Project Pickle](https://github.com/containers/toolbox/labels/Project%20Pickle%20%F0%9F%A5%92) label.

> as the images are pulled via `podman pull`

That's only true if you look at the code base *today*.  It might not hold true in the future.

> and if I remember correctly last time we spoke
> about this you mentioned the subuid/subgid support but then adding that
> was complicated by the binary patching itself.

Umm... I can't remember adding that last part, but I may be misremembering.

> > Are you proposing that Toolbx go back to the POSIX shell implementation?
> 
> No, I am proposing that of trying to use CGO we would just shell
> out to the respective utilities - the same way it is currently done for
> `podman` (https://github.com/containers/toolbox/blob/main/src/cmd/create.go#L473)
> 
> For example that would mean using option number 1 in #1180

Shelling out (or more correctly, fork-exec) is inherently slower than calling a function within the same process.  The need to potentially [migrate](https://github.com/containers/toolbox/commit/9fc44b4eac4233c9ec407d11787fc7b852dfba27) already forces us to do some, otherwise needless, file operations on *every* invocation of `toolbox(1)`.  If Podman was updated, it will also involve shelling out to `podman system migrate`.  Shelling out `getsubids(1)` will add to this load.

Currently we have people still complaining about [performance](https://github.com/containers/toolbox/issues/1070); and there is an urge to push Toolbx closer to being the default shell on OSes like Fedora Silverblue.  So, performance matters, and we need to be careful about what we add to the common code paths.

Secondly, Toolbx's target audience doesn't want to know anything about containers.  They just want an interactive command line environment that's familiar to them.  So, keeping the UI/UX clean matters.

We need to somehow balance those two.  One option is to just not check for anything at all and let Podman error out, except most users will be facing an error message that's hard to understand.  It can be argued that we can assume everybody has users created with a shadow-utils that's new enough to always define the subordinate ID ranges, but then you can't say that about [--preserve-fds](https://github.com/containers/toolbox/issues/1066).
Plus, this is only true for the code base *today*.  Nobody knows what other use-case will show up in the future.

In general, it does seem limiting to constrain ourselves to only shelling out to other binaries, parsing their outputs and use exit codes for error handling.  Using a static binary or not using CGO eventually leads to cases where Toolbx might need a custom helper binary?  How will we manage the compatibility between the different parts then?

mhjacks:
From my perspective, I'm not certain that toolbx is going to add an enormous amount of value to explain podman errors. If a user has a situation where podman is going to error out, they're going to have to deal with containers, pure and simple; toolbx can at best reference a man page; but over time even that may provide different advice than podman's documented error handling.

I'm not sure how common these kinds of cases are - generally, I would expect that anyone using FreeIPA subid's (since they're not the default, or if they are they're new enough that they require some intervention to set up) would be a bit more advanced, since the kinds of users that want FreeIPA need to get more in the weeds about UIDs, GIDs anyway. I'm very leery of penalizing the whole project for what feels like a very niche use case - but at the same time, the current behavior is a blocker for FreeIPA users, because toolbx is erroring when it shouldn't.

To me, that says that toolbx's approach of "hmm, there was an error, run podman {verbosely} to learn more" is the right one. It's forward portable, and backward portable, and it doesn't require toolbx to keep state on those situations.

debarshiray:
Let's not fragment the discussion across too many different places.  It will be good to consolidate the discussion around subordinate IDs and enterprise set-ups in https://github.com/containers/toolbox/pull/1180

TL;DR: `dlopen(3)` solves the current problem, worst case we call `getsubids(1)`.

Issue #1183, opened 2022-11-30
Improve the error message when running nested inside a non-Toolbx container


debarshiray:
Toolbx can run nested inside Toolbx containers:
```
[rishi@topinka ~]$ toolbox enter
⬢[rishi@toolbox ~]$ toolbox list
IMAGE ID      IMAGE NAME                                      CREATED
e085805ade4a  registry.access.redhat.com/ubi8/toolbox:latest  5 months ago
4537d20059a4  registry.access.redhat.com/ubi9/toolbox:latest  5 months ago
1c8862ef037f  registry.fedoraproject.org/fedora-toolbox:35    4 months ago
2110dbbc33d2  registry.fedoraproject.org/fedora-toolbox:36    4 months ago

CONTAINER ID  CONTAINER NAME     CREATED      STATUS   IMAGE NAME
6167f42745fe  fedora-toolbox-36  3 weeks ago  running  registry.fedoraproject.org/fedora-toolbox:36
```

However, it's not supposed to run inside random non-Toolbx containers.  Regardless, people still try to run it (see https://github.com/containers/toolbox/issues/989) and the error message in such cases isn't very good.

This is what happens:
```
[rishi@topinka ~]$ podman run -it --rm registry.fedoraproject.org/fedora:37 /bin/bash
[root@bcc4a951f32c /]# sudo dnf -y install toolbox
...
...
[root@bcc4a951f32c /]# ln -s ../ /run/host
[root@bcc4a951f32c /]# toolbox 
Error: TOOLBOX_PATH not set
[root@bcc4a951f32c /]# export TOOLBOX_PATH=/usr/bin/toolbox
[root@bcc4a951f32c /]# toolbox list
Error: this is not a toolbox container
```

It will be good if we could improve the error handling so that the problem is more obvious (eg., the problem with `TOOLBOX_PATH` is secondary here) and self-explanatory like our other error messages (see `src/cmd/utils.go`).

Something like this:
```
[rishi@topinka ~]$ podman run -it --rm registry.fedoraproject.org/fedora:37 /bin/bash
[root@bcc4a951f32c /]# sudo dnf -y install toolbox
...
...
[root@bcc4a951f32c /]# ln -s ../ /run/host
[root@bcc4a951f32c /]# toolbox 
Error: this is not a Toolbox container
Toolbox is only meant to work with Toolbox containers.
Run 'toolbox --help' for usage.
[root@bcc4a951f32c /]# toolbox list
Error: this is not a Toolbox container
[...]

PR #1191, opened 2022-12-07
Improve the error message when running nested inside a non-Toolbx container

nievesmontero:
The following files have been modified: list.go and root.go 
Fixed issue https://github.com/containers/toolbox/issues/1183

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -142,6 +142,10 @@ func preRun(cmd *cobra.Command, args []string) error {
 		if _, err := validateSubIDRanges(cmd, args, currentUser); err != nil {
 			return err
 		}
+	} else {
+		if !utils.IsInsideToolboxContainer() {
+			return createErrorNonToolboxContainer()
+		}
 	}
 
 	toolboxPath := os.Getenv("TOOLBOX_PATH")
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -138,6 +138,16 @@ func createErrorInvalidRelease(hint string) error {
 	return errors.New(errMsg)
 }
 
+func createErrorNonToolboxContainer() error {
+	var builder strings.Builder
+	fmt.Fprintf(&builder, "this is not a Toolbox container.\n")
+	fmt.Fprintf(&builder, "Toolbox is only meant to work with Toolbox containers.\n")
+	fmt.Fprintf(&builder, "Run '%s --help' for usage.", executableBase)
+
+	errMsg := builder.String()
+	return errors.New(errMsg)
+}
+
 func getUsageForCommonCommands() string {
 	var builder strings.Builder
 	fmt.Fprintf(&builder, "create    Create a new toolbox container\n")
--- a/test/system/001-version.bats
+++ b/test/system/001-version.bats
@@ -24,6 +24,7 @@ setup() {
 }
 
 @test "version: Check version using option --version" {
+  bats_require_minimum_version 1.5.0
   run --separate-stderr $TOOLBOX --version
 
   assert_success
--- a/test/system/002-help.bats
+++ b/test/system/002-help.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers.bash'
 
 setup() {
+  bats_require_minimum_version 1.5.0
   _setup_environment
 }
 
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  bats_require_minimum_version 1.5.0
   _setup_environment
   cleanup_containers
 }
--- a/test/system/103-container.bats
+++ b/test/system/103-container.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  bats_require_minimum_version 1.5.0
   _setup_environment
   cleanup_containers
 }
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  bats_require_minimum_version 1.5.0
   _setup_environment
   cleanup_containers
 }
--- a/test/system/105-enter.bats
+++ b/test/system/105-enter.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  bats_require_minimum_version 1.5.0
   _setup_environment
   cleanup_containers
 }
--- a/test/system/106-rm.bats
+++ b/test/system/106-rm.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  bats_require_minimum_version 1.5.0
   _setup_environment
   cleanup_containers
 }
--- a/test/system/107-rmi.bats
+++ b/test/system/107-rmi.bats
@@ -20,6 +20,7 @@ load 'libs/bats-assert/load'
 load 'libs/helpers'
 
 setup() {
+  bats_require_minimum_version 1.5.0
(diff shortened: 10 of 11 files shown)


src/cmd/list.go
```diff
@@ -88,7 +88,7 @@ func init() {
 func list(cmd *cobra.Command, args []string) error {
 	if utils.IsInsideContainer() {
 		if !utils.IsInsideToolboxContainer() {
-			return errors.New("this is not a toolbox container")
+			return errors.New("this is not a toolbox container \n Toolbox is only meant to work with Toolbox containers.")
```
debarshiray, line 91: There is one extra spurious space on each side of that `\n`.

src/cmd/root.go
```diff
@@ -162,7 +162,7 @@ func preRun(cmd *cobra.Command, args []string) error {
 				return err
 			}
 
-			return errors.New("TOOLBOX_PATH not set")
+			return errors.New("this is not a Toolbox container \n Toolbox is only meant to work with Toolbox containers.")
```
debarshiray, line 165: Instead of duplicating the error message, and running the risk of copy-paste mistakes, we should have a single function (in `src/cmd/utils.go`) that returns the error.

debarshiray (changes requested):
Thanks for working on this!  There's still some work to be done to address https://github.com/containers/toolbox/issues/1183

The main problem in https://github.com/containers/toolbox/issues/1183 is that instead of immediately pointing out that it's not a Toolbox container, `toolbox` complains about `TOOLBOX_PATH` which is a secondary issue.  To solve this, we would need to re-arrange our start-up code so that the most important problems are flagged first.

debarshiray:
recheck

debarshiray:
Due to the recent expansion of the test suite, we might have to increase the timeout for stable Fedoras: https://github.com/containers/toolbox/pull/1212

We recently had to do that for Fedora Rawhide.

debarshiray (changes requested):
Does this actually work as described in https://github.com/containers/toolbox/issues/1183 ?

I tried:
```
$ podman run --privileged -it --rm --userns=keep-id --volume /home/rishi:/home/rishi:rslave --volume /:/run/host:rslave registry.fedoraproject.org/fedora:37 /bin/bash
$ /path/to/compiled/toolbox
Error: TOOLBOX_PATH not set
```

Also, could you please squash the two commits into one?  You can use `git rebase -i` for that.

debarshiray (changes requested):
Thanks for rebasing this on top of https://github.com/containers/toolbox/pull/1212

We still need to address the review comments about the code, though.

nievesmontero:
When testing the code we realised that after establishing and exporting the ```TOOLBOX_PATH``` we obtain the following error messages. 
Do you think we should modify the error message for the ```TOOLBOX_PATH not set``` and add also the error message for the ```Error: this is not a Toolbox container.```?
```
$ TOOLBOX_PATH=`pwd`
$ export TOOLBOX_PATH
$ ./toolbox
Error: missing command

create    Create a new toolbox container
enter     Enter an existing toolbox container
list      List all existing toolbox containers and images

$ ./toolbox list
Error: this is not a Toolbox container.
Toolbox is only meant to work with Toolbox containers.
```

debarshiray:
> When testing the code we realised that after establishing and exporting the `TOOLBOX_PATH` we obtain the following error messages. Do you think we should modify the error message for the `TOOLBOX_PATH not set` and add also the error message for the `Error: this is not a Toolbox container.`?

There are two distinct error conditions.  One is a missing `TOOLBOX_PATH`, and the other is that `toolbox` is being used inside a container that's not a Toolbx container.

The `TOOLBOX_PATH` environment variable is an implementation detail of Toolbx that's used to make `toolbox` work inside a Toolbx container.  It holds the path to the `toolbox` binary on the host, and makes it known to the container.  When `toolbox` is invoked inside the container it uses `flatpak-spawn --host $TOOLBOX_PATH ...` to forward the call to the host.

Beyond these implementation details, there's the more high-level conceptual aspect that Toolbx is not meant to be used with random OCI containers that were not created by Toolbx itself.

There's, of course, some overlap between the two, but in some sense they are two separate things.

Currently, we check the presence of `TOOLBOX_PATH` early during start-up in `src/cmd/root.go`, but check whether the container is a Toolbx container later, separately, in the implementation of each separate Toolbx command.  This leads to https://github.com/containers/toolbox/issues/1183

In most (all?) cases, the user-visible problem is that Toolbx is being used inside a container that's not a Toolbx container.  `TOOLBOX_PATH` is not something the user needs to know about, and it's unlikely to go missing, unless the user explicitly goes and unsets it.  So, we should check for the most common user-visible problem first, before checking details needed by the implementation.

This might have the positive side effect of removing some duplicated code from each command, which is always good.

src/cmd/list.go
```diff
@@ -82,7 +82,7 @@ func init() {
 func list(cmd *cobra.Command, args []string) error {
 	if utils.IsInsideContainer() {
 		if !utils.IsInsideToolboxContainer() {
-			return errors.New("this is not a toolbox container")
+			return createErrorNonToolboxContainer()
```
debarshiray, line 85: If the code in `preRun()` works as intended, then these duplicated checks in each of the commands might not be necessary.

debarshiray (changes requested):

debarshiray:
Do you think you could write a system test to ensure that the improved error handling continues to work?

See the files inside `test/system`.  There's roughly one file per command.  We could add a test for each command or so.

src/cmd/utils.go
```diff
@@ -194,6 +204,14 @@ func resolveContainerAndImageNames(container, containerArg, distroCLI, imageCLI,
 		} else if errors.As(err, &errParseRelease) {
 			err := createErrorInvalidRelease(errParseRelease.Hint)
 			return "", "", "", err
+		} else if errors.As(err, &errContainer) {
+			if errors.Is(err, utils.ErrNonToolboxContainer) {
```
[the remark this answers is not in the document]
nievesmontero: You're right, I don't think this piece of code is even necessary.

src/cmd/root.go
```diff
@@ -158,13 +158,16 @@ func preRun(cmd *cobra.Command, args []string) error {
 
 	if toolboxPath == "" {
 		if utils.IsInsideContainer() {
-			if err := preRunIsCoreOSBug(); err != nil {
-				return err
+			if utils.IsInsideToolboxContainer() {
```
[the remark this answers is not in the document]
nievesmontero: I moved it right after the `if !utils.IsInsideContainer()` inside an `else`, the idea is to first check if it is inside a container and, if it is inside a container, then check if it is a toolbox container.

src/cmd/root.go
```diff
@@ -158,13 +158,16 @@ func preRun(cmd *cobra.Command, args []string) error {
...
+				if err := preRunIsCoreOSBug(); err != nil {
+					return err
+				}
+
+				return errors.New("TOOLBOX_PATH not set")
+			} else {
+				return errors.New("this is not a Toolbox container.\nToolbox is only meant to work with Toolbox containers.\nRun '%s --help' for usage.")
```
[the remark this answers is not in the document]
nievesmontero: Yes

src/cmd/list.go
nievesmontero: Check supressed.

nievesmontero:
> Do you think you could write a system test to ensure that the improved error handling continues to work?
> 
> See the files inside `test/system`. There's roughly one file per command. We could add a test for each command or so.

I can try and see if it works

nievesmontero:
> Ping, @nievesmontero
> 
> Are you still working on this? I am asking because the tests are failing.

I have put it aside while working on https://github.com/containers/toolbox/pull/1274 but will to try solve this as soon as possible.

nievesmontero:
recheck

debarshiray:
Umm... why did you close this one, @nievesmontero ?  :)

nievesmontero:
> Umm... why did you close this one, @nievesmontero ? :)

As all the tests were failing, I decided to start from scratch and create a new pull request, which didn't give any result because the tests are still failing.

PR #1171, opened 2022-11-18
Test -Dmigration_path_for_coreos_toolbox on CentOS Stream 9

debarshiray:
The code behind `-Dmigration_path_for_coreos_toolbox` is meant for Red Hat
Enterprise Linux 8 and 9.  Hence, it makes sense to use CentOS Stream 9
to run the 'unit tests' on the corresponding code path.

PR #1198, opened 2022-12-13
.zuul: Use the latest version of the tools for static analysis

debarshiray:
The 'unit tests' are no longer just unit tests.  They also run a bunch of static analysis tools like ShellCheck, codespell, gofmt and 'go vet'.  Since newer versions of these tools are generally better at catching problems in the codebase, it will be better to run the 'unit tests' on Fedora Rawhide with the latest versions than older stable Fedoras.

The timeout for the 'unit tests' need to be increased because Fedora Rawhide is slower is than stable Fedoras.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -18,22 +18,22 @@
 - job:
     name: unit-test
     description: Run Toolbox's unit tests declared in Meson
-    timeout: 600
+    timeout: 1200
     nodeset:
       nodes:
-        - name: ci-node-36
-          label: cloud-fedora-36
+        - name: fedora-rawhide
+          label: cloud-fedora-rawhide
     pre-run: playbooks/setup-env.yaml
     run: playbooks/unit-test.yaml
 
 - job:
     name: unit-test-migration-path-for-coreos-toolbox
     description: Run Toolbox's unit tests declared in Meson when built with -Dmigration_path_for_coreos_toolbox
-    timeout: 600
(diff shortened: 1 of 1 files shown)


debarshiray:
This is not a good idea, after all.

As seen in https://github.com/containers/toolbox/pull/1171, the Fedora Rawhide repositories can at times be busted.  Therefore, we can't use Fedora Rawhide to test anything which isn't at least being tested elsewhere.

The system tests are already being run on different OSes, and Fedora Rawhide is only one of them.  So that's fine.

https://github.com/containers/toolbox/pull/1171 aims to run `meson test` with `-Dmigration_path_for_coreos_toolbox` on CentOS Stream 9.  So, we need something reliable to run `meson test` without that option, and Fedora Rawhide isn't that option.

Once we get rid of the somewhat artificial division between the `unit-test*` and `system-test-*` jobs, we can run all the tests across all OSes, and everything will be hunky dory.

In the meantime, it's not so bad, because CentOS Stream 9 can have sufficiently newer versions of the tools, which can expose interesting problems in the code base.  eg., https://github.com/containers/toolbox/pull/1199

debarshiray:
It's worth revisiting this once again because we managed to make the Fedora Rawhide builds more robust against stale DNF caches in https://github.com/containers/toolbox/pull/1201  Therefore, I have resurrected this branch in https://github.com/containers/toolbox/pull/1206

Issue #1157, opened 2022-11-15
`toolbox run` command isn't terminated properly when terminal emulator is exited


ghost:
**Describe the bug**

`toolbox run` command isn't terminated properly when terminal emulator is exited, causing processes to linger around without user's awareness.

**Steps how to reproduce the behaviour**

1. Open System Monitor / Task Manager equivilent in your desktop environment
2. Search for `sleep`
3. Execute the following command in a terminal emulator (GNOME Terminal):

```bash
toolbox run sleep 30
```

4. Before `sleep` runs out of time, try to close the terminal emulator
5. It'll prompt for something like this:

![image](https://user-images.githubusercontent.com/40300931/191938419-18d5f1e0-7daf-4946-beeb-f2aa06cf07c1.png)

Insist closing it by pressing `Close Terminal`.

**Expected behaviour**

`sleep` process is terminated.

**Actual behaviour**

`sleep` keeps lingering around.

**Output of `toolbox --version` (v0.0.90+)**

`toolbox version 0.0.99.3`

**Toolbox package info (`rpm -q toolbox`)**

`toolbox-0.0.99.3-6.fc36.x86_64`

**Output of `podman version`**

```
Client:       Podman Engine
Version:      4.2.1
API Version:  4.2.1
Go Version:   go1.18.5
Built:        Thu Sep  8 03:58:19 2022
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**

`podman-4.2.1-2.fc36.x86_64`

**Info about your OS**

Fedora Silverblue 36

**Additional context**

> This is a screenshot which represent the issue better:
> 
> ![image](https://user-images.githubusercontent.com/40300931/201854868-f89a75dd-323d-4bbb-81bf-34dc05b60c82.png)
> 
> If I try to close the terminal emulator, it'll prompt the following:
> 
> ![image](https://user-images.githubusercontent.com/40300931/191938419-18d5f1e0-7daf-4946-beeb-f2aa06cf07c1.png)
> 
> If I press "Close Terminal", `sh`, `toolbox` and `podman` (which runs `exec` command) will be terminated because they're child process of the vte session.
> 
> However, notice the `conmon` and its child process `sleep` aren't part of `gnome-terminal-server`. When `sh` is terminated, `podman` (`exec` command) will be terminated but the corresponding `conmon` process will be kept intact. As a result, `sleep 30` isn't terminated properly.
> 
[...]

Issue #1204, opened 2022-11-15
Optionally make exec session terminate with parent


thjderjktyrjkt:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Add a flag to `podman exec` to make exec session terminate with parent, similar to [bubblewrap](https://github.com/containers/bubblewrap)'s [`bwrap --die-with-parent`](https://github.com/containers/bubblewrap/commit/b6370de0fc4be6bb801206e39437ad1f2f5c0be7).

Immutable distributions make use of `toolbox` / `distrobox` to provide a mutable environment. A common use is to run commands directly within container (`toolbox run [COMMAND]` / `distrobox enter -- [COMMAND]`), since they use exec session, they have the same limitation of not terminating child-proceess when terminal emulator is closed.

**Steps to reproduce the issue:**

1. Open System Monitor / Task Manager equivilent in your desktop environment, search for `sleep`

2. Run the following command in your terminal emulator (either one will work):

podman:

```bash
podman run --rm -it \
    --name debian \
    --entrypoint /bin/sh \
    docker.io/library/debian:11

# In a new terminal emulator window
podman exec debian sleep 30
```

toolbox:

```bash
toolbox create
toolbox run sleep 30
```

distrobox:

```bash
distrobox create
distrobox enter -- sleep 30
```

3. Then try to close terminal emulator, it'll prompt something like this:

![image](https://user-images.githubusercontent.com/40300931/191938419-18d5f1e0-7daf-4946-beeb-f2aa06cf07c1.png)

4. Insist closing it, then look at System Monitor

**Describe the results you received:**

`sleep 30` still runs within container.

**Describe the results you expected:**

Nah, this is expected, hence this feature request.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
Client:       Podman Engine
Version:      4.2.1
API Version:  4.2.1
Go Version:   go1.18.5
Built:        Thu Sep  8 03:58:19 2022
OS/Arch:      linux/amd64
```

**Output of `podman info`:**

<details>
  <summary>Click me</summary>

```
host:
  arch: amd64
  buildahVersion: 1.27.0
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.4-3.fc36.x86_64
    path: /usr/bin/conmon
[...]

Issue #908, opened 2021-10-29
Cannot change shell to zsh


alvarlagerlof:
**Describe the bug**
Shell does not change from bash to zsh with neither chsh or usermod -s 

**Steps how to reproduce the behaviour**
```
toolbox create --distro fedora --release 33
toolbox enter --release 33
echo $SHELL
sudo dnf install zsh
sudo usermod -s /bin/zsh alvar (or chsh, using `util-linux-user`)
exit
toolbox enter --release 33
echo $SHELL
```

**Expected behaviour**
$SHELL changes from bash to zsh

**Actual behaviour**
$SHELL stays at bash

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.99.1`

**Toolbox package info (`rpm -q toolbox`)**
`toolbox-0.0.99.1-1.fc34.x86_64`

**Output of `podman version`**
```
Version:      3.1.0
API Version:  3.1.0
Go Version:   go1.16
Built:        Tue Mar 30 15:29:36 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-3.1.0-1.fc34.x86_64`

**Info about your OS**
Fedora Silverblue 34

junjieyuan:
```toolbox enter``` [read host's ```SHELL``` environment variable](https://github.com/containers/toolbox/blob/3641a0032fff04ed50874de52416a75d1ab63597/src/cmd/enter.go#L129) as container's default shell.

Just a workaround:
```sh
SHELL=/bin/zshtoolbox enter
```

alvarlagerlof:
> `toolbox enter` [read host's `SHELL` environment variable](https://github.com/containers/toolbox/blob/3641a0032fff04ed50874de52416a75d1ab63597/src/cmd/enter.go#L129) as container's default shell.
> 
> Just a workaround:
> 
> ```shell
> SHELL=/bin/zsh toolbox enter
> ```

Thank you, this works well for now and fits with the alias I already made. I would still like it to just work as expected in the future though.

debarshiray:
I wonder if we need to add a `toolbox shell` command that:
* Is meant to be run only inside the container, like `init-container`.
* Looks at the default shell set in the container's `/etc/passwd` and tries to run it. If that's absent, falls back to whatever is specified by the host's `SHELL` environment variable, and finally to `/bin/bash`.

[...]

PR #1207, opened 2022-12-22
cmd/run: Ensure underlying container is stopped when toolbox is killed

halfline:
Right now "toolbox enter" creates a container on the fly, but then lets it linger after the foreground toolbox process is killed (for instance, from a terminal hangup).

Not killing the underlying container has the negative side effect of stalling shutdown if a toolbox shell is running.

This commit addresses that problem by detecting when the toolbox process is signaled, and then in response, kills off the entire cgroup associated with the underlying container.

Closes https://github.com/containers/toolbox/issues/1157


halfline:
@matthiasclasen i think this is probably a fix for the shutdown issue you told me about yesterday

debarshiray:
Just for the sake of posterity ...

> @matthiasclasen i think this is probably a fix for the shutdown issue you told me about yesterday

I think the blocked shutdown problem was reported as:
* https://github.com/containers/podman/issues/14531
* https://bugzilla.redhat.com/show_bug.cgi?id=2081664
* https://bugzilla.redhat.com/show_bug.cgi?id=2084498

... and might have been fixed by:
https://github.com/containers/podman/pull/17025

Note that https://github.com/containers/podman/issues/14531 points at https://github.com/containers/podman/pull/16785 but I don't think it had anything to do with fixing blocked shutdowns caused by lingering `podman exec` sessions.

debarshiray (changes requested):
Thanks for working on this @halfline - my apologies for not looking at it sooner.

I took the liberty to rebase the branch against `main` to play with it.  I don't think it is quite right, and the commit message suggests that there might be some misunderstanding.

Longer explanation follows ...

First, `toolbox enter` doesn't quite create a container on the fly.  It is confusing because there's some syntactic sugar where `toolbox create` will offer to create a container if you had none.  You could think of it as an embedded `toolbox create`.

The canonical flow is that a container is created with `toolbox create`, which wraps `podman create`.  Then that container is entered with `toolbox enter`, which wraps `podman start` and `podman exec`.  The `podman start` launches the container's entry point process, which would have been PID 1 inside the container had we been using a PID namespace, but we don't.  For Toolbx containers, this entry point is `toolbox init-container`.  The `podman exec` launches the process that the user actually interacts with.  eg., a CLI shell, Emacs, Vim, etc..

The same container can be entered multiple times from multiple terminals at the same time.  When that happens the `podman start` is basically a NOP because a container can have only one entry point and it's already running.

Neither the entry point nor the foreground processes are parented by `podman start` or `podman exec`.  Each of them are parented by a separate `conmon(8)` process which is double forked from `podman(1)` and parented by the systemd user instance.

The [blocked](https://github.com/containers/podman/issues/14531) [shutdown](https://bugzilla.redhat.com/show_bug.cgi?id=2081664) [problem](https://bugzilla.redhat.com/show_bug.cgi?id=2084498) was that when `SIGTERM` was sent to the container's entry point, either through `podman stop` or the shutdown sequence, the processes launched by `podman exec` wouldn't be terminated.  This would end up blocking shutdown.  My understanding is that this was caused by Toolbx containers not having a separate PID namespace and was fixed by https://github.com/containers/podman/pull/17025

This uncovers another problem.  How did those processes launched by `podman exec` manage to outlive their respective terminals?  This is https://github.com/containers/toolbox/issues/1204  I suspect this happens because the signal sent by the terminal only reaches upto the wrapper `podman exec` and doesn't get to the corresponding `conmon(8)` process.

With the changes in this branch, if I have multiple graphical terminals open with `toolbox enter` against the same container, then closing any one of those terminals with the cross button terminates the container's entry point and the foreground container process in all the other terminals.  This shouldn't happen.  Closing a terminal should only terminate its own foreground container process.

I don't understand cgroups very well, so I don't know why this is happening.

Toolbx containers also use the host's cgroup namespace, and, as far as I know, each terminal created by GNOME Terminal has its own cgroup.  For each of the foreground container processes, I get:
```
$ cat /proc/56151/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-2b4111a9-f4a9-4f2b-8acd-4adc9e673bac.scope
$ 
$ cat /proc/55919/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-9bf0516d-bd6f-4094-9504-0be37b1ce6bf.scope
$ 
$ cat /proc/56363/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-6ecb9e7e-aa06-42d6-b9b6-bb9d6cdb44f4.scope
```

So, I have no idea why terminating the processes in one cgroup is affecting the others.

halfline:
sorry it's been a while since i looked at this and even when I was working on this I only put it together in a few hours after Matthias visited and he mentioned the problem off hand, so I may be off base here... I haven't done much deep diving and I may have made wrong assumptions.

Having said that, my impression from your message is you believe I'm sending the hangup signal to the terminal's cgroup, but that's not the intent. The idea was to send the signal to the cgroup of the bash running in the container.

I got (at least) two things wrong I think:

- I had assumed each conmon would create a new cgroup for each podman-exec invocation. It seems that assumption was perhaps wrong.  Indeed, every shell seems to be using a shared cgroup. I get:

╎❯ cat /proc/self/cgroup
0::/user.slice/user-4153.slice/user@4153.service/user.slice/libpod-4e141e468f1a756b43407f64f0273b5c18fcf5b784ea2a13050d0a9f4e1528ab.scope/container

in all my terminals. I find this really surprising. I do see crun has a --cgroup option, but indeed looking in ps output, it's not getting used.

- I looking at the wrong pid to find the cgroup.  I'm erroneously using the pid of the toolbox init-container container to find the cgroup.

> This is https://github.com/containers/toolbox/issues/1204 I suspect this happens because the signal sent by the terminal only reaches upto the wrapper podman exec and doesn't get to the corresponding conmon(8) process

Right, when a terminal is closed, a SIGHUP is sent to all processes in the terminal "session" (in the setsid() sense of the word). These sessions are inherited from their parents, unless a child creates a fresh session. There's no way in unix afaik to move a child into a pre-existing session; it either uses the one it started with, or it starts its own.  conmon isn't part of the child hierarchy of podman-exec, so it won't inherit the terminal session. This means the hangup signal needs to be ferried either from podman-exec  or toolbox to the process started by conmon (or if children had their own cgroup like I thought, we could just send the signal to the whole cgroup)

debarshiray:
> Having said that, my impression from your message is you believe I'm sending the hangup signal to the terminal's cgroup, but that's not the intent. The idea was to send the signal to the cgroup of the bash running in the container.

I managed to confuse myself as well.

GNOME Terminal puts every `VteTerminal` into a separate systemd scope or cgroup.  That's what I was showing here, because these process IDs are of the `toolbox enter` processes, and not of `conmon(8)` or the processes inside the container:
```
$ cat /proc/56151/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-2b4111a9-f4a9-4f2b-8acd-4adc9e673bac.scope
$ 
$ cat /proc/55919/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-9bf0516d-bd6f-4094-9504-0be37b1ce6bf.scope
$ 
$ cat /proc/56363/cgroup 
0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-6ecb9e7e-aa06-42d6-b9b6-bb9d6cdb44f4.scope
```

That's why I didn't realize that all the `conmon(8)` processes for the same container share the same cgroup.  However, they don't share it across different containers.

 >     * I had assumed each conmon would create a new cgroup for each podman-exec invocation. It seems that assumption was perhaps wrong.  Indeed, every shell seems to be using a shared cgroup. I get:
> 
> 
> ╎❯ cat /proc/self/cgroup 0::/user.slice/user-4153.slice/[user@4153.service](mailto:user@4153.service)/user.slice/libpod-4e141e468f1a756b43407f64f0273b5c18fcf5b784ea2a13050d0a9f4e1528ab.scope/container
> 
> in all my terminals. I find this really surprising. I do see crun has a --cgroup option, but indeed looking in ps output, it's not getting used.

Right.  Thanks for the hint, because I had missed it before.
 
>     * I looking at the wrong pid to find the cgroup.  I'm erroneously using the pid of the toolbox init-container process to find the cgroup.

*nod*
 
> > This is #1204 I suspect this happens because the signal sent by the terminal only reaches upto the wrapper podman exec and doesn't get to the corresponding conmon(8) process
> 
> Right, when a terminal is closed, a SIGHUP is sent to all processes in the terminal "session" (in the setsid() sense of the word). These sessions are inherited from their parents, unless a child creates a fresh session. There's no way in unix afaik to move a child into a pre-existing session; it either uses the one it started with, or it starts its own. conmon isn't part of the child hierarchy of podman-exec, so it won't inherit the terminal session. This means the hangup signal needs to be ferried either from podman-exec or toolbox to the process started by conmon (or if children had their own cgroup like I thought, we could just send the signal to the whole cgroup)

*nod*

I don't see an obvious way to get the PID of the `conmon(8)` process or the process inside the container started by it.  That was my [question](https://github.com/containers/toolbox/issues/1204#issuecomment-1411291162) in https://github.com/containers/toolbox/issues/1204

Otherwise, we can insert a process that we control between `conmon(8)` and the user-requested process inside the container that will relay the PID inside the container back to the `toolbox enter` process.  Then we can ferry the `SIGHUP` from the terminal to the container process.  I don't know what would be the best way to do it.  Maybe we can send file descriptor?

This intermediate process under our control sounds a lot like the [toolbox shell](https://github.com/containers/toolbox/issues/908#issuecomment-968945503) idea that I had before.  We don't have to call it `toolbox shell`, though.  It's just that I am really bad at naming.  So, suggestions welcome.

halfline:
I think it probably is possible to get the right pid. just as an experiment:

╎❯ bundle_path=$(podman inspect 4e141e468f1a756b43407f64f0273b5c18fcf5b784ea2a13050d0a9f4e1528a  | jq -r '.[0].StaticDir')
╎❯ exec_id=$(podman exec -d 4e141e468f1a756b43407f64f0273b5c18fcf5b784ea2a13050d0a9f4e1528a sleep 300)
╎❯ pid=$(podman exec 4e141e468f1a756b43407f64f0273b5c18fcf5b784ea2a13050d0a9f4e1528a cat $bundle_path/$exec_id/exec_pid)
╎❯ echo $pid
193097

halfline:
I guess the complication is the exec_id is only printed for background exec invocations.  You can fish it out of podman inspect output, but you would have to diff the list of execids before and after to find the right one, which is a little racy.

debarshiray:
> I think it probably is possible to get the right pid. just as an experiment:

Cool.  I didn't know that exec IDs are a thing.  I think I never tried `podman exec --detach` until now.

> though it looks like toolbx doesn't use libpod ?

When we rewrote Toolbx in Go, we wanted to use libpod and the other backend Go packages directly instead of going through `podman(1)`.  However, the Podman folks discouraged us from doing that, so we didn't.

These days, for other reasons, I am considering implementing some small things directly inside `toolbox(1)` to avoid going through `podman(1)`, where we can rely on a stable interface.  I have no idea if this is such a case.

If we can't get the PID inside the container through a Podman interface, do you think the alternative of inserting a shim is sane?

debarshiray:
`podman exec --interactive --tty`, which is what `toolbox enter` does, creates a nested pseudo-terminal device and wires it up to the terminal device that the user is interacting with.  If the outer terminal goes away, then shouldn't it break the connection and somehow trigger the inner terminal?

halfline:
> `podman exec --interactive --tty`, which is what `toolbox enter` does, creates a nested pseudo-terminal device and wires it up to the terminal device that the user is interacting with. If the outer terminal goes away, then shouldn't it break the connection and somehow trigger the inner terminal?

Interesting. Right, when the terminal is closed, podman exec would be receiving a hang up signal (SIGHUP) except:

```
╎❯ echo $$
203855

╎❯ ps -eocomm,pid,ppid,tpgid,sess,pgrp,tty |grep -E 'podman|bash|toolbox|conmon|COMM'
COMMAND             PID    PPID   TPGID    SESS    PGRP TT
bash             203855    3069  207753  203855  203855 pts/18


╎❯ toolbox enter f39
╎❯ ps -eocomm,pid,ppid,tpgid,sess,pgrp,tty |grep -E 'podman|bash|toolbox|conmon|COMMAND|pts.18'
COMMAND             PID    PPID   TPGID    SESS    PGRP TT
bash             203855    3069  207768  203855  203855 ?
toolbox          207768  203855  207768  203855  207768 ?
podman           207888  207768  207768  203855  207768 ?
conmon           207903    1899      -1  207903  207903 ?
bash             207909  207903  207947  207909  207909 pts/3
```

You can see `podman exec` is in the toolbox process group and toolbox doesn't have a controlling tty. Neither conmon. In fact, nothing has /dev/pts/18 as the controlling tty anymore. This means when the terminal is closed, the hangup signal doesn't get sent. i guess maybe something became the foreground process group and then exited? or maybe there was a TIOCNOTTY ioctl?

regardless whatever is splicing input between the two ttys (podman? conmon?), should be getting a hang up in poll so it should know to tear things down i guess.

halfline:
just browsing around a little, i think the splicing happens here:

```
func setupStdioChannels(streams *define.AttachStreams, conn *net.UnixConn, detachKeys []byte) (chan error, chan error) {
        receiveStdoutError := make(chan error)
        go func() {
                receiveStdoutError <- redirectResponseToOutputStreams(streams.OutputStream, streams.ErrorStream, streams.AttachOutput, streams.AttachError, conn)
        }()

        stdinDone := make(chan error)
        go func() {
                var err error
                if streams.AttachInput {
                        _, err = util.CopyDetachable(conn, streams.InputStream, detachKeys)
                }
                stdinDone <- err
        }()

        return receiveStdoutError, stdinDone
}
```

So maybe something like this would fix it

```
@@ -180,20 +181,25 @@ func (c *Container) attachToExec(streams *define.AttachStreams, keys *string, se
        }
        defer func() {
                if err := conn.Close(); err != nil {
                        logrus.Errorf("Unable to close socket: %q", err)
                }
        }()
 
        // start listening on stdio of the process
        receiveStdoutError, stdinDone := setupStdioChannels(streams, conn, detachKeys)
 
+       go func() {
+               err := <-stdinDone
+               c.Kill(uint(syscall.SIGHUP))
+       }()
+
        // 4: send start message to child
        if err := writeConmonPipeData(startFd); err != nil {
                return err
        }
 
        return readStdio(conn, streams, receiveStdoutError, stdinDone)
 }
```

(untested, I don't know go, etc etc)

Issue #752, opened 2021-04-08
Report pulled image size


HarryMichal:
**Is your feature request related to a problem? Please describe.**

When Toolbox offers to pull an image, it estimates the size of the pulled image will be 500MB. This number is hard-coded and in most cases does not represent the reality. Toolbox images are usually less then 400MB big and that is after decompression. Compressed images are even smaller.

**Describe the solution you'd like**

Have a mechanism for reporting the size of the pulled image.

**Additional context**

This will, probably, require some work in https://github.com/containers/skopeo. Relevant upstream [issue](https://github.com/containers/skopeo/issues/641).

debarshiray:
We should be careful not to add a hard dependency on skopeo.  Toolbx is part of size-constrained operating systems like Fedora CoreOS, where an extra mandatory dependency on skopeo might be problematic.  We should attempt to use it, but if it's absent, we should have a fallback of some sort.

debarshiray:
That commit from @juhp that are showing up against this issue can be ignored.  It was addressed earlier in https://github.com/containers/toolbox/pull/875#issuecomment-970721430

nievesmontero:
In order to add this feature, I'm going through the create.go file. The variable %s is used all along this file and it gets the following value ```registry.fedoraproject.org/fedora-toolbox:36```. Where does it take that value? In the create.go file or in some other file? I think I need that information to implement the feature.

debarshiray:
> In order to add this feature, I'm going through the create.go file. The variable %s is used all along this file and it gets the following value `registry.fedoraproject.org/fedora-toolbox:36`. Where does it take that value?

That `%s` is a format specifier (or format verb as Go calls it) that's used in [printf format strings](https://en.wikipedia.org/wiki/Printf_format_string).  Can you show me the exact line(s) that you are looking at?

nievesmontero:
The main lines I've been looking at are from the 700-705 of the create.go file. The variable %s is the one from which I think we should get the image size.

debarshiray:
You mean these lines in [src/cmd/create.go](https://github.com/containers/toolbox/blob/main/src/cmd/create.go):
```
	if promptForDownload {
                fmt.Println("Image required to create toolbox container.")

                prompt := fmt.Sprintf("Download %s (500MB)? [y/N]:", imageFull)
                shouldPullImage = askForConfirmation(prompt)
        }
```

[...]

PR #1208, opened 2023-01-09
Report pulled image size

nievesmontero:
https://github.com/containers/toolbox/issues/752

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -27,7 +27,9 @@ import (
 	"github.com/briandowns/spinner"
 	"github.com/containers/toolbox/pkg/podman"
 	"github.com/containers/toolbox/pkg/shell"
+	"github.com/containers/toolbox/pkg/skopeo"
 	"github.com/containers/toolbox/pkg/utils"
+	"github.com/docker/go-units"
 	"github.com/godbus/dbus/v5"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -573,6 +575,30 @@ func getFullyQualifiedImageFromRepoTags(image string) (string, error) {
 	return imageFull, nil
 }
 
+func getImageSizeFromRegistry(imageFull string) (string, error) {
+	image, err := skopeo.Inspect(imageFull)
+	if err != nil {
+		return "", err
+	}
+
+	if image.LayersData == nil {
+		return "", errors.New("'skopeo inspect' did not have LayersData")
+	}
+
+	var imageSizeFloat float64
+
+	for _, layer := range image.LayersData {
+		if layerSize, err := layer.Size.Float64(); err != nil {
+			return "", err
+		} else {
+			imageSizeFloat += layerSize
+		}
+	}
+
+	imageSizeHuman := units.HumanSize(imageSizeFloat)
+	return imageSizeHuman, nil
+}
+
 func getServiceSocket(serviceName string, unitName string) (string, error) {
 	logrus.Debugf("Resolving path to the %s socket", serviceName)
 
@@ -687,7 +713,15 @@ func pullImage(image, release, authFile string) (bool, error) {
 	if promptForDownload {
 		fmt.Println("Image required to create toolbox container.")
 
-		prompt := fmt.Sprintf("Download %s (500MB)? [y/N]:", imageFull)
+		var prompt string
+
+		if imageSize, err := getImageSizeFromRegistry(imageFull); err != nil {
+			logrus.Debugf("Getting image size failed: %s", err)
+			prompt = fmt.Sprintf("Download %s? [y/N]:", imageFull)
+		} else {
+			prompt = fmt.Sprintf("Download %s (%s)? [y/N]:", imageFull, imageSize)
+		}
+
 		shouldPullImage = askForConfirmation(prompt)
 	}
 
--- a/src/pkg/skopeo/skopeo.go
+++ b/src/pkg/skopeo/skopeo.go
@@ -1,0 +1,50 @@
+/*
+ * Copyright © 2023 Red Hat Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package skopeo
+
+import (
+	"bytes"
+	"encoding/json"
+
+	"github.com/containers/toolbox/pkg/shell"
+)
+
+type Layer struct {
+	Size json.Number
+}
+type Image struct {
+	LayersData []Layer
+}
+
+func Inspect(target string) (*Image, error) {
+	var stdout bytes.Buffer
+
+	targetWithTransport := "docker://" + target
+	args := []string{"inspect", "--format", "json", targetWithTransport}
+
+	if err := shell.Run("skopeo", nil, &stdout, nil, args...); err != nil {
+		return nil, err
+	}
+
+	output := stdout.Bytes()
+	var image Image
+	if err := json.Unmarshal(output, &image); err != nil {
+		return nil, err
+	}
+
+	return &image, nil
+}
--- a/meson.build
+++ b/meson.build
@@ -84,10 +84,6 @@ if shellcheck.found()
   test('shellcheck toolbox (deprecated)', shellcheck, args: [toolbox_sh])
 endif
 
-if not skopeo.found()
-    message('Running system tests requires Skopeo for OCI image manipulation.')
-endif
-
 install_subdir(
   'test',
   install_dir: get_option('datadir') / meson.project_name(),
--- a/playbooks/dependencies-centos-9-stream.yaml
+++ b/playbooks/dependencies-centos-9-stream.yaml
@@ -52,7 +52,7 @@
     chdir: '{{ zuul.project.src_dir }}'
 
 - name: Check versions of crucial packages
-  command: rpm -qa ShellCheck codespell *kernel* gcc *glibc* golang shadow-utils-subid-devel podman conmon containernetworking-plugins containers-common container-selinux crun fuse-overlayfs flatpak-session-helper
+  command: rpm -qa ShellCheck codespell *kernel* gcc *glibc* golang shadow-utils-subid-devel podman conmon containernetworking-plugins containers-common container-selinux crun fuse-overlayfs flatpak-session-helper skopeo
 
 - name: Show podman versions
   command: podman version
--- a/playbooks/dependencies-fedora-restricted.yaml
+++ b/playbooks/dependencies-fedora-restricted.yaml
@@ -54,6 +54,14 @@
     state: absent
     update_cache: "{{ true if zuul.attempts > 1 else false }}"
 
+- name: Ensure that skopeo(1) is absent
+  become: yes
+  package:
+    name:
+      - skopeo
+    state: absent
+    update_cache: "{{ true if zuul.attempts > 1 else false }}"
+
 - name: Download Go modules
   shell: |
     go mod download -x
--- a/playbooks/dependencies-fedora.yaml
+++ b/playbooks/dependencies-fedora.yaml
@@ -52,7 +52,7 @@
     chdir: '{{ zuul.project.src_dir }}'
 
 - name: Check versions of crucial packages
-  command: rpm -qa ShellCheck codespell *kernel* gcc *glibc* shadow-utils-subid-devel golang podman conmon containernetworking-plugins containers-common container-selinux crun fuse-overlayfs flatpak-session-helper
+  command: rpm -qa ShellCheck codespell *kernel* gcc *glibc* shadow-utils-subid-devel golang podman conmon containernetworking-plugins containers-common container-selinux crun fuse-overlayfs flatpak-session-helper skopeo
 
 - name: Show podman versions
   command: podman version
--- a/src/meson.build
+++ b/src/meson.build
@@ -21,6 +21,7 @@ sources = files(
   'cmd/utils.go',
   'pkg/podman/podman.go',
   'pkg/shell/shell.go',
+  'pkg/skopeo/skopeo.go',
   'pkg/utils/libsubid-wrappers.c',
   'pkg/utils/errors.go',
   'pkg/utils/utils.go',


debarshiray:
recheck

debarshiray:
The previous `RETRY_LIMIT` on Fedora 35 seems to have been a transient error:
```
TASK [Install RPM packages]
fedora-35 | ERROR
fedora-35 | {
fedora-35 |   "msg": "Failed to download metadata for repo 'fedora': Cannot prepare internal mirrorlist: Status code: 404 for https://mirrors.fedoraproject.org/metalink?repo=fedora-35&arch=x86_64 (IP: 38.145.60.21)",
fedora-35 |   "rc": 1,
fedora-35 |   "results": []
fedora-35 | }
```

src/pkg/podman/podman.go
```diff
@@ -280,6 +282,47 @@ func Inspect(typearg string, target string) (map[string]interface{}, error) {
...
+
+	if err := shell.Run("skopeo", nil, &stdout, nil, args...); err != nil {
+		return nil, err
+	}
+
+	output := stdout.Bytes()
+
+	var info interface{}
+
+	if err := json.Unmarshal(output, &info); err != nil {
```
debarshiray, line 299: ... then we can have a custom type called `Image` that implements the [json.Unmarshaler](https://pkg.go.dev/encoding/json#Unmarshaler) interface.  [This](https://go.dev/blog/json) gives a good introduction to handling JSON in Go.

There's a similar `Image` type in `src/pkg/podman/podman.go` that's used to parse JSON in `GetImages()`.  We could have used this same type to parse Skopeo's JSON, but it's different from Podman's JSON, so it might be better to keep them separate.

debarshiray (changes requested):
Thanks for working on this, Nieves!  Some comments below:

src/cmd/create.go
```diff
@@ -646,6 +647,7 @@ func getServiceSocket(serviceName string, unitName string) (string, error) {
 }
 
 func pullImage(image, release string) (bool, error) {
+	logrus.Debugf("This is the image we're looking for %s", image)
```
debarshiray, line 650: Nitpick: "Looking for image %s" will be better because it matches the logs elsewhere.

src/cmd/create.go
```diff
@@ -666,18 +668,19 @@ func pullImage(image, release string) (bool, error) {
 	}
 
 	var imageFull string
+	var imageSize string = "docker://"
```
debarshiray, line 671: The `docker://` prefix looks like an implementation detail of the `skopeo inspect ...` wrapper.  So, I would suggest moving it there.

debarshiray (changes requested):

debarshiray:
Due to the recent expansion of the test suite, we might have to increase the timeout for stable Fedoras: https://github.com/containers/toolbox/pull/1212

We recently had to do that for Fedora Rawhide.

src/cmd/create.go
```diff
@@ -554,6 +554,7 @@ func getFullyQualifiedImageFromRepoTags(image string) (string, error) {
 		imageFull = image
 	} else {
 		info, err := podman.Inspect("image", image)
+
```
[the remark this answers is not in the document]
debarshiray: Unresolving, because I still see the stray newline in the latest version.

src/cmd/create.go
debarshiray: Unresolving, because the log message is still the same.

src/cmd/create.go
```diff
@@ -666,18 +668,19 @@ func pullImage(image, release string) (bool, error) {
 	}
 
 	var imageFull string
+	var imageSize string = "docker://"
 
 	if hasDomain {
 		imageFull = image
+
```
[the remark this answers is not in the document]
debarshiray: Unresolving, because I still see the stray newline in the latest version.

src/pkg/podman/podman.go
```diff
@@ -261,6 +261,7 @@ func ImageExists(image string) (bool, error) {
 //
 // Parameter 'typearg' takes in values 'container' or 'image' that is passed to the --type flag
 func Inspect(typearg string, target string) (map[string]interface{}, error) {
+
```
debarshiray, line 264: Nitpick: stray newline.

src/pkg/podman/podman.go
```diff
@@ -271,6 +272,7 @@ func Inspect(typearg string, target string) (map[string]interface{}, error) {
 	}
 
 	output := stdout.Bytes()
+
```
debarshiray, line 275: Ditto.

debarshiray (changes requested):
Thanks for rebasing this on top of https://github.com/containers/toolbox/pull/1212

We still need to address the review comments about the code, though.

debarshiray:
recheck

src/pkg/skopeo/skopeo.go
```diff
@@ -0,0 +1,65 @@
...
+package skopeo
+
+import (
+	"bytes"
+	"encoding/json"
+
+	"github.com/containers/toolbox/pkg/shell"
+)
+
+func Inspect(typearg string, target string) (interface{}, error) {
```
debarshiray, line 26: The `typearg` argument is unused, so it might be better to not have it.

The reason `podman.Inspect` has a `typearg` argument is because one can use `podman inspect` with both OCI containers and images.  However, that's not the case with `skopeo inspect` because it can only be used with images.

src/pkg/skopeo/skopeo.go
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright © 2019 – 2022 Red Hat Inc.
```
debarshiray, line 2: Nitpick: should be `Copyright © 2023 Red Hat Inc.`.

src/cmd/create.go
```diff
@@ -698,9 +699,15 @@ func pullImage(image, release string) (bool, error) {
 	}
 
 	if promptForDownload {
+
+		infoSize, err := podman.Inspect("image", imageSize)
```
debarshiray, line 703: Shouldn't it be `skopeo.Inspect`, not `podman.Inspect`?

debarshiray (changes requested):
Thanks for the updates, @nievesmontero !  Could you please squash the commits into one? You can use git rebase -i for that.

Further comments below:

src/pkg/podman/podman.go
```diff
@@ -280,6 +282,47 @@ func Inspect(typearg string, target string) (map[string]interface{}, error) {
 	return info[0], nil
 }
 
+func InspectSkopeo(typearg string, target string) (interface{}, error) {
```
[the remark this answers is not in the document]
debarshiray: This has been done in the latest iteration of changes.

debarshiray (changes requested):
Thanks for the updates, @nievesmontero !  It looks like there are some remnants from a `git rebase` or `git merge` or something similar that had conflicts and wasn't fully cleaned up.

src/cmd/create.go
```diff
@@ -646,6 +647,7 @@ func getServiceSocket(serviceName string, unitName string) (string, error) {
 }
 
 func pullImage(image, release string) (bool, error) {
+	logrus.Debugf("Looking for image %s", image)
```
debarshiray, line 650: Nitpick: a newline after the `logrus.Debugf` will be good because it will separate the log message from the actual implementation of the function.

debarshiray (changes requested):
Some of the comments from the past reviews still need to be addressed.

debarshiray (changes requested):

debarshiray:
The main things that are left to do here are:

* Avoid a hard dependency on Skopeo by ensuring that the errors from `skopeo inspect ...` are not hard failures.

* Use a custom `Image` type that implements the [json.Unmarshaler](https://pkg.go.dev/encoding/json#Unmarshaler) interface for parsing JSON.

* Move the `docker://` prefix into `skopeo.Inspect`.

See above for details on each of these points.

src/pkg/podman/podman.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 – 2022 Red Hat Inc.
+ * Copyright © 2023 Red Hat Inc.
```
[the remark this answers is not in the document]
nievesmontero: In one of the previous comments you told me it should be 2023 instead of 2019-2022 https://github.com/containers/toolbox/pull/1208#discussion_r1087708243
Should I leave it then just the way it was?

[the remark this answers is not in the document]
debarshiray: Yes, that was for `src/pkg/skopeo/skopeo.go`, but this is `src/pkg/podman/podman.go`.

Usually, you only update the copyright notices in the files that you touch.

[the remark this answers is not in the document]
nievesmontero: My fault, I didn't read the file correctly, thank you!

src/cmd/create.go
```diff
@@ -699,9 +699,14 @@ func pullImage(image, release, authFile string) (bool, error) {
 	}
 
 	if promptForDownload {
+		infoSize, err := skopeo.Inspect(imageFull)
```
debarshiray, line 702: I see that you are now passing `imageFull`, which doesn't have the `docker://` prefix.  That's better.  However, I don't see the prefix being added inside `skopeo.Inspect`.

src/pkg/podman/podman.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 – 2022 Red Hat Inc.
+ * Copyright © 2019-2022 Red Hat Inc.
```
debarshiray, line 2: Stray unrelated change.

debarshiray (changes requested):

src/pkg/podman/podman.go
nievesmontero: This change was made by mistake after fixing some problems I had.

src/cmd/create.go
nievesmontero: Now I am passing the string ```docker://``` inside the ```skopeo.Inspect()``` method, I hope that works better.

src/pkg/podman/podman.go
debarshiray: Unresolving, because it's still there.  :)

src/cmd/create.go
```diff
@@ -678,8 +679,6 @@ func pullImage(image, release, authFile string) (bool, error) {
 		}
 	}
 
-	logrus.Debugf("Looking for image %s", imageFull)
-
```
debarshiray, line 682: This debug message shouldn't be removed.

debarshiray (changes requested):

src/cmd/create.go
nievesmontero: Included again :)

src/pkg/podman/podman.go
nievesmontero: Modified :)

src/cmd/create.go
debarshiray: No, it should be inside the body of the `skopeo.Inspect` function.

Like I said [before](https://github.com/containers/toolbox/pull/1208#discussion_r1073452041), the `docker://` prefix is something that `skopeo inspect` needs.  The caller of `skopeo.Inspect` doesn't need it and doesn't need to know about it.  So, it should be inside `skopeo.Inspect`.

nievesmontero:
recheck

src/cmd/create.go
```diff
@@ -699,9 +700,14 @@ func pullImage(image, release, authFile string) (bool, error) {
 	}
 
 	if promptForDownload {
+		infoSize, err := skopeo.Inspect(imageFull)
+		if err != nil {
+			logrus.Debug("cannot get the size of the image")
```
debarshiray, line 705: Since we are suppressing the original error, it will be good to dump it in the logs.  Something like:
```
  logrus.Debugf("Getting image size failed: %s", err)
```

src/cmd/create.go
```diff
@@ -699,9 +700,14 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+		}
+
 		fmt.Println("Image required to create toolbox container.")
 
-		prompt := fmt.Sprintf("Download %s (500MB)? [y/N]:", imageFull)
+		prompt := fmt.Sprintf("Download %s (%vMB)? [y/N]:", imageFull, infoSize)
```
debarshiray, line 710: If we fail to get the image size, then this message will look odd:
```
Download <fully-qualified-image-name> [0MB]? [y/N]:
```

In many cases, it will fail because the `skopeo(1)` binary was absent from the system, which I think should be a [valid](https://github.com/containers/toolbox/issues/752) use-case.  So we should try to cover those situations elegantly.

debarshiray (changes requested):
Thanks for the updates, @nievesmontero !  This is looking a lot better, well done.  Some minor details:

debarshiray (commented):
One quick note about the commit message ...

Unless the change is massive, it's not so important to write things like:
```
The following package and file have been created src/pkg/skopeo/skopeo.go
```

... because it's obvious from looking at the commit (with `git show`) and is of no interest to downstream distributors.  What's more important is that it adds an optional dependency on Skopeo, because that's going to be important for downstream distributors, and to clarify why it's optional.  Those details aren't so obvious until someone carefully looks at the commit, but they are still important.

src/cmd/create.go
nievesmontero: I tried to solve this by adding an else on the code. Check if it works for you or if it could be done in a more elegantly way.

src/cmd/create.go
```diff
@@ -699,10 +700,20 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+			fmt.Println("Image required to create toolbox container.")
 
-		prompt := fmt.Sprintf("Download %s (500MB)? [y/N]:", imageFull)
-		shouldPullImage = askForConfirmation(prompt)
+			prompt := fmt.Sprintf("Download %s (500MB)? [y/N]:", imageFull)
+			shouldPullImage = askForConfirmation(prompt)
+		} else {
+			fmt.Println("Image required to create toolbox container.")
+
+			prompt := fmt.Sprintf("Download %s (%sMB)? [y/N]:", imageFull, infoSize)
```
debarshiray, line 714: The `MB` is getting duplicated because it's part of the human-readable size.  :)

src/pkg/skopeo/skopeo.go
```diff
@@ -0,0 +1,55 @@
...
+	var image Image
+	if err := json.Unmarshal(output, &image); err != nil {
+		return "", err
+	}
+
+	var totalSize int64 = 0
+	for _, layer := range image.LayersData {
+		totalSize += layer.Size
+	}
+	sizeInMB := units.HumanSize(float64(totalSize))
```
debarshiray, line 53: Nitpick: see above.

The size might not necessarily be in MB.  It's just in a human-readable form.  eg., It could also be in GB for sufficiently large images.

debarshiray (changes requested):

src/pkg/skopeo/skopeo.go
nievesmontero: You're right

src/pkg/skopeo/skopeo.go
```diff
@@ -0,0 +1,55 @@
...
+		return "", err
+	}
+
+	output := stdout.Bytes()
+	var image Image
+	if err := json.Unmarshal(output, &image); err != nil {
+		return "", err
+	}
+
+	var totalSize int64 = 0
```
debarshiray, line 49: Nitpick: just `var totalSize int64` is enough.  There's no need for the `= 0`.

src/pkg/skopeo/skopeo.go
```diff
@@ -0,0 +1,55 @@
...
+	targetWithTransport := "docker://" + target
+
+	args := []string{"inspect", targetWithTransport}
+	if err := shell.Run("skopeo", nil, &stdout, nil, args...); err != nil {
+		return "", err
+	}
+
+	output := stdout.Bytes()
+	var image Image
+	if err := json.Unmarshal(output, &image); err != nil {
```
debarshiray, line 45: Like we were discussing on chat, were you able to test the size calculation with a Skopeo older than `v1.10.0`?  That's when the LayersData array was [added](https://github.com/containers/skopeo/pull/1738) to `skopeo inspect` so, I am curious how it handles older versions.

debarshiray: I tested with an older Skopeo and found that the prompt was:
```
Download <fully-qualified-image-name> [0B]? [y/N]:
```

I took the liberty to make the necessary changes to not show that `0B`, as we discussed [before](https://github.com/containers/toolbox/pull/1208#discussion_r1126577835).

PR #1209, opened 2023-01-11
test/system: Tighten the regular expression used to detect the image ID

debarshiray:
An OCI image's identifier (or ID) is a SHA-256 hash [1] whose length is 64 hexadecimal digits.  The ID can be abbreviated to a smaller number of digits but the minimum is 6.  The regular expression being used to parse the image ID wasn't checking the number of digits, and would even accept an empty string as a valid ID, which was clearly wrong.

The format of a image reference is defined in:
https://github.com/moby/moby/blob/master/image/spec/v1.1.md https://github.com/distribution/distribution/blob/main/reference/reference.go https://github.com/distribution/distribution/blob/main/reference/regexp.go

Fallout from c6c2e426e0ad801ec06ea9c4fbfd1a9a1bbe7c99

[1] https://en.wikipedia.org/wiki/SHA-2


halfline:
is it conceivable that the hex digits could ever be capitalized ?

debarshiray:
> is it conceivable that the hex digits could ever be capitalized ?

I was thinking about it too.  Currently the reference [grammar](https://github.com/distribution/distribution/blob/main/reference/reference.go) and the reference [regular expression](https://github.com/distribution/distribution/blob/main/reference/regexp.go) defines `identifier` to have only lowercase alphabets.

I have at times thought about dropping our handcrafted image parsing code (eg., [ImageReferenceCanBeID](https://github.com/containers/toolbox/blob/main/src/pkg/utils/utils.go)) in favour of [github.com/distribution/distribution](https://github.com/distribution/distribution) but @alexlarsson 's [observations](https://github.com/containers/podman/issues/16205) about these libraries makes me hesitate.  However, that's just the actual Go code for the `toolbox` binary, not the shell scripts in the test suite.

debarshiray:
Anyway, I went and merged this pull request because regardless of whether we accommodate capitalized hexadecimal digits or not, we can't accept empty strings as a valid image ID.  :)

PR #1211, opened 2023-01-12
Silence ShellCheck-0.9.0

debarshiray:
Fedora Rawhide now has ShellCheck-0.9.0, which flags these new problems, while so far it only had ShellCheck-0.8.0.


debarshiray:
recheck

debarshiray:
recheck

PR #1213, opened 2023-01-18
cmd/run, pkg/podman: Stop container once the last session finishes

debarshiray:
Currently, once a toolbox container gets started with 'podman start', as part of the 'toolbox enter' command, it doesn't stop unless the host is shut down or someone explicitly calls 'podman stop'. This becomes annoying if someone tries to remove the container because commands like 'podman rm' and such don't work without the '--force' flag, even if all active 'toolbox enter' and 'toolbox run' sessions have terminated.

A crude form of reference counting has been set up that depends on 'podman stop' failing as long there's any active 'podman exec' session left.  Every invocation of 'podman exec' in 'enter' and 'run' is followed by 'podman stop', so that the container gets stopped once the last session finishes.

While this approach looks very crude at first glance, it does have the advantage of being ridiculously simple to implement.  Thus, it's a lot more robust and easier to verify than setting up some custom reference counting or synchronization using other means like POSIX signals or file locks.

Based on the implementation in github.com/coreos/toolbox.

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 – 2022 Red Hat Inc.
+ * Copyright © 2019 – 2023 Red Hat Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -284,6 +284,8 @@ func runCommand(container string,
 	}
 
 	logrus.Debugf("Container %s is initialized", container)
+
+	defer stopContainer(container)
 
 	if err := runCommandWithFallbacks(container,
 		preserveFDs,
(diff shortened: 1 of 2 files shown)


halfline:
I guess this is an alternative to https://github.com/containers/toolbox/pull/1207 ? If so, feel free to close https://github.com/containers/toolbox/pull/1207

debarshiray:
> I guess this is an alternative to #1207 ? If so, feel free to close #1207

Umm... no, I don't think they address the same problem.

This pull request was meant to address https://github.com/containers/toolbox/issues/114 which is about stopping the container (ie., exiting the entry point spawned by `podman start`) when the last `podman exec` session goes away during normal use, so that `--force` is not necessary with `podman rm` and the output of `toolbox list` is more intuitive.

https://github.com/containers/toolbox/pull/1207 is about addressing https://github.com/containers/toolbox/issues/1204 , which is about stopping the process started by `podman exec` because the terminal disappeared (ie., `SIGHUP`) or the machine is getting shut down (ie., `SIGTERM`).

src/cmd/run.go
```diff
@@ -285,6 +285,8 @@ func runCommand(container string,
 
 	logrus.Debugf("Container %s is initialized", container)
 
+	defer stopContainer(container)
```
debarshiray, line 288: This [doesn't](https://github.com/containers/toolbox/issues/114#issuecomment-1387701140) work with current implementations of `podman stop`.

Issue #1059, opened 2022-05-22
VSCode/Codium freeze whole system (does recover)


KirmesBude:
**Describe the bug**
When using VSCode/Codium inside the toolbox, the whole system regularly freezes for >=1 second. 
Especially easily reproducible if hovering over any of the menus. 
The VSCode/Codium themeing also does not look themed correct, so maybe this is part of the problem.
I do not face this problem with my laptop running silverblue 35.

**Steps how to reproduce the behaviour**
1. Create and enter a toolbox
2. install vscode/codium
3. run vscode/codium

**Expected behaviour**
I can use vscode/codium from within the toolbox no problem.

**Actual behaviour**
The whole system freezes. Sometimes for multiple seconds.

**Screenshots**
https://youtu.be/byqXiMcQ-bs

**Output of `toolbox --version` (v0.0.90+)**
toolbox version 0.0.99.3

**Toolbox package info (`rpm -q toolbox`)**
toolbox-0.0.99.3-2.fc35.x86_64

**Output of `podman version`**
```
Version:      3.4.7
API Version:  3.4.7
Go Version:   go1.16.15
Built:        Thu Apr 21 15:14:26 2022
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
podman-3.4.7-1.fc35.x86_64

**Info about your OS**
Kinoite 35

**Additional context**
Host Flatpacked codium works normally.
I am not sure if this is the right place to put this bug report. 
Seems to be Kinoite specific. I would be interested in any tips on how to proceed with troubleshooting.

HarryMichal:
Hi @KirmesBude! This sounds like a tricky bug.

> I do not face this problem with my laptop running silverblue 35.

Is that on the same machine? As far as I know the only difference between Silverblue and Kinoite should be the desktop environment.

> Host Flatpacked codium works normally.

This probably tells us it's not because of the graphics stack? *And it's Flatpak :wink:*

Could you also try to run VSCode from an RPM, please?

> I am not sure if this is the right place to put this bug report.

One could go through Kinoite's bug tracker but myself am not actually sure where that is.

> I would be interested in any tips on how to proceed with troubleshooting.

I would suggest you watch the journal (`journalctl`) and run VSCode with debug logging enabled.

KirmesBude:
> > I do not face this problem with my laptop running silverblue 35.
> 
> Is that on the same machine? As far as I know the only difference between Silverblue and Kinoite should be the desktop environment.

[...]

PR #1215, opened 2023-01-21
Set hostname to toolbox.{host_hostname}

jnohlgard:
Fixes https://github.com/containers/toolbox/issues/1059 when running X11 apps under kwin/Plasma desktop on Fedora Kinoite 37.

The random freezes experienced in #1059 are caused by DNS lookups of the unqualified host name 'toolbox' that the container uses by default, at least on my machine. Distrobox (https://github.com/89luca89/distrobox) uses the same style of container host name as this patch and has never had any issues for me under KDE.

An alternative solution which does not modify the host name is to explicitly add 'toolbox' to the hosts file or DNS, but it is more user friendly to have something that just works without any changes to the host configuration.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -409,8 +409,13 @@ func createContainer(container, image, release string, showCommandToEnter bool) 
 
 	createArgs = append(createArgs, xdgRuntimeDirEnv...)
 
+    localHostname, err := os.Hostname()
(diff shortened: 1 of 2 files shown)


jkemp814:
https://github.com/containers/toolbox/issues/98#issuecomment-1422038509

PR #1231, opened 2023-02-07
images: Synchronize README.md

debarshiray:
The canonical copy of README.md contains banners and labels in the header that aren't useful when the file is shipped as part of the images.  Hence, those were removed.

Only the images for currently maintained Fedoras (ie., 36, 37 and 38) were updated.


debarshiray:
recheck

Issue #1246, opened 2023-02-26
Completion generation with `meson` requires valid subgid/subuid


Foxboron:
**Describe the bug**
During building meson will try to create completions and this fails as the user doesn't have a set subuid/subgid in the build environment.

**Steps how to reproduce the behaviour**
Build toolbox without a subuid/subgid for the user.

**Expected behaviour**
Building toolbox in a restrictive build environment should work.

**Actual behaviour**
Building toolbox in a restrictive build environment doesn't work.

**Screenshots**

**Output of `toolbox --version` (v0.0.90+)**
Latest toolbox release.

Foxboron:
I think we can do something like this.

```diff
λ src main» git diff cmd/completion.go
diff --git a/src/cmd/completion.go b/src/cmd/completion.go
index 26ed378..c71135b 100644
--- a/src/cmd/completion.go
+++ b/src/cmd/completion.go
@@ -32,6 +32,7 @@ var completionCmd = &cobra.Command{
        ValidArgs:             []string{"bash", "fish", "zsh"},
        Args:                  cobra.ExactValidArgs(1),
        RunE:                  completion,
+       PersistentPreRunE:     func(cmd *cobra.Command, args []string) error { return nil },
 }

 func init() {
```

It would ignore some of the setup done with regard to the configuration and logging, but do we need that for the completion sub command?

antonc42:
I'm building toolbox with meson in a Python virtualenv and getting this error. Is it related to this bug? Is there a workaround? I'm not familiar with subuid/subgid. 

```
meson setup -Dprofile_dir=/etc/profile.d builddir
The Meson build system
Version: 1.0.1
Source dir: /home/user/scripts/update.d/toolbox-build/0.0.99.4
Build dir: /home/user/scripts/update.d/toolbox-build/0.0.99.4/builddir
Build type: native build
Project name: toolbox
Project version: 0.0.99.4
C compiler for the host machine: cc (gcc 9.4.0 "cc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0")
C linker for the host machine: cc ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64
Compiler for C supports arguments -print-file-name=libc.so: YES 

meson.build:17:0: ERROR: C header 'shadow/subid.h' not found

A full log can be found at /home/user/scripts/update.d/toolbox-build/0.0.99.4/builddir/meson-logs/meson-log.txt
```

[...]

PR #1249, opened 2023-03-01
If podman is not needed do not check its config

JanZerebecki:
to be able to build on OpenSUSE.

Following up on fc5f568, which added a check to not require migration if podman will never be run. Move the check earlier to also not check for the subuid config.

Note: I see no DCO or definition of what Signed-off-by means in this
project and the CI requires it, so I just sign off on this commit
without it meaning anything.

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -127,6 +127,7 @@ func init() {
 }
 
 func preRun(cmd *cobra.Command, args []string) error {
+	migrationNeeded := true
 	cmd.Root().SilenceUsage = true
 
 	if err := setUpLoggers(); err != nil {
@@ -136,7 +137,12 @@ func preRun(cmd *cobra.Command, args []string) error {
 	logrus.Debugf("Running as real user ID %s", currentUser.Uid)
 	logrus.Debugf("Resolved absolute path to the executable as %s", executable)
 
-	if !utils.IsInsideContainer() {
+	if cmdName, completionCmdName := cmd.Name(), completionCmd.Name(); cmdName == completionCmdName {
+		logrus.Debugf("Migration not needed: command %s doesn't need it", cmdName)
+		migrationNeeded = false
+	}
+
+	if migrationNeeded && !utils.IsInsideContainer() {
 		logrus.Debugf("Running on a cgroups v%d host", cgroupsVersion)
 
 		if currentUser.Uid != "0" {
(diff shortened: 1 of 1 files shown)


src/cmd/root.go
```diff
@@ -136,7 +137,12 @@ func preRun(cmd *cobra.Command, args []string) error {
 	logrus.Debugf("Running as real user ID %s", currentUser.Uid)
 	logrus.Debugf("Resolved absolute path to the executable as %s", executable)
 
-	if !utils.IsInsideContainer() {
+	if cmdName, completionCmdName := cmd.Name(), completionCmd.Name(); cmdName == completionCmdName {
+		logrus.Debugf("Migration not needed: command %s doesn't need it", cmdName)
+		migrationNeeded = false
+	}
+
+	if migrationNeeded && !utils.IsInsideContainer() {
```
debarshiray, line 145: This code that the `migrationNeeded` variable is guarding isn't actually related to migrating to newer Podman versions at all.  So, the code becomes a bit misleading.

It might be better to wrap `utils.ValidateSubIDRanges` in (another) convenience function that encapsulates all these details of whether we are generating the shell completions or inside a container or running as root.

src/cmd/root.go
```diff
@@ -218,11 +226,6 @@ func migrate(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	if cmdName, completionCmdName := cmd.Name(), completionCmd.Name(); cmdName == completionCmdName {
```
debarshiray, line 221: If this is moved out of the function, then the `cmd` and `args` parameters should also be removed.

However, I had deliberately kept this condition inside the function to make the code a bit more robust against future refactorings, and keeping the clutter out of the upper level functions.

debarshiray (changes requested):
Thanks for the patch, @JanZerebecki !  Seems like our friends at Arch Linux also had the [same](https://github.com/containers/toolbox/issues/1246) problem.

debarshiray:
@JanZerebecki , just before I noticed your pull request, I had just done pretty much the same thing in https://github.com/containers/toolbox/pull/1252

Since https://github.com/containers/toolbox/pull/1252 also has tests, I am going to close this one in favour of https://github.com/containers/toolbox/pull/1252 but I took the liberty to attribute this subset of the changes to you.

src/cmd/root.go
debarshiray: Changed in https://github.com/containers/toolbox/pull/1252

src/cmd/root.go
debarshiray: https://github.com/containers/toolbox/pull/1252

debarshiray:
Closing.  Thanks for taking care of Toolbx in openSUSE!

Issue #1263, opened 2023-03-16
Run 'skopeo inspect' asynchronously in 'create' to not block the user interaction


debarshiray:
This is a continuation from https://github.com/containers/toolbox/issues/752

We now use `skopeo inspect` to query the size of the image from the registry, which is wonderful.  However, `skopeo inspect` takes a few seconds to run, which leads to a non-ideal user experience, because the command seems stuck for a bit:
```
$ toolbox create
Image required to create toolbox container.
<wait for a few seconds>
Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)? [y/N]: 
```

Instead, it will be better if we run `skopeo inspect` asynchronously in the background, while waiting for the user to respond to the download prompt, and later update the prompt in-place with the size.  Something like this in the beginning, while `skopeo inspect` is running:
```
$ toolbox create
Image required to create toolbox container.
Download registry.fedoraproject.org/fedora-toolbox:39 (...)? [y/N]: 
```
*where the ... is some spinner to indicate activity*

... and then once we have the size:
```
$ toolbox create
Image required to create toolbox container.
Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)? [y/N]: 
```

If the `skopeo inspect` fails we don't show the size, like we do now:
```
$ toolbox create
Image required to create toolbox container.
Download registry.fedoraproject.org/fedora-toolbox:39? [y/N]: 
```

debarshiray:
I think that [Goroutines](https://go.dev/tour/concurrency/1) and [channels](https://go.dev/tour/concurrency/2) can help with this.

debarshiray:
Fixed by https://github.com/containers/toolbox/pull/1415

PR #1274, opened 2023-03-21
Run 'skopeo inspect' asynchronously in 'create' to not block the user interaction

nievesmontero:
Fixes https://github.com/containers/toolbox/issues/1263

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -711,20 +711,47 @@ func pullImage(image, release, authFile string) (bool, error) {
 	}
 
 	if promptForDownload {
+		imageSizeChan := make(chan string)
+		confirmationChan := make(chan bool)
+
 		fmt.Println("Image required to create toolbox container.")
 
-		var prompt string
+		prompt := fmt.Sprintf("Download %s (...)? [y/N]:", imageFull)
 
-		if imageSize, err := getImageSizeFromRegistry(imageFull); err != nil {
-			logrus.Debugf("Getting image size failed: %s", err)
-			prompt = fmt.Sprintf("Download %s? [y/N]:", imageFull)
-		} else {
-			prompt = fmt.Sprintf("Download %s (%s)? [y/N]:", imageFull, imageSize)
+		go func() {
+			shouldPullImage := askForConfirmation(prompt)
+			confirmationChan <- shouldPullImage
+		}()
+
+		// Launch a goroutine to call getImageSizeFromRegistry and send the size to the channel
+		go func() {
+			imageSize, err := getImageSizeFromRegistry(imageFull)
+			if err != nil {
+				logrus.Debugf("Getting image size failed: %s", err)
+				imageSize = ""
+			}
+			imageSizeChan <- imageSize
+		}()
+
+		var imageSize string
+		var userResponded = false
+
+		for !userResponded {
+			select {
+			case shouldPullImage = <-confirmationChan:
+				userResponded = true
+
+			case imageSize = <-imageSizeChan:
+				if imageSize != "" {
+					prompt = fmt.Sprintf("\rDownload %s (%s)? [y/N]:", imageFull, imageSize)
+				} else {
+					prompt = fmt.Sprintf("\rDownload %s? [y/N]:     ", imageFull)
+				}
+				fmt.Printf("%s", prompt)
+			}
 		}
 
-		shouldPullImage = askForConfirmation(prompt)
 	}
-
 	if !shouldPullImage {
 		return false, nil
 	}


src/cmd/create.go
```diff
@@ -710,16 +710,33 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+		// Show the download question immediately
+		prompt := fmt.Sprintf("Download %s (...)? [y/N]:", imageFull)
+		fmt.Println(prompt)
 
-		if imageSize, err := getImageSizeFromRegistry(imageFull); err != nil {
-			logrus.Debugf("Getting image size failed: %s", err)
-			prompt = fmt.Sprintf("Download %s? [y/N]:", imageFull)
-		} else {
+		// Wait until the size is received
+		imageSize := <-imageSizeChan
```
debarshiray, line 733: This means that the user can't respond to the prompt before the `skopeo inspect` returns, which is odd.  The user should be able to respond `yes` or `no` regardless of whether `skopeo inspect` has returned or not.

src/cmd/create.go
```diff
@@ -710,16 +710,33 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+
+		// Update the string with the image size if it was received or skip the size if it was not obtained
+		if imageSize != "" {
 			prompt = fmt.Sprintf("Download %s (%s)? [y/N]:", imageFull, imageSize)
+		} else {
+			prompt = fmt.Sprintf("Download %s? [y/N]:", imageFull)
```
debarshiray, line 739: This doesn't replace the prompt in-place, but instead adds a new prompt on the next line, which is odd.

debarshiray (changes requested):
Thanks for working on this, @nievesmontero   It's a good start but there are some problems.  See the comments below.

There are two big parts to this.  First is to ensure that the user can respond to the prompt (showing a spinner) while `skopeo inspect` is running.  Then, if Skopeo returns before the user has responded, to update the prompt in-place (with the size).

For the first part, there are two concurrent actions that need to be co-ordinated.  First is `getImageSizeFromRegistry()` returning either a `string` or an `error`, and then `askForConfirmation()` returning a `bool`.  Both those concurrent actions should run in separate new goroutines, and each of those three possible return values should have a corresponding channel (ie., a `string`, an `error` and a `bool`) to send them back from their goroutines to the main goroutine.  Reading from the three channels should be co-ordinated using a [select](https://go.dev/tour/concurrency/5).  This should be enough to ensure that the user can respond to the prompt while `skopeo inspect` is running.

Then there's the issue of updating the prompt in place.  For that we need a way to tell `askForConfirmation()` to update it.  This means that we need to wake it up from waiting for user input, delete the current prompt, print the new one, and then go back to waiting for user input.  So, we need a way to wake up `askForConfirmation()` and for that we might have to asynchronous reads.

src/cmd/create.go
```diff
@@ -710,16 +710,33 @@ func pullImage(image, release, authFile string) (bool, error) {
 		shouldPullImage = true
 	}
 
+	imageSizeChan := make(chan string)
+
 	if promptForDownload {
+
```
debarshiray, line 716: Nitpick: stray newline.

nievesmontero:
> Also, the commit message shouldn't exceed 72 characters per line.

I didnt't know that, I've already shortened the message

src/cmd/create.go
```diff
@@ -710,16 +710,33 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+				logrus.Debugf("Getting image size failed: %s", err)
+				imageSize = ""
+			}
+			imageSizeChan <- imageSize
+		}()
 
-		var prompt string
+		// Show the download question immediately
+		prompt := fmt.Sprintf("Download %s (...)? [y/N]:", imageFull)
+		fmt.Println(prompt)
```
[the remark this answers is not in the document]
nievesmontero: I think this was solved with my commit update :)

src/cmd/create.go
```diff
@@ -710,18 +710,42 @@ func pullImage(image, release, authFile string) (bool, error) {
...
 
+	imageSizeChan := make(chan string)
+
 	if promptForDownload {
 		fmt.Println("Image required to create toolbox container.")
+		// Launch a goroutine to call getImageSizeFromRegistry and send the size to the channel
+		go func() {
+			imageSize, err := getImageSizeFromRegistry(imageFull)
+			if err != nil {
+				logrus.Debugf("Getting image size failed: %s", err)
```
debarshiray, line 721: We should send the `error` back over a channel and have the main goroutine do the logging.  Otherwise, the logs will interfere with the confirmation prompt.

src/cmd/create.go
```diff
@@ -710,18 +710,42 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+				imageSize = ""
+			}
+			imageSizeChan <- imageSize
+		}()
+
+		// Show the download question immediately
+		prompt := fmt.Sprintf("Download %s (...)? [y/N]:", imageFull)
+		fmt.Printf("%s", prompt)
```
debarshiray, line 729: The prompt is being shown immediately, but the user can't interact with it.  It should use `askForConfirmation()` and run in a separate goroutine.

src/cmd/create.go
```diff
@@ -710,18 +710,42 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+
+		// Wait until the size is received or a timeout occurs
+		timeout := time.After(5 * time.Second)
```
debarshiray, line 732: Do we really need a timer?  We could have a timer that prevents against  `skopeo inspect` getting stuck for some reason, but it's probably not that important or necessary.  eg., we don't worry about `podman pull` getting stuck.  If there's some problem with the network, I would expect `skopeo inspect` to eventually time out on its own.

For the vast majority of cases, the user will act as the *timer*.  If `skopeo inspect` is too slow, then the user would probably decide `yes` or `no` before it fetches the size.  That's the thing that should really work.

debarshiray (changes requested):
Did you see the [hints](https://github.com/containers/toolbox/pull/1274#pullrequestreview-1352607257) in my previous review?  Did they make sense?

src/cmd/create.go
nievesmontero: I agree, I better remove the timer.

src/cmd/create.go
nievesmontero: I think I've solved this.

src/cmd/create.go
debarshiray: This is resolved because the user can respond to the prompt before the `skopeo inspect` returns.

src/cmd/create.go
```diff
@@ -710,19 +710,50 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+				imageSize = ""
+			}
+			imageSizeChan <- imageSize
+		}()
+
+		var imageSize string
+		select {
+		case imageSize = <-imageSizeChan:
+			if imageSize != "" {
+				prompt = fmt.Sprintf("\rDownload %s (%s)? [y/N]:", imageFull, imageSize)
```
debarshiray, line 740: The race is here.  We have 3 threads of execution - the two goroutines above and the *main* thread.  By the time we reach here, we can only be sure that the second goroutine attempted to get the size from the registry with `skopeo inspect`, but there's no guarantee that the first goroutine has shown the prompt for the first time already.

This is unlikely to happen, because in most cases running `skopeo inspect` will be so slow that the first goroutine to show the prompt will get scheduled to run on the CPU before `skopeo inspect` returns.  However, it can still happen if the stars align against us.  :)

To avoid this uncertainty, we need to cancel the first `askForConfirmation` call, and then start a second one.  This is why I wrote the *cancellable* asynchronous read [example](https://rishi.fedorapeople.org/async-read.go) for you.  Here, we need to create an `askForConfirmationAsync` analogous to the `scanFileAsync` in the example.

src/cmd/create.go
```diff
@@ -710,19 +710,50 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+			} else {
+				prompt = fmt.Sprintf("\rDownload %s? [y/N]:", imageFull)
+			}
+			fmt.Printf("%s", prompt)
```
debarshiray, line 744: If we have an `askForConfirmationAsync` then the second updated prompt will be displayed by it, and it won't lose the image size.

debarshiray (changes requested):
Thanks for updating the branch, @nievesmontero !  This is a lot better, but there is a problem with unexpected input and a race condition.

Here's the problem with unexpected input:
```
$ toolbox create
Download registry.fedoraproject.org/fedora-toolbox:36 (294.1MB)? [y/N]:d
Download registry.fedoraproject.org/fedora-toolbox:36 (...)? [y/N]: d
Download registry.fedoraproject.org/fedora-toolbox:36 (...)? [y/N]: d
Download registry.fedoraproject.org/fedora-toolbox:36 (...)? [y/N]: N
```

Notice how the prompt loses the image size.

By the way, did you study the *cancellable* asynchronous read [example](https://rishi.fedorapeople.org/async-read.go) that I showed you some time ago?  It shows how to add cancellation to a read operation (eg., `fmt.Scanf`).  `scanFileAsync` is like `askForConfirmation`, with the extra feature that it can be cancelled.  I think it will be important here.

debarshiray:
Could you please rebase this branch against `main`?  It's missing [this](https://github.com/containers/toolbox/commit/7269547af923a701d0b6f70f2c04e101071bf52c) bug-fix in `askForConfirmation` that's relevant here.

nievesmontero:
> Could you please rebase this branch against `main`? It's missing [this](https://github.com/containers/toolbox/commit/7269547af923a701d0b6f70f2c04e101071bf52c) bug-fix in `askForConfirmation` that's relevant here.

Done :)

debarshiray (commented):
@nievesmontero is this ready for review?  Did you address the above comments?

nievesmontero:
> @nievesmontero is this ready for review? Did you address the above comments?

Yes, it is ready for review. I tried to make the changes you asked me to and tried implementing the `askForConfirmationAsync`.

src/cmd/create.go
```diff
@@ -710,19 +717,48 @@ func pullImage(image, release, authFile string) (bool, error) {
...
+				select {
+				case shouldPullImage = <-confirmationChan:
+					userResponded = true
+				case <-time.After(time.Second):
+					break
+				}
+			}
+		case shouldPullImage = <-confirmationChan:
+			break
+		}
```
HarryMichal, line 761: Just like @debarshiray pointed out, there's a race condition here. Additionally, we have an unresolved design decision. First, about the race.

Rishi already summed it up well, so I'll just highlight some concrete cases when things can go wrong:

1. The first example is simple and is more of a logical error. The dialog is first printed by `askForConfirmation()` and once the image size is received, we call `fmt.Printf()` outside of it. But the prompt in `askForConfirmation()` does not fail fast and is forgiving when the user provides unknown input. Hence, you can end up with output like:
  
  ```
  ; toolbox create -r 40
  Download registry.fedoraproject.org/fedora-toolbox:40 (1.014GB)? [y/N]:foo
  Download registry.fedoraproject.org/fedora-toolbox:40 (...)? [y/N]: foo
  Download registry.fedoraproject.org/fedora-toolbox:40 (...)? [y/N]: 
  ```
  
  The `1.014GB` is replaced by `...` because the value was only printed but not kept around and definitely not communicated to `askForConfirmation()` which still has the old prompt in it. Thus, the value is lost.

2. The second case is a bit harder to reproduce but in the current implementation is possible. Let the scenario play out like this:

  ```
start askForConfirmation()
   \
    \            return getImageSizeFromRegistry()  
     \                \
  ---------------------------------------------------------------------
         /                   /                  /
        /                print new prompt    print old prompt
       /
  start getImageSizeFromRegistry()
  ```

This scenario is extremely unlikely but currently can happen due to two unsynchronized interactions with the prompt.

Both issues are a reason to create a fully async `askForConfirmation()` function which handles updates to the prompt in a synchronized and predictable manner.

HarryMichal (changes requested):
This looks like a good start and in local testing this seems to be doing the job its supposed to. Nice job!

Still, just like @debarshiray noted, I see as well some potential issues with how the goroutines are handled. See the notes attached to the patch for details on what the potential issues are.

Also, I don't see anywhere a discussion on how the dynamic CLI content should be handled from a design perspective. There are multiple approaches and all of them have their pros and cons.

Currently, our CLI interaction flow is generally the following:

```
do -> [write (user reads) -> read (user interacts)] -> do
     /
    /
interactive
```

With this patch we're changing this flow significantly by making it look like this:

```
do -> [print (user reads) -> read? (user interacts) -> update? (change the printed content) -> read (user interacts)] -> do
     /
    / 
interactive
```

The key difference is that until now we did not update the content we once printed. Updating CLI content is tricky (mainly if we're expecting input from the user) as overwriting existing content (and thus confusing the user) is extremely easy.

Currently the patch upon updating the prompt overwrites my prompt with several characters, leaving me guessing what characters I had already written. I see several approaches to this issue:

1. Update after submitting input.

This approach sacrifices the convenience by requiring the user to interact to get the size of the downloaded image. This could look something like:

```bash
Download registry.fedoraproject.org/fedora-toolbox:40 (press S to get image size)? [y/N/S]:
```

2. Block user input until the size available.

This approach would allow to use [briandowns/spinner](https://github.com/briandowns/spinner) to simplify updating the prompt but it would almost completely negate the original intent of making the operation non-blocking. Still, it fits our CLI experience the best as it makes use of existing capabilities.

3. Make use of a TUI toolkit (e.g., [rivo/tview](https://github.com/rivo/tview), [charmbracelet/bubbletea](https://github.com/charmbracelet/bubbletea),..) to build our CLI experience.

A significantly more involved and in my eyes quite unrealistic approach where we'd rebuild the experience using a TUI. Long project, low benefit (for a single interactive prompt).

Also, having a strategy for caching the results could help us reduce bandwidth and make the results available much quicker.

src/cmd/create.go
nievesmontero: Thanks for reviewing @HarryMichal ! I'll continue working on it and try to achieve all the requirements.

HarryMichal:
@nievesmontero, any progress on this?

@debarshiray, any ideas about the design we want for the CLI experience?

nievesmontero:
recheck

nievesmontero:
> @nievesmontero, any progress on this?
> 
> @debarshiray, any ideas about the design we want for the CLI experience?

I made changes on the structure of the go routine and rechecked to fix the two tests that were failing. It's pending for review.

HarryMichal:
@nievesmontero, I still see some unresolved topics in this PR. Namely:

- https://github.com/containers/toolbox/pull/1274#discussion_r1153642798
- https://github.com/containers/toolbox/pull/1274#discussion_r1361241590
- https://github.com/containers/toolbox/pull/1274#discussion_r1361266343

Could you address these first? Then we can do another round of review.

debarshiray (commented):
The changes here don't look that much different from previous iterations.

It doesn't look like you studied the cancellable asynchronous read [example](https://rishi.fedorapeople.org/async-read.go) that I showed you some time ago and mentioned [twice](https://github.com/containers/toolbox/pull/1274#pullrequestreview-1470642505) [before](https://github.com/containers/toolbox/pull/1274#discussion_r1223473875) in my reviews.  I was hoping that you would at least critique it, if you thought that it's not the best way to approach this problem.

Unfortunately, the [problem](https://github.com/containers/toolbox/issues/1263) that this pull request is meant to solve is important because it's holding us from making a new release.  :/

Issue #1073, opened 2022-07-04
Inconsistent permissions to access mounted partitions (`nobody`)


nelsonaloysio:
**Describe the bug**
I have an encrypted (with BitLocker) partition labeled `Data` unlocked automatically on boot time via `gio mount -d /dev/nvme0nXpY`. This is accomplished by storing the password in the GNOME keyring, plus granting unlocking and mounting permissions to regular users via `udisks` by writing the following contents to a file located in `/etc/polkit-1/localauthority/60-user.d/enable-mount.pkla`:

```
[Enable mount]
Action=org.freedesktop.udisks2*
Identity=*
ResultActive=yes
ResultAny=yes
ResultInactive=yes
```

Most times, it works wonderfully. Others, though, toolbox returns a `Permission denied` when trying to access the mounted folder `Data` in `/run/media/$USER` -- then, running `toolbox --log-level debug run sh -c 'ls -l /run/media/$USER/Data'` returns:

```
drwx------. 2 nobody nobody 40 Jul  4 18:01 Data
```

Weirdly, though, **this doesn't happen (ever) on the host machine itself** - the folder is always accessible to the user, on every boot.

Umounting and remounting the partition doesn't help; only "workaround" so far seems to be trying my luck rebooting the system.

**Steps how to reproduce the behaviour**
1. Boot system
2. Try to access folder
3. See error

**Expected behaviour**
Either the directory is always accessible (most times it is), or never accessible (sometimes it's not) to the user inside toolbox.

**Actual behaviour**
Most times, the folder is accessible to the user who mounted it; others, it isn't and `nobody` has permissions to access it (not even root).

**Screenshots**
Not applicable, I guess.

**Output of `toolbox --version` (v0.0.90+)**
toolbox version 0.0.99.3

**Toolbox package info (`rpm -q toolbox`)**
toolbox version 0.0.99.3

**Output of `podman version`**
```
Client:       Podman Engine
Version:      4.1.1
API Version:  4.1.1
Go Version:   go1.18.3
Built:        Wed Jun 15 15:31:58 2022
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
podman-4.1.1-1.fc36.x86_64

**Info about your OS**
Fedora Silverblue 36

nelsonaloysio:
[...]

Issue #1001, opened 2022-01-25
toolbox unable to start container fedora-toolbox-35


patriziobrunops:
**Describe the bug**
When trying to enter the container `fedora-toolbox-35` created using
```shell
toolbox create --distro fedora --release 35
```
I get the error
```shell
$ toolbox enter fedora-toolbox-35
Error: invalid entry point PID of container fedora-toolbox-35
```
To have a better error message I tried to attach to the container using `podman` instead of `toolbox` and got the error:
```shell
$ podman start --attach --interactive fedora-toolbox-35
Error: failed to get the current user: user: lookup userid 0: invalid argument
```
This error is returned by `toolbox` whenever Golang's `user.Current()` [gets an error when calling `getpwuid_r`](https://github.com/containers/toolbox/blob/73c53a347b402748b6a12ef9f74cecb7e8528c38/src/cmd/root.go#L316).

**Steps how to reproduce the behaviour**
1. toolbox create --distro fedora --release 35
2. toolbox enter fedora-toolbox-35

**Expected behaviour**
toolbox should start the default user shell inside a container.

**Actual behaviour**
toolbox exits with error code `1` and prints the error `Error: invalid entry point PID of container fedora-toolbox-35`.

**Screenshots**

**Output of `toolbox --version` (v0.0.90+)**
```
toolbox version 0.0.99.3
```
**Toolbox package info (`rpm -q toolbox`)**
```
toolbox-0.0.99.3-2.fc34.x86_64
```
**Output of `podman version`**
```
Version:      3.4.2
API Version:  3.4.2
Go Version:   go1.16.8
Built:        Sun Nov 14 00:16:48 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
```
podman-3.4.2-1.fc34.x86_64
```

**Info about your OS**
Fedora Silverblue 34

**Additional context**
When did the issue start occurring? After an update (what packages were updated)?
If the issue is about operating with containers/images (creating, using, deleting,..), share here what image you used. If you're unsure, share here the output of `toolbox list -i` (shows all toolbox images on your system).
```
IMAGE ID      IMAGE NAME                                    CREATED
ab8bc106d4a7  <none>                                        2 months ago
d8a734db8c5c  registry.fedoraproject.org/fedora-toolbox:34  7 months ago
40b181c70b73  registry.fedoraproject.org/fedora-toolbox:35  4 weeks ago
```

[...]

PR #1277, opened 2023-03-21
initContainer: Use unix.Mount syscall instead of mount external command

angiglesias:
Use golang unix syscalls package to perform mount operations instead of invoking in a shell the mount command.
Performing this operation programmatically should provide more observability debugging the code. It also reduces dependencies on external commands that may change or break functionality between versions.

### DIFF
--- a/.codespellexcludefile
+++ b/.codespellexcludefile
@@ -13,3 +13,4 @@
 		{"/var/lib/flatpak", "/run/host/var/lib/flatpak", "ro"},
 		{"/var/lib/systemd/coredump", "/run/host/var/lib/systemd/coredump", "ro"},
 		{"/var/log/journal", "/run/host/var/log/journal", "ro"},
+		"ro":            {false, unix.MS_RDONLY},
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -32,6 +32,7 @@ import (
 	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"golang.org/x/sys/unix"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


angiglesias:
> Thanks for working on this, @angiglesias ! Was there an immediate problem that inspired you to make this change?
> 
> In principle, it _will_ be good to use the `mount(2)` system call through `golang.org/x/sys/unix`, instead of going through the `mount(8)` command, but it's not that simple a replacement. :)
> 
> For example, I think Podman makes mistakes using the `mount(2)` system call, which leads to #1073 There's this pitfall that `mount(2)` doesn't propagate the mount flags to existing sub-mounts when performing a recursive bind mount. Notice how the bug can't be reproduced when the bind mounts are handled by `mount(8)` instead of Podman.
> 
> A good example of using `mount(2)` to do bind mounts is the [bind_mount()](https://github.com/containers/bubblewrap/blob/main/bind-mount.c) function in `bwrap(1)`. Especially the part where it works around the mount flags not getting applying to existing sub-mounts for recursive bind mounts.
> 
> So, before we change how bind mounts are done in Toolbx itself, we first need to solve #1073 so that we know what we are doing. Then, if we implement a replacement for `mount --rbind` in Toolbx itself, we would need good test coverage for it.

While I was investigating #1001 I thought it would be easier to debug the program reducing shell invocations to utilities such as mount using programmatic alternatives when possible. For the mounts, as the syscall is available through the already imported `golang.org/x/sys/unix` seemed a reasonable change. I didn't know the problems mentioned in those issues, I'll take  a careful look, thanks!

debarshiray (commented):
Thanks for squashing the commits, @angiglesias !  Did you manage to find some time for a more complete replacement of the `mount(8)` command using the `mount(2)` system call and/or a solution for https://github.com/containers/toolbox/issues/1073 and/or increasing our test coverage so that we can be sure that we are not regressing anything when we make changes?

debarshiray:
I am going to close this because this still doesn't have a complete replacement of the `mount(8)` command using `mount(2)` nor does it increase our test coverage so that we can be sure that we are not regressing anything when we make the change.

Please feel free to submit a new merge request when you have some updates.

debarshiray:
Thanks for playing with Toolbox @angiglesias !

angiglesias:
@debarshiray hey sorry for the unresponsiviness! I will try to find a moment to redo this PR properly to address those issues

PR #1279, opened 2023-03-22
cmd/utils: Handle space-separated input when asking for confirmation

debarshiray:
fmt.Scanf [1] is fragile when it comes to space-separated input.  It stores successive space-separated values into successive arguments as determined by the format string.  This breaks with untrusted input that can have an unknown number of space-separated values.

Here are some examples:
```
  $ toolbox create
  Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)?
    [y/N]: no no not at all
  $ no not at all
  bash: no: command not found...

  $ toolbox create
  Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)?
    [y/N]: foo bar
  Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)?
    [y/N]: Download registry.fedoraproject.org/fedora-toolbox:39
    (294.8MB)? [y/N]:
```

Instead this is what should happen:
```
  $ toolbox create
  Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)?
    [y/N]: no no not at all
  Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)?
    [y/N]: foo bar
  Download registry.fedoraproject.org/fedora-toolbox:39 (294.8MB)?
    [y/N]:
```

Fallout from 936f22ff152c424ce89b1454a52f1901de59d328

[1] https://pkg.go.dev/fmt#Scanf

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2020 – 2022 Red Hat Inc.
+ * Copyright © 2020 – 2023 Red Hat Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


HarryMichal:
Good catch! Maybe you could also make a test case for it? :stuck_out_tongue_winking_eye:.

debarshiray:
> Good catch! Maybe you could also make a test case for it? stuck_out_tongue_winking_eye.

Thanks for the review, @HarryMichal   We don't have good coverage for these interactive code paths.  eg., also the `enter` command.  I will try to think of something.

debarshiray:
I merged this because it touches upon the same areas of the code as https://github.com/containers/toolbox/issues/1263  It will make life simpler if we could serialize the work.

I will keep working on the tests.

PR #1281, opened 2023-03-25
Simplify descriptions for a few examples

nilslindemann:


debarshiray (commented):
Looking at the example again, I can see that it can be improved, but I don't know your specific complaint about it.  :)

Is it that the `toolbox enter foo` isn't directly related to the image, other than the possibility that the Toolbx container was not created from one of the built-in images?  If so, how about:

nilslindemann:
> ```
> Enter the default toolbox container matching the host OS
> ...
> Enter the default toolbox container for Fedora 36
> ...
> Enter a toolbox container with a custom name
> ...
> ```


This makes sense. It removes the confusion, whether the "foo" in `toolbox enter foo` is the name for the image or for the container.

Shall I make this change, and also have a look at the docs for the other toolbox commands? (A few use the wording ` using the default image`)

nilslindemann:
There are still ambiguities, like

```
Create a custom toolbox container from a custom image
    $ toolbox create --image bar foo   # what is the image name? What is the container name?
```

... which could be resolved by doing ...

```
Create a custom toolbox container from a custom image
    $ toolbox create --image myimage mycontainer
```

... or by doing ...

```
Create a custom toolbox container "foo" from a custom image "bar"
    $ toolbox create --image bar foo
```

What is your opinion about this?

doc/toolbox-enter.1.md
```diff
@@ -42,19 +42,19 @@ host.
...
 
-### Enter a toolbox container using the default image for Fedora 36
+### Enter the default toolbox container for Fedora 36
 
 ```
 $ toolbox enter --distro fedora --release f36
 ```
 
-### Enter a custom toolbox container using a custom image
+### Enter a toolbox container with a custom name
```
debarshiray, line 57: There's a similar example in `doc/toolbox-run.1.md` that could do with the same treatment.

debarshiray (commented):
Thanks for the updates, @nilslindemann !

doc/toolbox-enter.1.md
debarshiray: I took the liberty to update the custom container example in `doc/toolbox-run.1.md` too.

debarshiray:
Thanks for your contribution!

nilslindemann:
Thank you, too, for considering it.

PR #1291, opened 2023-04-15
Ubuntu image improvements

Jmennius:
* images: ubuntu: Move flatpak-xdg-utils installation to Containerfile
  Since `flatpak-xdg-utils` is essential to images, not 'extra'.
* images: ubuntu: Remove APT ESM hook
  This was recently introduced with `ubuntu-advantage-tools` and it tries to poke at some system services introducing annoying delay and messages.
  Even if the services are present (on Ubuntu host) and systemd is accessible (rootful container) - that wouldn't be appropriate still.


debarshiray:
Thanks, @Jmennius !

PR #1292, opened 2023-04-15
images: ubuntu: Add 23.04 Lunar image

Jmennius:
Which is the next non-LTS to be released on Apr 20th 2023 (and supported for 9 months).


debarshiray:
Thanks, @Jmennius !

PR #1293, opened 2023-05-01
profile.d/toolbox.sh: Add welcome message for Fedora Sericea

kuba3351:
Since Sericea is an official variant of Fedora, it should have an official welcome message like another ones.


debarshiray:
Merged, after testing on Fedora Silverblue with Bash and Z shell.  Thanks, @kuba3351

PR #1298, opened 2023-05-18
tests: Add an obvious error message

matthiasclasen:
We wasted some time trying to get the tests
running locally, when all we were missing was
git submodules init.

Add some more obvious hints about this possible
stumbling block.

### DIFF
--- a/test/system/README.md
+++ b/test/system/README.md
@@ -22,7 +22,8 @@ Running them won't remove any existing containers or images.
 These tests use a few standard libraries for `bats` which help with clarity
 and consistency. The libraries are [bats-support](https://github.com/bats-core/bats-support)
 and [bats-assert](https://github.com/bats-core/bats-assert). These libraries are
-provided as git submodules in the `libs` directory. Make sure both are present.
+provided as git submodules in the `libs` directory. Make sure both are present
+by running `git submodule init` and `git submodule update`.
 
 ## How to run the tests
 
--- a/test/system/setup_suite.bash
+++ b/test/system/setup_suite.bash
@@ -15,11 +15,23 @@
 # limitations under the License.
 #
 
-load 'libs/helpers'
+missing_dependencies=false
+
+if [ -f test/system/libs/bats-assert/load.bash ] && [ -f test/system/libs/bats-support/load.bash ]; then
+  load 'libs/helpers'
+else
+  missing_dependencies=true
+fi
 
 setup_suite() {
   bats_require_minimum_version 1.7.0
   echo "# test suite: Set up" >&3
+
+  if $missing_dependencies; then
+    echo "# Missing dependencies" >&3
+    echo "# Forgot to run 'git submodule init' and 'git submodule update' ?" >&3
+    return 1
+  fi
 
   local os_release="$(find_os_release)"
   local system_id="$(get_system_id)"
@@ -49,6 +61,10 @@ setup_suite() {
 teardown_suite() {
   bats_require_minimum_version 1.7.0
   echo "# test suite: Tear down" >&3
+
+  if $missing_dependencies; then
+    return 0
+  fi
 
   _setup_environment
 


nievesmontero:
@matthiasclasen please sign off the commit for the DCO not to fail

test/system/000-setup.bats
```diff
@@ -15,6 +15,12 @@
 # limitations under the License.
 #
 
+if [ ! -f test/system/libs/helpers.bash ]; then
+  echo "test/system/libs/helpers.hash is missing"
+  echo "Forgot to run git submodule init ?"
```
debarshiray, line 20: Is there no way to wire it up to Meson?

debarshiray (commented):
Thanks for catching this, @matthiasclasen !  One fly-by question above.

test/system/000-setup.bats
```diff
@@ -15,6 +15,12 @@
 # limitations under the License.
 #
 
+if [ ! -f test/system/libs/helpers.bash ]; then
```
debarshiray, line 18: `test/system/libs/helpers.bash` doesn't come from the `bats-assert` and `bats-support` submodules - it's part of Toolbx itself.  So, I think this check needs to look for things like `test/system/libs/bats-assert/load.bash` and such.

debarshiray (changes requested):
I was playing with this branch on a fresh `toolbox.git` tree without the Git submodules and noticed that the system tests don't fail with the expected error.

test/system/000-setup.bats
debarshiray: Done.  Changed to `test/system/libs/bats-assert/load.bash` and `test/system/libs/bats-support/load.bash`

test/system/000-setup.bats
```diff
@@ -15,6 +15,18 @@
 # limitations under the License.
 #
 
+if ! [ -f test/system/libs/bats-assert/load.bash ]; then
```
debarshiray, line 18: It might be better to move this to the [setup_suite](https://bats-core.readthedocs.io/en/stable/writing-tests.html) function.

Currently, it fails as:
```
$ TOOLBOX=/opt/bin/toolbox bats ./test/system
000-setup.bats
 ✗ setup_file failed
   (in test file test/system/000-setup.bats, line 21)
     `exit 1' failed with status 0
   test/system/libs/bats-assert/load.bash is missing
   Forgot to run 'git submodule init' and 'git submodule update' ?
001-version.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
002-help.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
101-create.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
102-list.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
103-container.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
104-run.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
105-enter.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
106-rm.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
107-rmi.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
108-completion.bats
 ✗ setup_file failed
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
999-teardown.bats
 ✗ setup_file failed
   (in test file test/system/999-teardown.bats, line 18)
     `load 'libs/helpers'' failed with status 0
   bats_load_safe: Could not find '/tmp/toolbox/test/system/libs/bats-support/load'[.bash]
   bats warning: Executed 12 instead of expected 193 tests

193 tests, 12 failures, 181 not run
```

With `setup_suite`, it seems to fail as:
```
$ TOOLBOX=/opt/bin/toolbox bats ./test/system
 ✗ setup_suite
   Forgot to run 'git submodule init' and 'git submodule update' ?
   (from function `setup_suite' in test file test/system/setup_suite.bash, line 5)
     `return 1' failed
   bats warning: Executed 1 instead of expected 193 tests

193 tests, 1 failure, 192 not run
```

... which is cleaner.  Although, I have no idea why it doesn't show the `... load.bash is missing` error message.

In fact, the entire `000-setup.bats` should probably go into `setup_suite`.

debarshiray: Our Git [history](https://github.com/containers/toolbox/commit/54a2ca1ead343c3db2c7443cc58b9d9e896d3c46) indicates that our `000-setup.bats` and `999-teardown.bats` were added before Bats started to offer `setup_suite` and `teardown_suite`.  Those [became](https://github.com/bats-core/bats-core/commit/fb467ec3f04e322aee335c92c374158edfff0344) a thing in Bats 1.7.0, which we now require anyway.

test/system/000-setup.bats
debarshiray: Also, it should be `if ! [ -f ...`, not `if [ ! -f ...`, because `[` is actually a shell command that takes `-f ...` as its arguments and we need to negate the whole command.  Yes, shell scripts are weird.  :)

debarshiray:
recheck

test/system/000-setup.bats
debarshiray: The move to `setup_suite` and `teardown_suite` was done in https://github.com/containers/toolbox/pull/1317

debarshiray: >  I have no idea why it doesn't show the `... load.bash is missing` error message.

Moving the error output to `setup_suite` fixes the problem of the missing error message.  We [shouldn't](https://bats-core.readthedocs.io/en/stable/writing-tests.html) be printing outside the `@test`, `setup*` or `teardown*` functions in general.

danpawlik:
recheck

danpawlik:
sorry for "recheck" comment, but I need to verify Zuul CI job.

debarshiray:
> sorry for "recheck" comment, but I need to verify Zuul CI job.

No need to apologize.  Thanks for working on Zuul.

*I might start badgering you with my Zuul questions in future.  :P*

PR #1303, opened 2023-06-07
pkg/utils: Support host operating systems without VERSION_ID

debarshiray:
The VERSION_ID field in os-release(5) is optional [1].  It's absent on Arch Linux, which follows a rolling-release model and uses the BUILD_ID field instead:
  BUILD_ID=rolling

A subsequent commit will add built-in support for Arch Linux.  Hence, the code to get the default release from the host operating system can no longer assume the presence of the VERSION_ID field in os-release(5).

[1] https://www.freedesktop.org/software/systemd/man/os-release.html

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -39,12 +39,14 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+type GetDefaultReleaseFunc func() (string, error)
 type GetFullyQualifiedImageFunc func(string, string) string
 type ParseReleaseFunc func(string) (string, error)
 
 type Distro struct {
 	ContainerNamePrefix    string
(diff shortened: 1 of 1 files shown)


src/pkg/utils/utils.go
```diff
@@ -273,6 +277,25 @@ func getDefaultImageForDistro(distro, release string) string {
...
+		panicMsg := fmt.Sprintf("failed to find %s in the list of supported distributions", distro)
+		panic(panicMsg)
+	}
+
+	osRelease, err := osrelease.Read()
+	if err != nil {
+		return "", err
+	}
+
+	return osRelease[distroOj.ReleaseDefaultField], nil
```
debarshiray, line 296: Ugh, typo.  :(

PR #1308, opened 2023-06-08
.github/workflows: Publish the Arch Linux image at quay.io/toolbx/...

debarshiray:
Until now, the Arch Linux image was being published at quay.io/toolbx-images/archlinux-toolbox:latest.  This renames the image to arch-toolbox [1] to match the os-release(5) ID on Arch, and changes the location to quay.io/toolbx/arch-toolbox:latest.

Build and push when there are changes in the 'images/arch' directory or in the GitHub workflow itself, as well as every other week (7th and 21st days of a month to be precise).

[1] Commit 2568528cb7f52663
    https://github.com/containers/toolbox/pull/861

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,6 @@
 *    @HarryMichal @debarshiray
+/.github/workflows/arch-images.yaml       @Foxboron
+/.github/workflows/arch-images-pr.yaml    @Foxboron
 /data/gfx/*.gif    @jimmac
 /images/arch       @Foxboron
 /images/rhel       @debarshiray @olivergs
--- a/.github/workflows/arch-images-pr.yaml
+++ b/.github/workflows/arch-images-pr.yaml
@@ -1,0 +1,31 @@
+name: Build the arch-toolbox image for PRs
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - images/arch/**
+      - .github/workflows/arch-images-pr.yaml
+
+jobs:
+  build-and-push-images:
+    name: Build the arch-toolbox image for PRs
+
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build the arch-toolbox image
+        uses: docker/build-push-action@v3
+        with:
+          context: images/arch
+          file: images/arch/Containerfile
+          platforms: linux/amd64
+          push: false
+          no-cache: true
+          tags: quay.io/toolbx/arch-toolbox:latest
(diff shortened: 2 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -1,4 +1,5 @@
 *    @HarryMichal @debarshiray
+/.github/workflows/arch-images.yaml    @Foxboron
```
travier, line 2: Should you add that as well for the ubuntu folks?

travier:
Note that this is missing the PR workflow, like the ubuntu one, which lets us validate change to the images on PRs while making sure that the images are not pushed.

travier:
https://github.com/toolbx-images/images/blob/main/.github/workflows/archlinux-pr.yaml

debarshiray:
> Note that this is missing the PR workflow, like the ubuntu one, which lets us validate change to the images on PRs while making sure that the images are not pushed.

Yes, I was trying to limit my copy-paste cargo cult to the minimum.  :)

For example, I was a bit suspicious that the actual `cron` schedule is *every Monday at 00:00*, not *every other week* as in the original commit message from https://github.com/containers/toolbox/pull/483  I had to re-read the `crontab` [manual](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/crontab.html#tag_20_25_07) to be sure.  Did I get that right?

Did you see my question about `permissions: read-all` above?

Anyway, let me try to add a PR workflow.

travier:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

I had added it as a precaution. Not sure how much that matters here.

Foxboron (approved):

debarshiray:
> About `read-all`, I'm not sure how that works exactly

Okay.  I will leave it alone since it seems to be working as is.
 
> For the cron syntax, you can use https://crontab.guru/ (linked in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) to validate your config

Thanks, https://crontab.guru/ is amazing.  I think my understanding of the schedule above was [correct](https://crontab.guru/#0_0_*_*_MON).

debarshiray:
> > Note that this is missing the PR workflow, like the ubuntu one, which lets us validate change to the images on PRs while making sure that the images are not pushed.
>
> Anyway, let me try to add a PR workflow.

Done.

debarshiray:
At least the workflow for PRs seems to be failing with:
```
Invalid workflow file: .github/workflows/arch-images-pr.yaml#L17
The workflow is not valid. .github/workflows/arch-images-pr.yaml (Line: 17, Col: 11): Unrecognized named-value: 'env'. Located at position 1 within expression: env.distro_pretty
```

... and I am getting lost in the [documentation](https://docs.github.com/en/actions/learn-github-actions/variables) about whether this is meant to be a valid syntax or not:
```
name: Build the ${{ env.distro }}-toolbox image for PRs
```

At this point, I want to ditch the variables and keep it simple, and leave it to somebody more knowledgeable to parametrize things.

debarshiray:
Let's hope the `arch-toolbox` image shows up at `quay.io/toolbx/arch-toolbox:latest`.  Fingers crossed.  :)

PR #1321, opened 2023-06-23
fix syntax

sumantro93:
Removed dashes

### DIFF
--- a/doc/toolbox.1.md
+++ b/doc/toolbox.1.md
@@ -87,40 +87,18 @@ by option **log-level**.
 Same as `--log-level=debug`. Use `-vv` to include `--log-podman`.
 
 ## COMMANDS
+The following commands correspond to markdown files which will help you with execution.
 
-Commands for working with toolbox containers and images:
-
-**toolbox create(1)**
-
-Create a new toolbox container.
-
-**toolbox enter(1)**
-
-Enter a toolbox container for interactive use.
-
-**toolbox help(1)**
-
-Display help information about Toolbox.
-
-**toolbox init-container(1)**
-
-Initialize a running container.
-
-**toolbox list(1)**
-
-List existing toolbox containers and images.
-
-**toolbox rm(1)**
-
-Remove one or more toolbox containers.
-
-**toolbox rmi(1)**
-
-Remove one or more toolbox images.
-
-**toolbox run(1)**
-
-Run a command in an existing toolbox container.
+| Command                                          | Description                                                                 |
+| ------------------------------------------------ | --------------------------------------------------------------------------- |
+| [toolbox-create(1)](toolbox-create.1.md)         | Create a new toolbox container                                              |
+| [toolbox-enter(1)](toolbox-enter.1.md)           | Enter a toolbox container for interactive use                               |
+| [toolbox-help(1)](toolbox-help.1.md)             | Display help information about Toolbox                                      |
+| [toolbox-init-container(1)](toolbox-init-container.1.md) |Initialize a running container                                       |
+| [toolbox-list(1)](toolbox-list.1.md)             | List existing toolbox containers and images                                 |
+| [toolbox-rm(1)](toolbox-rm.1.md)                 | Remove one or more toolbox containers                                       |
+| [toolbox-rmi(1)](toolbox-rmi.1.md)               | Remove one or more toolbox images                                           |
+| [toolbox-run(1)](toolbox-run.1.md)               | Run a command in an existing toolbox container                              |
 
 ## FILES ##
 
--- a/doc/toolbox.1.md
+++ b/doc/toolbox.1.md
@@ -48,6 +48,9 @@ falls back to a Fedora image. Supported host operating systems are:
 
 * Fedora
 * Red Hat Enterprise Linux >= 8.5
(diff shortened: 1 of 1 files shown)


doc/toolbox.1.md
```diff
@@ -90,35 +90,35 @@ Same as `--log-level=debug`. Use `-vv` to include `--log-podman`.
 
 Commands for working with toolbox containers and images:
 
-**toolbox-create(1)**
+**toolbox create(1)**
```
debarshiray, line 93: The thing is that `toolbox-create(1)` is pointing to the *manual* for the `toolbox create` command, and not the command itself.  Those parentheses indicate the section within the entire collection of manuals where the `toolbox-create` manual page exists.

It's a convention that's mentioned in [man(1)](https://man7.org/linux/man-pages/man1/man.1.html), and is used in `podman(1)` and [git(1)](https://git-scm.com/docs/git) to list the manuals for each command.

If you want to provide some very brief example of how to use Toolbx, then we can replicate the *Usage* section from the [website](https://containertoolbx.org/install/).

debarshiray (changes requested):
Thanks for working on this, @sumantro93 !  I had this vague memory that you had found some other problem in the past, which isn't this.

debarshiray:
```
fedora-rawhide | not ok 106 container: Check container starts without issues in 43565ms
2023-06-23 12:24:37.882397 | fedora-rawhide | # (from function `assert_success' in file test/system/libs/bats-assert/src/assert.bash, line 114,
fedora-rawhide | #  in test file test/system/103-container.bats, line 38)
fedora-rawhide | #
fedora-rawhide | # status : 1
fedora-rawhide | # output :
fedora-rawhide | # --
fedora-rawhide | #
```

debarshiray:
recheck

debarshiray:
recheck

doc/toolbox.1.md
```diff
@@ -87,40 +87,18 @@ by option **log-level**.
...
-Remove one or more toolbox images.
-
-**toolbox-run(1)**
-
-Run a command in an existing toolbox container.
+The following commands correspond to markdown files which will help you with execution.
+
+| Command                                          | Description                                                                 |
+| ------------------------------------------------ | --------------------------------------------------------------------------- |
+| [toolbox-create(1)](toolbox-create.1.md)         | Create a new toolbox container                                              |
```
debarshiray, line 94: These Markdown files are primarily used to generate troff formatted manuals with `go-md2man(1)`, and these changes generate broken files.

It will be good if we can link to these Markdown files from the Toolbx [website](https://containertoolbx.org/), but that will require figuring out how to add links without breaking the generated manuals.

doc/toolbox.1.md
```diff
@@ -87,40 +87,18 @@ by option **log-level**.
...
-Remove one or more toolbox containers.
-
-**toolbox-rmi(1)**
-
-Remove one or more toolbox images.
-
-**toolbox-run(1)**
-
-Run a command in an existing toolbox container.
+The following commands correspond to markdown files which will help you with execution.
```
debarshiray, line 90: This looks a bit odd.  :)

The user doesn't need to know if the manuals are written in Markdown or something else.  When the manuals are consumed through `man(1)` they will be in the troff format, and when they are consumed through the web browser they will be in HTML.

Notice how the [flatpak(1)](https://man7.org/linux/man-pages/man1/flatpak.1.html), [git(1)](https://git-scm.com/docs/git) or `podman(1)` manuals don't say anything like this.

debarshiray (changes requested):
Thanks for updating the branch!  However, I am not sure where you are going with this pull request.  :)

Like I said [before ](https://github.com/containers/toolbox/pull/1321#discussion_r1242000664), the existing text follows established conventions.  If you want to provide some very brief example of how to use Toolbx, then we can replicate the Usage section from the [website](https://containertoolbx.org/install/).  Otherwise, there isn't much to gain from moving the text around.

debarshiray:
Could you please correct your Git authorship information with your full name and a real email address, and squash the commits to get rid of the old discarded changes?

Also, try rebasing against `main` to get the CI fixed.

doc/toolbox.1.md
```diff
@@ -48,6 +48,9 @@ falls back to a Fedora image. Supported host operating systems are:
 
 * Fedora
 * Red Hat Enterprise Linux >= 8.5
+* CentOS Stream
+* Arch Linux
```
debarshiray, line 52: We don't support CentOS Stream, because we don't have built-in images wired with `--distro` for it, but we do support Ubuntu.  There's a table below that needs updating too.

Also, please keep the lists alphabetically ordered.  :)

debarshiray (changes requested):

doc/toolbox.1.md
debarshiray: Here's a screenshot showing the breakage:

image

debarshiray:
This hasn't seen any activity for a couple of months now, and we need to start preparing for a new release.  https://github.com/containers/toolbox/pull/1379 adds a usage section, and  documents Arch Linux and Ubuntu as the newly supported operating system distributions.

Hence, I am closing this one.

PR #1332, opened 2023-07-04
test/system: Simplify the check for Fedora Rawhide

debarshiray:
First, it's not a good idea to use awk(1) as a grep(1) replacement. Unless one really needs the AWK programming language, it's better to stick to grep(1) because it's simpler.

Secondly, it's better to look for a specific os-release(5) field instead of looking for the occurrence of 'rawhide' anywhere in the file, because it lowers the possibility of false positives.


debarshiray:
This was merged.  I had typoed the Git remote and it confused GitHub.

PR #1334, opened 2023-07-07
test/system: Silence SC2154

debarshiray:
Otherwise https://www.shellcheck.net/ would complain:
```
  Line 202:
  run echo "$name"
            ^---^ SC2154 (warning): name is referenced but not assigned.
```

See: https://www.shellcheck.net/wiki/SC2154

PR #1336, opened 2023-07-09
test/system: Simplify the check for Fedora Rawhide

debarshiray:
First, it's not a good idea to use awk(1) as a grep(1) replacement. Unless one really needs the AWK programming language, it's better to stick to grep(1) because it's simpler.

Secondly, it's better to look for a specific os-release(5) field instead of looking for the occurrence of 'rawhide' anywhere in the file, because it lowers the possibility of false positives.

### DIFF
--- a/playbooks/dependencies-fedora-restricted.yaml
+++ b/playbooks/dependencies-fedora-restricted.yaml
@@ -14,6 +14,11 @@
 # limitations under the License.
 #
 
+- name: Install python3-dnf
+  become: true
+  command: dnf5 --assumeyes install python3-dnf
+  when: ansible_distribution_major_version | int >= 39
+
 - name: Ensure that subordinate group ID ranges are absent
   become: yes
   file:
--- a/test/system/setup_suite.bash
+++ b/test/system/setup_suite.bash
@@ -33,7 +33,6 @@ setup_suite() {
     return 1
   fi
 
-  local os_release="$(find_os_release)"
   local system_id="$(get_system_id)"
   local system_version="$(get_system_version)"
 
@@ -49,8 +48,7 @@ setup_suite() {
   _pull_and_cache_distro_image ubuntu 20.04 || false
   _pull_and_cache_distro_image busybox || false
   # If run on Fedora Rawhide, cache 2 extra images (previous Fedora versions)
-  local rawhide_res="$(awk '/rawhide/' $os_release)"
-  if [ "$system_id" = "fedora" ] && [ -n "$rawhide_res" ]; then
+  if is_fedora_rawhide; then
     _pull_and_cache_distro_image fedora "$((system_version-1))" || false
     _pull_and_cache_distro_image fedora "$((system_version-2))" || false
   fi
--- a/playbooks/dependencies-fedora.yaml
+++ b/playbooks/dependencies-fedora.yaml
@@ -14,6 +14,11 @@
 # limitations under the License.
 #
 
+- name: Install python3-dnf
+  become: true
+  command: dnf5 --assumeyes install python3-dnf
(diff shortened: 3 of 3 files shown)


debarshiray:
The system tests failed on Ubuntu 22.04:
```
not ok 1 setup_suite
# (from function `_clean_temporary_storage' in file test/system/libs/helpers.bash, line 63,
#  from function `teardown_suite' in test file test/system/setup_suite.bash, line 71)
#   `_clean_temporary_storage' failed with status 0
# /home/runner/work/toolbox/toolbox/containers/toolbox/test/system/setup_suite.bash: line 52: 22.04: syntax error: invalid arithmetic operator (error token is ".04")
# Error: no container with name or ID "docker-registry" found: no such container
# A storage.conf file exists at /home/runner/.cache/toolbox/system-test-storage/storage.conf
# You should remove this file if you did not modified the configuration.
# bats warning: Executed 23 instead of expected 22 tests
Error: Process completed with exit code 1.
```

test/system/setup_suite.bash
```diff
@@ -49,8 +48,7 @@ setup_suite() {
   _pull_and_cache_distro_image ubuntu 20.04 || false
   _pull_and_cache_distro_image busybox || false
   # If run on Fedora Rawhide, cache 2 extra images (previous Fedora versions)
-  local rawhide_res="$(awk '/rawhide/' $os_release)"
-  if [ "$system_id" = "fedora" ] && [ -n "$rawhide_res" ]; then
+  if ! is_fedora_rawhide; then
```
debarshiray, line 51: Bingo!  The condition got flipped.  :(

debarshiray:
@danpawlik @TristanCacqueray All the tests running on Fedora Rawhide hosts are failing in the same way as https://github.com/containers/toolbox/pull/1334 :
```
TASK [Install RPM packages]
fedora-rawhide | ERROR
fedora-rawhide | {
fedora-rawhide |   "msg": "Could not import the dnf python module using /usr/bin/python3 (3.12.0b3 (main, Jun 21 2023, 00:00:00) [GCC 13.1.1 20230614 (Red Hat 13.1.1-4)]). Please install `python3-dnf` or `python2-dnf` package or ensure you have specified the correct ansible_python_interpreter. (attempted ['/usr/libexec/platform-python', '/usr/bin/python3', '/usr/bin/python2', '/usr/bin/python'])",
fedora-rawhide |   "results": []
fedora-rawhide | }
```

This pull requests actually makes changes to Rawhide specific code paths, so I am a bit hesitant to override the CI.  :/

debarshiray:
recheck

danpawlik:
I added a workaround https://github.com/containers/toolbox/pull/1338
It should be  removed after we update the Zuul to new version (it should happen probably next year).

playbooks/dependencies-fedora-restricted.yaml
```diff
@@ -14,6 +14,11 @@
 # limitations under the License.
 #
 
+- name: Install python3-dnf
```
TristanCacqueray, line 17: Perhaps add a note why this was added for rawhide, so that the task can be removed/adjusted at the next fedora release.

TristanCacqueray (approved):

playbooks/dependencies-fedora-restricted.yaml
debarshiray: Good point.  I already added a detailed note to the commit message of the [commit](https://github.com/containers/toolbox/commit/96791726a35d9a19c53064bbb17204f4aba92e46) from @danpawlik 

I got a bit impatient while his pull request https://github.com/containers/toolbox/pull/1338 was going through CI, and rebased this PR on top of his.  That's why it looks a bit confusing.

PR #1338, opened 2023-07-11
playbooks: Add workaround for Fedora Rawhide

danpawlik:
The Ansible version that is installed in Zuul is 2.13.7 that is expecting python3-dnf library to be installed on the host.
I have done a test on the rawhide host and it seems that the ansible-core package that is available in Fedora repo (2.15) is fine and it does not require python3-dnf package. It means, that if the pull request will be merged, we should revert it later, after Zuul update.

### DIFF
--- a/playbooks/dependencies-fedora-restricted.yaml
+++ b/playbooks/dependencies-fedora-restricted.yaml
@@ -14,6 +14,11 @@
 # limitations under the License.
 #
 
+- name: Install python3-dnf
+  become: true
+  command: dnf5 --assumeyes install python3-dnf
+  when: ansible_distribution_major_version | int >= 39
+
 - name: Ensure that subordinate group ID ranges are absent
   become: yes
   file:
--- a/playbooks/dependencies-fedora.yaml
+++ b/playbooks/dependencies-fedora.yaml
@@ -14,6 +14,11 @@
 # limitations under the License.
 #
 
+- name: Install python3-dnf
+  become: true
+  command: dnf5 --assumeyes install python3-dnf
+  when: ansible_distribution_major_version | int >= 39
+
 - name: Install RPM packages
   become: yes
   package:


debarshiray:
> The Ansible version that is installed in Zuul is 2.13.7 that is expecting python3-dnf library to be installed on the host. I have done a test on the rawhide host and it seems that the ansible-core package that is available in Fedora repo (2.15) is fine and it does not require python3-dnf package. It means, that if the pull request will be merged, we should revert it later, after Zuul update.

I see.  Thanks for investigating this, @danpawlik !

I wonder if this is somehow related to the DNF5 [Change](https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5) for Fedora 39.

On Fedora 38, `dnf` pulls in `python3-dnf`:
```
$ rpm -q dnf
dnf-4.15.0-1.fc38.noarch
$ rpm --requires -q dnf
/bin/sh
/bin/sh
/bin/sh
python3-dnf = 4.15.0-1.fc38
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
rpmlib(RichDependencies) <= 4.12.0-1
```

... but on Fedora 39:
```
# rpm -qa | grep dnf
libdnf5-5.0.15-3.fc39.x86_64
libdnf5-cli-5.0.15-3.fc39.x86_64
dnf5-5.0.15-3.fc39.x86_64
dnf5-plugins-5.0.15-3.fc39.x86_64
```

Maybe older versions of Ansible needs to learn how to cope with operating systems using DNF5?  I don't know if that makes any sense.  :)

I also found these [two](https://src.fedoraproject.org/rpms/ansible-core/c/b7db1fa79b42af59bd9552a9dfce9ec132838216) [commits](https://src.fedoraproject.org/rpms/ansible-core/c/71c43a9b1c1c32febaff7dfc6cb641d775245b08) related to DNF5 in the `ansible-core` package in Fedora Rawhide/39.  I suppose they are not the root cause of this, because the Ansible in Fedora Rawhide/39 is not being used here?

debarshiray:
For anybody who might be following along, this is about fixing the recent CI failures on Fedora Rawhide/39:
```
TASK [Install RPM packages]
fedora-rawhide | ERROR
fedora-rawhide | {
fedora-rawhide |   "msg": "Could not import the dnf python module using /usr/bin/python3 (3.12.0b3 (main, Jun 21 2023, 00:00:00) [GCC 13.1.1 20230614 (Red Hat 13.1.1-4)]). Please install `python3-dnf` or `python2-dnf` package or ensure you have specified the correct ansible_python_interpreter. (attempted ['/usr/libexec/platform-python', '/usr/bin/python3', '/usr/bin/python2', '/usr/bin/python'])",
fedora-rawhide |   "results": []
fedora-rawhide | }
```

See https://github.com/containers/toolbox/pull/1334 ,  https://github.com/containers/toolbox/pull/1336 , etc..

playbooks/dependencies-centos-9-stream.yaml
```diff
@@ -1,3 +1,9 @@
+---
+- name: Install python3-dnf
+  become: true
+  command: dnf install -y python3-dnf
+  when: ansible_distribution_major_version | int > 38
```
debarshiray, line 5: This playbook is only used for CentOS Stream 9, so I think we don't need the workaround here.

Issue #911, opened 2021-11-02
'init-container' fails when /var/lib/flatpak, /var/lib/systemd/coredump or /var/log/journal on the host are mounted with nodev, noexec or nosuid


AbsolutelyLudicrous:
**Describe the bug**
`toolbox enter` fails when `/var/log` is a mounted btrfs volume. Of note is that unmounting, then remounting, `/var/log` does not prevent an already started container from being entered. Having /var/log mounted as a tmpfs seems to work just fine, however.

**Steps how to reproduce the behaviour**
Mount a btrfs volume on `/var/log` and try to start a container.

**Expected behaviour**
Toolboxes run just fine

**Actual behaviour**
`toolbox enter` fails with a cryptic error message about an invalid entry point PID.

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.99.2`

**Toolbox package info (`rpm -q toolbox`)**
`toolbox-0.0.99.2^3.git075b9a8d2779-4.fc35.x86_64`

**Output of `podman version`**
```
Version:      3.4.1
API Version:  3.4.1
Go Version:   go1.16.8
Built:        Wed Oct 20 10:31:56 2021
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-3.4.1-1.fc35.x86_64`

**Info about your OS**
Fedora Silverblue 35; recently upgraded from F34, if that matters

**Additional context**
I am 80% sure this worked at one point, and that the Silverblue wiki did not mention `/var/log` as a forbidden mount.

> If you see an error message saying: `Error: invalid entry point PID of container <name-of-container>`, add to the ticket output of command `podman start --attach <name-of-container>`.

Attached below. The relevant part is:
```
level=debug msg="Creating directory /var/log/journal"
level=debug msg="Binding /var/log/journal to /run/host/var/log/journal"
mount: /var/log/journal: filesystem was mounted, but any subsequent operation failed: Unknown error 5005.
Error: failed to bind /var/log/journal to /run/host/var/log/journal
```

[tb-enter-dev.txt](https://github.com/containers/toolbox/files/7464503/tb-enter-dev.txt)

[podman-start.txt](https://github.com/containers/toolbox/files/7464507/podman-start.txt)

The relevant part of my fstab is:

`/dev/sda4               /var/log                btrfs   subvol=@varlog,compress=zstd:1,nosuid,nodev 0 0`

debarshiray:
[...]

PR #1331, opened 2023-07-04
build: Enforce shellcheck(1) on all Bats tests

debarshiray:

PR #1340, opened 2023-07-13
Be aware of security hardened mount points

debarshiray:
Sometimes locations such as `/var/lib/flatpak`, `/var/lib/systemd/coredump` and `/var/log/journal` sit on security hardened mount points that are marked as `nosuid,nodev,noexec` [1].  In such cases, when Toolbx is used rootless, an attempt to bind mount these locations read-only at runtime with `mount(8)` fails because of permission problems:
```
  # mount --rbind -o ro <source> <containerPath>
  mount: <containerPath>: filesystem was mounted, but any subsequent
      operation failed: Unknown error 5005.
```

(Note that the above error message from mount(8) was subsequently improved to show something more meaningful than 'Unknown error' [2].)

The problem is that `init-container` is running inside the container's mount and user namespace and the source paths were mounted inside the host's namespace with `nosuid,nodev,noexec`.  The above `mount(8)` call tries to remove the `nosuid,nodev,noexec` flags from the mount point and replace them with only `ro`, which is something that can't be done from a child namespace.

Note that this doesn't fail when Toolbx is running as root.  This is because the container uses the host's user namespace and is able to remove the `nosuid,nodev,noexec` flags from the mount point and replace them with only `ro`.  Even though it doesn't fail, the flags shouldn't get replaced like that inside the container, because it removes the security hardening of those mount points.

There's actually no benefit in bind mounting these paths as read-only. It was historically done this way 'just to be safe' because a user isn't expected to write to these locations from inside a container.  However, Toolbx doesn't intend to provide any heightened security beyond what's already available on the host.

Hence, it's better to get out of the way and leave it to the permissions on the source location from the host operating system to guard the castle.  This is accomplished by not passing any file system options to `mount(8)` [1].

Note that this isn't a problem when Toolbx is running as root, because the container uses the host's user namespace.

Based on an idea from Si.

[1] https://man7.org/linux/man-pages/man8/mount.8.html

[2] util-linux commit 9420ca34dc8b6f0f
    https://github.com/util-linux/util-linux/commit/9420ca34dc8b6f0f
    https://github.com/util-linux/util-linux/pull/2376

https://github.com/containers/toolbox/issues/911

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -62,10 +62,10 @@ var (
 		{"/run/udev/data", "/run/host/run/udev/data", ""},
 		{"/run/udev/tags", "/run/host/run/udev/tags", ""},
 		{"/tmp", "/run/host/tmp", "rslave"},
-		{"/var/lib/flatpak", "/run/host/var/lib/flatpak", "ro"},
+		{"/var/lib/flatpak", "/run/host/var/lib/flatpak", ""},
 		{"/var/lib/libvirt", "/run/host/var/lib/libvirt", ""},
-		{"/var/lib/systemd/coredump", "/run/host/var/lib/systemd/coredump", "ro"},
-		{"/var/log/journal", "/run/host/var/log/journal", "ro"},
+		{"/var/lib/systemd/coredump", "/run/host/var/lib/systemd/coredump", ""},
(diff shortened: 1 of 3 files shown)


debarshiray:

PR #1344, opened 2023-08-03
Revert "playbooks: Add workaround for Fedora Rawhide"

debarshiray:
The DNF5 Change [1] was dropped from Fedora 39 (and Rawhide) [2] and postponed for a later Fedora.  Therefore, there's no need for this workaround.

This reverts commit 96791726a35d9a19c53064bbb17204f4aba92e46.

[1] https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5

[2] https://pagure.io/fesco/issue/3039

PR #1347, opened 2023-08-11
test/system: Silence some problems pointed out by ShellCheck 

debarshiray:


debarshiray:
recheck

debarshiray:

debarshiray:
I have no idea why `unit-test-migration-path-for-coreos-toolbox` says `TIMED_OUT`.  It did succeed in https://github.com/containers/toolbox/pull/1340 which is rebased on top of this pull request.  Let's consider it a flake.

Issue #585, opened 2020-10-18
Can't enter toolbox with domain user


fallendusk:
**Describe the bug**
Toolbox fails to initialize while run as domain/enterprise user. The /run/user/809201000/toolbox/container-initialized file is never created. I manually added this user to /etc/subuid and /etc/subgid. Toolbox works on this VM with non-enterprise login. I can manually run podman to launch the fedora-toolbox container and it also works sans all the toolbox magic.

**Steps how to reproduce the behaviour**
Try to run toolbox enter with an enterprise user

**Expected behaviour**
Toolbox should work as it does with a normal linux user. 

**Actual behaviour**
```
 greg  ~  toolbox enter -vv
DEBU Running as real user ID 809201000            
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU Checking if /etc/subgid and /etc/subuid have entries for user greg 
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Toolbox config directory is /home/greg@ad.fallendusk.com/.config/toolbox 
INFO[0000] podman filtering at log level debug          
DEBU[0000] Called version.PersistentPreRunE(podman --log-level debug version --format json) 
DEBU[0000] Reading configuration file "/usr/share/containers/containers.conf" 
[...]

PR #1349, opened 2023-08-15
cmd/initContainer: Simplify removing the user's password

debarshiray:
It's one less invocation of an external command, which is good because
spawning a new process is generally expensive.

One positive side-effect of this is that on some Active Directory
set-ups, the entry point no longer fails with:
```
  Error: failed to remove password for user login@company.com: failed
      to invoke passwd(1)
```
... because of:
```
  # passwd --delete login@company.com
  passwd: Libuser error at line: 210 - name contains invalid char `@'.
```

This is purely an accident, and isn't meant to be an intential change to
support Active Directory.  Tools like `useradd(8)` and `usermod(8)` from
Shadow aren't meant to work with Active Directory users, and, hence, it
can still break in other ways.  For that, one option is to expose `$USER`
from the host operating system to the Toolbx container through a Varlink
interface that can be used by nss-systemd inside the container.

Based on an idea from Si.

https://github.com/containers/toolbox/issues/585

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -393,6 +393,7 @@ func configureUsers(targetUserUid int, targetUser, targetUserHome, targetUserShe
 			"--groups", sudoGroup,
 			"--home-dir", targetUserHome,
 			"--no-create-home",
+			"--password", "",
 			"--shell", targetUserShell,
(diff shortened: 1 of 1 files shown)


debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

PR #1354, opened 2023-08-19
cmd/initContainer: Do not try to remount /etc/machine-id as ro

alatiera:
Followup to 1cc9e07b7c36fe9f9784b40b58f0a2a3694dd328

Sometimes the parent location might be mounted with nosuid,nodev,noexec and trying to remount it as ro would remove those and thus fails.

See commit mentioned above for more details.

https://github.com/containers/toolbox/issues/911

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -52,7 +52,7 @@ var (
 		source        string
 		flags         string
 	}{
-		{"/etc/machine-id", "/run/host/etc/machine-id", "ro"},
+		{"/etc/machine-id", "/run/host/etc/machine-id", ""},
 		{"/run/libvirt", "/run/host/run/libvirt", ""},
 		{"/run/systemd/journal", "/run/host/run/systemd/journal", ""},
 		{"/run/systemd/resolve", "/run/host/run/systemd/resolve", ""},
--- a/test/system/211-dbus.bats
+++ b/test/system/211-dbus.bats
@@ -1,0 +1,89 @@
+# shellcheck shell=bats
+#
+# Copyright © 2023 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+load 'libs/bats-support/load'
+load 'libs/bats-assert/load'
+load 'libs/helpers'
+
+setup() {
+  bats_require_minimum_version 1.7.0
+  _setup_environment
+  cleanup_containers
+}
+
+teardown() {
+  cleanup_containers
+}
+
+@test "dbus: session bus" {
+  local expected_response
+  expected_response="$(gdbus call \
+                         --session \
+                         --dest org.freedesktop.DBus \
+                         --object-path /org/freedesktop/DBus \
+                         --method org.freedesktop.DBus.Peer.Ping)"
+
+  create_default_container
+
+  run --keep-empty-lines --separate-stderr "$TOOLBOX" run sh -c 'gdbus call \
+                                                                   --session \
+                                                                   --dest org.freedesktop.DBus \
+                                                                   --object-path /org/freedesktop/DBus \
+                                                                   --method org.freedesktop.DBus.Peer.Ping'
+
+  assert_success
+  assert_line --index 0 "$expected_response"
+
+  if check_bats_version 1.10.0; then
+    assert [ ${#lines[@]} -eq 1 ]
+  else
+    assert [ ${#lines[@]} -eq 2 ]
+  fi
+
+  # shellcheck disable=SC2154
+  assert [ ${#stderr_lines[@]} -eq 0 ]
+}
+
+@test "dbus: system bus" {
+  local expected_response
+  expected_response="$(gdbus call \
+                         --system \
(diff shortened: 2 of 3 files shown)


alatiera:
Here's the error I was getting ftr:

```
level=debug msg="Resolved /etc/localtime to /run/host/usr/share/zoneinfo/Europe/Athens"
level=debug msg="Creating regular file /etc/machine-id"
level=debug msg="Binding /etc/machine-id to /run/host/etc/machine-id"
mount: /etc/machine-id: filesystem was mounted, but any subsequent operation failed: Unknown error 5005.
Error: failed to bind /etc/machine-id to /run/host/etc/machine-id
```

and here's what my mounts look like:

```
TARGET                        SOURCE           FSTYPE          OPTIONS
/                             /dev/dm-1        btrfs           rw,nodev,relatime,space_cache=v2,subvolid=5,subvol=/
├─/tmp                        tmpfs            tmpfs           rw,nosuid,nodev,nr_inodes=1048576
├─/dev                        devtmpfs         devtmpfs        rw,nosuid,size=4096k,nr_inodes=8215431,mode=755
│ ├─/dev/hugepages            hugetlbfs        hugetlbfs       rw,nosuid,nodev,relatime,pagesize=2M
│ ├─/dev/mqueue               mqueue           mqueue          rw,nosuid,nodev,noexec,relatime
│ ├─/dev/shm                  tmpfs            tmpfs           rw,nosuid,nodev
│ └─/dev/pts                  devpts           devpts          rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000
├─/sys                        sysfs            sysfs           rw,nosuid,nodev,noexec,relatime
│ ├─/sys/fs/selinux           selinuxfs        selinuxfs       rw,nosuid,noexec,relatime
│ ├─/sys/kernel/tracing       tracefs          tracefs         rw,nosuid,nodev,noexec,relatime
│ ├─/sys/kernel/debug         debugfs          debugfs         rw,nosuid,nodev,noexec,relatime
│ ├─/sys/kernel/security      securityfs       securityfs      rw,nosuid,nodev,noexec,relatime
│ ├─/sys/fs/cgroup            cgroup2          cgroup2         rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot
│ ├─/sys/firmware/efi/efivars efivarfs         efivarfs        rw,nosuid,nodev,noexec,relatime
│ ├─/sys/fs/bpf               bpf              bpf             rw,nosuid,nodev,noexec,relatime,mode=700
│ └─/sys/fs/fuse/connections  fusectl          fusectl         rw,nosuid,nodev,noexec,relatime
├─/proc                       proc             proc            rw,nosuid,nodev,noexec,relatime
│ └─/proc/sys/fs/binfmt_misc  systemd-1        autofs          rw,relatime,fd=30,pgrp=1,timeout=0,minproto=5,maxproto=5,direct,pipe_ino=1150
├─/run                        tmpfs            tmpfs           rw,nosuid,nodev,size=13154028k,nr_inodes=819200,mode=755
│ ├─/run/user/1000            tmpfs            tmpfs           rw,nosuid,nodev,relatime,size=6577012k,nr_inodes=1644253,mode=700,uid=1000,gid=1000
│ │ ├─/run/user/1000/gvfs     gvfsd-fuse       fuse.gvfsd-fuse rw,nosuid,nodev,relatime,user_id=1000,group_id=1000
│ │ └─/run/user/1000/doc      portal           fuse.portal     rw,nosuid,nodev,relatime,user_id=1000,group_id=1000
│ └─/run/media/alatiera/disk  /dev/sda5        squashfs        ro,nosuid,nodev,relatime,errors=continue
├─/boot                       systemd-1        autofs          rw,relatime,fd=48,pgrp=1,timeout=120,minproto=5,maxproto=5,direct,pipe_ino=8369
├─/home                       /dev/mapper/home btrfs           rw,nodev,relatime,space_cache=v2,subvolid=5,subvol=/
└─/usr                        /dev/mapper/usr  squashfs        ro,relatime,errors=continue
  └─/usr                      sysext           overlay         ro,nodev,relatime,lowerdir=/run/systemd/sysext/meta/usr:/run/systemd/sysext/extensions/toolbox/usr:/run/systemd/sysext/extensions/shell_dot
```

And a link to the [issue](https://gitlab.gnome.org/GNOME/gnome-build-meta/-/issues/718) when found out

alatiera:
I am not entirely sure about the fix, since `/etc/machine-id` is a bit different and gets mounted in `/run/host` but seems to be working.

debarshiray:
That's odd.  The CentOS Stream 9 and Fedora tests on [Zuul CI](https://softwarefactory-project.io/zuul/t/local/status) didn't run.

debarshiray:
recheck

debarshiray (commented):
Thanks for tracking this down, @alatiera !  The patch looks good to me.

I have only one question about the root cause.  It seems to me that on this particular build of GNOME OS, the entire `/` on the host is mounted with `...,nodev,...`.  Since `/etc` isn't a separate mount point, `/etc` is also `nodev`, and this runs into problem with the `ro` flag as it's used here.  ie., `mount --rbind -o ro /run/host/etc/machine-id /etc/machine-id` tries to remove that `nodev` and runs into problems.  Did I understand that right?

alatiera:
That's my understanding as well. Only `/usr` is a separate mount in this build yes. Both `/var` and `/etc` are just under `/`

debarshiray:
> That's my understanding as well. Only `/usr` is a separate mount in this build yes. Both `/var` and `/etc` are just under `/`

Thanks for confirming that!  I took the liberty to add that to the commit message, because I will surely mix up the details very soon.  :)

Let's wait for the CI to finish.

Note that the tests run on Fedora Rawhide nodes are failing because of the same reasons as in https://github.com/containers/toolbox/pull/1344 and https://github.com/containers/toolbox/pull/1331 , and the root cause appears to be rsync: https://bugzilla.redhat.com/show_bug.cgi?id=2229654  So, I am going to temporarily ignore them.

PR #1355, opened 2023-08-22
test/system: Test that group and user IDs work

debarshiray:
These tests assume that the group and user information on the host operating system can be provided by different plugins for the GNU Name Service Switch (or NSS) functionality of the GNU C Library.  eg., on enterprise FreeIPA set-ups.  However, it's expected that everything inside the Toolbx container will be provided by `/etc/group`, `/etc/passwd`, `/etc/shadow`, etc..

While `/etc/group` and `/etc/passwd` can be read by any user, `/etc/shadow` can only be read by root.  However, it's awkward to use `sudo(8)` in the test cases involving `/etc/shadow`, because they ensure that `root` and `$USER` don't need passwords to authenticate inside the container, and `sudo(8)` itself depends on that.  If `sudo(8)` is used, the test suite can behave unexpectedly if Toolbx didn't set up the container correctly. eg., it can get blocked waiting for a password.

Hence, `podman unshare` is used instead to enter the container's initial user namespace, where `$USER` from the host appears as `root`.  This is sufficient because the test cases only need to read `/etc/shadow` inside the Toolbx container.

### DIFF
--- a/test/system/206-user.bats
+++ b/test/system/206-user.bats
@@ -1,0 +1,520 @@
+# shellcheck shell=bats
+#
+# Copyright © 2023 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


debarshiray:

PR #1356, opened 2023-08-22
cmd/initContainer: Simplify code by removing a function parameter

debarshiray:
Until now, `configureUsers()` was pushing the burden of deciding whether
to add a new user or modify an existing one on the callers, even though
it can trivially decide itself.  Involving the caller loosens the
encapsulation of the user configuration logic by spreading it across
`configureUsers()` and it's caller, and adds an extra function parameter
that needs to be carefully set and is vulnerable to programmer errors.

Fallout from 9ea6fe5852ea8f5225114d825e8e6813e2a3cfea

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -236,24 +236,12 @@ func initContainer(cmd *cobra.Command, args []string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


debarshiray:

PR #1363, opened 2023-09-14
test/system: Replace the shebangs with 'shell' directives

debarshiray:
These files aren't marked as executable, and shouldn't be, because they
aren't meant to be standalone executable scripts.  They're meant to be
part of a test suite driven by Bats.  Therefore, it doesn't make sense
for them to have shebangs, because it gives the opposite impression.

The shebangs were actually being used by external tools like Coverity to
deduce the shell when running shellcheck(1).  Shellcheck's inline
'shell' directive is a more obvious way to achieve that.

### DIFF
--- a/test/system/001-version.bats
+++ b/test/system/001-version.bats
@@ -1,4 +1,4 @@
-#!/usr/bin/env bats
+# shellcheck shell=bats
 #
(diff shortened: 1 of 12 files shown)


debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

PR #1367, opened 2023-09-19
test/system: Simplify checking if the container started or not

debarshiray:
Bats' `run` helper is not necessary to merely check if a command succeeded or not [1].  In this case, it's idiomatic to pipe the *output* directly to `grep(1)` and use it as the condition for an `if` branch.

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -439,10 +439,9 @@ function container_started() {
   for TRIES in 1 2 3 4 5
   do
     run "$PODMAN" logs "$container_name"
-    container_output="$output"
+
     # Look for last line of the container startup log
-    run grep 'Listening to file system and ticker events' <<< "$container_output"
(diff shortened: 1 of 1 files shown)


HarryMichal (approved):
Straightforward and nicer to read :+1:. But makes me wonder whether this function should be merged with `start_container()` as it does the same with the difference of also ensuring the container did, in fact, start.

debarshiray:
> Straightforward and nicer to read 👍. But makes me wonder whether this function should be merged with `start_container()` as it does the same with the difference of also ensuring the container did, in fact, start.

Thanks for the review, @HarryMichal !  I have wondered that too.  I didn't want to change too many things here, because my main [objective](https://github.com/containers/toolbox/pull/1331) here is to be able to enforce `shellcheck(1)` on all the Bats tests.

debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

PR #1368, opened 2023-09-19
test/system: Silence SC1090

debarshiray:
Otherwise https://www.shellcheck.net/ would complain:
```
  Line 505:
  . "$os_release"
    ^-----------^ SC1090 (warning): ShellCheck can't follow non-constant
                  source. Use a directive to specify location.
```

See: https://www.shellcheck.net/wiki/SC1090

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -502,7 +502,9 @@ function get_system_id() (
     return
   fi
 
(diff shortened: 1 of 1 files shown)


debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

PR #1373, opened 2023-09-21
test/system: Avoid future problems caused by Bat's 'run' overwriting 'i'

debarshiray:
Until Bats 1.10.0, `run` with options had a bug where it would overwrite
the value of the `i` variable even outside `run` [1].

In these particular instances, no options are being passed to 'run',
and, hence, currently there's no problem.  However, in case a future
commit adds an option, then it could lead to hard-to-debug problems.
eg., `--separate-stderr` sets `i` to 1, `--show-output-of-passing-tests`
sets it to 2, etc..  Therefore, depending on the flag and the loop, the
loop might get terminated prematurely or run infinitely or something
else.

Moreover, Bats 1.10.0 is only available in Fedora >= 39 and is absent
from Fedoras 37 and 38.  Therefore, it's not possible to consider this
bug fixed.

Hence, it's better to preemptively work around it to avoid any future
issues.

[1] Bats commit 502dc47dd063c187
    https://github.com/bats-core/bats-core/commit/502dc47dd063c187
    https://github.com/bats-core/bats-core/issues/726

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -107,6 +107,8 @@ function _pull_and_cache_distro_image() {
     assert_success
   fi
 
+  local -i j
+
   for ((j = 0; j < num_of_retries; j++)); do
     run "$SKOPEO" copy --dest-compress "docker://${image}" "dir:${IMAGE_CACHE_DIR}/${image_archive}"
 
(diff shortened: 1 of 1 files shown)


debarshiray:
recheck

debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

PR #1375, opened 2023-09-22
test/system: Ensure failure if an invalid distribution is specified

debarshiray:
Contrary to what the documentation might seem to imply [1], Bats' `fai`' helper only aborts a test case under certain circumstances.  eg., when called from `setup_suite()`, but not from within a child function, and a `@test` case, but not from within the `run` helper.

If `fail` is called from within `run`, then the code after it will continue to execute.  The test case will only fail if `run` eventually catches a non-zero exit code that's caught by `assert_success` [2]. Similarly, it doesn't abort if called from within a child function in `setup_suite()`.

Currently, `_pull_and_cache_distro_image()` is a child function called from `setup_suite()`.  So `fail` won't abort if an invalid distribution is specified.

Fortunately, `pull_distro_image()` is being called from within `@test` cases, but outside `run`.  So, there's no problem with it now.  However, some future code changes can unknowingly alter this reality and it too can run into unexpected behaviour.

Therefore, it's better to be safe, and explicitly specify a non-zero exit code after `fail`.  It will ensure that it works as expected under all circumstances.

[1] https://github.com/bats-core/bats-support

[2] https://github.com/bats-core/bats-assert

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -88,6 +88,7 @@ function _pull_and_cache_distro_image() {
 
   if [ -z "${IMAGES[$distro]+x}" ]; then
     fail "Requested distro (${distro}) does not have a matching image"
+    return 1
   fi
 
   image="${IMAGES[$distro]}"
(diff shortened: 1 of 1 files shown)


debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

Issue #1359, opened 2023-08-25
Using toolbox can destroy months or years of bash history


AdamWill:
**Describe the bug**
I use a [profile.d snippet](https://www.happyassassin.net/posts/2015/01/16/bash-history-with-multiple-sessions/) based on [one from bkuhn](https://lists.gnu.org/archive/html/bug-bash/2011-06/msg00040.html) to keep a very long and constantly updated bash history. It's surprisingly valuable to me (I say "surprisingly" because I didn't realize how valuable it was till this bug happened and I found myself swearing constantly about all the commands I've lost from my history...)

It seems that when you use a toolbox container, it shares your user's history file from your home directory, but does *not* source any configuration of the history environment variables you do in `/etc/profile.d`. So...inside the toolbox container, the `HISTSIZE` and `HISTFILESIZE` values are the defaults, and when you enter the container, it truncates your massive history to a...very short one.

**Steps how to reproduce the behaviour**
1. Set up a host system (I hit the bug using Fedora 38 and Fedora 39, and carefully reproduced it in an F38 VM for this bug report) to have large `HISTSIZE` and `HISTFILESIZE` configured via `/etc/profile.d` (I suspect the bug won't happen if you do this in `~/.bash_profile` or `~/.bashrc`, as the container probably gets those), and a history of more than 1000 entries (I copied my `profile.d` snippet and current history file into a VM)
2. Before entering the container, check the size of the history file and how many items you see when you run `history`
3. Run `toolbox enter`
4. Check the history file size within and outside the container: it will now be much smaller. Note that `HISTSIZE` and `HISTFILESIZE` are both 1000 inside the container and running `history` in the container shows 1000 items, plus however many commands you've run in the container
5. Exit the container. Run `history` from outside the container and see you now have only ~1000 entries in it

**Expected behaviour**
The container should not eat thousands of entries and years of accumulated knowledge from my history file.

**Actual behaviour**
It did.

**Screenshots**

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.99.4`

**Toolbox package info (`rpm -q toolbox`)**
toolbox-0.0.99.4-3.fc39.x86_64

**Output of `podman version`**
```
Client:       Podman Engine
Version:      4.6.0
API Version:  4.6.0
Go Version:   go1.21rc3
Built:        Mon Jul 24 05:44:03 2023
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
[...]

PR #1380, opened 2023-09-29
pkg/utils: Preserve environment variables for Bash's history facility

debarshiray:
Any system-wide customization to Bash's history facilities done through a custom /etc/profile.d configuration snippet on the host operating system gets lost inside the Toolbx container.

This is because Toolbx doesn't know what name to expect for the custom /etc/profile.d snippet on the host, and, hence, can't give access to it through a bind mount or symbolic link inside the container.  The user can definitely set up their own symbolic link inside the container to a snippet inside /run/host/etc/profile.d.  However, it's tedious to do that for all containers, and the user may not even know that they are missing the customization until they notice something wrong with the history, which is shared across all containers and the host, and at that point they might have already lost commands that they can't easily reconstruct.

Therefore, it's worth trying to improve the situation by default.

This tries to preserve the environment variables used to customize Bash's history facilities [1] across the host operating system and Toolbx container.  It assumes that the Bash start-up scripts inside the container won't overwrite any of the propagated variables, which might not always be the case [2].

[1] https://www.gnu.org/software/bash/manual/html_node/Bash-History-Facilities.html
    https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

[2] https://pagure.io/setup/pull-request/48

https://github.com/containers/toolbox/issues/1359


debarshiray:
```
fedora-rawhide | #
fedora-rawhide | # --
```

Issue #603, opened 2020-10-29
Toolbox containers missing man pages for many packages


lambda:
**Describe the bug**

The fedora-toolbox containers from https://registry.fedoraproject.org/, which are installed by default by `toolbox`, are missing a number of man pages, such as those for utilities in `coreutils`, which are packaged in `coreutils-common`.

It looks like each release has a hand-maintained list of packages to restore docs, since docs are not included in the parent fedora image. For example, https://github.com/containers/toolbox/blob/master/images/fedora/f33/missing-docs list packages to restore docs for in the f33/fedora-toobox container. But this list is incomplete, there are a number more packages which are missing docs.

**Steps how to reproduce the behaviour**

This has been tested on both f32 and f33, as I was in the process of upgrading when I discovered this.

1. Run `toolbox create` to create the default toolbox (or `toolbox create --release f33` if you want to be specific about it)
2. Run `toolbox enter` to start a shell in the default toolbox (or `toolbox enter --release f33`)
3. Run `man mktemp` within the toolbox shell to look up docs on the `mktemp` utility

**Expected behaviour**

I expected to read the docs on the `/usr/bin/mktemp` utility form the `coreutils` project.

**Actual behaviour**

I got only `man(3) mktemp`, on the `mktemp` standard library call.

**Output of `toolbox --version` (v0.0.90+)**

```
[lambda@eiffel ~]$ toolbox --version
toolbox version 0.0.96
```

**Toolbox package info (`rpm -q toolbox`)**

```
[lambda@eiffel ~]$ rpm -q toolbox
warning: Found bdb Packages database while attempting sqlite backend: using bdb backend.
toolbox-0.0.96-1.fc33.x86_64
```

**Output of `podman version`**

```
[lambda@eiffel ~]$ podman version
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Built:        Sun Oct  4 19:25:24 2020
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**

```
[lambda@eiffel ~]$ rpm -q podman
warning: Found bdb Packages database while attempting sqlite backend: using bdb backend.
podman-2.1.1-10.fc33.x86_64
```

**Info about your OS**

```
[lambda@eiffel ~]$ cat /etc/os-release 
NAME=Fedora
VERSION="33.20201028.0 (Silverblue)"
ID=fedora
VERSION_ID=33
VERSION_CODENAME=""
PLATFORM_ID="platform:f33"
[...]

PR #1396, opened 2023-11-01
images/fedora: Ensure that documentation and translations are present

debarshiray:
This should finally ensure that the fedora-toolbox image doesn't have any package that had its content, such as documentation or translations, stripped out by the fedora base image.

Until now, missing-docs had a hand-maintained list of packages that had their content stripped out by the fedora base image.  These packages are reinstalled when building the fedora-toolbox image to restore the lost content.  Unfortunately, this list was incomplete because it was only updated when someone noticed that something is missing.

Now, the list is generated with:
```
  $ rpm --all --query --state --queryformat "PACKAGE: %{NAME}\n"
```

... to ensure that it's always complete.

The existing built-in test to ensure that the desired files are actually present in the final image was extended to cover some of those that were absent.  A new built-in test, based on the above rpm(1) command, was added as a fallback to ensure that the final image doesn't have any package with missing content.

Only the images for currently maintained Fedoras (ie., 37, 38 and 39) were updated.

As suggested by Brian Campbell.

https://github.com/containers/toolbox/issues/603


debarshiray:
These failures are because of [some](https://lwn.net/SubscriberLink/947941/3717a3cefa2cef72/) changes in GNU roff 1.23:
```
fedora-rawhide | not ok 3 help: Command 'help' in 222ms
fedora-rawhide | #  in test file test/system/002-help.bats, line 50)
fedora-rawhide | #   `assert_line --index 3 --partial "toolbox - Tool for containerized command line environments on Linux"' failed
fedora-rawhide | #
fedora-rawhide | # index     : 3
fedora-rawhide | # substring : toolbox - Tool for containerized command line environments on Linux
fedora-rawhide | # line      :        toolbox ‐ Tool for containerized command line environments on Linux
fedora-rawhide | # --
fedora-rawhide | #
```

PR #1403, opened 2023-11-15
utils.go: bump fedora fallback version to 38

juhp:
37 will EOL soon

### DIFF
--- a/src/pkg/utils/utils.go
+++ b/src/pkg/utils/utils.go
@@ -64,7 +64,7 @@ const (
 	containerNamePrefixFallback = "fedora-toolbox"
 	distroFallback              = "fedora"
 	idTruncLength               = 12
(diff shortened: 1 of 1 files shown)


juhp:
The versions in .zuul.yaml are also getting really old.
It would be good to keep them up to date.
I can open a PR I guess if it helps.

debarshiray (approved):
Thanks!  Looks good to me.

Rebased against `main`.  Let's wait for the CI to finish.

debarshiray:
> The versions in .zuul.yaml are also getting really old. It would be good to keep them up to date. I can open a PR I guess if it helps.

Yes, of course.  We need to drop Fedora 37 from the CI once it's goes EOL.  Please feel free to submit a pull request.

PR #1404, opened 2023-11-16
test/system: add support for no ~/.bash_profile

pennbauman:
The test "run: Ensure that a login shell is used to invoke the command" assumes `~/.bash_profile` is present the host system and fails without it.

### DIFF
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -132,12 +132,20 @@ teardown() {
 @test "run: Ensure that a login shell is used to invoke the command" {
   create_default_container
 
-  cp "$HOME"/.bash_profile "$HOME"/.bash_profile.orig
-  echo "echo \"$HOME/.bash_profile read\"" >>"$HOME"/.bash_profile
+  local bash_backup
+  if [ -f "$HOME/.bash_profile" ]; then
+    bash_backup="$(mktemp "$HOME/.bash_profile.orig-XXX")"
+    cp "$HOME/.bash_profile" "$bash_backup"
+  fi
+  echo "echo \"$HOME/.bash_profile read\"" >> "$HOME/.bash_profile"
 
   run "$TOOLBOX" run true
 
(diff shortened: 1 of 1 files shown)


debarshiray (commented):
Thanks for your contribution!

Purely in terms of code, the change looks alright.  However, I am curious about the context.  Why do you not have a `~/.bash_profile`?  Is it because you are running a Linux distribution which doesn't use `~/.bash_profile` as the start-up file for login shells?  Or is it because you reconfigured your own OS to be different?

I am asking because this test is a bit hacky.  It's written with the assumption that the host operating system is Fedora or something that uses `~/.bash_profile`.  On top of that, as I learnt later, the test suite doesn't use a separate isolated `$HOME` different from the user's actual `$HOME`, which is dangerous.  We should really fix that.

So, I am trying to understand your perspective.

pennbauman:
I don't have a `~/.bash_profile` because I don't use bash (at least for my regular user) and the file don't really do anything by default (`~/.bashrc` will be loaded without it), so I just deleted it. If you think this patch is too situational to include, that perfectly reasonable.

debarshiray:
I forgot to get back to this pull request.  Sorry about that!!

> On top of that, as I learnt later, the test suite doesn't use a separate isolated $HOME different from the user's actual $HOME, which is dangerous. We should really fix that.

This was finally addressed in commit [4251a0409ffcb9131d](https://github.com/containers/toolbox/commit/4251a0409ffcb9131dcf150104ae5fe91237cc18) (or https://github.com/containers/toolbox/pull/1668) by isolating the user's actual `$HOME` on the host from the system tests.  As part of that the tests inject their own `~/.bash_profile`, which is carried as data for the tests, into the isolated `$HOME`.

I believe that it also addresses the problem you were facing here because the tests should no longer depend on the contents of the user's actual `$HOME` on the host.  It's a slightly better approach than the patch here because it doesn't touch the user's `$HOME` at all, and messing with the user's data without their knowledge is always bad and dangerous.

PR #1407, opened 2023-11-18
cmd/create: Support passing --device option to podman-create

Jmennius:
This allows to use CDI infrastructure, which often does more then just mapping devices in `/dev` -
for NVIDIA this will additionally map a set of libraries into container (which are essential to use the device without a hassle).

Since this is only a pass-through, maybe instead of having a `--device` specific option  we should have an ability to pass arbitrary option to `podman-create`? Like after `toolbox create -c my-container -- --device foo:bar --other-podman-option`?

Fixes: #116 (although it is possible to use nvidia devices inside toolboxes - this change improves the usability significantly when using NVIDIA CTK+CDI with toolbox)

### DIFF
--- a/doc/toolbox-create.1.md
+++ b/doc/toolbox-create.1.md
@@ -8,6 +8,7 @@ toolbox\-create - Create a new Toolbx container
                [*--distro DISTRO* | *-d DISTRO*]
                [*--image NAME* | *-i NAME*]
                [*--release RELEASE* | *-r RELEASE*]
+               [*--device DEVICE*]
                [*CONTAINER*]
 
 ## DESCRIPTION
@@ -110,6 +111,12 @@ remote registry.
 Create a Toolbx container for a different operating system RELEASE than the
 host. Cannot be used with `--image`.
 
+**--device** DEVICE
+
+Add a host device to the container. Passed as is to podman-create(1).
+This option can be specified multiple times or
+multiple comma-separated arguments can be specified.
+
 ## EXAMPLES
 
 ### Create the default Toolbx container matching the host OS
@@ -134,6 +141,12 @@ $ toolbox create --image bar foo
 
 ```
 $ toolbox create --authfile ~/auth.json --image registry.example.com/bar
+```
+
+### Create the default toolbox container with nvidia CDI device
+
+```
+$ toolbox create --device=nvidia.com/gpu=all
 ```
 
 ## SEE ALSO
--- a/doc/toolbox.conf.5.md
+++ b/doc/toolbox.conf.5.md
@@ -31,6 +31,11 @@ remote registry.
 Create a Toolbx container for a different operating system RELEASE than the
 host. Cannot be used with `image`.
 
+**devices** = ["DEVICE", ...]
+
+Add a host device to the container. Accepts a list of devices.
+Each device specification is assed as is to podman-create(1).
+
 ## FILES
 
 The following locations are looked up in increasing order of priority:
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -35,6 +35,7 @@ import (
 	"github.com/godbus/dbus/v5"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 type promptForDownloadError struct {
@@ -54,6 +55,7 @@ var (
 		distro    string
 		image     string
 		release   string
+		devices   []string
 	}
 
 	createToolboxShMounts = []struct {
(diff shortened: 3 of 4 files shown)


Jmennius:
There was a related issue with `nvidia-ctk`: NVIDIA/nvidia-container-toolkit#143 and the fix was merged.
Until it is released there is a workaround - the chmod hook can be just removed from nvidia CDI spec file.

Jmennius:
@debarshiray can I attract your attention, get some opinions on this? 😁

debarshiray (commented):
Sorry for the delay, @Jmennius  I finally got myself some NVIDIA hardware to play with this.

I see that the [Container Device Interface](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html) requires installing the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html).

However, as far as I can make out, the `nvidia-container-toolkit` or `nvidia-container-toolkit-base` packages are only available from NVIDIA's own repositories right now.  For example, I am on Fedora 39, and neither the RPMFusion free nor the non-free repositories have it, but they do have NVIDIA's proprietary driver.

Is there anything else other than NVIDIA that uses the Container Device Interface?

I would like to understand the situation a bit better.  Ultimately I want to make it as smooth as possible for the user to enable the NVIDIA proprietary driver.  That becomes a problem if one needs to enable multiple different unofficial repositories, at least on Fedora.

Jmennius:
> Sorry for the delay, @Jmennius  I finally got myself some NVIDIA hardware to play with this.
> 
> I see that the [Container Device Interface](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html) requires installing the [NVIDIA Container Toolkit](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/install-guide.html).
> 
> However, as far as I can make out, the `nvidia-container-toolkit` or `nvidia-container-toolkit-base` packages are only available from NVIDIA's own repositories right now.  For example, I am on Fedora 39, and neither the RPMFusion free nor the non-free repositories have it, but they do have NVIDIA's proprietary driver.

Yeah, you are right - it's only available in Nvidia repos for Fedora. It would be nice if it was repackaged somehow on rpmfusion... I saw that some distroa package it.

> Is there anything else other than NVIDIA that uses the Container Device Interface?

I am not aware of other CDI implementations :(

> I would like to understand the situation a bit better.  Ultimately I want to make it as smooth as possible for the user to enable the NVIDIA proprietary driver.  That becomes a problem if one needs to enable multiple different unofficial repositories, at least on Fedora.

I guess this is a way to go for the best experience.
Another option with Nvidia is to basically reinstall Nvidia driver libraries inside the container with each upgrade of the kernel driver on the host.

debarshiray:
> For Silverblue, something that I've come up with to handle regenerating the CDI spec after an upgrade: `/etc/systemd/system/nvidia-cdi-update.service`:

That's a really neat hack, indeed.  :)

debarshiray:
> > Is there anything else other than NVIDIA that uses the Container Device Interface?
> 
> I am not aware of other CDI implementations :(

I see commits from Intel in [github.com/cncf-tags/container-device-interface](https://github.com/cncf-tags/container-device-interface/commits/main/), which is great.

debarshiray (commented):
Thanks to @Jmennius and @owtaylor I changed my mind about how to enable the proprietary NVIDIA driver in Toolbx containers.  Since Intel, NVIDIA and several container tools, including Podman, have embraced the Container Device Interface, it's a better path to take than the unmanaged Flatpak extension approach that I had mentioned [before](https://github.com/containers/toolbox/issues/116).

However, I want to be a bit careful when using the CDI.  The way it's widely advertised requires root privileges, because `podman run --device nvidia.com/gpu...` expects the CDI file to be present in either `/etc/cdi` or `/var/run/cdi`.  It's not possible to create the file with `nvidia-ctk cdi generate` and put it in those locations without root access.  It will be good if we could make it work entirely rootless.

One option is to use the Go packages from [tags.cncf.io/container-device-interface](https://github.com/cncf-tags/container-device-interface) and [github.com/NVIDIA/nvidia-container-toolkit](https://github.com/NVIDIA/nvidia-container-toolkit) to create the Container Device Interface file ourselves during `enter` and `run`, make it available to `init-container`, and let it parse and apply it when the container starts.  The CDI file is ultimately a bunch of environment variables, bind mounts and hooks to call `ldconfig(8)`, so it shouldn't be that hard.  Since Toolbx already makes the entire `/dev` from the host available to the container, we don't need to worry about the devices.

This avoids the need for root privileges, and has the extra benefit of enabling the driver in existing Toolbx containers.

I have a working proof-of-concept using this approach in https://github.com/containers/toolbox/pull/1497 that seems to work with the NVIDIA Quadro P600 GPU on my ThinkPad P72 laptop.

PR #1409, opened 2023-11-21
test/system: switch HOME fallback to tmp dir (fix for Ubuntu hosts)

pennbauman:
The test "run: Ensure that $HOME is used as a fallback working directory" assumes `/etc/kernel` will not be present in containers, but this is not the case for Ubuntu (and Debian) containers. This patch switches the test to using a temporary directory (under `/var/tmp` which is not mounted by toolboxes) to ensure fallback is activated.

### DIFF
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -255,15 +255,21 @@ teardown() {
 
   create_default_container
 
-  pushd /etc/kernel
+  local host_only_dir
+  host_only_dir="$(mktemp --directory /var/tmp/toolbox-test-XXXXXXXXXX)"
+
+  pushd "$host_only_dir"
   run --separate-stderr "$TOOLBOX" run pwd
   popd
+
+  rm --force --recursive "$host_only_dir"
 
   assert_success
   assert_line --index 0 "$HOME"
   assert [ ${#lines[@]} -eq 1 ]
   lines=("${stderr_lines[@]}")
-  assert_line --index $((${#stderr_lines[@]}-2)) "Error: directory /etc/kernel not found in container $default_container_name"
+  assert_line --index $((${#stderr_lines[@]}-2)) \
+    "Error: directory $host_only_dir not found in container $default_container_name"
   assert_line --index $((${#stderr_lines[@]}-1)) "Using $HOME instead."
   assert [ ${#stderr_lines[@]} -gt 2 ]
 }


debarshiray (commented):
Thanks for your contribution!

We should definitely try to get the tests ready to be run on Ubuntu hosts, so that we can enable them in our CI.  The tests do have lots of assumptions about running on Fedora hosts at the moment, which isn't good.  So thanks for looking into that.

However, do Ubuntu hosts really not have `/etc/kernel`?  So far, I tried Ubuntu 22.04, 22.10, 23.04 & 23.10 Desktop, and they all had `/etc/kernel.

I am not really opposed to the change, other than someone wanting to share `/var/tmp` between the host and container in future, but I want to understand the problem that you found.

test/system/104-run.bats
```diff
@@ -255,15 +255,20 @@ teardown() {
 
   create_default_container
 
-  pushd /etc/kernel
+  local testdir
+  testdir="$(mktemp -d '/var/tmp/toolbox-test-XXXX')"
```
debarshiray, line 259: Nitpick: the single quotes aren't necessary.

debarshiray (commented):
The commit message could be slightly better.  Something like:
"test/system: Don't assume that all host OSes will have /etc/kernel"

debarshiray:
I took the liberty to address the nits and updated the branch by rebasing against `main`.  I am waiting for your response to my [earlier](https://github.com/containers/toolbox/pull/1409#pullrequestreview-1763477568) question.

pennbauman:
Ubuntu hosts do have the `/etc/kernel` directory, but so do Ubuntu containers. So when the test is run using an Ubuntu container it just enters the container at `/etc/kernel` without being forced to fallback.

debarshiray:
@pennbauman Since you spent some time running the test on non-Fedora hosts, would you mind submitting a new PR that runs this "fallback working directory" test on other images, and not just the default one?  eg., Arch Linux, RHEL, Ubuntu, etc. like we do elsewhere.

pennbauman:
@debarshiray I was thinking that would be a good next step. I will look into that when I have some time.

debarshiray:
> @debarshiray I was thinking that would be a good next step. I will look into that when I have some time.

Thanks!

PR #1413, opened 2023-11-30
cmd: Track the active container on Fedora Linux Asahi Remix

debarshiray:
Christian Hergert requested this.  He is working on improving the integration of Toolbx with the terminal emulation stack in GNOME and Fedora, and he is using Fedora Linux Asahi Remix for his work.


debarshiray:
> `$SHELLOPTS` is already a thing, maybe we could do `$VTEOPTS`
> which has the VTE feature string? We could add something to that
> if the app supports the necessary API.

Sounds good to me.  :)

Issue #1410, opened 2023-11-22
fedora-toolbox:39 has a dangling /etc/resolv.conf symlink on hosts without systemd-resolved(8)


ianw:
The 39 (and 40) tag appears to have `/etc/resolv.conf` in the container as a dangling symlink to `systemd/resolve/stub-resolv.conf`

```
# in 39 toolbox
$ podman run --dns=none -it registry.fedoraproject.org/fedora-toolbox:39 /bin/bash -c "ls -l /etc/resolv.conf; rpm -qa | grep systemd-resolv"
lrwxrwxrwx. 1 root root 39 Nov  7 07:55 /etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
systemd-resolved-254.5-2.fc39.x86_64

# not in 38 toolbox
$ podman run --dns=none -it registry.fedoraproject.org/fedora-toolbox:38 /bin/bash -c "ls -l /etc/resolv.conf; rpm -qa | grep systemd-resolv"
-rw-r--r--. 1 root root 21 Oct  6 06:48 /etc/resolv.conf

# still in 40 toolbox
$ podman run --dns=none -it registry.fedoraproject.org/fedora-toolbox:40 /bin/bash -c "ls -l /etc/resolv.conf; rpm -qa | grep systemd-resolv"
lrwxrwxrwx. 1 root root 39 Nov 20 07:34 /etc/resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
systemd-resolved-254.5-2.fc40.x86_64
```

This breaks name resolution on toolbox entry for people on systems like rhel9 that are not using systemd-resolved

ianw:
I imagine this is also the cause of the problems in https://github.com/containers/toolbox/issues/1406 although it's not specified what system the toolbox is being run on there

debarshiray:
I added some tests to ensure that DNS works inside the containers: https://github.com/containers/toolbox/pull/1414

It's not as exhaustive as I would like it to be, because we aren't currently running all the tests on CentOS Stream 9 and Ubuntu 22.10.  However, it's a start.

We should also probably check `/etc/resolv.conf` itself, but that would require this bug to be fixed first.

debarshiray:
One way to fix this bug would be to always reset the container's `/etc/resolv.conf` to a known good value on container start.  This could cause problems in cases where the user deliberately changed it inside the container, because the customization will get overwritten whenever the container is restarted.

[...]

PR #1425, opened 2023-12-17
Fix DNS queries in images with systemd-resolved on hosts  without it

debarshiray:
On some Toolbx images with `systemd-resolved(8)`, like the `fedora-toolbox`
images for Fedora 39 onwards, `/etc/resolv.conf` can end up being a
symbolic link inside the container that expects the host operating
system to also use `systemd-resolved(8)`:
```ShellSession
  $ ls -l /etc/resolv.conf
  lrwxrwxrwx. 1 root root 39 Nov 28 08:50 /etc/resolv.conf ->
    ../run/systemd/resolve/stub-resolv.conf
```

This happens because `systemd-resolved(8)` already makes `/etc/resolv.conf`
a symbolic link inside the image, and, hence, the container's entry
point doesn't change it to point at the host's copy of the file at
`/run/host/etc/resolv.conf`.  Instead, it's left pointing at the host's
copy of the files maintained by `systemd-resolved(8)` under
`/run/systemd/resolve`, which happen to be also available inside the
container [1].

If the host OS doesn't use `systemd-resolved(8)`, like Red Hat Enterprise
Linux 9, then this leads to a dangling symbolic link and breaks DNS
queries.

Note that the presence of `systemd-resolved(8)` in the recent
`fedora-toolbox` images is a regression caused by the ToolbxReleaseBlocker
Change [2] for Fedora 39 where the image was rewritten in terms of
`fedora-kickstarts` and `pungi-fedora` instead of a Container/Dockerfile.
By mistake, `systemd` crept in as an RPM needed by the image [3], which
in turn pulled in the `systemd-resolved` RPM as a weak dependency [4].

Hopefully, that will get fixed.  However, it's also not practical to
keep track of all the Toolbx images out there in the wild, so it's
wise to make `toolbox(1)` more resilient to such things.

This will have the downside of overwriting some custom user-made
modifications to the container's `/etc/resolv.conf`.  While that's
unfortunate, it's more important to have Toolbx images produce working
containers on a wide range of host OSes.  It will be better to come up
with a more explicit way to support custom user-made modifications to
the container's configuration.  Perhaps with a persistent stamp file.

[1] Commit af602c7d227617d2
    https://github.com/containers/toolbox/commit/af602c7d227617d2
    https://github.com/containers/toolbox/pull/707

[2] https://fedoraproject.org/wiki/Changes/ToolbxReleaseBlocker

[3] fedora-kickstarts commit 48e2c3b5598de32f
    https://pagure.io/fedora-kickstarts/c/48e2c3b5598de32f

[4] fedora-kickstarts commit 49306cb6eada8777
    https://pagure.io/fedora-kickstarts/c/49306cb6eada8777

https://github.com/containers/toolbox/issues/1410

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -200,7 +200,8 @@ func initContainer(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
-		if _, err := os.Readlink("/etc/resolv.conf"); err != nil {
+		if resolvConfTarget, err := os.Readlink("/etc/resolv.conf"); err != nil ||
+			resolvConfTarget != "/run/host/etc/resolv.conf" {
 			if err := redirectPath("/etc/resolv.conf",
 				"/run/host/etc/resolv.conf",
 				false); err != nil {
--- a/test/system/203-network.bats
+++ b/test/system/203-network.bats
@@ -62,6 +62,132 @@ teardown() {
   assert [ ${#stderr_lines[@]} -eq 0 ]
 }
 
+@test "network: /etc/resolv.conf inside the default container" {
+  create_default_container
+
+  run --keep-empty-lines --separate-stderr "$TOOLBOX" run readlink /etc/resolv.conf
+
+  assert_success
+  assert_line --index 0 "/run/host/etc/resolv.conf"
+
+  if check_bats_version 1.10.0; then
+    assert [ ${#lines[@]} -eq 1 ]
+  else
+    assert [ ${#lines[@]} -eq 2 ]
+  fi
+
+  # shellcheck disable=SC2154
+  assert [ ${#stderr_lines[@]} -eq 0 ]
+}
+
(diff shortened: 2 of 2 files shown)


debarshiray:
Cool, the tests do fail if `/etc/resolv.conf` inside the Toolbx container is not a symbolic link:
```
fedora-rawhide | not ok 195 network: /etc/resolv.conf inside the default container in 1724ms
fedora-rawhide | # (from function `assert_success' in file test/system/libs/bats-assert/src/assert.bash, line 114,
fedora-rawhide | #  in test file test/system/203-network.bats, line 70)
fedora-rawhide | #
fedora-rawhide | # status : 1
fedora-rawhide | # output :
fedora-rawhide | # --
```

I had replaced `/etc/resolv.conf` with `/etc/bashrc` to test the tests.  :P

PR #1427, opened 2023-12-19
cmd/utils: Increase buffer size when discarding terminal input

HarryMichal:
Without a sufficient buffer size the discard function does not read fast/efficiently enough causing multiple lines indicating "passed and discarded input" to show up.

I used an already defined constant[0] for the buffer size to prevent the use of a yet-another magical constant.

[0] https://pkg.go.dev/bytes#pkg-constants

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"bufio"
+	"bytes"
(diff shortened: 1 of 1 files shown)


debarshiray (approved):
Thanks, @HarryMichal  Looks good to me.

PR #1428, opened 2023-12-19
cmd/create: Require valid terminal when prompting to pull an image

HarryMichal:
To limit the possible misbehaviour of Toolbx when ued in scripting, require the presence of valid terminal on stdin when an image needs to be pulled to create a new toolbx.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -730,6 +730,16 @@ func pullImage(image, release, authFile string) (bool, error) {
 	}
 
 	if promptForDownload {
+		if !term.IsTerminal(os.Stdin) || !term.IsTerminal(os.Stdout) {
+			var builder strings.Builder
+			fmt.Fprintf(&builder, "image required to create toolbox container.\n")
+			fmt.Fprintf(&builder, "Use option '--assumeyes' to download the image.\n")
+			fmt.Fprintf(&builder, "Run '%s --help' for usage.", executableBase)
+
+			errMsg := builder.String()
+			return false, errors.New(errMsg)
+		}
+
 		shouldPullImage = showPromptForDownload(imageFull)
 	}
 
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -1,3 +1,4 @@
+
 # shellcheck shell=bats
 #
 # Copyright © 2019 – 2023 Red Hat, Inc.
@@ -25,7 +26,7 @@ setup() {
 }
 
 teardown() {
-  cleanup_containers
+  cleanup_all
 }
 
 @test "create: Smoke test" {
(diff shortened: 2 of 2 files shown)


src/cmd/create.go
```diff
@@ -219,6 +221,15 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 
 	pulled, err := pullImage(image, release, authFile)
 	if err != nil {
+		if errors.Is(err, errorNotTerminal) {
+			var builder strings.Builder
+			fmt.Fprintf(&builder, "User interaction required to pull a new image. Stopping now.\n")
+			fmt.Fprintf(&builder, "Run '%s --help' for usage.", executableBase)
```
debarshiray, line 227: Nitpick: the error could hint at the potential solution.

For example, if we re-use the text from the image download prompt, then:
```
Error: image required to create toolbox container.
The '--assumeyes' option is needed.
```

This would fit with our pattern for mentioning the immediate cause of the error, followed by the potential solution or the cause of the error.

debarshiray (commented):
Cool!

debarshiray (changes requested):
This test failure:
```
not ok 77 create: With a custom image that needs an authentication file in 589ms
fedora-rawhide | #  in test file test/system/101-create.bats, line 685)
fedora-rawhide | #   `assert_line --index 0 "Error: failed to pull image $DOCKER_REG_URI/$image"' failed
fedora-rawhide | #
fedora-rawhide | # expected : Error: failed to pull image localhost:50000/fedora-toolbox:34
fedora-rawhide | # actual   : Error: User interaction required to pull a new image. Stopping now.
fedora-rawhide | # --
```

... seems to say that even when `toolbox create` was run non-interactively with the `--assumeyes` option, it still gave the error about requiring user interaction, instead of saying that it failed to pull the image.  Once `--assumeyes` has been given, it shouldn't worry about the standard input stream not being connected to a terminal device.

src/cmd/create.go
```diff
@@ -63,6 +63,8 @@ var (
 		{"/etc/profile.d/toolbox.sh", "/etc/profile.d/toolbox.sh"},
 		{"/etc/profile.d/toolbox.sh", "/usr/share/profile.d/toolbox.sh"},
 	}
+
+	errorNotTerminal = errors.New("terminal is not connected")
```
debarshiray, line 67: We can use [unix.ENOTTY](https://pkg.go.dev/golang.org/x/sys/unix) instead of inventing our own.  It's the same error that `term.GetState` returns when the `*os.File` isn't connected to a terminal device.

Or we can create the final user facing error directly inside `pullImage` because it's a sufficiently high-level function in the `cmd` package.

debarshiray:
> This test failure:
>
> [...]
> 
> ... seems to say that even when `toolbox create` was run non-interactively with the `--assumeyes` option, it still gave the error about requiring user interaction, instead of saying that it failed to pull the image.
> 
> Once `--assumeyes` has been given, it shouldn't worry about the standard input stream not being connected to a terminal device.

I took the liberty to fix this to get the tests to pass, and addressed the above nits.

The fact that this tripped up the tests suggests that we should add some new tests for this new error as well.  :)

debarshiray:
> The fact that this tripped up the tests suggests that we should add some new tests for this new error as well. :)

Added some tests and merged.  Thanks, @HarryMichal !

PR #1429, opened 2023-12-20
cmd: Let the terminal know the active container on all host OSes

debarshiray:
Prompt is a new terminal emulation application [1] designed for a container-oriented desktop that implements the OSC 777 escape sequence to track and preserve the user's current Toolbx container [2].  Hence, Fedora's fork of GNOME Terminal is no longer the only one to offer this.

Therefore, it's time let users of all host operating systems enjoy this feature.

This reverts commits a3e8d8d12bac6dd63010b71c6e091486fb585f37 and 3a96feba478448cd0c5628238619c3805c6aca1e.

[1] https://gitlab.gnome.org/chergert/prompt

[2] https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/17

https://github.com/containers/toolbox/issues/218

### DIFF
--- a/src/cmd/enter.go
+++ b/src/cmd/enter.go
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 – 2022 Red Hat Inc.
+ * Copyright © 2019 – 2023 Red Hat Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


HarryMichal:
Do we know how does this behave in non-VTE/non-patched terminal emulators? Does it change the behaviour of the emulator or is it just ignored?

debarshiray:
> Do we know how does this behave in non-VTE/non-patched terminal emulators? Does it change the behaviour of the emulator or is it just ignored?

I believe VTE just ignores the unknown escape sequence.

This is this branch on Ubuntu 23.10 with a GNOME Terminal that doesn't preserve the current Toolbx container:

image

Issue #558, opened 2020-09-19
HOSTNAME is not set on f33 image


A6GibKm:
## OS
Fedora Silverblue 33
toolbox 0.0.95

## Issue

The enviromental variable `HOSTNAME` is not being set on toolbox, `echo $HOSTNAME` returns only a new line. This does not correspond to the behaviour in previous images.

Related issue: https://github.com/fedora-silverblue/issue-tracker/issues/84

HarryMichal:
Hi @A6GibKm! I quickly tried to reproduce this but could not.

My steps were:

- `toolbox create` (I run Fedora Silverblue 33 -> the `f33/fedora-toolbox` image was used)
- `toolbox enter`
- `echo $HOSTNAME`

The output of the last command was `toolbox` as currently expected. Is there something specific to your use-case I missed? What shell do you run?

A6GibKm:
Not that I know of, I removed my `.bashrc` for testing purposes and enter a newly created toolbox. Also,

```
$ podman images
REPOSITORY                                     TAG     IMAGE ID      CREATED       SIZE
registry.fedoraproject.org/f33/fedora-toolbox  33      95dc9d4b0f27  2 weeks ago   327 MB
registry.fedoraproject.org/f32/fedora-toolbox  32      f7cf4b593fc1  6 months ago  360 MB
docker.io/library/fedora                       32      15fd256292f6  7 months ago  206 MB
```

garrett:
I don't see it here on my workstation, but I do see the missing `$HOSTNAME` in my test VM, even after a `podman system reset`.

```
$ podman images
REPOSITORY                                     TAG     IMAGE ID      CREATED      SIZE
registry.fedoraproject.org/f33/fedora-toolbox  33      95dc9d4b0f27  3 weeks ago  327 MB
```

The VM is completely vanilla (except for overlaying gitk for testing other toolbox/podman related issues), upgraded to the latest F33 beta Silverblue compose.

```
● ostree://fedora:fedora/33/x86_64/silverblue
                   Version: 33.20200929.n.0 (2020-09-29T08:06:48Z)
                BaseCommit: 215677333515cc11110cf016a4cbb021eb929b552f5ed25bb6d403c9f19c1a4a
              GPGSignature: Valid signature by 97A1AE57C3A2372CCA3A4ABA6C13026D12C944D0
           LayeredPackages: fedora-workstation-repositories gitk
```

A6GibKm:
[...]

PR #1436, opened 2024-01-18
test/system: Test that the HOSTNAME environment variable is set

debarshiray:
Bash automatically sets the `HOSTNAME` environment variable to the name of
the current host [1] as returned by `gethostname(2)`, which is the same as
`hostname(1)`.

However, on Fedora, from Fedora 33 onwards, `/etc/profile` sets the
`HOSTNAME` environment variable to `hostnamectl --transient` [2], and,
from Fedora 35 onwards, it has a fallback to `hostname(1)` [3].  These two
approaches return different values when used inside a Toolbx container.
The former picks up the hostname of the host operating system, while the
fallback gets the name that was set when creating the container with
`podman create --hostname toolbox ...`.

Hence, the value of `HOSTNAME` inside a Toolbx container for Fedora
depends on whether the corresponding version of the `fedora-toolbox` image
contained `hostnamectl(1)` or not.

[1] https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

[2] setup commit eb9cc4dce89be24f
    https://pagure.io/setup/c/eb9cc4dce89be24f
    https://bugzilla.redhat.com/show_bug.cgi?id=1745245

[3] setup commit ddd74b5d971a734c
    https://pagure.io/setup/c/ddd74b5d971a734c
    https://pagure.io/setup/pull-request/28
    https://bugzilla.redhat.com/show_bug.cgi?id=1938223

https://github.com/containers/toolbox/issues/558

### DIFF
--- a/test/system/220-environment-variables.bats
+++ b/test/system/220-environment-variables.bats
@@ -439,3 +439,129 @@ teardown() {
   # shellcheck disable=SC2154
   assert [ ${#stderr_lines[@]} -eq 0 ]
 }
+
+@test "environment variables: HOSTNAME inside the default container" {
+  create_default_container
+
+  # shellcheck disable=SC2016
+  run --keep-empty-lines --separate-stderr "$TOOLBOX" run bash -c 'echo "$HOSTNAME"'
+
+  assert_success
+  assert_line --index 0 --regexp "^(toolbox|$HOSTNAME)$"
+
+  if check_bats_version 1.10.0; then
+    assert [ ${#lines[@]} -eq 1 ]
+  else
+    assert [ ${#lines[@]} -eq 2 ]
+  fi
+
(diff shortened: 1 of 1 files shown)


debarshiray:
Fedora's `/etc/profile` has it's own mechanism to set the `HOSTNAME` environment variable, which leads to:
```
fedora-rawhide | not ok 313 environment variables: HOSTNAME inside the default container in 3188ms
fedora-rawhide | #  in test file test/system/220-environment-variables.bats, line 450)
fedora-rawhide | #   `assert_line --index 0 "toolbox"' failed
fedora-rawhide | #
fedora-rawhide | # expected : toolbox
fedora-rawhide | # actual   : np0004424111.novalocal
fedora-rawhide | # --
```

PR #1437, opened 2024-01-19
test/system: Replace the RHEL toolbox:8.7 image with toolbox:8.9

debarshiray:
Red Hat Enterprise Linux 8.7 reached End of Life on 31st May 2023: https://access.redhat.com/articles/4038291
https://access.redhat.com/support/policy/updates/errata

Since the tests are intended for Toolbx, not the Red Hat infrastructure, it will be better to use a newer image, because it will be closer to what the users are seeing.

### DIFF
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -1,4 +1,3 @@
-
 # shellcheck shell=bats
 #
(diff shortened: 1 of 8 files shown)


debarshiray:
recheck

debarshiray:
recheck

PR #1441, opened 2024-01-22
Fix pacman cache removal in Arch image

Antiz96:
The `--noconfirm` pacman flag automatically accepts the default answer for each questions and the default answer to removing cached packages when using `pacman -Scc` is "N" (no).
That means that `pacman -Scc --noconfirm` does not actually remove cached packages.

Since pacman doesn't have a way to choose to answer yes or no automatically *(yet?)*, a simple workaround is to use `yes | pacman -Scc ` instead. However, while it is [recommended to set `pipefail` when using pipes](https://docs.docker.com/develop/develop-images/instructions/#using-pipes), it is not possible with the `yes` command (see [this comment](https://stackoverflow.com/a/70852402) for more details).  
Although, it should be safe to just use a piped `yes` without setting `pipefail` as I don't see how it could be a cause of failure in that particular case.

If using a pipe without setting `pipefail` is a concern, one can simply run `rm -f /var/cache/pacman/pkg/*` instead.  
I can adapt this PR accordingly if needed.

I remain available if needed :slightly_smiling_face:

### DIFF
--- a/images/arch/Containerfile
+++ b/images/arch/Containerfile
@@ -13,7 +13,7 @@ RUN pacman -Syu --needed --noconfirm - < extra-packages
 RUN rm /extra-packages
 
 # Clean up cache
-RUN pacman -Scc --noconfirm
+RUN yes | pacman -Scc
 
 # Enable sudo permission for wheel users
 RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/toolbox


Foxboron (approved):

debarshiray:
@Antiz96 thanks for noticing this!

I wonder if it will be possible to add a test for this.  Maybe based on:
```
$ ls -1 /var/cache/pacman/pkg
```

In the past, I managed to hack in some tests for the `fedora-toolbox` images into their Container files.  It gives some assurance that future changes to the images or the underlying OS won't regress anything.

Antiz96:
@debarshiray You're welcome :)

We could add a line like this for instance:

```
RUN [ -z "$(ls /var/cache/pacman/pkg)" ] || echo "non empty" # or exit, or whatever is needed
```

Although, to be honest, I'm not so worried about `yes | pacman -Scc` failing to clean the cache, but you can never be too careful I guess :smile:  
I can open another PR for the above test if you want to.

debarshiray:
> @debarshiray You're welcome :)
> 
> Could add a line like this for instance:
> 
> ```
> RUN [ -z "$(ls /var/cache/pacman/pkg)" ] || echo "non empty" # or exit, or whatever is needed
> ```

Sounds good to me.
 
> Although, to be honest, I'm not so worried about `yes | pacman -Scc` failing to clean the cache, but you can never be too careful I guess 😄 I can open another PR for the above test if you want to.

Ultimately @Foxboron is the maintainer of the Arch Linux images.  So, I leave it to him.

I personally tend to err on the side of caution.  Someone might casually change something without checking, and then who knows?  After all, you did discover this problem where the existing attempt wasn't working as expected.  :)

PR #1442, opened 2024-01-22
build: Do not use use auto dependencies for shell completion scripts

valentindavid:
It is difficult for downstream to track dependencies when they are automatic. This results in shell completion scripts not being installed in packages because builders do not have the right dependency. This commit adds meson feature arguments to guard those dependencies so that downstream distributions can use `-Dauto_features=enabled`.

For more explanation, see rule #3 of:
https://blogs.gnome.org/mcatanzaro/2022/07/15/best-practices-for-build-options/

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,7 +1,19 @@
 option(
+  'bash_completions',
+  description: 'Install Bash completion scripts',
+  type: 'feature',
+)
+
+option(
   'bash_completions_dir',
   description: 'Directory for Bash completion scripts',
   type: 'string',
+)
+
+option(
+  'fish_completions',
+  description: 'Install fish completion scripts',
+  type: 'feature',
 )
 
 option(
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -1,9 +1,21 @@
+option(
+  'bash_completions',
+  description: 'Install bash completion scripts',
```
debarshiray, line 3: Nitpick: Bash, not bash.

debarshiray (commented):
Thanks, @valentindavid

I had this suspicion that these auto dependencies were not correct, but I was too lazy to carefully read Michael's blog.  So, thanks for correcting them.

meson_options.txt
debarshiray: I took the liberty to fix it.

debarshiray:
Rebased, added a link to this pull request and merged.  Thanks, again.

Issue #1399, opened 2023-11-05
What is this project's name?


pmaddams:
Is it Toolbox or Toolbx? Please don't tell me the O is silent. The [project website](https://containertoolbx.org/) uses both. If I search for software that uses Toolbo?x, should I search for both spellings? Please provide an FAQ.

debarshiray:
> I suggest that you change the name back to Toolbox, 

That's unlikely to happen.  :)

The decision to drop one *o* and rename the project as *Toolbx* happened in October 2021: https://github.com/containers/containertoolbx.org/pull/1

Renaming a live project with many different pieces is a deceptively large amount of work, which is why we never got around to getting it fully done.

> because Toolbx is unpronounceable.

The pronunciation is still *toolbox*, but it's spelt without the final *o*.  It's similar to how Flatpak is flatpack without the *c*.

The problem with spelling the name as Toolbox is that it's a very generic word.  More
and more people have been pointing out that "toolbox" is terribly difficult to search for, and it's impossible to find any decent Internet real estate by that name.

> Imagine people at a conference talking about container tools. I think they will pronounce it Toolbox, so it's confusing for the name to be spelled different. It's not too late to change it back.

Most people will likely have seen the slides at conference talks and blogs referring to the project as Toolbx, and they will probably use tab completion when using the actual command.  Those should reduce the confusion a bit.

> Do you want help changing the name in various places?

You could help finishing the rename to Toolbx.  :)

debarshiray:
Pull requests for:
* the code: https://github.com/containers/toolbox/pull/1446
* the website: https://github.com/containers/containertoolbx.org/pull/39

... to advance with the renaming some more.

debarshiray:
> > Do you want help changing the name in various places?
> 
> You could help finish the rename to _Toolbx_. :)

The above commits and pull requests contain pointers to work that's left to be done.  I think the easiest next step would be to add create a `/run/.toolbxenv` stamp file and migrate our own use of the stamp file to the new name.

debarshiray:
Closing.  Thanks for stopping by.

PR #1446, opened 2024-02-03
Drop one "o" and rename the project as "Toolbx"

debarshiray:
This is meant to make the project more searchable on the Internet.  More and more people have been pointing out that "toolbox" is terribly difficult to search for, and it's impossible to find any decent Internet real estate by that name.

Some exceptions:

  * The code repository is still https://github.com/containers/toolbox.  It will be renamed after giving a heads-up to other contributors.

  * The name of the binary is still `toolbox`.  The name is embedded into existing Toolbx containers as their entry point, which is bind mounted from the host operating system when the containers are started.  Trivially renaming the binary will prevent these containers from starting.

  * For similar reasons, the `TOOLBOX_PATH` environment variable is still the same.

  * For similar reasons, the `profile.d` file to be read by the shell on start-up is still called `toolbox.sh`.

  * The label used to identify Toolbx containers and images is still called `com.github.containers.toolbox`.  There are many existing Toolbx containers, and many Toolbx images beyond the control of the Toolbx project that use this label to identity themselves.  Simply renaming the label will prevent these containers and images from being recognized.

  * The names of the built-in Toolbx images still retain the word `toolbox`.  Images under the new name need to be published on the OCI registries and the `toolbox(1)` binary needs to be taught to handle both old and new names, wherever necessary, for backwards compatibility.

  * The stamp file used to identify Toolbx containers is still called `/run/.toolboxenv` because it's used by various external programs and users to identify Toolbx containers.

  * The OSC 777 escape sequence to track and preserve the user's current Toolbx container [1] still emits `toolbox` as the name of the container runtime.  Changing the escape sequence can break terminal emulation applications, like Prompt [2], that consume it.  Hence, it needs to be done carefully.

  * The runtime directories at `/run/toolbox`, when used as root, and `$XDG_RUNTIME_DIR/toolbox`, when used rootless, weren't renamed.

    When used as root, `/run/toolbox` is embedded into existing Toolbx containers as a bind mount from the host.  Trivially renaming the path will prevent these containers from starting.

    Secondly, both these paths are used to synchronize container start-up.  If the paths are trivially renamed, and the `toolbox(1)` binary is updated and used without stopping all existing containers, then it won't be able to enter containers that were already started.  Strictly speaking, this scenario isn't supported, since updates are always expected to be *offline* [3].  However, it's worth noting because solving the previous problem might also address this.

  * The configuration file for RPM is still called `/usr/lib/rpm/macros.d/macros.toolbox`.

[1] https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/17

[2] https://gitlab.gnome.org/chergert/prompt

[3] https://www.freedesktop.org/software/systemd/man/latest/systemd.offline-updates.html

https://github.com/containers/toolbox/issues/1399

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,6 +1,6 @@
 ---
 name: Bug report
-about: Toolbox's bug report template
+about: Toolbx's bug report template
 title: ''
 labels: 1. Bug
 assignees: ''
@@ -28,7 +28,7 @@ If applicable, add screenshots to help explain your problem.
 **Output of `toolbox --version` (v0.0.90+)**
 e.g., `toolbox version 0.0.90`
 
-**Toolbox package info (`rpm -q toolbox`)**
+**Toolbx package info (`rpm -q toolbox`)**
 e.g., `toolbox-0.0.18-2.fc32.noarch`
 
 **Output of `podman version`**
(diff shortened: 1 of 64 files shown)


debarshiray:
recheck

Issue #608, opened 2020-10-29
Cannot change group ownership of files within a toolbox container


Iolaum:
**Describe the bug**
When the primary user tries to change the group ownership of a file within a toolbox container it fails with the following error:
`chown: changing group of '/tmp/test.1': Operation not permitted`

**Steps how to reproduce the behaviour**

1. Create and enter container from Fedora Silverblue 33 host:
```
$ toolbox create --container test
$ toolbox enter test
```
Reproduce error:
```
$ sudo groupadd group1
$ sudo usermod -a -G group1 $USER
$ echo test > /tmp/test.1
$ chown :group1 /tmp/test.1 
chown: changing group of '/tmp/test.1': Operation not permitted
```

**Expected behaviour**
The change in group ownership would succeed (as it does on the host and on a "normal" podman fedora container).

**Actual behaviour**
Changing the group ownership fails.

**Screenshots**

**Output of `toolbox --version` (v0.0.90+)**
```
$ toolbox --version
toolbox version 0.0.96
```

**Toolbox package info (`rpm -q toolbox`)**
```
$ rpm -q toolbox
warning: Found bdb Packages database while attempting sqlite backend: using bdb backend.
toolbox-0.0.96-1.fc33.x86_64
```

**Output of `podman version`**
```
$ podman version
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Built:        Mon Oct  5 02:25:24 2020
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
```
$ rpm -q podman
warning: Found bdb Packages database while attempting sqlite backend: using bdb backend.
podman-2.1.1-10.fc33.x86_64
```

**Info about your OS**
```
$ cat /etc/os-release 
NAME=Fedora
VERSION="33.20201028.0 (Silverblue)"
...
```

**Additional context**
When did the issue start occurring? After an update (what packages were updated)?
If the issue is about operating with containers/images (creating, using, deleting,..), share here what image you used. If you're unsure, share here the output of `toolbox list -i` (shows all toolbox images on your system).

```
$ toolbox list -i
IMAGE ID      IMAGE NAME                                        CREATED
4f81af0fcbc0  registry.fedoraproject.org/f33/fedora-toolbox:33  2 days ago
```

debarshiray:
[...]

PR #1447, opened 2024-02-08
test/system: Ensure that the user is part of a group with the same name, and the process started by 'podman exec' has all groups

debarshiray:
https://github.com/containers/toolbox/issues/608

### DIFF
--- a/test/system/206-user.bats
+++ b/test/system/206-user.bats
@@ -525,3 +525,234 @@ teardown() {
   # shellcheck disable=SC2154
   assert [ ${#stderr_lines[@]} -eq 0 ]
 }
+
+@test "user: id(1) for $USER inside the default container" {
+  create_default_container
+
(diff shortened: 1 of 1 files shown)


test/system/206-user.bats
```diff
@@ -512,9 +518,241 @@ teardown() {
...
   assert_success
+  assert_line --regexp "^$USER:x:[[:digit:]]+:$USER$"
   assert_line --regexp "^sudo:x:[[:digit:]]+:$USER$"
   assert [ ${#lines[@]} -gt 1 ]
 
   # shellcheck disable=SC2154
   assert [ ${#stderr_lines[@]} -eq 0 ]
 }
+
+@test "user: id(1) for $USER inside the default container" {
```
debarshiray, line 529: These tests fail if I revert the crun [fix](https://github.com/containers/crun/commit/9effaebb429a1aed).  So, I suppose they serve their purpose.

PR #1462, opened 2024-02-28
playbooks/system-test, test/system: Avoid running out of storage space

debarshiray:
The system tests download several images when setting up the test suite, and cache them for later use by the tests [1].  This saves time and avoids hitting rate limits imposed by OCI registries by not downloading the same images repeatedly for several tests, but at the cost of increased use of storage space to cache the images.

The images are cached under `BATS_TMPDIR`.  It defaults to the `TMPDIR` environment variable, and if that's not set then to `/tmp` [2].  Normally, `TMPDIR` isn't set, and the images end up getting cached under `/tmp`.  Now, `/tmp` is typically on `tmpfs` backed by RAM or swap, which means that it should be used for smaller size-bounded files only, and `/var/tmp` should be used for everything else [3].

The images are big enough that a collection of them can't be described as smaller and size-bounded, and it led to:
```
  1..306
  # test suite: Set up
  # test suite: Tear down
  not ok 1 setup_suite
  # (from function `setup_suite' in test file ./setup_suite.bash, line
      55)
  #   `_pull_and_cache_distro_image fedora "$((system_version-1))" ||
      false' failed
  # Failed to cache image registry.fedoraproject.org/fedora-toolbox:40
      to /tmp/bats-run-IPz4Cn/image-cache/fedora-toolbox-40
  # time="2024-02-19T11:41:43Z" level=fatal msg="copying system image
      from manifest list: writing blob: write
      /tmp/bats-run-IPz4Cn/image-cache/fedora-toolbox-40/dir-put-blob607392514:
      no space left on device"
  # bats warning: Executed 1 instead of expected 306 tests
```

So, change the default location of the `BATS_TMPDIR` environment variable to `/var/tmp` by setting `TMPDIR`.

[1] Commit 50683c9d9a78adc9
    https://github.com/containers/toolbox/commit/50683c9d9a78adc9
    https://github.com/containers/toolbox/pull/375

[2] https://bats-core.readthedocs.io/en/stable/writing-tests.html

[3] https://systemd.io/TEMPORARY_DIRECTORIES/

### DIFF
--- a/.github/workflows/ubuntu-tests.yaml
+++ b/.github/workflows/ubuntu-tests.yaml
@@ -1,5 +1,5 @@
 #
-# Copyright © 2023 Red Hat, Inc.
+# Copyright © 2023 – 2024 Red Hat, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/NEWS
+++ b/NEWS
@@ -527,7 +527,7 @@ Overview of changes in 0.0.1
 
 ----
 
-Copyright © 2018 – 2023 Red Hat, Inc.
+Copyright © 2018 – 2024 Red Hat, Inc.
 All rights reserved.
 
 Copying and distribution of this file, with or without modification,
--- a/gen-docs-list
+++ b/gen-docs-list
@@ -1,6 +1,6 @@
 #!/bin/sh
 #
-# Copyright © 2019 – 2022 Red Hat, Inc.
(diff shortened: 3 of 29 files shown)


HarryMichal:
Nice catch. I just wonder whether we should set the variable only in the CI or make it a requirement when invoking the tests (therefore also update the README).

debarshiray:
> I just wonder whether we should set the variable only in the CI or make it a requirement when invoking the tests (therefore also update the README).

Good point.  Yes, let's mention it in `test/system/README.md`.

Issue #1397, opened 2023-11-02
Allow Local Builds of Images from Containerfile


wbrawner:
**Is your feature request related to a problem? Please describe.**
I don't necessarily want to have to push my development images to a remote registry.

**Describe the solution you'd like**
 It would be more convenient to simply provide a `Containerfile` in a repository alongside the source code that contributors could use to build a local image. `toolbox create` could then offer functionality to use this `Containerfile` instead of needing an image name.

**Describe alternatives you've considered**
One can just run `podman create` prior to running `toolbox create`, though it would be nice to have `toolbox` handle that automatically.

**Additional context**
VS Code has [development containers](https://containers.dev/) that allow for maintainers to ship `Dockerfile`s alongside their code, so that contributors don't need to spend time configuring their machines with the necessary dependencies in order to be able to contribute code. Dev containers configure much more than the container images though, such as VS Code extensions and forwarded ports, and they aren't as tightly integrated with the desktop environment, so they aren't a great substitute.

wbrawner:
I've been doing a little thinking here. `toolbox create` can just have a `-f` flag added to point to a `Containerfile` to build the image from, but what about other commands? Also, what about the naming of the container? Should we come up with some sort of naming convention based on the directory name to distinguish containers between various projects or should the `Containerfile` be required to include a `name` label? I think if we could agree on a naming convention for containers created from a `Containerfile`, we could have the other commands check for the existence of a `Containerfile` in the current directory and use that naming convention to check for an existing container for that directory and use it. Alternatively, we could leave the other commands unaffected and require a `name` argument be passed when `toolbox create` is called with the `-f` flag. Thoughts?

boolzy:
I was just working on an alias/wrapper to do this by creating a shell function to call `podman build`, so it's nice to see plans for native functionality.

My approach was to have `toolbox build` call `podman build`, as to more logically separate CLI args, where `toolbox create` is more closely associated with `podman run`, so having build added under create seemed to cross responsibilities per say because I don't believe it's possible to build an image with `podman run`.

Ideally I'd like to see something like `toolbox build -t localhost/custom-toolbox:latest /path/to/Containerfile`, which more closely matches the `podman build` workflow.

[...]

PR #1469, opened 2024-03-11
Add a flag to build the image from a file

KilianHanich:
Hello,

this PR is about #1397. It adds two flags:

- `--build` to specify a build-context which is then passed to `podman build`
- `--build-tag` to optionally specify how the built image should be named; it passes the argument to `podman build --tag`

I changed the files under doc to reflect the changed and added a few bats test cases.

I can add more if these are wanted, but I currently don't think more are needed for this feature.


KilianHanich:
Would be nice to know how this error wasn't caught by building it locally, but whatever.

KilianHanich:
Ok, I have no idea what the problem with this error is.

It's calling "podman images" that fails, but it works on my machine for some reason.

debarshiray (commented):
Wow, thanks for picking this up again.

One quick question: other than the discussion on the original [issue](https://github.com/containers/toolbox/issues/1397), did you see the discussion on the previously attempted [pull request](https://github.com/containers/toolbox/pull/1401)?

KilianHanich:
> One quick question: other than the discussion on the original [issue](https://github.com/containers/toolbox/issues/1397), did you see the discussion on the previously attempted [pull request](https://github.com/containers/toolbox/pull/1401)?

Yes, I did.

On the one hand, yes, filesystems are racy and one could ofc guard against stuff.

On the other hand, one just passed a path to the directory with the Containerfile in and guarding against it can be quite annoying.

So, we get a few questions:
- How likely is it to actually be a problem?
- Does this project want the complexity from it? It would require quite a bit more changes to properly do it.
- This option assumes that the Containerfile is at the root of the build context. If the other files in there actually matter but the Containerfile itself changed (or is removed), we don't have any guarantee that these files didn't change either. So, what would such a guard actually guard against? If it's malicious, one can do it just as well via these files. And if it isn't, one has no guarantee that the other files wouldn't be broken (for the build) too.
- Shouldn't this problem be delegated to better builders? (e.g. podman-build, buildah etc.)

Personally I don't think dealing with that is worth it. But ofc if you want that, I would implement it.

KilianHanich:
On a different note, could it be that podman can't pull something in the tests? Because that would explain the failures...

KilianHanich:
I think I finally figured out what fails.
podman build fails thanks to SELinux...

To be exact, podman outputs at step 6 (`RUN rm /etc/rpm/macros.image-language-conf`):
`/bin/sh: error while loading shared libraries: /lib64/libc.so.6: cannot apply additional memory protection after relocation: Permission denied`
The AVC has as scontext `system_u:system_r:container_t:s0:c123,c347` and as tcontext `unconfined_u:object_r:cache_home_t:s0`.

@debarshiray Do you have an idea about how to fix this? I am not really well versed when it comes to SELinux.

KilianHanich:
Don't approve of it yet, please.

KilianHanich:
Ok, so, there is still one problem here with the build pipeline:

Since bats create a temporary image repository under a tmpdir which are labeled with e.g. `unconfined_u:object_r:cache_home_t:s0` which SELinux doesn't like since podman is labeled as `system_u:system_r:container_t:s0:` with some Multi-Category-Security label.

I tried to locally fix this by running `XDG_CACHE_HOME=$HOME/podmanreg bats test/system` to make the test suit place the stuff it works on somewhere else and gets a different label. That didn't work since it still can't access files with type `user_home_t`.

This implementation so far works with the default placement of the images and with SELinux in permissive mode. But I am unsure where to go from here.

I also tested this by creating a VM and set `XDG_CACHE_HOME` to `$HOME/.local/share` and creating a symlink from `~/.local/share/toolbx/system-test-storage` to `~/.local/share/containers` (I didn't want to loose my actual toolboxes after all and cleaned up the contents of `~/.local/share/containers`). That worked.

So, the remaining issue is one with the pipeline and I don't know how to proceed with that. @debarshiray, do you have an idea?

PR #1475, opened 2024-04-02
profile.d: don't try to source non-existent os-release file

ptalbert:
When /etc/os-release does not exist we directly try to source /usr/lib/os-release. This causes a 'No such file or directory' error when entering a toolbox which has neither.

Fix this by checking for the presence of /usr/lib/os-release before sourcing it, just like what is done for /etc/os-release.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -13,7 +13,7 @@ toolbox_welcome_stub="$toolbox_config/toolbox-welcome-shown"
 eval $(
           if [ -f /etc/os-release ]; then
               . /etc/os-release
-          else
+          elif [ -f /usr/lib/os-release ]; then
               . /usr/lib/os-release
           fi
 


mh21:
@ptalbert should it set some defaults for the expected variables that now are unset?

ptalbert:
On Tue, Apr 2, 2024 at 5:49 PM Michael Hofmann ***@***.***>
wrote:

> @ptalbert <https://github.com/ptalbert> should it set some defaults for
> the expected variables that now are unset?
>

I don't see this script doing anything special which requires some default
value. In the case where neither os-release file exists the behaviour of
the script will not be any different after this change.

debarshiray (commented):
Welcome to Toolbx!  :)

It's true that the complete absence of `os-release(5) makes no difference to the `profile.d/toolbox.sh` snippet as it's today.  However, the `toolbox(1)` binary does make extensive use of `os-release(5)` for some critical logic.

The `toolbox(1)` binary currently uses the [github.com/acobaugh/osrelease](https://github.com/acobaugh/osrelease) Go module to parse `os-release(5)` and it has the same structure as the `profile.d/toolbox.sh` snippet.  ie., `/etc/os-release` may be missing, but it somewhat assumes that `/usr/lib/os-release` is present.  eg., there's no fallback to `linux` if `ID` is absent.

Therefore, if  we are going to make changes to support set-ups without any `os-release(5)`, we need to ensure that the whole thing, including the binary and the snippet, works.  This brings me to the question: how did you end up without any `os-release(5)`?

mh21:
> This brings me to the question: how did you end up without any os-release(5)?
By using a RHEL6-based container image 😂

ptalbert:
@debarshiray like Michael said, I noticed this when using a rhel6 container image with toolbox. Every time you enter the toolbox (or source your profile in the container) there is this warning about a missing /usr/lib/os-release.

For our purposes I am happy to take Michael's solution but maybe there are a few unrelated people out there using toolbox with containers that predate os-release 🤷. If there is something else I can test please let me know.

PR #1484, opened 2024-05-02
test/system: Drop unnecessary port configuration from registry container

debarshiray:
The specific port used by the local temporary Docker registry inside the
container doesn't matter.  The container is running only one service and
its users only see the corresponding port on the host operating system.
The default port within the container is 5000 [1] and there's no reason
to change it.

[1] https://github.com/distribution/distribution-library-image

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -190,20 +190,21 @@ function _setup_docker_registry() {
   assert_success
 
   # Create a Docker registry
-  run "$PODMAN" --root "${DOCKER_REG_ROOT}" run -d \
+  run "$PODMAN" --root "${DOCKER_REG_ROOT}" run \
+    --detach \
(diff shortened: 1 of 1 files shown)


debarshiray:
recheck

debarshiray:
Note that we can't change the local temporary Docker registry container to:

* Use plain text HTTP instead of encrypted HTTP, because it would require changing all the `podman(1)` and `skopeo(1)` calls everywhere to include flags like `--tls-verify=false` and `--dest-tls-verify=false`.  It's not worth doing because no real registry out in the wild should be using plain text HTTP.

* Not use `--privileged` because otherwise the container can't read the certificate and keys inside the `/certs` directory.  The other option is to use `--security-opt label=disable`.  We can make that change, if for some reason that's considered better, but it will take some research.

PR #1495, opened 2024-05-20
Optimize 'enter' and 'run' for a container getting initialized

debarshiray:
Currently, the `enter` and `run` commands poll at one second intervals
to check if the Toolbx container's entry point has created the
initialization stamp file to indicate that the container has been
initialized.  This came from the POSIX shell implementation [1], where
it was relatively easier to poll than to use `inotify(7)` to monitor the
file system.

The problem with polling is that the interval is always going to be
either too short and waste resources or too long and cause delays.  The
current one second interval is sufficiently long to add a noticeable
delay to the `enter` and `run` commands.

It will be better to use `inotify(7)` to monitor the file system, which is
quite easy to do with the Go implementation, so that the commands can
proceed as soon as the initialization stamp file is available, instead
of waiting for the polling interval to pass.

There's a fallback to polling, as before, when the operating system is
suffering from a shortage of resources needed for `inotify(7)`.  This code
path can be forced through the `TOOLBX_RUN_USE_POLLING` environment
variable for testing.  Setting this environment variable disables some
code to ensure that the polling ticker is actually used, because,
otherwise, the race between the creation and detection of the
initialization stamp file makes it difficult to test the fallback.

[1] Commit d3e0f3df06d3f5ac
    https://github.com/containers/toolbox/commit/d3e0f3df06d3f5ac
    https://github.com/containers/toolbox/pull/305

https://github.com/containers/toolbox/issues/1070

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -287,7 +288,8 @@ func runCommand(container string,
 		return err
 	}
 
-	initializedStamp := fmt.Sprintf("%s/container-initialized-%d", toolboxRuntimeDirectory, entryPointPID)
+	initializedStampBase := fmt.Sprintf("container-initialized-%d", entryPointPID)
+	initializedStamp := filepath.Join(toolboxRuntimeDirectory, initializedStampBase)
 
 	logrus.Debugf("Checking if initialization stamp %s exists", initializedStamp)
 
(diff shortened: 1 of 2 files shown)


debarshiray:
> Build failed. https://softwarefactory-project.io/zuul/t/local/buildset/31353efeaf6a40719897ddb35175c13e
> 
> ❌ [unit-test-migration-path-for-coreos-toolbox ](https://softwarefactory-project.io/zuul/t/local/build/b7c0545ca11546c781e3ad97bd4d10c6) **POST_FAILURE** in 3m 47s

I will waive that `POST_FAILURE` because the only change between the last two runs was this addition to `test/system/104-run/bats`:
```
+  assert_output --partial "Handling polling tick"
```

PR #1497, opened 2024-05-22
cmd, pkg/nvidia: Enable the proprietary NVIDIA driver

debarshiray:
This uses the NVIDIA Container Toolkit [1] to generate a Container
Device Interface specification [2] on the host during `enter` and `run`
commands.  The specification is saved as JSON in the runtime directories
at `/run/toolbox` or `$XDG_RUNTIME_DIR/toolbox` to make it available to the
Toolbx container's entry point.  The environment variables mentioned in
the specification are directly passed to `podman exec`, while the hooks
and mounts are handled by the entry point.

Toolbx containers already have access to all the devices in the host
operating system's `/dev`, and containers share the kernel space driver
with the host.  So, this is only about making the user space driver
available to the container.  It's done by bind mounting the files
mentioned in the generated CDI specification from the host to the
container, and then updating the container's dynamic linker cache.

This neither depends on `nvidia-ctk cdi generate` to generate the
Container Device Interface specification nor on `podman create --device`
to consume it.

The main problem with `nvidia-ctk` and `podman create` is that the
specification must be saved in `/etc/cdi` or `/var/run/cdi`, both of which
require root access, for it to be visible to `podman create --device`.
Toolbx containers are often used rootless, so requiring root privileges
for hardware support, something that's not necessary on the host, will
be a problem.

Secondly, updating the `toolbox(1)` binary won't let existing containers
use the proprietary NVIDIA driver, because `podman create` only affects
new containers.

Therefore, `toolbox(1)` uses the Go APIs used by `nvidia-ctk cdi generate`
and `podman create --device` to generate, save, load and apply the CDI
specification itself.  This removes the need for root privileges due to
`/etc/cdi` or `/var/run/cdi`, and makes driver avaiable to existing
containers.

Based on an idea from Ievgen Popovych.

[1] https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/
    https://github.com/NVIDIA/nvidia-container-toolkit

[2] https://github.com/cncf-tags/container-device-interface

https://github.com/containers/toolbox/issues/116


debarshiray:
```
TASK [Install RPM packages]
fedora-rawhide | ERROR
fedora-rawhide | {
fedora-rawhide |   "failures": [],
fedora-rawhide |   "msg": "Could not import the libdnf5 python module using /usr/bin/python3 (3.12.3 (main, Apr 17 2024, 00:00:00) [GCC 14.0.1 20240411 (Red Hat 14.0.1-0)]). Please install python3-libdnf5 package or ensure you have specified the correct ansible_python_interpreter. (attempted ['/usr/libexec/platform-python', '/usr/bin/python3', '/usr/bin/python2', '/usr/bin/python'])"
fedora-rawhide | }
```

I believe this because the Zuul executor got updated from Ansible 2.13.7 to 2.15.10, which now [has](https://github.com/ansible/ansible/issues/78898) [support](https://github.com/ansible/ansible/commit/a81b787a05100986) for DNF5, and the DNF5 [Change](https://fedoraproject.org/wiki/Changes/SwitchToDnf5) is now being aimed at Fedora 41 (and Rawhide).  I am trying to fix the CI in https://github.com/containers/toolbox/pull/1509

Issue #1481, opened 2024-04-29
Add distribution: Ubuntu 24.04 LTS


marknsikora:
**Is your feature request related to a problem? Please describe.**
Newest ubuntu release has dropped, not yet available as a toolbox image. 

**Describe the solution you'd like**
Create a new container image

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

Jmennius:
I expect to tackle this on the weekend, expect a PR relatively soon.

znmeb:
Thanks! I've been working with the Docker Hub image `docker.io/library/ubuntu:24.04` and "out of the box", the `unminimize`script doesn't work.I took a look at the shell script and found a work-around, but is it normal for `unminimze` to just exit without actually doing anything?

PR #1499, opened 2024-06-02
Images: Add Ubuntu 24.04 (new LTS) and remove Ubuntu 23.04 (EOL interim)

Jmennius:
Quite simple, it does seem to work :)

Fixes: #1481


Foxboron (approved):

travier (approved):

travier:
Thanks!

travier:
Ah, my bad, I missed that this was the general toolbox repo and I think here we keep all versions of the images (i.e. we don't remove EOL'ed ones) in the repo.

travier:
Could you make a follow-up PR to re-add the EOL'ed 23.04 ones but not build them? Thanks

Jmennius:
> Ah, my bad, I missed that this was the general toolbox repo and I think here we keep all versions of the images (i.e. we don't remove EOL'ed ones) in the repo.

oh really? I think there was one that we have removed as well before... (22.10 in #1416)

travier:
Ah, good point. I'll let @debarshiray decide. I don't mind either way, it's all in the history.

debarshiray:
Thanks for taking care of this, @Jmennius @travier @Foxboron !

> Ah, my bad, I missed that this was the general toolbox repo and I think here we keep all versions of the images (i.e. we don't remove EOL'ed ones) in the repo.

I don't have a strong opinion.  I keep the sources of the EOL:ed `fedora-toolbox` images around because I sometimes find them useful to track down some historical factoid.  If some image maintainer wants to remove them, then so be it.  :)

PR #1514, opened 2024-06-28
profile.d/toolbox.sh: set PS1 as explicit string constants

juhp:
Since the PS1 strings are constants there is no need to use printf and
embed binary color codes, which also makes it hard to match on toolbox PS1.

I tested these in bash and zsh locally and they seem to work fine.


juhp:
(I dropped the zsh PS1 change, since shellcheck doesn't like `$'...'` quoting :man_shrugging:)

juhp:
Closing this for now in favor of #1517

Issue #1357, opened 2023-08-25
do not override PS1


juhp:
**Describe the bug**
Toolbox currently overrides the shell prompt (PS1).
I feel it would be better if it just prepended ⬢ to the existing PS1 rather than overwriting the setting.

**Steps how to reproduce the behaviour**
For example for Fedora 39 the default bash prompt (now green) is:

```
petersen@localhost-live:~$ 
```
which on `toolbox enter` then changes to:
```
⬢[petersen@toolbox ~]$ 
```

**Expected behaviour**
I think it would be good if the prompt just modified to:
```
⬢petersen@toolbox:~$ 
```
Then the prompt would also be closer to (derived from) the default system prompt for the toolbox container OS I think.

**Actual behaviour**
```
⬢[petersen@toolbox ~]$ 
```

debarshiray:
This is a good point!

I remember trying to prepend to `PS1` when I first did this, and for some reason it didn't work.  It might have been due to the order in which `PS1` is set by the start-up scripts on Fedora for Bash as a login shell.  I don't remember.

juhp:
It is a little tricky since the setup order of PS1 is different for a bash login shell, but I got it working now with bash-color-prompt after changing `toolbox.sh` to:
```bash
    [ "${BASH_VERSION:-}" != "" ] && PS1="\[\033[35m\]⬢\[\033[0m\] ${PS1@P}"
```

debarshiray:
You seem to have jogged my memory a bit.  I think it had something to do with the order in which the snippets in `/etc/profile.d` are invoked depending on whether Bash is invoked as a login shell or a non-login shell.  The order isn't stable.  My memory is still very vague.  :)

Anyway, if you have already found a way to improve the situation, then feel free to file a pull request against Toolbx or Fedora's shell start-up [scripts](https://pagure.io/setup/), whichever is appropriate.  Or, I can dig into it later myself.

swick:
Just to add one more data point: for developing mutter we use a script to create a nested D-Bus session and it modifies `PS1` to show the prompt is in this nested session. This is broken currently on toolbox and I often am not sure if a terminal is in this nested D-Bus context or not.

glemco:
This may have side effects, but what about not touching PS1 at all and just overriding the `PROMPT_COMMAND`:
```sh
PROMPT_COMMAND='printf "\033[35m⬢\033[0m"'
```

debarshiray:
[...]

PR #1517, opened 2024-07-07
profile.d: prepend purple hexagon to PS1 rather than overwriting it

juhp:
Importantly this means the traditional Red Hat style prompt is not forced onto all other OS toolboxes as well.

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -48,8 +48,8 @@ fi
 
 if [ -f /run/.containerenv ] \
    && [ -f /run/.toolbxenv ]; then
-    [ "${BASH_VERSION:-}" != "" ] && PS1=$(printf "\[\033[35m\]⬢ \[\033[0m\]%s" "[\u@\h \W]\\$ ")
-    [ "${ZSH_VERSION:-}" != "" ] && PS1=$(printf "\033[35m⬢ \033[0m%s" "[%n@%m]%~%# ")
+    [ "${BASH_VERSION:-}" != "" ] && PS1="\[\033[35m\]⬢ \[\033[0m\]$PS1"
+    [ "${ZSH_VERSION:-}" != "" ] && PS1=$(printf "\033[35m⬢ \033[0m%s" "$PS1")
 
     if ! [ -f "$toolbox_welcome_stub" ]; then
         echo ""


juhp:
I tested this locally in a VM and it works for me both with bash and zsh in fedora-toolbox:41
and ubuntu-toolbox:24.04

glemco:
This doesn't work if one overrides the PS1 in a user file (e.g. bashrc), as those are sourced after the ones in /etc/profile.d and would wipe the PS1 change introduced here.
A more robust way could be to override the `PROMPT_COMMAND` (https://github.com/containers/toolbox/issues/1357#issuecomment-2343235837), although I'm not fully sure about the implications on shells not supporting it as an array

debarshiray (commented):
Thanks for working on this, @juhp !

I didn't mean to ignore this.  I have been trying to get a new dot-zero release out, and keep hitting various obstacles, and I am always paranoid about anything that touches the shell start-up scripts because every operating system does things so differently.

It will be good to try it out on host and container combinations of the operating systems that we claim to support.  ie., Arch, Fedora, RHEL and Ubuntu.  I know, the matrix isn't small.  :)

What do you think of the `PROMPT_COMMAND` idea, and perhaps falling back to the current approach if PROMPT_COMMAND is not an array?

The current approach with the hexagon doesn't work on some host/container combinations.  eg., Arch and Ubuntu, but I forgot the details.  I wonder if this improves that situation or something can be done to improve it.

juhp:
> This doesn't work if one overrides the PS1 in a user file (e.g. bashrc), as those are sourced after the ones in /etc/profile.d and would wipe the PS1 change introduced here.

@glemco that is already the case today and I might even call it a feature: users should be able to override toolbox PS1 if they want, and it is particularly necessary currently since the current toolbox prompt is rather restrictive and insufficient (the biggest problem being that it is does not distinguish different toolboxes clearly).

juhp:
> I have been trying to get a new dot-zero release out, and keep hitting various obstacles, and I am always paranoid about anything that touches the shell start-up scripts because every operating system does things so differently.

I would love to see a 0.1 release finally :-)

> It will be good to try it out on host and container combinations of the operating systems that we claim to support. ie., Arch, Fedora, RHEL and Ubuntu. I know, the matrix isn't small. :)

I feel a better approach might be to put out snapshots and test them in Rawhide, etc (early in the development cycle) for broader testing: it is not really practical for an individual to test all possible host × container matrices realistically.
Or ideally it should be done in CI or automated of course. But I will try to do some testing later.

> What do you think of the `PROMPT_COMMAND` idea, and perhaps falling back to the current approach if PROMPT_COMMAND is not an array?

I reopened #1514 which was actually the original motivation or wart I wanted to fix, since currently the binary PS1 makes it harder than necessary to override PS1 for users or `bash-color-prompt` in Fedora, etc. That PR may be a simpler first step since it doesn't change the current PS1 much (it just changes from a binary to ascii string).

I need to think more about PROMPT_COMMAND in this context.

> The current approach with the hexagon doesn't work on some host/container combinations. eg., Arch and Ubuntu, but I forgot the details. I wonder if this improves that situation or something can be done to improve it.

I see, maybe there should be an issue to track that?

juhp:
Rebased and I tried to minimize the changes further.

PR #1523, opened 2024-08-08
Add support for loongarch

wuruilong01:
The following error was encountered compiling in the debian community, and this PR solves the problem.[https://buildd.debian.org/status/fetch.php?pkg=golang-github-containers-toolbox&arch=loong64&ver=0.0.99.3%2Bgit20230118%2B446d7bfdef6a-2.1&stamp=1722443115&raw=0](url)


src/meson.build
```diff
@@ -56,6 +56,8 @@ elif cpu_family == 'x86_64' and endian == 'little'
   dynamic_linker = '/lib64/ld-linux-x86-64.so.2'
 elif cpu_family == 'riscv64' and endian == 'little'
   dynamic_linker = '/lib/ld-linux-riscv64-lp64d.so.1'
+elif cpu_family == 'loongarch64' and endian == 'little'
+  dynamic_linker = '/lib/ld-linux-loongarch-lp64d.so.1'
```
debarshiray, line 60: Is there a typo here?

According to [this](https://sourceware.org/glibc/wiki/ABIList#loongarch) ABI specification, the dynamic linker should be `/lib64/ld-linux-loongarch-lp64d.so.1`, but here it's only `/lib/...`.

debarshiray (changes requested):

debarshiray:
Meanwhile, I took the liberty to rebase against `main`.

src/meson.build
wuruilong01: It was a clerical error, you're right, thanks for the code review!

wuruilong01:
> Thanks for working on this, @wuruilong01 !
> 
> We need your full name in the Git authorship information. After searching for a bit, I found [this](https://github.com/wuruilong01/snapd/commit/d56ca1664d232ab7d1578787ba5ba06500964084) commit. May I assign your commit to "Ruilong Wu"?

It's okay.

src/meson.build
debarshiray: Okay!  Thanks for confirming that.

debarshiray:
Thank you for your contribution, @wuruilong01 !

PR #1526, opened 2024-08-14
test/system: Consolidate teardown() to remove both containers & images

debarshiray:
It's far more consistent and understandable if all tests start with a
clean state without any containers or images present.  Otherwise, the
subtle side-effects of having some image left behind from a previous
test can lead to surprises, and there's no need to spend time wondering
whether some tests should only clean up the containers or both
containers and images.

PR #1527, opened 2024-08-14
.zuul: Drop testing on Fedora 38

debarshiray:
Fedora 38 reached End of Life on 21st May 2024:
https://docs.fedoraproject.org/en-US/releases/eol/

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -81,24 +81,12 @@
     pre-run: playbooks/setup-env.yaml
     run: playbooks/system-test.yaml
 
-- job:
-    name: system-test-fedora-38
-    description: Run Toolbx's system tests in Fedora 38
-    timeout: 3600
-    nodeset:
-      nodes:
-        - name: fedora-38
-          label: cloud-fedora-38
-    pre-run: playbooks/setup-env.yaml
-    run: playbooks/system-test.yaml
-
 - project:
     periodic:
       jobs:
         - system-test-fedora-rawhide
         - system-test-fedora-40
         - system-test-fedora-39
-        - system-test-fedora-38
     check:
       jobs:
         - unit-test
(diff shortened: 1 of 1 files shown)


debarshiray:
This is odd:
```
1..368
# test suite: Set up
# test suite: Tear down
not ok 1 setup_suite
# (from function `setup_suite' in test file test/system/setup_suite.bash, line 48)
#   `_pull_and_cache_distro_image rhel 8.10 || false' failed
# Failed to cache image registry.access.redhat.com/ubi8/toolbox:8.10 to /var/tmp/bats-run-bgPYrs/suite/image-cache/rhel-toolbox-8.10
# time="2024-08-14T22:20:13Z" level=error msg="Rolling back transaction: sql: transaction has already been committed or rolled back"
# time="2024-08-14T22:20:13Z" level=fatal msg="copying system image from manifest list: writing blob: write /var/tmp/bats-run-bgPYrs/suite/image-cache/rhel-toolbox-8.10/dir-put-blob370975523: no space left on device"
# bats warning: Executed 1 instead of expected 368 tests
```

*no space left on device*

@danpawlik @TristanCacqueray could this be something related to the Software Factory infrastructure?  Only Fedora Rawhide nodes are running out of storage space.  We hit the same problem in https://github.com/containers/toolbox/pull/1526 too.

debarshiray:
It doesn't look like the `registry.fedoraproject.org/fedora-toolbox:41` image is that much bigger than it's `fedora-toolbox:40` counterpart.  The over-the-wire compressed sizes are 377M versus 376.4M as per Skopeo, and the on-disk uncompressed sizes are 2.23G versus 2.19G.

I know that the images [grew](https://pagure.io/fedora-pgm/pgm_docs/pull-request/65) in size in the past few months for hitherto unknown reasons, but I can't understand why we are running out of space only on Fedora Rawhide nodes.

debarshiray:
I went ahead and tentatively merged this pull request, but the CI breakage looks unrelated to me.

PR #1533, opened 2024-09-03
cmd/root, cmd/run: Handle printing all errors but the CLI parsing ones

debarshiray:
This will make it easier to propagate the exit codes of subordinate
processes through an `exitError` instance, when `toolbox(1)` is invoked
inside a container, and invocation is forwarded to the host.

Cobra doesn't honour the root command's `SilenceErrors`, if an error
occurred when parsing the command line for a command, even though the
command was found.  However, Cobra does honour `SilenceErrors`, if the
error occurred afterwards.

Therefore, to avoid setting `SilenceErrors` in each and every command, it
was set in the `PersistentPreRunE` hook (ie., `preRun`), which is called
after all command line parsing has been successfully completed.

https://github.com/containers/toolbox/issues/957

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -76,11 +76,14 @@ func (e *exitError) Error() string {
 
 func Execute() {
 	if err := rootCmd.Execute(); err != nil {
+		if rootCmd.SilenceErrors {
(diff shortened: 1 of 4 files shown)


debarshiray:
recheck

Issue #1438, opened 2024-01-21
Arch Linux test runner


klausenbusk:
Hi

It has been [brought to my attention](https://debarshiray.wordpress.com/2024/01/20/toolbx-now-offers-built-in-support-for-arch-linux-and-ubuntu/) that you would like some test runners for Arch Linux and Debian.

We (Arch Linux) might be able to help and as part of Arch Linux's DevOps team, I would like some more details on what exactly you need.

I assume we are talking about a [self-hosted GitHub runner](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners) for running a job workflow similar to the existing [`ubuntu-tests.yaml` workflow](https://github.com/containers/toolbox/blob/6dfd2294295a18e228e18ca13c7b8d0f44ab8fb6/.github/workflows/ubuntu-tests.yaml)?

How powerful must the runner be? Would a [Hetzner CX11 VM](https://www.hetzner.com/cloud) be sufficient? My understanding is that this would allow one concurrent job.

debarshiray:
First of all, let me thank you deeply for responding to the call for help!  It's much appreciated.

I suspect that we don't need a particularly powerful runner.

We are currently using an instance of [Zuul CI](https://zuul-ci.org/) called [Software Factory](https://softwarefactory-project.io/) for Fedora and CentOS Stream 9 runners, and GitHub Actions [workflow](https://github.com/containers/toolbox/blob/main/.github/workflows/ubuntu-tests.yaml) for a Ubuntu 22.04 runner.  Software Factory has CI runners (or nodes) with different hardware capabilities.  We are using the [ones](https://github.com/containers/toolbox/blob/main/.zuul.yaml) called `cloud-fedora-rawhide`, `cloud-fedora-39`, etc., but I don't know their exact configuration off hand.

I suppose we have two ways to add Arch Linux to the Toolbx CI.  We can either try to add Arch Linux images to Software Factory, which, I think, will then run on Software Factory's CI runners (or nodes), or we can try to add a runner for GitHub Actions workflows.  [Here](https://softwarefactory-project.io/cgit/config/tree/nodepool/virt_images) are the existing image definitions known to Software Factory.

I am not an expert in either, so I am happy to leave it to your tastes and preferences.  @danpawlik and @TristanCacqueray are my contacts for questions about Zuul CI and Software Factory.  I hope they will correct me if I said anything wrong above.

TristanCacqueray:
[...]

PR #1535, opened 2024-09-05
test/system: Update Ubuntu versions used for tests (16.04, 18.04, 20.04 -> 20.04, 22.04, 24.04)

pennbauman:
The Ubuntu versions used for testing are updated from 16.04, 18.04, and 20.04 to 20.04, 22.04, and 24.04. This brings all Ubuntu releases used into the current standard support window, and removes tests using Ubuntu 16.04 and 18.04 which have failed on every machine I've tried them on. 

It may also/alternatively make sense to reduce the number of Ubuntu versions tested (probably to just 24.04), as no other distro is tested with multiple releases. Other distros could also have the versions used updated, for example Fedora 34 is used which is far out of date.

### DIFF
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -181,36 +181,6 @@ teardown() {
   assert_output --regexp "Created[[:blank:]]+rhel-toolbox-8.10"
 }
 
-@test "create: Ubuntu 16.04" {
-  pull_distro_image ubuntu 16.04
-
-  run "$TOOLBX" --assumeyes create --distro ubuntu --release 16.04
-
-  assert_success
-  assert_output --partial "Created container: ubuntu-toolbox-16.04"
-  assert_output --partial "Enter with: toolbox enter ubuntu-toolbox-16.04"
-
-  run $PODMAN ps --all
-
-  assert_success
-  assert_output --regexp "Created[[:blank:]]+ubuntu-toolbox-16.04"
-}
-
-@test "create: Ubuntu 18.04" {
-  pull_distro_image ubuntu 18.04
-
-  run "$TOOLBX" --assumeyes create --distro ubuntu --release 18.04
-
-  assert_success
-  assert_output --partial "Created container: ubuntu-toolbox-18.04"
-  assert_output --partial "Enter with: toolbox enter ubuntu-toolbox-18.04"
-
-  run $PODMAN ps --all
-
-  assert_success
-  assert_output --regexp "Created[[:blank:]]+ubuntu-toolbox-18.04"
-}
-
 @test "create: Ubuntu 20.04" {
   pull_distro_image ubuntu 20.04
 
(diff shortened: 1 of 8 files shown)


debarshiray (commented):
Thanks for working on this, @pennbauman !  My apologies for not responding sooner.  I was caught up with getting a new release out.

I see that the pull request has two commits that are somewhat unrelated.  So, let me start with a high level comment before getting into too many details.

The `test/system: Fix create tests for host distro/version` commit is certainly welcome.  As we add built-in support for more and more distributions, the combinatorial explosion of host and container operating system pairs makes it important to run the tests on as many different host OSes as possible.  Some months ago, I had made an aborted [attempt](https://github.com/containers/toolbox/pull/1505) at this.  I am hoping that your effort will finally push this over the finishing line.

So, thanks for that.

The `test/system: Update Ubuntu versions used for tests` commit makes me curious.  Were you experiencing failures from the tests for Ubuntus 16.04 and 18.04?  The tests seem to be passing when run on the CI.

Other than the need to test as many host and container pairs as possible, there's one technical need to test containers with older operating systems.  The `toolbox(1)` binary has a weird runtime ABI requirement.  Binaries built on newer host OSes must run on containers with older OSes, so must binaries across different distributions.  This is because `toolbox(1)` from the host is bind mounted into the containers as their entry point.

The [toolbox-create(1)](https://github.com/containers/toolbox/blob/main/doc/toolbox-create.1.md) and [toolbox-init-container(1)](https://github.com/containers/toolbox/blob/main/doc/toolbox-init-container.1.md) manuals talk more about this in more detail.  I once wrote a [blog](https://debarshiray.wordpress.com/2022/10/02/toolbx-running-the-same-binary-on-arch-linux-fedora-ubuntu-etc-containers/) to describe the other options that we had explored.

This is generally unsupported - the inverse is supported.

So, we do various tricks when building the `toolbox(1)` binary to ensure that it will meet the above requirements.  See commit 6063eb27b98939942e316771224c5653a9b2e59b for the full details.

While this was implemented after discussion with the glibc developers, it's also not something that's officially supported by them.  Toolbx is clearly pushing the boundaries here.  So, it's important to keep testing against containers with older OSes to ensure that it still works.  This is the main reason for still testing against `fedora-toolbox:34`.

Now, specifically about Ubuntus 16.04 and 18.04, I see that they are still [covered](https://ubuntu.com/about/release-cycle) by the Pro subscription.  So, they are not completely unsupported.  They could be useful as representatives of other similarly old OSes out there, like Debian.  However, if they are causing some real problems that would be too expensive to solve given the benefits, then we can consider dropping them too.

test/system/libs/helpers.bash
```diff
@@ -551,6 +551,11 @@ function get_system_id() (
 function get_system_version() (
   local os_release
 
+  if [ "$(get_system_id)" = "arch" ]; then
+    echo "latest"
```
debarshiray, line 555: I needed to solve the same problem that this hunk addresses for https://github.com/containers/toolbox/issues/1438 which reminded me again of this pull request.

Instead of calling `get_system_id`, we can directly use `VERSION_ID` because we also source `os-release(5)` here.

debarshiray:
> The `test/system: Fix create tests for host distro/version` commit is certainly welcome. As we add built-in support for more and more distributions, the combinatorial explosion of host and container operating system pairs makes it important to run the tests on as many different host OSes as possible.

We need this commit for https://github.com/containers/toolbox/issues/1438

So, let me try to rebase this pull request, so that we can get this commit merged.

PR #1540, opened 2024-09-10
build: Bump go-nvlib to 0.6.0 and nvidia-container-toolkit to 1.16.0

debarshiray:
Note that `github.com/NVIDIA/go-nvlib` > 0.2.0 isn't API compatible with
`github.com/NVIDIA/nvidia-container-toolkit` 1.15.0.  The next release of
`nvidia-container-toolkit` is 1.16.0 and it requires `go-nvlib` 0.6.0.

Therefore, these two Go modules need to be updated together.

The src/go.sum file was updated with `go mod tidy`.


debarshiray:
recheck

PR #1541, opened 2024-09-13
cmd/run, pkg/nvidia: Detect mismatched NVIDIA kernel & user space driver

debarshiray:
The proprietary NVIDIA driver has a kernel space part and a user space
part, and they must always have the same matching version.  Sometimes,
the host operating system might end up with mismatched parts.  One
reason could be that the different third-party repositories used to
distribute the driver might be incompatible with each other.  eg., in
the case of Fedora it could be RPM Fusion and NVIDIA's own repository.

This shows up in the systemd journal as:
```
  $ journalctl --dmesg
  ...
  kernel: NVRM: API mismatch: the client has the version 555.58.02, but
          NVRM: this kernel module has the version 560.35.03.  Please
          NVRM: make sure that this kernel module and all NVIDIA driver
          NVRM: components have the same version.
  ...
```

Without any special handling of this scenario, users would be presented
with a very misleading error:
```
  $ toolbox enter
  Error: failed to get Container Device Interface containerEdits for
      NVIDIA
```

Instead, improve the error message to be more self-documenting:
```
  $ toolbox enter
  Error: the proprietary NVIDIA driver's kernel and user space don't
      match
  Check the host operating system and systemd journal.
```

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -269,7 +269,14 @@ func runCommand(container string,
 
 	cdiSpecForNvidia, err := nvidia.GenerateCDISpec()
 	if err != nil {
-		if !errors.Is(err, nvidia.ErrPlatformUnsupported) {
+		if errors.Is(err, nvidia.ErrNVMLDriverLibraryVersionMismatch) {
+			var builder strings.Builder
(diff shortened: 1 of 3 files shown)


debarshiray:
recheck

debarshiray:
recheck

PR #1542, opened 2024-09-15
Add whitespace padding to the PS1 profile

alatiera:
Add an extra space after the ⬢ so things are less
smooshed together.


alatiera:
Before and after

image

debarshiray (commented):
Thanks, @alatiera !  Let me rebase against `main` and restart the CI.  It should run a little faster now and hopefully avoid the timeout.

PR #1546, opened 2024-09-18
.zuul, test/system: Simplify line count checks by using Bats >= 1.10.0

debarshiray:
Commit 87eaeea6f0c2b245 already added a dependency on Bats >= 1.10.0,
which is present on Fedora >= 39.  Therefore, it should be exploited
wherever possible to simplify things.

Currently, the CI has been frequently timing out on stable Fedora nodes.
So, increase the timeout from 1 hour 50 minutes to 2 hours to avoid
that.

For what it's worth, the timeout for Fedora Rawhide nodes is 2 hours 10
minutes and it seems enough.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -62,7 +62,7 @@
 - job:
     name: system-test-fedora-40
     description: Run Toolbx's system tests in Fedora 40
-    timeout: 6600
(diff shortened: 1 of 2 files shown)


debarshiray:
recheck

PR #1552, opened 2024-09-27
test/system: Optimize the resource limits tests

debarshiray:
The system tests can be very I/O intensive, because many of them copy
OCI images from the test suite's image cache directory to its local
container/storage store, create containers, and then delete everything
to run the next test with a clean slate.  This makes them slow.

The runtime environment tests, which includes the resource limit tests,
are particularly slow because they don't skip the I/O even when testing
error handling.  This makes them a good target for optimizations.

The resource limit tests query the values for different resources from
the same default container without changing its state.  Therefore, a lot
of disk I/O can be avoided by creating the default container only once
for all the tests.

This can save even 30 minutes.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -49,71 +49,141 @@
     run: playbooks/unit-test.yaml
 
 - job:
-    name: system-test-fedora-rawhide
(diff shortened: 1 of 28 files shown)


debarshiray:
recheck

PR #1563, opened 2024-10-09
Add recipe for Fedora 40 image

alexdesiqueira:
Hi @debarshiray, everyone,
I'm experiencing a few "issues" when trying to use `fedora-toolbox` on `distrobox`: `fedora-toolbox:latest` pulls Fedora 39, while `fedora-toolbox:40` pulls Rawhide.
When investigating, I saw we missed a recipe for a Fedora 40 image here. This is an attempt to create this image.
Please feel free to close this if it's not necessary, or let me know how I could improve it.
Thanks!


debarshiray (changes requested):
Thanks for your interest in Toolbx.  A few things:

There are certainly `registry.fedoraproject.org/fedora-toolbox` images tagged with `39`, `40`, `41` and `42`.  Toolbx doesn't use the `latest` tag by default to avoid any ambiguities, and these do work as expected:
```
$ toolbox create --release f39
...
$ toolbox create --release f40
...
```

I can see that `registry.fedoraproject.org/fedora-toolbox:latest` now points to the `fedora-toolbox:40` image.

If you really want it fixed, you will need to work it out with the [Fedora infrastructure](https://pagure.io/fedora-infrastructure/) team.  There's nothing that can be done in this GitHub project.  Only the Arch and Ubuntu images are built and distributed from this GitHub project.  Other than that, we carry a copy of the image definitions to make it easier to do development and testing across multiple distributions without having to look for the images in different places.  For example, unless you work for Red Hat, you wouldn't be able to see the defintion of the UBI-based RHEL images.

Finally, the `fedora-toolbox` images are no longer defined as Containerfile:s, so the image defintions in this pull request are invalid.

The `fedora-toolbox:39` image was defined as a Kickstart:
https://fedoraproject.org/wiki/Changes/ToolbxReleaseBlocker
https://pagure.io/fedora-kickstarts/

... and the `fedora-toolbox:40` and newer images are defined as KIWI descriptions:
https://fedoraproject.org/wiki/Changes/KiwiBuiltCloudImages
https://pagure.io/fedora-kiwi-descriptions

We can ignore the Kickstart for the purposes for this GitHub project, because it was short lived and it's difficult to build them outside the Fedora infrastructure.

So, what we need to do is add the KIWI descriptions to this GitHub project.

debarshiray:
Closing.

PR #1565, opened 2024-10-17
Added ubuntu-toolbox version 24.10, removed 23.10 (eol)

davidscholberg:
This PR adds an ubuntu-toolbox image for ubuntu 24.10 and removes version 23.10 as it has reached EOL.

I copied the files from 24.04 and made a few slight changes (obviously updated the version numbers but also had to explicitly install `unminimize` since it doesn't seem to be included with the image anymore). Here's everything that I changed from 24.04:

```diff
diff -urN images/ubuntu/24.04/Containerfile images/ubuntu/24.10/Containerfile
--- images/ubuntu/24.04/Containerfile   2024-10-16 21:41:33.928967339 -0400
+++ images/ubuntu/24.10/Containerfile   2024-10-16 22:09:08.875131964 -0400
@@ -1,8 +1,8 @@
-FROM docker.io/library/ubuntu:24.04
+FROM docker.io/library/ubuntu:24.10

 LABEL com.github.containers.toolbox="true" \
       name="ubuntu-toolbox" \
-      version="24.04" \
+      version="24.10" \
       usage="This image is meant to be used with the toolbox command" \
       summary="Base image for creating Ubuntu Toolbx containers" \
       maintainer="Ievgen Popovych <jmennius@gmail.com>"
@@ -21,9 +21,10 @@
 # Install ubuntu-minimal & ubuntu-standard
 # Install extra packages as well as libnss-myhostname
 COPY extra-packages /
-RUN sed -Ei '/apt-get (update|upgrade)/s/^/#/' /usr/local/sbin/unminimize && \
-    apt-get update && \
-    yes | /usr/local/sbin/unminimize && \
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get -y install unminimize && \
+    sed -Ei '/apt-get (update|upgrade)/s/^/#/' /usr/bin/unminimize && \
+    yes | /usr/bin/unminimize && \
     DEBIAN_FRONTEND=noninteractive apt-get -y install \
         ubuntu-minimal ubuntu-standard \
         libnss-myhostname \
```

I'm not sure if any of the other steps in the Containerfile are out of date, but the image does work fine with toolbox from my testing of it.

PR #1568, opened 2024-10-21
Ubuntu workflow: Disable 16.04 and 18.04 builds, do not fail early

Jmennius:
16.04 and 18.04 are not updated anymore (2 years and over 1 year respectively) so there is no point in rebuilding them all the time. Keep image recipes.

With changes by @tom-reinders from #1567 to not stop the workflow early in case one of the versions fails.

### DIFF
--- a/.github/workflows/ubuntu-images.yaml
+++ b/.github/workflows/ubuntu-images.yaml
@@ -37,6 +37,7 @@ jobs:
     strategy:
       matrix:
         release: ['16.04', '18.04', '20.04', '22.04', '24.04', '24.10']
+      fail-fast: false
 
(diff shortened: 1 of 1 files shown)


tom-reinders:
@Jmennius In #1565 when removing 23.10 the files where also removed.

Is there a reason to not do that for 16.04 and 18.04?

Jmennius:
> @Jmennius In #1565 when removing 23.10 the files where also removed.
> 
> Is there a reason to not do that for 16.04 and 18.04?

Yeah, we keep image recipes in general. I opt to remove interim releases because they are not so important in the long run.

travier (approved):

Issue #1573, opened 2024-10-24
Entering fails if Nouveau is being used while the proprietary NVIDIA driver is still installed


picsel2:
**Describe the bug**
A clear and concise description of *what the bug is*. If possible, re-run the command(s) with `--log-level debug` and put the output here.

I can not enter the toolbox:

```
$ toolbox --log-level debug enter
DEBU Running as real user ID 1000                 
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU Looking up sub-GID and sub-UID ranges for user sebastian 
DEBU TOOLBX_DELAY_ENTRY_POINT is                  
DEBU TOOLBX_FAIL_ENTRY_POINT is                   
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Migrating to newer Podman                    
DEBU Toolbx config directory is /home/sebastian/.config/toolbox 
DEBU Current Podman version is 5.2.3              
DEBU Creating runtime directory /run/user/1000/toolbox 
DEBU Old Podman version is 5.2.3                  
DEBU Migration not needed: Podman version 5.2.3 is unchanged 
DEBU Setting up configuration                     
DEBU Setting up configuration: file /home/sebastian/.config/containers/toolbox.conf not found 
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-40'               
DEBU Image: 'fedora-toolbox:40'                   
DEBU Release: '40'                                
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-40'               
DEBU Image: 'fedora-toolbox:40'                   
DEBU Release: '40'                                
DEBU Checking if container fedora-toolbox-40 exists 
DEBU Inspecting container fedora-toolbox-40       
DEBU Entry point of container fedora-toolbox-40 is toolbox (PID=0) 
DEBU Inspecting mounts of container fedora-toolbox-40 
DEBU Generating Container Device Interface for NVIDIA 
[...]

PR #1575, opened 2024-10-24
pkg/nvidia: Unbreak 'enter' if proprietary driver is there but not used

debarshiray:
If the proprietary NVIDIA driver is installed, particularly `libnvidia-ml.so.1`, but the kernel driver is not being used, then `enter` fails with:
```
  $ toolbox enter
  Error: failed to initialize NVIDIA Management Library
```

This was tested on Fedora 39 Workstation with the proprietary NVIDIA driver from RPM Fusion, which makes it possible to easily disable the driver without uninstalling it [1].

Note that, with and without this change, there's a delay of a few seconds inside `nvmlInit()` from the NVIDIA Management Library.

[1] https://rpmfusion.org/Howto/NVIDIA

https://github.com/containers/toolbox/issues/1573

### DIFF
--- a/src/pkg/nvidia/nvidia.go
+++ b/src/pkg/nvidia/nvidia.go
@@ -68,7 +68,10 @@ func GenerateCDISpec() (*specs.Spec, error) {
 			logrus.Debugf("Generating Container Device Interface for NVIDIA: failed to initialize NVML: %s",
 				err)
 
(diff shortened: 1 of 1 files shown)


picsel2:
I have tested this PR and it works well: 
[toolbox-tests.md](https://github.com/user-attachments/files/17530580/toolbox-tests.md)

picsel2 (approved):

PR #1584, opened 2024-11-01
build: Bump github.com/briandowns/spinner to 1.23.1 (part 1)

debarshiray:
Note that github.com/briandowns/spinner 1.18.1 introduced an undesired dependency on github.com/mattn/go-isatty for the IsTerminal() API, which was later removed in 1.23.1 [1].  Fewer dependencies are always good because it reduces the amount of code in use.

The src/go.sum file was updated with 'go mod tidy'.

[1] github.com/briandowns/spinner commit 8f269dd04fbfe236
    https://github.com/briandowns/spinner/commit/8f269dd04fbfe236
    https://github.com/briandowns/spinner/pull/156


debarshiray:
recheck

PR #1587, opened 2024-11-05
playbooks/dependencies-centos-9-stream: Show the Bash version

debarshiray:
Fallout from 6626b11e1565412e411f585657ebe9615ec58cad

### DIFF
--- a/playbooks/dependencies-centos-9-stream.yaml
+++ b/playbooks/dependencies-centos-9-stream.yaml
@@ -54,7 +54,7 @@
     chdir: '{{ zuul.project.src_dir }}'
 
 - name: Check versions of crucial packages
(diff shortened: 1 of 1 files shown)


debarshiray:
Oops, I forgot that we still need `flatpak-session-helper` because it provides the `org.freedesktop.Flatpak.Development` interface that `flatpak-spawn(1)` uses:
```
not ok 223 create: Try with an invalid custom name (using positional argument, forwarded to host)
fedora-rawhide | #  in test file test/system/501-create.bats, line 44)
fedora-rawhide | #   `assert_line --index 0 "Error: invalid argument for 'CONTAINER'"' failed
fedora-rawhide | # ~ ~/src/github.com/containers/toolbox
fedora-rawhide | #
fedora-rawhide | # expected : Error: invalid argument for 'CONTAINER'
fedora-rawhide | # actual   : Portal call failed: The name is not activatable
fedora-rawhide | # --
```

I am glad we have tests!

debarshiray:
recheck

PR #1593, opened 2024-11-12
playbooks: Restore the generation & installation of the Bash completions

debarshiray:
The package providing `/usr/share/pkgconfig/bash-completion.pc` changed
from `bash-completion` to `bash-completion-devel` in Fedora 41 onwards [1].
This meant that the Bash completions were no longer being generated and
installed by the CI.

[1] Fedora bash-completion commit d1f5dc48c0440cc6
    https://src.fedoraproject.org/rpms/bash-completion/c/d1f5dc48c0440cc6
    https://bugzilla.redhat.com/show_bug.cgi?id=1457164

### DIFF
--- a/playbooks/dependencies-centos-9-stream.yaml
+++ b/playbooks/dependencies-centos-9-stream.yaml
@@ -2,7 +2,6 @@
   become: yes
   package:
     name:
-      - bash-completion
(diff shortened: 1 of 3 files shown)


debarshiray:
recheck

PR #1598, opened 2024-11-29
test/system: Fix issue with python resolver IP type

pennbauman:
The existing code would never properly select the IPv4 address instead a fallback value was selected when the IPv4 option was used, the script is updated to that both IPv4 and IPv6 will be properly selected.

### DIFF
--- a/test/system/203-network.bats
+++ b/test/system/203-network.bats
@@ -24,9 +24,8 @@ load 'libs/helpers'
 readonly RESOLVER_PYTHON3='\
 import socket; \
 import sys; \
-family = socket.AddressFamily.AF_INET if sys.argv[1] == "A" else 0; \
-family = socket.AddressFamily.AF_INET6 if sys.argv[1] == "AAAA" else 0; \
-addr = socket.getaddrinfo(sys.argv[2], None, family, socket.SocketKind.SOCK_RAW)[0][4][0]; \
+family = {"A": socket.AddressFamily.AF_INET, "AAAA": socket.AddressFamily.AF_INET6}; \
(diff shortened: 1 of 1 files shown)


test/system/203-network.bats
```diff
@@ -24,9 +24,8 @@ load 'libs/helpers'
 readonly RESOLVER_PYTHON3='\
 import socket; \
 import sys; \
-family = socket.AddressFamily.AF_INET if sys.argv[1] == "A" else 0; \
-family = socket.AddressFamily.AF_INET6 if sys.argv[1] == "AAAA" else 0; \
```
debarshiray, line 28: Aha, `family` was getting overwritten to `0` in the second line if `sys.argv[1]` is `"A"`, right?

test/system/203-network.bats
```diff
@@ -24,9 +24,8 @@ load 'libs/helpers'
...
-addr = socket.getaddrinfo(sys.argv[2], None, family, socket.SocketKind.SOCK_RAW)[0][4][0]; \
+family = {"A": socket.AddressFamily.AF_INET, "AAAA": socket.AddressFamily.AF_INET6}
```
debarshiray: Are we missing a trailing semi-colon?

debarshiray: It looks like the tests don't fail without the trailing semi-colon.  Maybe they are getting skipped internally and not getting reflected as a failure?

debarshiray: I restored the semi-colon for the time being.  The tests pass.  Please let me know if I got that wrong.

debarshiray:
Thanks again, @pennbauman !

PR #1599, opened 2024-11-29
.github/workflows, test/system: Enable 210-ulimit.bats on Ubuntu 22.04

pennbauman:
ulimit is a shell builtin, so the shell it is run from changes the version that is used. Ubuntu (and Debian) use dash as their system shell, instead of bash. The means that toolbox run also used dash with Ubuntu containers, caused an unexpected version of ulimit to be used in tests. Tests were updated to force the use of bash and its version of ulimit.

### DIFF
--- a/.github/workflows/ubuntu-tests.yaml
+++ b/.github/workflows/ubuntu-tests.yaml
@@ -161,6 +161,7 @@ jobs:
             test/system/108-completion.bats \
             test/system/201-ipc.bats \
             test/system/203-network.bats \
+            test/system/210-ulimit.bats \
             test/system/220-environment-variables.bats \
             test/system/230-cdi.bats \
             test/system/250-kerberos.bats \
--- a/test/system/210-ulimit.bats
+++ b/test/system/210-ulimit.bats
@@ -38,7 +38,7 @@ teardown_file() {
   local limit
   limit=$(ulimit -H -R)
 
-  run --keep-empty-lines --separate-stderr "$TOOLBX" run ulimit -H -R
+  run --keep-empty-lines --separate-stderr "$TOOLBX" run bash -c 'ulimit -H -R'
 
   assert_success
   assert_line --index 0 "$limit"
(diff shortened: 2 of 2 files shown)


debarshiray (changes requested):
My apologies that I failed to get to this sooner, and thanks for tracking this down!

The code changes in the patch look correct to me, but the commit message is wrong, and it's important because the underlying reasons aren't straightforward.

When `toolbox run ulimit` is invoked, it boils to running `bash --login -c 'exec ulimit'` inside the container.  If we try that command on Fedora and Ubuntu hosts, we will get different results.

On Fedora:
```ShellSession
$ bash --login -c 'exec ulimit'
unlimited
```

On Ubuntu:
```ShellSession
$ bash --login -c 'exec ulimit'
bash: line 1: exec: ulimit: not found
```

This is because Bash's `exec` built-in cannot execute another built-in and needs an external command; and Fedora ships a wrapper for the `ulimit` built-in as an external command to satisfy [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/toc.html):
```ShellSession
$ cat /usr/bin/ulimit
#!/bin/sh
builtin ulimit "$@"
```

See Fedora's [bash.spec](https://src.fedoraproject.org/rpms/bash/blob/rawhide/f/bash.spec) and bug [1297166](https://bugzilla.redhat.com/show_bug.cgi?id=1297166).

I don't know if it was ever discussed adding these wrappers to the Debian family (/cc @Jmennius ).  If not, then maybe it's worth discussing?

Anyway, wrapping the `ulimit` in a `bash -c 'ulimit'` solves this problem because then it becomes `bash --login -c 'exec bash -c ulimit'`, and the `exec` built-in can execute `bash(1)` which is an external command.

It will be good to change the commit message to reflect the correct reason to save our future selves some head-scratching.  :)

debarshiray (changes requested):
We should add `test/system/210-ulimit.bats` to `.github/workflows/ubuntu-tests.yaml` to enable them in the CI.

debarshiray:
> We should add `test/system/210-ulimit.bats` to `.github/workflows/ubuntu-tests.yaml` to enable them in the CI.

I took the liberty to do this, and rebased on top of `main`.  The commit message still needs fixing.

debarshiray:
> The commit message still needs fixing.

I took the liberty to update the commit message.

I didn't want to delay this any further because of relatively minor details.

PR #1608, opened 2025-01-17
Show the size in list like podman does it

afbjorklund:
Make the output more similar to `podman images`

It still uses 1000-based units, so no mebibytes here.


afbjorklund:
Ah there was already a PR, didn't check page two...

* #1035

PR #1612, opened 2025-01-23
build: Bump github.com/spf13/viper to 1.19.0

debarshiray:


danpawlik:
@debarshiray so it looks better now, or?

debarshiray:
> @debarshiray so it looks better now, or?

Hey!  Setting the `GOINSECURE` environment variable to `go.opencensus.io` worked around the problem.  Thanks for checking in.

debarshiray:
Let's only bump Viper to 1.17.0 for now, because Fedora 41 and 40 are in the process of updating it from 1.17.0 to 1.19.0.

PR #1614, opened 2025-01-29
build: Bump NVIDIA Container Toolkit to 1.17.3

debarshiray:
The `src/go.sum` file was updated with `go mod tidy`.

PR #1629, opened 2025-04-25
pkg/utils: Use errors.Is() instead of os.IsNotExist()

debarshiray:
The `os.IsNotExist()` function [1] predates the introduction of the
`errors.Is()` function [2] in Go 1.13 [3].  From Go >= 1.16, the
documentation explicitly recommends the use of `errors.Is()` instead of
`os.IsNotExist()` [4].

The Go implementation of Toolbx never used any Go older than 1.13 [5],
and currently it requires Go >= 1.21 [6].  So, there's no reason not to
use the more modern and recommended alternative.

[1] https://pkg.go.dev/os#IsNotExist

[2] https://pkg.go.dev/errors#Is

[3] https://go.dev/blog/go1.13-errors

[4] Go commit b641f0dcf48aa748
    https://github.com/golang/go/commit/b641f0dcf48aa748
    https://github.com/golang/go/issues/41122

[5] Commit d857471aa2f233e5
    https://github.com/containers/toolbox/commit/d857471aa2f233e5
    https://github.com/containers/toolbox/pull/318

[6] Commit 82e85bac9f5e69a5
    https://github.com/containers/toolbox/commit/82e85bac9f5e69a5
    https://github.com/containers/toolbox/pull/1614


debarshiray:
recheck

PR #1632, opened 2025-04-30
cmd: Simplify string building

debarshiray:
When the `fmt.Fprintf()` [1] function is used to write to a
`strings.Builder` [2] instance, it uses the `io.Writer` [3] interface, which
is the `strings.Builder.Write()` method.  This method is practically the
same as the `strings.Builder.WriteString()` method, other than the fact
that the former accepts a slice of bytes and the latter accepts a
string.  So, the difference is the initial call to `fmt.Fprintf()`.

Therefore, unless format verbs [4] are needed to build the string,
`fmt.Fprintf()` can be replaced with `strings.Builder.WriteString()`.  It
reduces one function call and is shorter to type.

Fallout from the following:
  * e390f15469959a4b7458fd37069f4a2c8e0b9a4c
  * 7542f5fc867b57bf3dc67bbae02cc09ccc0b5df2
  * e58992066f6a576b943987789b84a01894ef6193
  * 8dd2f8e80aad1b76659b44f20c346f75e647d65d
  * 063bdf965fe42468b6b69a333b28b475fa05436f

[1] https://pkg.go.dev/fmt#Fprintf

[2] https://pkg.go.dev/strings#Builder

[3] https://pkg.go.dev/io#Writer

[4] https://pkg.go.dev/fmt

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -552,7 +552,7 @@ func configureKerberos() error {
 	}
 
 	kcmConfigString := `# Written by Toolbx
-# https://github.com/containers/toolbox
+# https://containertoolbx.org/
 #
 # # To disable the KCM credential cache, comment out the following lines.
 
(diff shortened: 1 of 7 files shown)


debarshiray:
recheck

debarshiray:
recheck

PR #1635, opened 2025-05-01
test/system: Optimize the help, group and user tests

debarshiray:
The system tests can be very I/O intensive, because many of them copy
OCI images from the test suite's image cache directory to its local
container/storage store, create containers, and then delete everything
to run the next test with a clean slate.  This makes them slow.

A lot of disk I/O can be avoided by reducing the number of times this
happens.


debarshiray:
recheck

PR #1636, opened 2025-05-01
cmd/initContainer: Style fix

debarshiray:
Use a `strings.Builder` [1] instance instead of raw string literals for
consistency.
    
Fallout from 772b66bf3e962785730e55f6242fa74066de0743
    
[1] https://pkg.go.dev/strings#Builder


debarshiray:
recheck

Issue #626, opened 2020-11-17
Give access to the CA certificates from the host


debarshiray:
We need to do something to ensure that the certificates from certificate authorities (or CAs) that are available inside the container are kept synchronized with the host.

Normally, we'd do this by using symbolic links or bind mounts in `/etc`, but the layout of the files differs quite a bit across distributions. Debian uses `/etc/ssl`, Fedora uses `/etc/pki` and has `/etc/ssl/certs` as a symbolic link to `/etc/pki/tls/certs`, and so on. It's a mess.

Instead, we might have better luck using [p11-kit](https://github.com/p11-glue/p11-kit), just as Flatpak does. `p11-kit` offers an uniform interface for accessing certificates stored on the file system, smart cards and so on. Normally, `p11-kit` looks up the certificates that are present on the host's file system. However, it can also be configured to forward all look-ups to a socket, and a `p11-kit server ...` instance on the other side does the look-ups and returns the results over the socket.

In our case, we need to start a `p11-kit server ...` instance on the host, configure the container to forward everything to it over the socket, and ensure that the socket is present at the right place.

Here are the Flatpak counterparts of this issue:
https://github.com/flatpak/flatpak/pull/1757
https://github.com/p11-glue/p11-kit/issues/68

debarshiray:
I have some work-in-progress code to make this work with `p11-kit`, as discussed above.  Meanwhile, https://github.com/containers/toolbox/pull/1634 updates the `arch-toolbox` and `ubuntu-toolbox` images to include `p11-kit-client.so`.

debarshiray:
> I have some work-in-progress code to make this work with `p11-kit`, as discussed above. Meanwhile, [#1634](https://github.com/containers/toolbox/pull/1634) updates the `arch-toolbox` and `ubuntu-toolbox` images to include `p11-kit-client.so`.

Those images needed some more work: https://github.com/containers/toolbox/pull/1643

The bulk of the code changes are beginning to appear at: https://github.com/containers/toolbox/pull/1644

The `fedora-toolbox` image is loosely blocking on:
https://src.fedoraproject.org/rpms/p11-kit/pull-request/53

... and it will be updated through the [Fedora KIWI descriptions](https://pagure.io/fedora-kiwi-descriptions).  The UBI based `toolbox` image for RHEL will hopefully be updated by Red Hat according to their internal processes.

debarshiray:
[...]

PR #1634, opened 2025-05-01
images/arch, images/ubuntu: Add p11-kit-client.so

debarshiray:
A subsequent commit will use this to give Toolbx containers access to
the certificates from certificate authorities on the host.  This commit
was kept separate from the changes to `toolbox(1)` to ensure that the
`arch-toolbox` and `ubuntu-toolbox` images are ready before that
happens.

https://github.com/containers/toolbox/issues/626

PR #1643, opened 2025-05-07
images: Create the directory for configuring PKCS#11 modules

debarshiray:
It's better to ensure that the `/etc/pkcs11/modules` directory exists in the image, instead of having the Toolbx container's entry point create it at runtime, because it can be a confirmation that p11-kit was built to read the module configuration from this location.

### DIFF
--- a/images/arch/Containerfile
+++ b/images/arch/Containerfile
@@ -23,3 +23,6 @@ RUN yes | pacman -Scc
 
 # Enable sudo permission for wheel users
 RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/toolbox
+
+# Enable the use of p11-kit-client.so to access CA certificates from the host
+RUN mkdir --parents /etc/pkcs11/modules
--- a/images/ubuntu/18.04/Containerfile
+++ b/images/ubuntu/18.04/Containerfile
@@ -33,6 +33,9 @@ RUN rm /extra-packages
 # Allow authentication with empty password, promptless
 RUN sed -i '/^auth.*pam_unix.so/s/nullok_secure/try_first_pass nullok/' /etc/pam.d/common-auth
 
+# Enable the use of p11-kit-client.so to access CA certificates from the host
+RUN mkdir --parents /etc/pkcs11/modules
+
 # Fix empty bind-mount to clear selinuxfs (see #337)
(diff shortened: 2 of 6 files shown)


images/arch/Containerfile
```diff
@@ -23,3 +23,6 @@ RUN yes | pacman -Scc
 
 # Enable sudo permission for wheel users
 RUN echo "%wheel ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/toolbox
+
+# Enable the use of p11-kit-client.so to access CA certificates from the host
+RUN mkdir --parents /etc/pkcs11/modules
```
debarshiray, line 28: @Foxboron do you think the `libp11-kit` package should also create the `/etc/pkcs11/modules` directory?  It already creates `/usr/share/p11-kit/modules`, which isn't quite the same thing, but similar and related to the PKCS#11 modules that it ships.

Currently, the `p11-kit` package creates `/etc/pkcs11/pkcs11.conf.example`, and I wonder if it should be moved to the `libp11-kit` package.  I think the global configuration file (see `pkcs11.conf(5)`) affects both the libraries and executables.

images/ubuntu/18.04/Containerfile
```diff
@@ -33,6 +33,9 @@ RUN rm /extra-packages
 # Allow authentication with empty password, promptless
 RUN sed -i '/^auth.*pam_unix.so/s/nullok_secure/try_first_pass nullok/' /etc/pam.d/common-auth
 
+# Enable the use of p11-kit-client.so to access CA certificates from the host
+RUN mkdir --parents /etc/pkcs11/modules
```
debarshiray, line 37: @Jmennius I wonder if `libp11-kit0` or `p11-kit-modules` should create the `etc/pkcs11/modules` directory?

debarshiray:
The `ubuntu-toolbox:22.04` image is failing to build as [before](https://github.com/containers/toolbox/pull/1634#issuecomment-2844710808).

images/arch/Containerfile
Foxboron: I checked with the package maintainer and he doesn't have a strong opinion. I'll take a look at the packaging when I have time and see if we can harmonize this.

debarshiray: Thanks for following up!  :)

PR #1644, opened 2025-05-07
Give access to the CA certificates from the host

debarshiray:

### DIFF
--- a/src/cmd/root.go
+++ b/src/cmd/root.go
@@ -263,20 +263,27 @@ func migrate(cmd *cobra.Command, args []string) error {
 
 	migrateLock := toolboxRuntimeDirectory + "/migrate.lock"
 
-	migrateLockFile, err := os.Create(migrateLock)
+	migrateLockFile, err := utils.Flock(migrateLock, syscall.LOCK_EX)
 	if err != nil {
-		logrus.Debugf("Migrating to newer Podman: failed to create migration lock file %s: %s", migrateLock, err)
-		return errors.New("failed to create migration lock file")
+		logrus.Debugf("Migrating to newer Podman: %s", err)
+
+		var errFlock *utils.FlockError
+
+		if errors.As(err, &errFlock) {
+			if errors.Is(err, utils.ErrFlockAcquire) {
+				err = utils.ErrFlockAcquire
+			} else if errors.Is(err, utils.ErrFlockCreate) {
+				err = utils.ErrFlockCreate
+			} else {
+				panicMsg := fmt.Sprintf("unexpected %T: %s", err, err)
+				panic(panicMsg)
+			}
+		}
+
+		return err
 	}
 
 	defer migrateLockFile.Close()
-
-	migrateLockFD := migrateLockFile.Fd()
-	migrateLockFDInt := int(migrateLockFD)
-	if err := syscall.Flock(migrateLockFDInt, syscall.LOCK_EX); err != nil {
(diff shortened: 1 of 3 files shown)


src/pkg/utils/arch.go
```diff
@@ -25,6 +25,11 @@ func getFullyQualifiedImageArch(image, release string) string {
 	return imageFull
 }
 
+func getP11KitClientPathsArch() []string {
+	paths := []string{"/usr/lib/pkcs11/p11-kit-client.so"}
```
debarshiray, line 29: @Foxboron I hope I got the path to `p11-kit-client.so` right for Arch Linux.  I suppose we build the `arch-toolbox:latest` image only for `x86_64`, correct?

src/pkg/utils/ubuntu.go
```diff
@@ -38,6 +38,15 @@ func getFullyQualifiedImageUbuntu(image, release string) string {
 	return imageFull
 }
 
+func getP11KitClientPathsUbuntu() []string {
+	paths := []string{
+		"/usr/lib/aarch64-linux-gnu/pkcs11/p11-kit-client.so",
+		"/usr/lib/x86_64-linux-gnu/pkcs11/p11-kit-client.so",
```
debarshiray, line 44: @Jmennius I hope I got the path to `p11-kit-client.so` right for Ubuntu. I suppose we build the `ubuntu-toolbox` images only for `aarch64` and `x86_64`, correct?

src/pkg/utils/arch.go
Foxboron:
```
λ ~ » sudo pacman -F p11-kit-client.so
core/libp11-kit 0.25.5-1 [installed]
    usr/lib/pkcs11/p11-kit-client.so
```

The path is correct :)

> I suppose we build the arch-toolbox:latest image only for x86_64, correct?

Yes

debarshiray: Thanks for confirming that!

debarshiray:
Grr... looks like the system tests get stuck at:
```
...
# test suite: Tear down
```

debarshiray:
> Grr... looks like the system tests get stuck at:
> 
> ```
> ...
> # test suite: Tear down
> ```

That's [because](https://bats-core.readthedocs.io/en/stable/writing-tests.html#file-descriptor-3-read-this-if-bats-hangs) of the `p11-kit server` process not getting cleaned up.

debarshiray:
> > Grr... looks like the system tests get stuck at:
> > ```
> > ...
> > # test suite: Tear down
> > ```
> 
> That's [because](https://bats-core.readthedocs.io/en/stable/writing-tests.html#file-descriptor-3-read-this-if-bats-hangs) of the `p11-kit server` process not getting cleaned up.

To prevent the test suite from getting stuck, we need to clean up the `p11-kit server` process.

To do that we need to be careful we don't touch the `p11-kit server` instance that was invoked by _normal_ uses of Toolbx.  One way to do this is to [isolate](https://github.com/containers/toolbox/pull/1652) the host's `XDG_RUNTIME_DIR` from the test suite.  Unfortunately, that has turned out to be [easier said](https://github.com/containers/toolbox/pull/1652#issuecomment-2927237728) than done.

To avoid letting perfection become the enemy of the good, I will split the tests out of this pull request, so that they can be added later when the above problems are solved.

debarshiray (commented):
It was [pointed out](https://github.com/containers/toolbox/issues/1674) that the code here broke child sessions started by `sshd(8)`.  There's a fix for it beginning to appear at https://github.com/containers/toolbox/pull/1695

PR #1652, opened 2025-05-30
Isolate the host's XDG_RUNTIME_DIR from the system tests

debarshiray:
`XDG_RUNTIME_DIR` is needed for two groups of reasons when Toolbx is used rootless.

First, it's important for `toolbox(1)` itself to work rootless because it needs to place several files:

  * The 'lock' file to synchronize Podman migrations.

  * The initialization stamp file to synchronize the container's entry point with the user-facing `enter` and `run` commands running on the host operating system.

  * The generated Container Device Interface specification.

These files need to be separate for the `toolbox(1)` processes run by the system tests, those run by the user for 'normal' use, and concurrent invocations of the tests.

Therefore, it's better to use a custom `XDG_RUNTIME_DIR` that's within the sandbox offered by Bats [1].  The sandbox is clearly labelled as being used by Bats, is unique for each invocation, and Bats takes care of cleaning everything up once it has finished running.

Second, `XDG_RUNTIME_DIR` is used to propagate things like the user D-Bus, Pipewire and Wayland sockets from the host to the container.  These don't need to be separated.  However, if a custom `XDG_RUNTIME_DIR` is used then those sockets that are used by the system tests, such as the user D-Bus socket, have to be replicated.

Therefore, a custom D-Bus instance is run to offer the user D-Bus socket with a configuration similar to that of the host OS.

However, Podman can't use systemd as the cgroups manager with this D-Bus instance, as the bus wasn't started by the user systemd instance.  So, a custom `containers.conf(5)` is used to change the cgroups manager to cgroupfs.  The only other options in the `containers.conf(5)` are those that are common across Fedora 41 and 42, and Ubuntu 22.04 and 24.04.


debarshiray:
The test suite runs fine locally on my Fedora 42 Workstation.  However, it runs into problems when running on my Ubuntu 22.04 virtual machine, and on these F41, F42, Rawhide and Ubuntu 22.04 CI hosts.

As far as I can see, the problem is that `podman rm --force` can't kill the containers' entry points with `SIGTERM`.  I see the same problem with `podman stop`.

After tracking it down through the code, I found out that the `cgroup-path` is empty inside the `$XDG_RUNTIME_DIR/crun/$CONTAINER_ID/status` file.  This is likely causing `crun kill --all` to silently fail with a successful exit code.

Issue #1675, opened 2025-07-02
Add a --workdir option to 'run'


debarshiray:
[Ptyxis](https://gitlab.gnome.org/chergert/ptyxis) currently enters a Toolbx container in a new terminal by directly using Podman's `start` and `exec` commands.  We need to work out a way to make it use `toolbox(1)` because Toolbx's `enter` and `run` commands have become more than just those Podman commands and some recent Toolbx features might not work without going through `toolbox(1)`.

I think the `run` command is more appropriate for terminal emulation applications like Ptyxis because it already [supports](https://github.com/containers/toolbox/issues/1066) `--preserve-fds`, which is something these apps need.

Ptyxis needs a way to explicitly mention the working directory when entering a Toolbx container in a new terminal to ensure that the directory is preserved.  This is because the Ptyxis process that creates a new pseudoterminal and will spawn `toolbox(1)` against it, has a different current working directory than the one in the user's previous terminal.  Therefore, `toolbox(1)` cannot automatically find out what the working directory should be.

One way to address this would be to mirror the `--workdir` option of `podman exec` in `toolbox run`.

PR #1679, opened 2025-07-06
cmd, pkg/utils: Stop the container once the last session finishes

debarshiray:
Currently, once a Toolbx container gets started with `podman start`, as
part of the `enter` or `run` commands, it doesn't stop unless the host
is shut down or someone explicitly calls `podman stop`.  This becomes
annoying if someone tries to remove the container because commands like
`podman rm` and such don't work without the `--force` flag, even if all
active `enter` and `run` sessions have ended.

A system of reference counting based on advisory file locks has been
used to automatically exit the container's entry point once all the
active sessions have ended.  Two locks are used - a global lock that's
common for all containers, and a local lock that's specific to each
container.  The initialization stamp file is conveniently used as the
local lock.

The `enter` and `run` sessions acquire shared file locks and the
container's entry point acquires ones that are exclusive.  All attempts
at acquiring the locks are blocking unless otherwise noted.

The global lock is acquired at the beginning of `enter` and `run` before
they inspect the container, negotiate the path to the local lock (ie.,
the initialization stamp file) with the entry point and create it.  Once
the local lock is known by `enter` and `run`, they acquire it and only
then release the global.

The Toolbx container's entry point tries to acquire the global lock as
it creates the initialization stamp file (ie., the local lock).  This
waits for the `enter` and `run` invocations to receive the location of
the local lock, acquire it and release the global.  Once the entry point
acquires the global lock, it releases it, and waits trying to acquire
the local lock.

This sequence of acquiring and releasing the locks lets the entry point
track the state of the `enter` and `run` invocations.  It should only
try to acquire the local lock after the `enter` and `run` invocations
have acquired it before invoking `podman exec`.

The entry point is able to acquire the local lock after all `enter` and
`run` sessions end and release their local locks.

At this point, a new `enter` or `run` invocation might be in the process
of starting.  Both sides need to be careful not to race against each
other and up in an invalid state.  eg., a `podman start` being invoked
against a container whose entry point is just about to exit, or a
`podman exec` being invoked against a container whose entry point is
about to exit or has already exited.

Therefore, the entry point makes a non-blocking attempt to acquire the
global lock while holding the local.  If it fails, then it's because a
new `enter` or `run` was invoked that is in the process of negotiating
the path to the local lock with the entry point.  In this case, the
entry point releases the local lock and goes back trying to acquire the
global lock, as it did when creating the initialization stamp file (ie.,
the local lock).  If it succeeds, then no new `enter` or `run` is in the
process of starting, and the entry point can exit.

If this system of reference counting is simplified to just the global
lock, then all the entry points of all Toolbx containers will exit only
after all the `enter` and 'run' sessions across all Toolbx containers
have ended.  The local lock makes it possible to do this for each
container separately.

This system will not work without the global lock.  It will cause a few
races if a new `enter` or `run` is invoked, just as the last of the
previous batch of sessions end, letting the entry point acquire the
local lock and prepare to exit.

Sometimes, a Toolbx container's entry point is started directly with
`podman start`, without going through the `enter` or `run` commands, for
debugging.  Care was taken to detect this case by making a non-blocking
attempt to acquire the global lock from the entry point before creating
the initialization stamp file (ie., the local lock).

If it fails, then it's because an `enter` or `run` is waiting for the
container to get initialized by the entry point, and things proceed as
described above.  If it succeeds, then it's because the entry point was
started directly.  In this case, the entry point releases the global
lock, and adds a timeout after creating the initialization stamp file
before trying to acquire any other locks to give the user time to invoke
`enter` or `run`.  A timeout of 25 seconds is used, as is the default
for D-Bus method calls [1] and when waiting for the entry point to
initialize the container.

[1] https://docs.gtk.org/gio/property.DBusProxy.g-default-timeout.html

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -396,6 +396,12 @@ func initContainer(cmd *cobra.Command, args []string) error {
 	for {
 		select {
 		case <-done:
+			logrus.Debugf("Removing initialization stamp %s", initializedStamp)
+			if err := os.Remove(initializedStamp); err != nil {
+				logrus.Debugf("Removing initialization stamp %s failed: %s", initializedStamp, err)
+				return errors.New("failed to remove initialization stamp")
+			}
+
 			cause := context.Cause(waitForExitCtx)
 			logrus.Debugf("Exiting entry point: %s", cause)
 			return nil


debarshiray:
This pull request can't be merged until the work to enable Ptyxis to use `toolbox(1)` to enter Toolbx containers is finished (eg., https://github.com/containers/toolbox/issues/1675) and Ptyxis is switched over.

debarshiray:
recheck

debarshiray:
Weird.  One of the tests for the `HOSTNAME` environment variable is randomly failing:

```
fedora-rawhide | not ok 86 environment variables: HOSTNAME inside the default container
fedora-rawhide | # tags: arch-fedora runtime-environment
fedora-rawhide | # (from function `assert_success' in file test/system/libs/bats-assert/src/assert.bash, line 114,
fedora-rawhide | #  in test file test/system/220-environment-variables.bats, line 383)
fedora-rawhide | #
fedora-rawhide | # status : 1
fedora-rawhide | # output :
fedora-rawhide | # --
fedora-rawhide | #
```
```
fedora-42 | not ok 89 environment variables: HOSTNAME inside RHEL 8.10
fedora-42 | # tags: arch-fedora runtime-environment
fedora-42 | # (from function `assert_success' in file test/system/libs/bats-assert/src/assert.bash, line 114,
fedora-42 | #  in test file test/system/220-environment-variables.bats, line 416)
fedora-42 | #   `assert_success' failed
fedora-42 | #
fedora-42 | # -- command failed --
fedora-42 | # status : 1
fedora-42 | # output :
fedora-42 | # --
fedora-42 | #
```

debarshiray:
recheck

debarshiray (commented):
This algorithm seems to be working as far as I can make out.  However, we need to teach [Ptyxis](https://gitlab.gnome.org/chergert/ptyxis) to enter Toolbx containers using `toolbox(1)` commands instead of directly using Podman's `start` and `exec` commands before we can merge this.  One of the blockers for Ptyxis is tracked in issue https://github.com/containers/toolbox/issues/1675

The algorithm adds a system of reference counting based on advisory file locks between the `enter` and `run` commands on one hand, and the container's entry point on the other.  If Ptyxis tries to enter a Toolbx container without following the algorithm then it risks causing deadlocks or other undefined behaviour.  I think it's easier and better to make Ptyxis use `toolbox(1)` than to reimplement the algorithm separately in Ptyxis.

Issue #1358, opened 2023-08-25
Possibility to "rebase" a container 


bilelmoussaoui:
**Is your feature request related to a problem? Please describe.**

Currently, when updating say,  from Fedora 38 to Fedora 39. You would usually stick with using your old  container until at some point, one of your projects starts depending on a too new release of a package forcing you to create a new container. The creation bits are pretty straightforward; the annoying part is copying all you had installed and configured inside the old container and it would be pretty handy if toolbox could just do some parts of it for me, optionally.

**Describe the solution you'd like**

When running `toolbox create`, if the user has an old version of the distribution container, toolbox could ask to re-install the existing packages from the old one maybe?

**Describe alternatives you've considered**

Weird local shell scripts to make my life easier

**Additional context**
Nothing to add

JakobDev:
You can actually update the Fedora version inside the Container

debarshiray:
> You can actually update the Fedora version inside the Container

Yeah, one option is to try to upgrade the existing Fedora container in-place to a newer Fedora release, just like one would upgrade a Fedora host.

However, sometimes people don't want to do that, because of too many local changes that can either break the upgrade or they want to start from as close to a pristine state as possible.  For those cases, I can think of two options.

One, you could maintain a separate Containerfile where you layer on the extra packages that you use on top of the `registry.fedoraproject.org/fedora-toolbox` image, and then always create containers from images built from this Containerfile.

Or, you could maintain an Ansible playbook that you always run to perform post-installation customization on a container.  I think Toolbx could grow some features to make it easier to do that.

debarshiray:
I found myself thinking about this today.

@sgallagher has a script that uses `rpm --all --query ...` inside the existing container to get a list of the installed RPMs, renames the container as backup, downloads the newer image, creates a new container, and installs the list of RPMs inside it.

The script currently updates containers to newer builds of the same image, not from Fedora N to N+1, but that's not a big deal.

I suspect we might have problems dealing with third-party RPM repositories, like RPM Fusion, Google Chrome, VS Code, etc..  It would be good if we could detect this in the older container and give the user a hint to somehow (eg., manually) add those repositories into the new container.

[...]

Issue #873, opened 2021-09-10
always offer to pull latest image when creating a toolbox


juhp:
**Is your feature request related to a problem? Please describe.**
We frequently see bug reports caused by people having an old toolbox image cached locally.

**Describe the solution you'd like**
`toolbox create` should either offer to pull or always pull down the latest image being requested .
For extra points it could check if a newer image exists first.

**Additional context**
I am imagining something like this:
```
$ toolbox create
A newer fedora-toolbox:34 image is available, do you want to download it? [Y/n]
```

I think this is the least surprising behavior (though personally I would always want to pull the latest image).
A more conservative solution could be to add a `--pull` flag.

debarshiray:
> We frequently see bug reports caused by people having an
> old toolbox image cached locally.

Umm... do you have some examples of these bug reports?

I know that people struggle to switch from Rawhide to a freshly branched unstable Fedora at the branching point. Anything else?

Have we ever pushed out badly broken images for stable Fedoras? Ultimately users can always 'dnf update' their containers, can't they?

juhp:
I will have to go hunting for examples: there should be enough around.

Conversely, why would you not want to create a new toolbox with the latest image?

debarshiray:
> (Also `filesystem` can't be dnf updated.)

This one is https://github.com/containers/toolbox/issues/643, which should be fixed now.

debarshiray:
> Conversely, why would you not want to create a new toolbox with
> the latest image?

If there were no other trade-offs, yes, I would always want to use the latest image.

However, if a user keeps accumulating a large number of images, then:
* It can negatively impact disk space.
* It can clutter up the `toolbox list` output and make it difficult to get a feel for what images are locally available.

These are just trade-offs, and we need to figure out what the right balance is. That's why I was curious about examples of bugs caused by stale images.

There could be other ways of looking at this too:
* Sharing the DNF caches to reduce network traffic.
* Prompting the user that a new image is now available, so that they can manually migrate.

oturpe:
> > We frequently see bug reports caused by people having an
> > old toolbox image cached locally.
> 
> Umm... do you have some examples of these bug reports?
> 
> I know that people struggle to switch from Rawhide to a freshly branched unstable Fedora at the branching point. Anything else?
> 
> Have we ever pushed out badly broken images for stable Fedoras? Ultimately users can always 'dnf update' their containers, can't they?

1. At a time when `toolbox:36` image is Rawhide,
[...]

Issue #580, opened 2020-10-14
Remind users about software updates


allanday:
People running toolboxes without the latest security updates is a Bad Thing™. It would therefore be good to have a mechanism by which they are reminded that their toolboxes are out of date, or require updates.

It would be good to investigate potential technical solutions to this. In UX terms, we probably want a minimal and non-intrusive reminder, like:

```
$ toolbox enter fedora-32
fedora-32: 7 days since the last update check. Please run 'dnf update'.
⬢ $
```

tpopela:
One idea here would be to use the output of `dnf updateinfo list --security --available` that can produce e.g.:

```
0 $ dnf updateinfo list --security --available
Fedora 32 openh264 (From Cisco) - x86_64                                                                                                                                                                        201  B/s | 543  B     00:02
Fedora Modular 32 - x86_64                                                                                                                                                                                       15 kB/s |  22 kB     00:01
Fedora Modular 32 - x86_64 - Updates                                                                                                                                                                             84 kB/s |  19 kB     00:00
Fedora Modular 32 - x86_64 - Updates                                                                                                                                                                            903 kB/s | 1.1 MB     00:01
Fedora Modular 32 - x86_64 - Test Updates                                                                                                                                                                        16 kB/s |  18 kB     00:01
Fedora Modular 32 - x86_64 - Test Updates                                                                                                                                                                       178 kB/s | 349 kB     00:01
Fedora 32 - x86_64 - Test Updates                                                                                                                                                                                33 kB/s |  13 kB     00:00
Fedora 32 - x86_64 - Test Updates                                                                                                                                                                               2.0 MB/s | 2.9 MB     00:01
Fedora 32 - x86_64 - Updates                                                                                                                                                                                     19 kB/s |  14 kB     00:00
Fedora 32 - x86_64 - Updates                                                                                                                                                                                    3.0 MB/s |  13 MB     00:04
Fedora 32 - x86_64                                                                                                                                                                                               30 kB/s |  22 kB     00:00
RCM Tools for Fedora 32 (RPMs)                                                                                                                                                                                  0.0  B/s |   0  B     00:00
Errors during downloading metadata for repository 'rcm-tools-fedora-rpms':
  - Curl error (6): Couldn't resolve host name for https://download.devel.redhat.com/rel-eng/RCMTOOLS/latest-RCMTOOLS-2-F-32/compose/Everything/x86_64/os/repodata/repomd.xml [Could not resolve host: download.devel.redhat.com]
[...]

PR #1683, opened 2025-07-14
Add Upgrade Functionality

silverhadch:
## toolbox upgrade: Add package manager detection and upgrade capability

This PR introduces a new `toolbox upgrade` command that:

### Key Features
- Automatically detects the container's package manager (supports 11 package managers)
- Performs full system upgrades using the native package manager
- Works with both individual containers (`--container`) and all containers (`--all`)
- Provides clear output showing detected package manager and upgrade progress

### Supported Package Managers
✔️ dnf/microdnf/yum (RHEL/Fedora)  
✔️ apt (Debian/Ubuntu)  
✔️ pacman (Arch)  
✔️ zypper (openSUSE)  
✔️ apk (Alpine)  
✔️ And 5 others (xbps, emerge, slackpkg, swupd)

### Usage Examples
```sh
# Upgrade specific container
toolbox upgrade --container fedora-toolbox-38

# Upgrade all containers
toolbox upgrade --all
```

### Implementation Details
- Uses existing `runCommand` infrastructure
- Gracefully handles errors during upgrade process
- Includes comprehensive man page documentation
- Follows established toolbox patterns

This provides users with a simple, unified way to keep all their toolbox containers updated regardless of distro.

### DIFF
--- a/doc/toolbox-upgrade.1.md
+++ b/doc/toolbox-upgrade.1.md
@@ -1,0 +1,61 @@
+
+% toolbox-upgrade 1
+
+## NAME
+toolbox\-upgrade - Upgrade packages in Toolbx containers
+
+## SYNOPSIS
+**toolbox upgrade** [*--all* | *-a*] [*--container* | *-c* *CONTAINER*]
+
+## DESCRIPTION
+Upgrades packages inside one or more Toolbx containers by automatically detecting
+the package manager (dnf, apt, pacman, etc.) and running the appropriate upgrade
+command. The container should have been created using the `toolbox create` command.
+
+This command will:
+1. Detect the available package manager in the container
+2. Run the appropriate update/upgrade commands
+3. Report any errors that occur during the process
+
+## OPTIONS
+**--all, -a**
+Upgrade all Toolbx containers. Cannot be used with *--container*.
+
+**--container, -c** *CONTAINER*
+Upgrade a specific Toolbx container. Cannot be used with *--all*.
+
+## EXAMPLES
+**Upgrade packages in a specific container:**
+```
+$ toolbox upgrade --container fedora-toolbox-38
+```
+
+**Upgrade packages in all containers:**
+```
+$ toolbox upgrade --all
+```
+
+**Typical output:**
+```
+Detected package manager: dnf
+Updating metadata...
+Upgrading packages...
+Complete!
+```
+
+## NOTES
+Supported package managers:
+- dnf (Fedora, RHEL)
+- microdnf (Minimal Fedora/RHEL)
+- yum (Older RHEL/Fedora)
+- apt (Debian, Ubuntu)
+- pacman (Arch Linux)
+- xbps (Void Linux)
+- zypper (openSUSE)
+- apk (Alpine Linux)
+- emerge (Gentoo)
+- slackpkg (Slackware)
+- swupd (Clear Linux)
+
+## SEE ALSO
+`toolbox(1)`, `toolbox-create(1)`, `toolbox-list(1)`
--- a/src/cmd/upgrade.go
+++ b/src/cmd/upgrade.go
@@ -1,0 +1,156 @@
+/*
+ * Copyright © 2025 Hadi Chokr <hadichokr@icloud.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/containers/toolbox/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var (
+	upgradeAll       bool
+	upgradeContainer string
+)
+
+var upgradeCmd = &cobra.Command{
+	Use:               "upgrade",
+	Short:             "Detect package manager and upgrade packages in toolbx containers",
+	RunE:              runUpgrade,
+	ValidArgsFunction: completionContainerNamesFiltered,
+}
+
+func init() {
+	upgradeCmd.Flags().BoolVar(&upgradeAll, "all", false, "Upgrade all Toolbx containers")
+	upgradeCmd.Flags().StringVar(&upgradeContainer, "container", "", "Name of the toolbox container to upgrade")
+
+	// Register container flag completion
+	if err := upgradeCmd.RegisterFlagCompletionFunc("container", completionContainerNames); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to register flag completion function: %v\n", err)
+		os.Exit(1)
+	}
+	upgradeCmd.SetHelpFunc(upgradeHelp)
+	rootCmd.AddCommand(upgradeCmd)
+}
+
+func runUpgrade(cmd *cobra.Command, args []string) error {
+	if !upgradeAll && upgradeContainer == "" {
+		return errors.New("must specify either --all or --container")
+	}
+
+	if upgradeAll && upgradeContainer != "" {
+		return errors.New("cannot specify both --all and --container")
+	}
+
+	if upgradeAll {
+		containers, err := getContainers()
+		if err != nil {
+			return err
+		}
+
+		if len(containers) == 0 {
+			return errors.New("no Toolbx containers found")
+		}
+
+		for _, container := range containers {
+			fmt.Printf("Upgrading container: %s\n", container.Name())
+			if err := execUpgradeInContainer(container.Name()); err != nil {
+				fmt.Fprintf(os.Stderr, "Error upgrading container %s: %v\n", container.Name(), err)
+				// Continue with next container instead of returning
+			}
+		}
+		return nil
+	}
+
+	return execUpgradeInContainer(upgradeContainer)
+}
+
+// execUpgradeInContainer runs detection and upgrade inside the specified container
+func execUpgradeInContainer(container string) error {
+	pkgs := []struct {
+		detect  string
+		upgrade string
+		name    string
+	}{
+		{"command -v dnf", "sudo dnf -y upgrade", "dnf"},
+		{"command -v microdnf", "sudo microdnf upgrade -y", "microdnf"},
+		{"command -v yum", "sudo yum -y upgrade", "yum"},
+		{"command -v apt", "sudo apt update && apt upgrade -y", "apt"},
+		{"command -v pacman", "sudo pacman -Syu --noconfirm", "pacman"},
+		{"command -v xbps-install", "sudo xbps-install -Su -y", "xbps"},
+		{"command -v zypper", "sudo zypper update -y", "zypper"},
+		{"command -v apk", "sudo apk update && apk upgrade", "apk"},
+		{"command -v emerge", "sudo emerge --sync && emerge -uDN @world", "emerge"},
+		{"command -v slackpkg", "sudo slackpkg update && slackpkg upgrade-all", "slackpkg"},
+		{"command -v swupd", "sudo swupd update", "swupd"},
+	}
+
+	for _, pkg := range pkgs {
+		// Use runCommand to check if package manager exists
+		err := runCommand(container,
+				  false,  // defaultContainer
+		    "",     // image
+		    "",     // release
+		    0,      // preserveFDs
+		    []string{"sh", "-c", pkg.detect},
+		    false,  // emitEscapeSequence
+		    false,  // fallbackToBash
+		    true)   // pedantic
+
+		if err == nil {
+			fmt.Printf("Detected package manager: %s\n", pkg.name)
+
+			// Use runCommand to execute the upgrade
+			upgradeErr := runCommand(container,
+						 false,
+			    "",
+			    "",
+			    0,
+			    []string{"sh", "-c", pkg.upgrade},
+			    false,
+			    false,
+			    true)
+
+			return upgradeErr
+		}
+	}
+	return errors.New("no supported package manager found")
+}
+
+func upgradeHelp(cmd *cobra.Command, args []string) {
+	if utils.IsInsideContainer() {
+		if !utils.IsInsideToolboxContainer() {
+			fmt.Fprintf(os.Stderr, "Error: this is not a Toolbx container\n")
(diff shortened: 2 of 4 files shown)


doc/toolbox-upgrade.1.md
```diff
@@ -0,0 +1,61 @@
+
+% toolbox-upgrade 1
+
+## NAME
+toolbox\-upgrade - Upgrade packages in Toolbx containers
+
+## SYNOPSIS
+**toolbox upgrade** [*--all* | *-a*] [*--container* | *-c* *CONTAINER*]
```
debarshiray: Wouldn't it be better to be able to `toolbox upgrade CONTAINER`, without the `--container` option?  It's less typing, and the pattern is already used in the `create`, `enter`, `rm` and `rmi` commands.

We would still need the `--all` option.

src/cmd/upgrade.go
```diff
@@ -0,0 +1,156 @@
...
+		{"command -v microdnf", "sudo microdnf upgrade -y", "microdnf"},
+		{"command -v yum", "sudo yum -y upgrade", "yum"},
+		{"command -v apt", "sudo apt update && sudo apt upgrade -y", "apt"},
+		{"command -v pacman", "sudo pacman -Syu --noconfirm", "pacman"},
+		{"command -v xbps-install", "sudo xbps-install -Su -y", "xbps"},
+		{"command -v zypper", "sudo zypper update -y", "zypper"},
+		{"command -v apk", "sudo apk update && sudo apk upgrade", "apk"},
+		{"command -v emerge", "sudo emerge --sync && sudo emerge -uDN @world", "emerge"},
+		{"command -v slackpkg", "sudo slackpkg update && sudo slackpkg upgrade-all", "slackpkg"},
+		{"command -v swupd", "sudo swupd update", "swupd"},
```
debarshiray: It will be better to encode the package manager metadata in the image.  That way any image that's not built into Toolbx can easily specify its own metadata without requiring changes to the Toolbx code, a new Toolbx release, etc..

One option is to express these through the image's labels.  eg., `com.github.containers.toolbox.package-manager.update="sudo dnf --assumeyes update"`.

If the label is missing, we error out by saying that the container doesn't support updates.  With passing time, as the newer images propagate, this will happen less and less.  If the labels are there, we can directly try the command without the `command -v ...` step.

debarshiray (changes requested):
Thanks for working on this, @silverhadch !

There are a few different use-cases for handling updates to Toolbx containers.  https://github.com/containers/toolbox/issues/580 is about a minimal command line interface banner.  https://github.com/containers/toolbox/issues/1358 is about rebasing an existing container over an image for a new operating system version, which is a variant of rebasing over an image for the same OS version, and they are both related to https://github.com/containers/toolbox/issues/873

So, we need to think through how all those different use-cases will fit together.  eg., the mechanism to show a CLI banner overlaps with the changes in this pull request.  Do you have any ideas for that?  If nothing else, it can impact the name of the new command.  :)

By the way, Git commits in the feature branch for this PR needs some fixing, because there's a merge commit within the branch itself.

silverhadch:
I agree it would be better if we use metadata instead. Will work on this soon.

silverhadch:
@debarshiray 
We could define in the metadate

1. Update Command
2. Install Command
3. List packages command
4. Query Security Updates

So update subcommand updates the box while upgrade tries to rebase by checking if there is a new image then list the packages then installs them into the new image. 

For the banner we check if there are security updates and if there are inform the user.

silverhadch:
@debarshiray 
We could define in the metadata

1. Update Command
2. Query Security Updates/Last updated

So update subcommand updates the box. 

For the banner we check if there are security updates and if there are inform the user. 

The Rebasing is a diffrent usecase that i could add a rebase subcommand in the future

doc/toolbox-upgrade.1.md
silverhadch: Done

src/cmd/upgrade.go
silverhadch: Also done

silverhadch:
@debarshiray  
I applied the suggestions. Now the upgrade command reads the update command from the container’s labels and no longer requires the `--container` flag if the container name is passed as a positional argument.

So the `upgrade` subcommand updates the container as intended.

Other use cases are out of scope for this PR and will be implemented later. This PR focuses purely on the updating functionality.

**Planned future PRs:**  
- Displaying a banner to notify the user if security updates are available.  
- Adding a `rebase` subcommand to handle rebasing containers.

silverhadch:
@debarshiray 
It would be nice if you take a look again. Since its slowing down progress in KDE Kontainer since we added a lot of Code for Downstream Toolbox Features which these PRs are trying to upstream.

src/cmd/utils.go
```diff
@@ -362,6 +362,7 @@ func getUsageForCommonCommands() string {
 	builder.WriteString("create    Create a new Toolbx container\n")
 	builder.WriteString("enter     Enter an existing Toolbx container\n")
 	builder.WriteString("list      List all existing Toolbx containers and images\n")
+	builder.WriteString("upgrade   Upgrade an existing Toolbx container via their Package Manager\n")
```
debarshiray: Nitpick: I'd leave it out from this list.

It's meant to be a starting point to orient the user with the most fundamental commands.  We don't list every command here (eg., `rm`, `rmi` and `run` are absent) because the full list can be obtained via tab completion or the manual pages.

silverhadch:
@debarshiray 
I added Tests and Updated the Images.

silverhadch:
@debarshiray 
The Tests will fail till the images are updated. But hey atleast the error path is getting indirectly tested.

doc/toolbox-upgrade.1.md
debarshiray: I meant that we shouldn't have the `--container` at all, and just use the positional argument:
```ShellSession
$ toolbox upgrade CONTAINER
```

The `--container` is used by the `run` command, because the command to be run takes up the positional argument; and it's used by the `create` and `enter` commands only for backwards compatibility with old POSIX shell implementation and is no longer documented in their manual pages.

@martymichal made this change when writing the Go implementation because it's a lot more natural to type `toolbox create CONTAINER` or `toolbox enter CONTAINER`, than `toolbox create --container CONTAINER` or `toolbox enter --container CONTAINER`; and in this case it lets you specify multiple containers with `toolbox upgrade CONTAINER1 CONTAINER2`.

So, let's not introduce a `--container` option here, and let's just stick to the new style.

test/system/111-upgrade.bats
```diff
@@ -0,0 +1,54 @@
...
+  cleanup_all
+}
+
+teardown() {
+  cleanup_all
+}
+
+@test "upgrade(Arch): Upgrade Arch container" {
+  create_distro_container arch latest arch-toolbox-test
+  assert_success
```
debarshiray, line 33: Is it really necessary to use [assert_success](https://github.com/bats-core/bats-assert) here?  It's supposed to be used to check the exit code of the command invoked by the `run` helper.

When we only need to check that a command succeeded, there's [no need](https://bats-core.readthedocs.io/en/stable/writing-tests.html) to use `run` or check anything, because Bats sets `set -e` for all tests and the test will automatically fail.

src/cmd/upgrade.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func runUpgrade(cmd *cobra.Command, args []string) error {
+	// Use positional argument as container name if --container not set
+	if upgradeContainer == "" && len(args) == 1 {
+		upgradeContainer = args[0]
+	}
+
+	if !upgradeAll && upgradeContainer == "" {
+		return errors.New("must specify either --all or a container name")
```
debarshiray, line 62: We should show:
```
Error: missing argument for "upgrade"
```

... as we do for the other commands.  eg., see [src/cmd/rm.go](https://github.com/containers/toolbox/blob/main/src/cmd/rm.go)

doc/toolbox-upgrade.1.md
```diff
@@ -0,0 +1,63 @@
+
+% toolbox-upgrade(1)
+
+## NAME
+toolbox-upgrade - Upgrade packages in Toolbx containers
+
+## SYNOPSIS
+**toolbox upgrade** [*--all* | *-a*] [*--container* | *-c* *CONTAINER*] [*CONTAINER*]
```
debarshiray, line 8: Don't you want to support specifying multiple containers by name:
```ShellSession
$ toolbox upgrade CONTAINER1 CONTAINER2
```

The `rm` and `rmi` commands support that.

src/cmd/upgrade.go
```diff
@@ -0,0 +1,139 @@
...
+
+	return execUpgradeInContainer(upgradeContainer)
+}
+
+// execUpgradeInContainer runs detection and upgrade inside the specified container
+func execUpgradeInContainer(container string) error {
+
+	inspectedcontainer, err := podman.InspectContainer(container)
+	if err != nil {
+		return errors.New("Failed to inspect Metadata.")
```
debarshiray, line 96: The error should be:
```
fmt.Errorf("failed to inspect container %s", container)
```

... as is used elsewhere in the code base.

src/cmd/upgrade.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	labels := inspectedcontainer.Labels()
+	pkgline := labels["com.github.containers.toolbox.package-manager.update"]
+	if pkgline == "" {
+		return errors.New("'com.github.containers.toolbox.package-manager.update' Label not set in Containers Metadata.")
```
debarshiray, line 102: We need an error message that a user, who doesn't know about the internal implementation details, can understand and suggests how to move forward.  :)

Something like:
```
Error: container CONTAINER does not support "upgrade"
Recreate it with a newer image or file a bug.
```

src/cmd/upgrade.go
```diff
@@ -0,0 +1,139 @@
...
+	} else {
+		// Use runCommand to execute the upgrade
+		upgradeErr := runCommand(inspectedcontainer.Name(),
+			false,
+			"",
+			"",
+			0,
+			[]string{"sh", "-c", pkgline},
```
debarshiray, line 110: I was thinking about this over the past few weeks in the context of the [NVIDIAScape](https://www.wiz.io/blog/nvidia-ai-vulnerability-cve-2025-23266-nvidiascape) security issue, and it struck me that someone can trick a user into using a malicious image that contains a harmful command in its `c.gh.containers.toolbx.package-manager.update` label.

It will be wise to prompt the user with the command that will be run before running it.  The user can use the `--assumeyes` if they want to override the prompt.  Look for the `askForConfirmation` function for an existing example of this.

This makes me wonder if we should have variants of the `c.gh.containers.toolbx.package-manager.update` label that use an _assume yes_ option for the package manager and don't.  It will be bad if a `dnf --assumeyes upgrade` breaks the container without giving the user a way to opt out of it as usual.

src/cmd/upgrade.go
debarshiray: Thanks a lot!

I realized that I should have written `toolbx`, not `toolbox` for the name of the label.  So: `com.github.containers.toolbx.package-manager.update`.

I should have checked my copy-paste.  My apologies for the trouble!

Labels are annoying to migrate, so let's go with the new name of the project from the beginning.

images/arch/Containerfile
```diff
@@ -1,6 +1,7 @@
 FROM docker.io/library/archlinux:base-devel
 
 LABEL com.github.containers.toolbox="true" \
+      com.github.containers.toolbox.package-manager.update="sudo pacman -Syu --noconfirm" \
```
Foxboron, line 4: The issue is that this is not going to work that reliably. `pacman` is going to prompt for some resolutions and that will lock things up. You *could* run something along the lines of `sudo sh -c 'yes | pacman -Syu --noconfirm'` but you might end up in a situation where the keyring is sufficiently outdated for this to work.

```
sudo sh -c 'sudo pacman -Sy archlinux-keyring; yes | pacman -Su --noconfirm'
```

or similar could work but there is no guarantees as Arch Linux is not built around the assumptions that unattended upgrades are supported.

silverhadch:
@debarshiray @Foxboron 
Thanks for the Feedback once i finished migrating my System to NixOS, i will apply the suggestions.

images/arch/Containerfile
debarshiray: I see, interesting.  Thanks for that insight, @Foxboron 

In that case, my vote is to not add the label to the `arch-toolbox` image, and leave it as explicitly unsupported.  It's better to not advertise things to users that are not meant to work.

src/cmd/upgrade.go
debarshiray: Immediately after I submitted the comment, it struck me that a malicious image can already trick Toolbx into doing bad things.

Such as, the entry point of Toolbx containers rely on the `mount(8)` binary.  So, if there's an image that has replaced that binary with something else, then `toolbox enter` can lead to bad things.

We could re-implement the parts of `mount(8)` that Toolbx needs in Go to avoid this, but then someone can replace the `bash(1)` binary with something else, and so on.

So, I suppose, the security argument doesn't really hold?

Even then, maybe we should offer a prompt and show what's about to happen just for the sake of politeness?  Just like we do before downloading an image.

I am still wondering if this means we should have a prompt before the package manager is invoked, or after it's invoked but before the packages are downloaded and changed, or both.

silverhadch: Probably the best approach would be to tell the user the update command then prompt them and if the user wants via a flag skip it but by default they will be told about it.
And to be honest we should probably have a vetted compatibility list like distrobox.

debarshiray: > Probably the best approach would be to tell the user the update command then prompt them and if the user wants via a flag skip it but by default they will be told about it.

You mean prompting the user both before and after invoking the package manager?  ie., prompting the user that `dnf upgrade` will be called, then prompting the user to proceed with what `dnf upgrade` wants to do?

I suspect we could avoid the first prompt and only inform the user that `dnf upgrade` is going to be run, if we are sure that it won't generate a lot of network traffic.  Otherwise, it will be bad for users with limited Internet connectivity.

I am beginning to think that having too many prompts is annoying, and it's better than the alternative of potentially rudely surprising the users by not letting them have a say.

> And to be honest we should probably have a vetted compatibility list like distrobox.

How do you want to implement that?

The first step for declaring full support for an OS is to run the CI on hosts with that OS, which isn't a trivial task.  Do you want to skip the `upgrade` command for those that are on the compatibility list?

Conan-Kudo: I think it makes sense to have a confirmation prompt before running the package manager if the command has some kind of `--assumeyes` argument. If it doesn't, then confirmation can happen after invocation.

silverhadch:
I am sorry but due to https://github.com/containers/toolbox/issues/1706 and other compile issues (linking issues and needing to run with linker itself), I am unable to continue working on this. If you really need it you could take the code and do the PR yourself. In the mean time you can use distrobox with toolbx images.

PR #1688, opened 2025-07-22
Add temporary workaround for missing subuid/subgid in Rawhide

calancha:
Some CI jobs on Fedora Rawhide fail with Podman rootless builds due to missing subuid/subgid ranges for the zuul-worker user:

    Error: cannot find mappings for user zuul-worker in /etc/subuid

Fedora 42 and other distros already ship default ranges (e.g., zuul-worker:524288:65536), but Rawhide seems to lack them.

Add pre-tasks to setup-env.yaml to:

- Dump current /etc/subuid and /etc/subgid contents for debug
- Ensure the required ranges exist without overwriting existing configs

Once the Rawhide base image is fixed, this can be reverted.


playbooks/dependencies-fedora.yaml
```diff
@@ -39,6 +39,40 @@
       - udisks2
     use: "{{ 'dnf' if zuul.attempts > 1 else 'auto' }}"
 
+- name: Dump current /etc/subuid contents
+  become: true
+  command: cat /etc/subuid
```
debarshiray: Nitpick: do we really need to run `cat /etc/subuid` as root?  The file is world-readable.

debarshiray (commented):
Thanks for the quick workaround, @calancha !  I would have had to spend some time expanding my Ansible skills to do this myself.  :)

This error seems unrelated, and is probably due to some change in Fedora Rawhide:
```
fedora-rawhide | #  in test file test/system/104-run.bats, line 814)
fedora-rawhide | #   `assert_line --index 1 "bash: line 1: exec: /etc: cannot execute: Is a directory"' failed
fedora-rawhide | # ~ /home/zuul-worker/src/github.com/containers/toolbox
fedora-rawhide | #
fedora-rawhide | # index    : 1
fedora-rawhide | # expected : bash: line 1: exec: /etc: cannot execute: Is a directory
fedora-rawhide | # actual   : Error: failed to invoke command /etc in container fedora-toolbox-43
fedora-rawhide | # --
```

playbooks/dependencies-fedora.yaml
calancha: You're right, running as root isn't necessary. 😳
The goal was just to show that it's empty. I'll drop the `become: true`

Luap99:
FYI this may not be temporary, seems like useradd no longer assign a subuid/gid range by default: https://bugzilla.redhat.com/show_bug.cgi?id=2382662

debarshiray:
> FYI this may not be temporary, seems like useradd no longer assign a subuid/gid range by default: https://bugzilla.redhat.com/show_bug.cgi?id=2382662

Thanks for the reference, @Luap99 !  It saved me a trip down the rabbit hole.  :)

playbooks/dependencies-fedora.yaml
```diff
@@ -39,6 +39,22 @@
       - udisks2
     use: "{{ 'dnf' if zuul.attempts > 1 else 'auto' }}"
 
+- name: Ensure subuid/subgid ranges exist for zuul-worker (workaround for Rawhide CI issue)
+  become: true
+  lineinfile:
+    path: "{{ item.file }}"
+    regexp: "^zuul-worker:"
+    line: "zuul-worker:524288:65536"
+    create: yes
```
debarshiray: Nitpick: maybe `true` instead of `yes`?  That's what the [documentation](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html) uses, and you have used elsewhere.  :)

debarshiray: I took the liberty to update the branch.

debarshiray:
I updated the commit message with references to the `useradd(8)` change and discussions, to clarify what's going on.

debarshiray:
Thank you for your contribution, @calancha !

PR #1689, opened 2025-07-22
WIP: Refresh Podman user namespace mappings on Rawhide

calancha:
Some CI jobs on Fedora Rawhide failed with rootless Podman builds due to stale or missing UID/GID mappings, even after installing shadow-utils-subid-devel:

    Error: cannot find mappings for user zuul-worker in /etc/subuid

Force a refresh of Podman’s user namespace mappings with `podman system migrate`, which re-reads /etc/subuid and /etc/subgid and applies default ranges when available.

Once the Rawhide base image includes proper mappings by default, this workaround can be removed.

### DIFF
--- a/playbooks/dependencies-fedora.yaml
+++ b/playbooks/dependencies-fedora.yaml
@@ -39,6 +39,11 @@
       - udisks2
     use: "{{ 'dnf' if zuul.attempts > 1 else 'auto' }}"
 
+- name: Force Podman to refresh user namespace mappings
+  become: true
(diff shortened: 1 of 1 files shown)


debarshiray (commented):
Thanks for looking into this, @calancha !  Is this pull request still needed?  https://github.com/containers/toolbox/pull/1688 looks more complete.

calancha:
Closing: We need the complete fix from https://github.com/containers/toolbox/pull/1688

PR #1693, opened 2025-08-01
toolbox.sh: use new TOOLBOX_DO_NOT_OVERRIDE_PS1

juhp:
This allows users to set `TOOLBOX_DO_NOT_OVERRIDE_PS1` in their shell configuration
to prevent toolbox from overriding `PS1`.

Addresses #1357

### DIFF
--- a/profile.d/toolbox.sh
+++ b/profile.d/toolbox.sh
@@ -48,8 +48,10 @@ fi
 
 if [ -f /run/.containerenv ] \
    && [ -f /run/.toolboxenv ]; then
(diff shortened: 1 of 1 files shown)


juhp:
(BTW this could be reverted once toolbox no longer overwrites PS1, 
but in the short-term this PR would help to bridge the gap.)

I don't understand the CI failure for [system-test-fedora-rawhide-commands-options](https://softwarefactory-project.io/zuul/t/local/build/13891a8a94804bf5a5fe92a60c3a8c28) - seems to me that should be unrelated to this change.

Issue #1674, opened 2025-07-02
Failure to load CA certificates when SSH'd into container


ptomato:
**Describe the bug**
Similar to #1665 but without interference from homebrew, and occurs when SSH'ing into the container.

**Steps how to reproduce the behaviour**
1. `podman pull registry.fedoraproject.org/fedora-toolbox:42`
2. `toolbox create test`
3. `toolbox enter test`
4. `sudo dnf -y install gnutls-utils openssh-server`
5. `sudo bash -c 'printf "Port 2233\nListenAddress localhost\nPermitEmptyPasswords yes\n" >> /etc/ssh/sshd_config'`
6. `sudo /usr/libexec/openssh/sshd-keygen ed25519`
7. `sudo /usr/sbin/sshd -D`
8. Switch to a host terminal
9. `ssh -p 2233 localhost`
10. `gnutls-cli gnome.org`
 
**Expected behaviour**
I would expect the same thing to happen as if I had done `toolbox enter test` in step 8 instead of `ssh`:
```
$ gnutls-cli gnome.org
Processed 376 CA certificate(s).
Resolving 'gnome.org:443'...
Connecting to '3.211.23.167:443'...
- Certificate type: X.509
- Got a certificate list of 2 certificates.
- Certificate[0] info:
 - subject `CN=gnome.org', issuer `CN=R11,O=Let's Encrypt,C=US', serial 0x0515a41201c750b2745c23fc968ddd77b73a, RSA key 2048 bits, signed using RSA-SHA256, activated `2025-06-11 11:17:01 UTC', expires `2025-09-09 11:17:00 UTC', pin-sha256="e/CiYqk51GX71LsXS+AnCW/BqFrf8s+vMOoeGkTixEQ="
	Public Key ID:
		sha1:52ea8b93c31ee191a8b7b5f03acba8729bf8106b
		sha256:7bf0a262a939d465fbd4bb174be027096fc1a85adff2cfaf30ea1e1a44e2c444
	Public Key PIN:
		pin-sha256:e/CiYqk51GX71LsXS+AnCW/BqFrf8s+vMOoeGkTixEQ=

- Certificate[1] info:
[...]

PR #1695, opened 2025-08-06
cmd/initContainer: Unbreak access to CA certificates in sshd(8) sessions

debarshiray:
When a Toolbx container is set up to use the `p11-kit-client.so` PKCS #&NoBreak;11
module instead of the usual `p11-kit-trust.so` module, the
`P11_KIT_SERVER_ADDRESS` environment variable must be set inside the
container, so that it can communicate with the host operating system.

Currently, this works as described above with the `enter` and `run`
commands, but not within child sessions started by an `sshd(8)` [1]
instance running inside a container, because `P11_KIT_SERVER_ADDRESS` is
absent.

To make this work, `sshd(8)` [1] must be configured [2] to set
`P11_KIT_SERVER_ADDRESS` in its child sessions.

If `sshd(8)` uses the `/etc/ssh/sshd_config.d` directory for configuration,
then the entry point will automatically do this from now on.  This
requires at least OpenSSH 8.2, which added support for the `Include`
directive in `sshd_config(5)` [2,3], and the directive must be used to
include the configuration from `/etc/ssh/sshd_config.d`.

Otherwise, the user will have to do it themself.  eg., Ubuntu 16.04
Xenial Xerus and 18.04 Bionic Beaver don't use `/etc/ssh/sshd_config.d`
because their OpenSSH is too old [4,5].

Note that the permissions of the `/etc/ssh/sshd_config.d` directory and
its contents differ across operating system distributions.  OSes within
the Fedora family use `0700` for the directory and `0600` for its contents.
Arch Linux and Ubuntu use `0755` and `0644`.  The entry point tries to
follow the permissions used by the distribution.

Fallout from 5ed2442214f0eb8d58d1bfc7df50f62328815cb2

[1] https://man7.org/linux/man-pages/man8/sshd.8.html

[2] https://man7.org/linux/man-pages/man5/sshd_config.5.html

[3] OpenSSH commit c2bd7f74b0e0f3a3
    https://github.com/openssh/openssh-portable/commit/c2bd7f74b0e0f3a3
    https://bugzilla.mindrot.org/show_bug.cgi?id=2468

[4] https://code.launchpad.net/~git-ubuntu-import/ubuntu/+source/openssh/+git/openssh/+ref/ubuntu/xenial-updates

[5] https://code.launchpad.net/~git-ubuntu-import/ubuntu/+source/openssh/+git/openssh/+ref/ubuntu/bionic-updates

https://github.com/containers/toolbox/issues/1674

### DIFF
--- a/src/cmd/initContainer.go
+++ b/src/cmd/initContainer.go
@@ -600,10 +600,11 @@ func configurePKCS11(targetUser *user.User) error {
 		}
 	}
 
-	if path, err := utils.GetP11KitServerSocket(targetUser); err != nil {
+	serverSocket, err := utils.GetP11KitServerSocket(targetUser)
+	if err != nil {
 		return err
-	} else if !utils.PathExists(path) {
-		logrus.Debugf("%s: socket %s not found", logPrefix, path)
+	} else if !utils.PathExists(serverSocket) {
+		logrus.Debugf("%s: socket %s not found", logPrefix, serverSocket)
 		logrus.Debugf("%s: skipping", logPrefix)
 		return nil
 	}
@@ -620,6 +621,10 @@ func configurePKCS11(targetUser *user.User) error {
 		pkcs11ConfigBytes,
 		0644); err != nil {
 		return fmt.Errorf("failed to configure PKCS #11 to read from the host: %w", err)
+	}
+
+	if err := configureSSHD(serverSocket); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


debarshiray:
> Build failed. https://softwarefactory-project.io/zuul/t/local/buildset/1fc2126b49fe48b1b30806799a9f44ed
> 
> ❌ [system-test-fedora-rawhide-commands-options ](https://softwarefactory-project.io/zuul/t/local/build/c1dbba6fbcee4040b6f8c10d54a1b0e0) **FAILURE** in 54m 42s

The error is the same as we have seen [recently](https://github.com/containers/toolbox/pull/1688#pullrequestreview-3045061238) on Fedora Rawhide:
```
fedora-rawhide | not ok 186 run: Try /etc as a command
fedora-rawhide | #  in test file test/system/104-run.bats, line 814)
fedora-rawhide | #   `assert_line --index 1 "bash: line 1: exec: /etc: cannot execute: Is a directory"' failed
fedora-rawhide | # ~ /home/zuul-worker/src/github.com/containers/toolbox
fedora-rawhide | #
fedora-rawhide | # index    : 1
fedora-rawhide | # expected : bash: line 1: exec: /etc: cannot execute: Is a directory
fedora-rawhide | # actual   : Error: failed to invoke command /etc in container fedora-toolbox-43
fedora-rawhide | # --
```

It seems unrelated, and is probably due to some change in Fedora Rawhide.

PR #1696, opened 2025-08-06
utils: preserve more KDE-related environment variables

bkoropoff:
I noticed that Qt applications run in my toolbox container on Fedora Kinoite 42 did not respect KDE's theme or display scaling settings.  It turns out this was due to the `KDE_SESSION_VERSION` environment variable missing.  I went through all environment variables which KDE and konsole set in my session which weren't preserved and added them to the list for completeness.  Let me know if you want some removed or split into separate commits.


debarshiray (commented):
Thanks for the patch, @bkoropoff !

I didn't have time to go through all the environment variables.  However, `KDE_FULL_SESSION` and `KDE_SESSION_VERSION` look obviously correct, based on your report and the KDE [documentation](https://userbase.kde.org/KDE_System_Administration/Environment_Variables).  Therefore, I split them into https://github.com/containers/toolbox/pull/1698 to be able to include them in the new Toolbx release that I am preparing, so that at least some of your problems are addressed soon.

Hey @grulja !  Do you have any comments on preserving the values of these environment variables inside Toolbx containers?

Sometimes I think that we should change Toolbx's current approach of preserving only some environment variables, particularly those that are not set by the user in their shell start-up files, to one where it preserves all variables.

debarshiray (commented):
Feel free to add [tests](https://github.com/containers/toolbox/blob/main/test/system/220-environment-variables.bats) for any of these environment variables that you think can be useful.

grulja:
For the record, I have the same setup (Kinoite 42) and I don't have any issue with Qt applications not being properly scaled or themed when running inside Toolbox. I really wonder whether `KDE_SESSION_VERSION` has any impact on this, I would rather say it's used so others can find out whether you run KDE Plasma 5 or KDE Plasma 6. Still, it won't definitely hurt preserving this env variable, same for `KDE_FULL_SESSION`.

bkoropoff:
I definitely see different behavior, e.g.:

`toolbox run qdbusviewer-qt6`:
image

`toolbox run env KDE_SESSION_VERSION=6 qdbusviewer-qt6`:
image

`qdbusviewer` seems to pick up scaling regardless, but `audacious` doesn't without the environment variable set.  This might have something to do with `audacious` running through XWayland.

bkoropoff:
If toolbox continues to preserve only some environment variables, it might be good to make the list configurable through files in XDG_CONFIG_HOME and XDG_CONFIG_DIRS (thus picking up system defaults in `/etx/xdg`).  This would allow both users and distributions/packages to add variables to the list.

debarshiray (commented):
I found myself waiting for CI to complete, so I decided to poke at some more of these environment variables.

I included `KONSOLE_DBUS_SESSION`, `KONSOLE_DBUS_WINDOW` and `KONSOLE_PROFILE_NAME` in https://github.com/containers/toolbox/pull/1698 because they are mentioned in the Konsole [documentation](https://userbase.kde.org/Konsole/en) and [Neovim](https://github.com/neovim/neovim) uses them.

debarshiray:
> If toolbox continues to preserve only some environment variables, it might be good to make the list configurable through files in XDG_CONFIG_HOME and XDG_CONFIG_DIRS (thus picking up system defaults in `/etx/xdg`). This would allow both users and distributions/packages to add variables to the list.

Toolbx has a [configuration file](https://github.com/containers/toolbox/blob/main/doc/toolbox.conf.5.md).  We can add a `preserved-environment-variables` option to it.  Do you want to submit a pull request implementing it?  :)

debarshiray:
> For the record, I have the same setup (Kinoite 42) and I don't have any issue with Qt applications not being properly scaled or themed when running inside Toolbox. I really wonder whether `KDE_SESSION_VERSION` has any impact on this, I would rather say it's used so others can find out whether you run KDE Plasma 5 or KDE Plasma 6. Still, it won't definitely hurt preserving this env variable, same for `KDE_FULL_SESSION`.

Thanks a lot for taking a look, @grulja !

grulja:
> `qdbusviewer` seems to pick up scaling regardless, but `audacious` doesn't without the environment variable set. This might have something to do with `audacious` running through XWayland.

Trying `qdbusviewer-qt6` with and without the env variable set produces same styling and everything.  Same goes for `qdbusviewer-qt6 -platform xcb` to enforce XWayland.

Also `audacious --qt` works for me in sense I don't see a difference between using and not using the `KDE_SESSION_VERSION=6`. I don't know why it runs with xcb backend (XWayland), but you can enforce it to run natively on Wayland with `QT_QPA_PLATFORM=wayland`.

debarshiray:
Fly-by comment: @bkoropoff do you want to pursue the rest of the variables, or are you satisfied with what's in Toolbx 0.2?

bkoropoff:
My immediate issue is resolved, but I might revisit this if I find something else that doesn't work

debarshiray:
> My immediate issue is resolved, but I might revisit this if I find something else that doesn't work

Thanks for the update, @bkoropoff !  I am glad that it got sorted out for you.

I am closing this pull request for the time being to keep the list of pending PRs under control.  Please feel free to open a new one if you need more environment variables to be preserved.

Thanks for your contribution!

Issue #1684, opened 2025-07-14
Deprecate containers that need the org.freedesktop.Flatpak D-Bus service


debarshiray:
Starting from Toolbx [0.0.97](https://github.com/containers/toolbox/releases/tag/0.0.97), released in November 2020, Toolbx containers no longer need the `org.freedesktop.Flatpak.SessionHelper` D-Bus interface provided by the `org.freedesktop.Flatpak` D-Bus service.  This change was made to [support](https://github.com/containers/toolbox/issues/267) rootful uses of Toolbx with `sudo toolbox ...`.

It is going to be five years in a few months, since this change was made, and it's becoming harder and harder to practically test backwards compatibility with these containers.  All the host operating systems where Toolbx is regularly tested offer newer versions:
* Arch Linux and Fedora offer the latest upstream releases
* RHEL 8.10 offers at least 0.0.99.4
* [Ubuntu](https://packages.ubuntu.com/search?keywords=podman-toolbox&searchon=names&suite=all&section=all) offers at least 0.0.99.2

Among those OSes that we might start testing on in the near future, even [Debian](https://packages.debian.org/search?keywords=podman-toolbox&searchon=names&suite=stable&section=all) offers at least 0.0.99.3.

Therefore, this is a good time to mark these older containers as deprecated, as we have [done](https://github.com/containers/toolbox/commit/9dc52814301c5a3f2e283401fcfae58bafd8c9ae) in the past in similar cases.  After a long enough deprecation window, we can finally stop supporting them.

Containers that use `org.freedesktop.Flatpak.SessionHelper` have a directory called `/run/host/monitor` that's a bind mount from the host operating system.  The host counterpart of this directory is maintained by `org.freedesktop.Flatpak`.  The D-Bus service needs to be started before starting these containers with `podman start`.

[Commit 71b5c8c0a235249b](https://github.com/containers/toolbox/commit/71b5c8c0a235249b4dca1ab0997b29525e819a8a) changed how the containers are created without depending on `org.freedesktop.Flatpak.SessionHelper` and [commit 82c32bea742621a3](https://github.com/containers/toolbox/commit/82c32bea742621a3dc744a6062632da39e07820f) stopped starting the D-Bus service for containers that don't need it.

debarshiray:
Closed by https://github.com/containers/toolbox/pull/1697

PR #1697, opened 2025-08-06
Deprecation message when using 0.0.96 and older Toolbx container + tests

DaliborKr:
Added deprecation message when running a container that was created with Toolbx 0.0.96 and older, as described in #1684

### DIFF
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -256,7 +256,7 @@ func runCommand(container string,
 	if entryPoint != "toolbox" {
 		var builder strings.Builder
 		fmt.Fprintf(&builder, "container %s is too old and no longer supported\n", container)
-		fmt.Fprintf(&builder, "Recreate it with Toolbx version 0.0.17 or newer.")
+		fmt.Fprintf(&builder, "Recreate it with Toolbx version 0.0.97 or newer.")
 
 		errMsg := builder.String()
 		return errors.New(errMsg)
@@ -528,6 +528,9 @@ func callFlatpakSessionHelper(container podman.Container) error {
 	if !needsFlatpakSessionHelper {
 		return nil
 	}
+
+	fmt.Fprintf(os.Stderr, "Warning: container %s uses deprecated features\n", name)
+	fmt.Fprintf(os.Stderr, "Consider recreating it with Toolbx version 0.0.97 or newer.\n")
 
 	if _, err := utils.CallFlatpakSessionHelper(); err != nil {
 		return err
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -316,6 +316,21 @@ teardown() {
   assert [ ${#stderr_lines[@]} -eq 0 ]
 }
 
+@test "run: Ensure that containers using o.fd.Flatpak.SessionHelper are deprecated" {
+  local container="deprecated"
+
+  create_container_flatpak_session_helper "$container"
+
+  run --keep-empty-lines --separate-stderr "$TOOLBX" run --container "$container" true
+
+  assert_success
+  assert [ ${#lines[@]} -eq 0 ]
+  lines=("${stderr_lines[@]}")
+  assert_line --index 0 "Warning: container $container uses deprecated features"
+  assert_line --index 1 "Consider recreating it with Toolbx version 0.0.97 or newer."
+  assert [ ${#stderr_lines[@]} -eq 2 ]
+}
+
 @test "run: Try the non-existent default container with none other present" {
   local default_container_name
   default_container_name="$(get_system_id)-toolbox-$(get_system_version)"
@@ -860,6 +875,6 @@ teardown() {
   assert [ ${#lines[@]} -eq 0 ]
   lines=("${stderr_lines[@]}")
   assert_line --index 0 "Error: container $container is too old and no longer supported"
-  assert_line --index 1 "Recreate it with Toolbx version 0.0.17 or newer."
+  assert_line --index 1 "Recreate it with Toolbx version 0.0.97 or newer."
   assert [ ${#stderr_lines[@]} -eq 2 ]
 }
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -577,3 +577,73 @@ function is_fedora_rawhide() (
 
   return 0
 )
+
+
+# Creates a container with org.freedesktop.Flatpak.SessionHelper D-Bus interface
+#
+# Pulling of an image is taken care of by the function
+#
+# Parameters:
+# ===========
+# - container - name of the container
+function create_container_flatpak_session_helper() (
+  local container="$1"
+
+  local default_image
+  default_image="$(get_default_image)"
+
+  pull_default_image
+
+  if ! gdbus call \
+         --session \
+         --dest org.freedesktop.Flatpak \
+         --object-path /org/freedesktop/Flatpak/SessionHelper \
+         --method org.freedesktop.Flatpak.SessionHelper.RequestSession; then
+    echo "failed to call org.freedesktop.Flatpak.SessionHelper.RequestSession" >&2
+    return 1
+  fi
+
+  local user_id_real
+  if ! user_id_real="$(id --real --user)"; then
+    echo "failed to get the real user ID" >&2
+    return 1
+  fi
+
+  if ! podman create \
+         --dns none \
+         --env TOOLBOX_PATH="$TOOLBX" \
+         --hostname toolbox \
+         --ipc host \
+         --label com.github.containers.toolbox=true \
+         --label com.github.debarshiray.toolbox=true \
+         --name "$container" \
+         --network host \
+         --no-hosts \
+         --pid host \
+         --privileged \
+         --security-opt label=disable \
+         --userns keep-id \
+         --user root:root \
+         --volume "$HOME":"$HOME":rslave \
+         --volume "$TOOLBX":/usr/bin/toolbox \
+         --volume "$XDG_RUNTIME_DIR":"$XDG_RUNTIME_DIR" \
+         --volume "$XDG_RUNTIME_DIR/.flatpak-helper/monitor":/run/host/monitor \
+         --volume /dev:/dev:rslave \
+         --volume /etc:/run/host/etc \
+         --volume /run:/run/host/run:rslave \
+         --volume /tmp:/run/host/tmp:rslave \
+         --volume /usr:/run/host/usr:rslave \
+         --volume /var:/run/host/var:rslave \
+         "$default_image" \
+         toolbox init-container \
+           --home "$HOME" \
+           --monitor-host \
+           --shell "$SHELL" \
+           --uid "$user_id_real" \
+           --user "$USER"; then
+    echo "failed to create container $container" >&2
+    return 1
+  fi
+
+  return 0
+)


test/system/libs/helpers.bash
```diff
@@ -577,3 +577,86 @@ function is_fedora_rawhide() (
...
+#Returns the fully-resolved (symlink-free) absolute path of the user's home directory
+get_home_dir_evaled_path(){
+    if ! home_dir_evaled_path=$(readlink -f "$HOME" 2>/dev/null); then
+        echo "failed to resolve symlinks of $HOME directory." >&2
+        return 1
+    fi
+
+    echo "$home_dir_evaled_path"
+    return 0
+}
```
debarshiray: Git and GitHub are complaining that there is no newline at the end the file.  eg., `git show ...` says:
```
\ No newline at end of file
```

You can hit ENTER so that the file ends with an empty line.

test/system/104-run.bats
```diff
@@ -857,3 +857,36 @@ teardown() {
...
+
+@test "run: Ensure that label 'com.github.containers.toolbox=true' is present in toolbx container" {
+  local container_name="toolbox-container"
+
+  create_container "$container_name"
+
+  run podman inspect --format '{{ index .Config.Labels "com.github.containers.toolbox" }}' "$container_name"
+
+  assert_success
+  assert_output --regexp "true"
```
debarshiray: Nitpick: there is a trailing whitespace here.

test/system/104-run.bats
```diff
@@ -857,3 +857,36 @@ teardown() {
...
+  assert_output --regexp "true" 
+
+  run "$TOOLBX" run --container "$container_name" true
+  assert [ ${#lines[@]} -eq 0 ]
+  assert_success
+}
+
+@test "run: Ensure that the warning message appears when using deprecated container (created before version 0.0.97)" {
+  local container_name="toolbox-container-deprecated"
+
```
debarshiray: Nitpick: there is a trailing whitespace here.

test/system/104-run.bats
```diff
@@ -857,3 +857,36 @@ teardown() {
...
+@test "run: Ensure that the warning message appears when using deprecated container (created before version 0.0.97)" {
+  local container_name="toolbox-container-deprecated"
+  
+  create_container_flatpak_dbus "$container_name"
+
+  in_container_command="true"
+
+  run --keep-empty-lines --separate-stderr "$TOOLBX" run --container "$container_name" "$in_container_command"
+
```
debarshiray: Nitpick: there is a trailing whitespace here.

test/system/104-run.bats
```diff
@@ -863,3 +863,36 @@ teardown() {
   assert_line --index 1 "Recreate it with Toolbx version 0.0.17 or newer."
   assert [ ${#stderr_lines[@]} -eq 2 ]
 }
+
+@test "run: Ensure that label 'com.github.containers.toolbox=true' is present in toolbx container" {
```
debarshiray: Some relatively minor points:

We should split this new test into a separate commit, because it's unrelated to the deprecation of old containers that use `org.freedesktop.Flatpak.SessionHelper`, as described in https://github.com/containers/toolbox/issues/1684

This is a test for the `create` command because it's checking if it added the `com.github.containers.toolbox` label to the container it created.  So, it should go in `test/system/101-create.bats`.

The new thing that this test adds is the check for the label in the containers created by `create`.  So, it can be merged with the existing tests for `create` by adding the `podman inspect` to them.  We already use `podman ps --all` to check if the container was actually created, and this can be the next assertion.

test/system/104-run.bats
```diff
@@ -863,3 +863,36 @@ teardown() {
...
+  local container_name="toolbox-container"
+
+  create_container "$container_name"
+
+  run podman inspect --format '{{ index .Config.Labels "com.github.containers.toolbox" }}' "$container_name"
```
debarshiray: Cool!  I didn't know that we can use `index` with a JSON dictionary.  It would have made my life easier in the past.  :)

test/system/104-run.bats
```diff
@@ -863,3 +863,36 @@ teardown() {
...
+
+  assert_success
+  assert_output --regexp "true"
```
debarshiray: We shouldn't use the `--regexp` option here, because we aren't using any regular expressions, and there's no reason to expect anything other than `true` in the output.  So, just:
```
  assert_output "true"
```

test/system/104-run.bats
debarshiray: I took the liberty to address these in https://github.com/containers/toolbox/pull/1716 because they were relatively minor issues.

test/system/104-run.bats
debarshiray: Done in https://github.com/containers/toolbox/pull/1716

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,86 @@ function is_fedora_rawhide() (
...
+#
+# Pulling of an image is taken care of by the function
+#
+# Parameters:
+# ===========
+# - container_name - name of the container
+function create_container_flatpak_dbus() (
+  local container_name
+
+  container_name="$1"
```
debarshiray: Nitpick: these two lines can be combined.

I know there are some places in the code base where the lines are split, but it is an anti-pattern caused by a misunderstanding.  It's because ShellCheck [complains](https://www.shellcheck.net/wiki/SC2155) about cases where [command substitution](https://tldp.org/LDP/abs/html/commandsub.html) is used.

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,86 @@ function is_fedora_rawhide() (
...
+  flatpak_helper_monitor_path=$(get_flatpak_helper_monitor_path) \
+    || fail "Error getting Flatpak Helper monitor path"
+
+  home_dir_evaled=$(get_home_dir_evaled_path) \
+    || fail "Error getting Home directory evaled path"
+
+  pull_default_image
+  distro="$(get_system_id)"
+  version="$(get_system_version)"
+  image_full="${IMAGES[$distro]}:$version"
```
debarshiray: There's already a `get_default_image` function that does this.  :)

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,86 @@ function is_fedora_rawhide() (
...
+
+    echo "$output" | grep -o "/.*/monitor"
+    return 0
+}
+
+
+#Returns the fully-resolved (symlink-free) absolute path of the user's home directory
+get_home_dir_evaled_path(){
+    if ! home_dir_evaled_path=$(readlink -f "$HOME" 2>/dev/null); then
+        echo "failed to resolve symlinks of $HOME directory." >&2
```
debarshiray: I forgot to mention that starting from commit [4251a0409ffcb913](https://github.com/containers/toolbox/commit/4251a0409ffcb9131dcf150104ae5fe91237cc18) the test suite overrides the user's `HOME` environment variable with a path that's private to the tests under `/tmp` or the value of `TMPDIR`.  This was done to isolate the user's real home directory from whatever the tests might be doing to it.

This means that there's no need to resolve the symbolic links that we see on Fedora Silverblue and other image-based OSes, because they are not present in this private path, unless the tests are run in an unusual environment.

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,86 @@ function is_fedora_rawhide() (
...
+    if ! output=$(gdbus call \
+                    --session \
+                    --dest org.freedesktop.Flatpak \
+                    --object-path /org/freedesktop/Flatpak/SessionHelper \
+                    --method org.freedesktop.Flatpak.SessionHelper.RequestSession); then
+        echo "failed to call org.freedesktop.Flatpak.SessionHelper.RequestSession" >&2
+        return 1
+    fi
+
+    echo "$output" | grep -o "/.*/monitor"
```
debarshiray: Interesting.  I didn't know about the `-o` option in `grep(1)`.  Looks very useful.  :)

src/cmd/run.go
```diff
@@ -529,6 +529,9 @@ func callFlatpakSessionHelper(container podman.Container) error {
 		return nil
 	}
 
+	fmt.Fprintf(os.Stderr, "Warning: container %s uses deprecated features\n", container.Name())
+	fmt.Fprintf(os.Stderr, "Consider recreating it with Toolbox version 0.0.97 or newer.\n")
```
debarshiray: We should also update the recommended version a few lines above from 0.0.17 to 0.0.97, because containers created with 0.0.17 to 0.0.96 are now deprecated.

test/system/libs/helpers.bash
debarshiray: I tried to replace this by directly using `HOME`.  Let's how the CI responds.

If the CI succeeds, then it will be better to use the simpler option, because you don't want to end up debugging and testing the tests too much.  :)

debarshiray: Okay, so the CI passed.

test/system/libs/helpers.bash
debarshiray: I see that the output from the D-Bus method is:
```
({'path': <'/run/user/1000/.flatpak-helper/monitor'>, 'pkcs11-socket': <'/run/user/1000/.flatpak-helper/pkcs11-flatpak-39343'>},)
```

The problem is that if we parse the path by looking for `"/.*/monitor"`, then we are assuming that the path will always end with `/monitor`.  This defeats the purpose of querying the path from the D-Bus service, instead of assuming it and hard coding it.

Interestingly, the last time the path [changed](https://github.com/flatpak/flatpak/commit/b4bb890516888fdeb3468bafa659997c400ce916) was in June 2018.  Before that it used to be `$XDG_RUNTIME_DIR/flatpak-monitor`, and that value wouldn't fit `"/.*/monitor"`.

I don't know what would be the best or easiest way to parse the path from our test suite.  It's in the GVariant [format](https://docs.gtk.org/glib/gvariant-text-format.html) and there are various ways to encode and decode it.  The Go implementation of Toolbx uses [github.com/godbus/dbus/v5](https://pkg.go.dev/github.com/godbus/dbus/v5) to parse it in [CallFlatpakSessionHelper()](https://github.com/containers/toolbox/blob/main/src/pkg/utils/utils.go).

Until we figure out a _correct_ way to parse it, let's just assume that it will be `$XDG_RUNTIME_DIR/.flatpak-helper/monitor` and hard code it.  It used to be hard coded in the POSIX shell implementation and it hasn't changed in the last seven years.  It's better to hard code, than have some extra code that misleadingly looks like it's doing it correctly, when it's not.  :)

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,50 @@ function is_fedora_rawhide() (
...
+          "$default_image" \
+          toolbox init-container \
+                  --home "$HOME" \
+                  --shell "$SHELL" \
+                  --uid "$(id -ru)" \
+                  --user "$USER"
+    )
+
+    # Print the full command
+    # printf "%q " "${podman_cmd[@]}" >> /tmp/toolbox_podman_command.log
```
debarshiray: We can use `bats --trace ...` for this.  It will show all the commands that get executed.

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,56 @@ function is_fedora_rawhide() (
...
+    --dest org.freedesktop.Flatpak \
+    --object-path /org/freedesktop/Flatpak/SessionHelper \
+    --method org.freedesktop.Flatpak.SessionHelper.RequestSession
+
+  podman_cmd=(
+      podman create \
+          --dns none \
+          --env TOOLBOX_PATH="$TOOLBX" \
+          --name "$container_name" \
+          --network host \
```
debarshiray: It might be good to add the labels, because you have been working on checking the labels, and it will let us add another test to `test/system/102-list.bats` to ensure that these deprecated containers are listed.

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,56 @@ function is_fedora_rawhide() (
...
+          --volume "$HOME":"$HOME":rslave \
+          --volume "$TOOLBX":/usr/bin/toolbox \
+          --volume "$XDG_RUNTIME_DIR":"$XDG_RUNTIME_DIR" \
+          --volume "$XDG_RUNTIME_DIR/.flatpak-helper/monitor":/run/host/monitor \
+          --volume /etc:/run/host/etc \
+          --volume /usr:/run/host/usr:rslave \
+          "$default_image" \
+          toolbox init-container \
+                  --home "$HOME" \
+                  --shell "$SHELL" \
```
debarshiray: Let's add the `--monitor-host` option to `init-container` here.

We [deprecated --monitor-host](https://github.com/containers/toolbox/commit/58638c594004adb8a093ca1c774c082a37011282) and stopped using it from Toolbx 0.0.99.5.  It means that containers that were created by Toolbx < 0.0.97 to use `org.freedesktop.Flatpak.SessionHelper` still have the option, and `init-container` should continue to handle it.  It is hard or rare to test deprecated code paths, so using it in this test is a good opportunity to test it.

test/system/libs/helpers.bash
```diff
@@ -577,3 +577,56 @@ function is_fedora_rawhide() (
...
+                  --uid "$(id -ru)" \
+                  --user "$USER"
+    )
+
+    # Print the full command
+    # printf "%q " "${podman_cmd[@]}" >> /tmp/toolbox_podman_command.log
+
+    "${podman_cmd[@]}" || fail "Podman couldn't create container '$container_name'"
```
debarshiray: Nitpick: we can use `echo` and `return 1`, like you had used when resolving the home directory and parsing the path used by `org.freedesktop.Flatpak.SessionHelper`.

I know `fail` is used in some other parts of the code base.  However, over time, it seems to me that it doesn't have that many benefits over printing a message to the standard error stream and returning with a non zero value.  The `fail` function isn't part of core Bats and comes from [bats-support](https://github.com/bats-core/bats-support), which is sometimes not packaged by distributions, and we have to carry it as a Git submodule in Toolbx.  If we stop using it, then we can drop this dependency.

test/system/104-run.bats
```diff
@@ -316,6 +316,21 @@ teardown() {
...
+
+  create_container_flatpak_session_helper "$container"
+
+  run --keep-empty-lines --separate-stderr "$TOOLBX" run --container "$container" true
+
+  assert_success
+  assert [ ${#lines[@]} -eq 0 ]
+  lines=("${stderr_lines[@]}")
+  assert_line --index 0 "Warning: container $container uses deprecated features"
+  assert_line --index 1 "Consider recreating it with Toolbox version 0.0.97 or newer."
```
debarshiray: It should be Toolbx without the _o_.  :)

debarshiray:
One thing that struck me while playing with this pull request, is that the deprecated container created by the test suite doesn't suffer from https://github.com/containers/crun/issues/1841 even though we are not using `podman create --mount type=devpts,destination=/dev/pts`.

I tried to compile Toolbx 0.0.96 on Fedora 42 and the container it created did hit https://github.com/containers/crun/issues/1841 and I failed to enter it.  I wonder what is it about the deprecated container created by the test suite that it doesn't have this problem.

debarshiray:
Anyway, merged!  Thanks for working on this.

PR #1699, opened 2025-08-07
test/system: Test fixes due to changed behavior in new bash version

DaliborKr:
In the new version of bash, the error output has been changed, and this caused a test failure in Toolbx CI on Fedora Rawhide.

- GNU bash, version 5.2.37 (bash-5.2.37-1.fc42.x86_64)
```
$ bash -c 'exec "$@"' bash /etc
bash: line 1: /etc: Is a directory
bash: line 1: exec: /etc: cannot execute: Is a directory
```

- GNU bash, version 5.3.0 (bash-5.3.0-2.fc43.x86_64)
````
$ bash -c 'exec "$@"' bash /etc
bash: line 1: /etc: Is a directory
````

 https://github.com/containers/toolbox/pull/1688#pullrequestreview-3045061238

### DIFF
--- a/test/system/104-run.bats
+++ b/test/system/104-run.bats
@@ -811,9 +811,15 @@ teardown() {
   assert [ ${#lines[@]} -eq 0 ]
   lines=("${stderr_lines[@]}")
   assert_line --index 0 "bash: line 1: /etc: Is a directory"
-  assert_line --index 1 "bash: line 1: exec: /etc: cannot execute: Is a directory"
-  assert_line --index 2 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
-  assert [ ${#stderr_lines[@]} -eq 3 ]
+
+  if [ ${#stderr_lines[@]} -eq 2 ]; then
+    assert_line --index 1 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
+  elif [ ${#stderr_lines[@]} -eq 3 ]; then
+    assert_line --index 1 "bash: line 1: exec: /etc: cannot execute: Is a directory"
+    assert_line --index 2 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
+  else
+    assert bash -c "[ ${#stderr_lines[@]} -eq 2 ] || [ ${#stderr_lines[@]} -eq 3 ]"
+  fi
 }
 
 @test "run: Try a non-existent command" {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/system/104-run.bats [hunk 1]
```diff
@@ -810,10 +810,17 @@ teardown() {
...
+
+  if [ ${#stderr_lines[@]} -eq 3 ]; then
+    assert_line --index 0 "bash: line 1: /etc: Is a directory"
+    assert_line --index 1 "bash: line 1: exec: /etc: cannot execute: Is a directory"
+    assert_line --index 2 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
+  elif [ ${#stderr_lines[@]} -eq 2 ]; then
+    assert_line --index 0 "bash: line 1: /etc: Is a directory"
+    assert_line --index 1 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
+  else
+    fail "Unexpected number of stderr lines: ${#stderr_lines[@]}"
```
debarshiray: It would be slightly better if we could somehow use the `assert` function here.  That way, if the test fails, it will be immediately obvious from the error message what the expected number of lines were without looking at the source code of the tests.

test/system/104-run.bats
```diff
@@ -810,10 +810,17 @@ teardown() {
   assert_failure
   assert [ ${#lines[@]} -eq 0 ]
   lines=("${stderr_lines[@]}")
-  assert_line --index 0 "bash: line 1: /etc: Is a directory"
-  assert_line --index 1 "bash: line 1: exec: /etc: cannot execute: Is a directory"
-  assert_line --index 2 "Error: failed to invoke command /etc in container $(get_latest_container_name)"
-  assert [ ${#stderr_lines[@]} -eq 3 ]
+
+  if [ ${#stderr_lines[@]} -eq 3 ]; then
+    assert_line --index 0 "bash: line 1: /etc: Is a directory"
```
debarshiray: It would be slightly better to move the first line out of the `if` branches.  It's common to both Bash versions, and will get asserted even if the number of lines is completely wrong.

test/system/104-run.bats
debarshiray: I managed to find a way to do it so that when the test fails, it shows as:
```
   -- assertion failed --
   expression : bash -c [ 4 -eq 2 ] || [ 4 -eq 3 ]
   --
```

test/system/104-run.bats
debarshiray: I took the liberty to make this minor change myself, so that I can include this in the upcoming release.

debarshiray:
> You made me curious and I went digging into the Bash sources to understand the exact change more.

I updated the commit message to add a reference to the Bash change.

debarshiray:
recheck

debarshiray:
Damn!  The CI timed out.  I'll merge once it passes.

PR #1703, opened 2025-08-08
Prepare 0.2

debarshiray:


debarshiray:
The regular expression for checking the version numbers needs to be updated:
```
fedora-rawhide | not ok 1 version: Check version using option --version
fedora-rawhide | # (from function `assert_output' in file test/system/libs/bats-assert/src/assert.bash, line 239,
fedora-rawhide | #  in test file test/system/001-version.bats, line 32)
fedora-rawhide | #   `assert_output --regexp '^toolbox version [0-9]+\.[0-9]+\.[0-9]+(\.[0-9]+)?$'' failed
fedora-rawhide | #
fedora-rawhide | # -- regular expression does not match output --
fedora-rawhide | # regexp : ^toolbox version [0-9]+\.[0-9]+\.[0-9]+(\.[0-9]+)?$
fedora-rawhide | # output : toolbox version 0.2
fedora-rawhide | # --
fedora-rawhide | #
```

Issue #1622, opened 2025-04-08
Cannot enter container created from non-Toolbx image because /home is not empty


bryce-carson:
**Describe the bug**
I cannot enter a new container created with docker.io/rocker/r-bspm:latest when using toolbox on Fedora 41 Silverblue (Fedora Linux 41.20250408.0 (Silverblue)):

`toolbox create bspm --image docker.io/rocker/r-bspm:latest`

The following error is returned when attempting `toolbox enter bspm`:

<details>
<summary>Error: failed to redirect /home to /var/home: remove /home: directory not empty</summary>
<pre>
DEBU Running as real user ID 1000                 
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU Looking up sub-GID and sub-UID ranges for user bryce 
DEBU TOOLBX_DELAY_ENTRY_POINT is                  
DEBU TOOLBX_FAIL_ENTRY_POINT is                   
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Migrating to newer Podman                    
DEBU Toolbx config directory is /var/home/bryce/.config/toolbox 
DEBU Current Podman version is 5.4.1              
DEBU Creating runtime directory /run/user/1000/toolbox 
DEBU Old Podman version is 5.4.1                  
DEBU Migration not needed: Podman version 5.4.1 is unchanged 
DEBU Setting up configuration                     
DEBU Setting up configuration: file /var/home/bryce/.config/containers/toolbox.conf not found 
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-41'               
DEBU Image: 'fedora-toolbox:41'                   
DEBU Release: '41'                                
DEBU Resolving container and image names          
DEBU Container: 'bspm'                            
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'bspm'                            
DEBU Image: 'fedora-toolbox:41'                   
DEBU Release: '41'                                
DEBU Checking if container bspm exists            
DEBU Inspecting container bspm                    
DEBU Entry point of container bspm is toolbox (PID=0) 
DEBU Inspecting mounts of container bspm          
[...]

PR #1707, opened 2025-08-15
Added a warning when using a non-Toolbx container

DaliborKr:
As mentioned in https://github.com/containers/toolbox/issues/1622, not all images are guaranteed to work with Toolbx. Therefore, every Toolbx verified image contains specific labels (`com.github.containers.toolbox="true"` or `com.github.debarshiray.toolbox="true"`), which means that the image was previously tested with Toolbx.

In this PR, the detection of usage of non-Toolbx images has been added to the `create`, `run`, and `enter` commands. If such an image is detected, the user is prompted with a message like:

```
$ toolbox create --image foo/bar:123 cont_foo_bar_123
Image 'foo/bar:123' is not a Toolbx image and may not work correctly (see https://containertoolbx.org/doc/). Continue anyway? [y/N]: 
...
```

Continuing may lead to an error while opening the container.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -222,23 +222,50 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 		return nil
 	}
 
-	imageFull, err := podman.GetFullyQualifiedImageFromRepoTags(image)
-	if err != nil {
-		var errImage *podman.ImageError
+	imageObj, inspectErr := podman.InspectImage(image)
 
-		if errors.As(err, &errImage) {
-			if errors.Is(err, podman.ErrImageRepoTagsEmpty) {
-				logrus.Debugf("Image %s has empty RepoTags, likely because it is without a name", image)
-				imageFull = image
-			} else if errors.Is(err, podman.ErrImageRepoTagsMissing) {
-				return fmt.Errorf("missing RepoTags for image %s", image)
-			} else {
-				panicMsg := fmt.Sprintf("unexpected %T: %s", err, err)
-				panic(panicMsg)
-			}
-		} else {
-			return err
+	logrus.Debugf("Resolving fully qualified name for image %s from RepoTags", image)
+	var imageFull string
+
+	if utils.ImageReferenceHasDomain(image) {
+		imageFull = image
+	} else {
+		if inspectErr != nil {
+			return fmt.Errorf("failed to inspect image %s", image)
 		}
+
+		var err error
+		imageFull, err = podman.GetFullyQualifiedImageFromRepoTags(image, imageObj)
+		if err != nil {
+			var errImage *podman.ImageError
+
+			if errors.As(err, &errImage) {
+				if errors.Is(err, podman.ErrImageRepoTagsEmpty) {
+					logrus.Debugf("Image %s has empty RepoTags, likely because it is without a name", image)
+					imageFull = image
+				} else if errors.Is(err, podman.ErrImageRepoTagsMissing) {
+					return fmt.Errorf("missing RepoTags for image %s", image)
+				} else {
+					panicMsg := fmt.Sprintf("unexpected %T: %s", err, err)
+					panic(panicMsg)
+				}
+			} else {
+				return err
+			}
+		}
+	}
+
+	logrus.Debugf("Resolved image %s to %s", image, imageFull)
+
+	var imageUnknownLabel []string
+
+	hasWarnings, shouldContinue := checkAndWarnImageCompatibility(imageFull, imageObj, true)
+	if !shouldContinue {
+		return nil
+	}
+
+	if hasWarnings {
+		imageUnknownLabel = []string{"--label", "com.github.containers.toolbx.image-unknown=true"}
 	}
 
 	var toolbxDelayEntryPointEnv []string
@@ -445,6 +472,8 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 		"--ipc", "host",
 		"--label", "com.github.containers.toolbox=true",
 	}...)
+
+	createArgs = append(createArgs, imageUnknownLabel...)
 
 	createArgs = append(createArgs, devPtsMount...)
 
--- a/src/pkg/podman/image.go
+++ b/src/pkg/podman/image.go
@@ -47,6 +47,8 @@ type imageImages struct {
 
 type imageInspect struct {
 	created      string
+	entrypoint   []string
+	env          []string
 	id           string
 	labels       map[string]string
 	namesHistory []string
@@ -279,7 +281,9 @@ func (image *imageInspect) UnmarshalJSON(data []byte) error {
 		Created interface{}
 		ID      string
 		Config  struct {
-			Labels map[string]string
+			Entrypoint []string
+			Env        []string
+			Labels     map[string]string
 		}
 		NamesHistory []string
 		RepoTags     []string
@@ -296,6 +300,8 @@ func (image *imageInspect) UnmarshalJSON(data []byte) error {
 		image.created = utils.HumanDuration(int64(value))
 	}
 
+	image.entrypoint = raw.Config.Entrypoint
+	image.env = raw.Config.Env
 	image.id = raw.ID
 	image.labels = raw.Config.Labels
 	image.namesHistory = raw.NamesHistory
--- a/src/pkg/podman/podman.go
+++ b/src/pkg/podman/podman.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/HarryMichal/go-version"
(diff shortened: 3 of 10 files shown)


DaliborKr:
recheck

DaliborKr:
I have updated this PR by adding detection of more aspects in the base images used for Toolbx containers, which users should know about. To the already existing Toolbx images detection (according to its labels `com.github.containers.toolbox="true"` or `com.github.debarshiray.toolbox="true" `), two additional checks were added:

**1. Presence of environmental variable `LD_PRELOAD`**
Images that have the environmental variable `LD_PRELOAD` set represent a potential threat in terms of container escape possibility (see https://www.wiz.io/blog/nvidia-ai-vulnerability-cve-2025-23266-nvidiascape).

**2. Already set image `entry point`**
Toolbx sets its own `entry point` while creating a new container, so the base image shouldn't have set any before it is used for this purpose.

src/pkg/podman/podman.go
```diff
@@ -506,3 +574,38 @@ func SystemMigrate(ociRuntimeRequired string) error {
 
 	return nil
 }
+
+func DoesImageFulfillRequirements(image string) (bool, string, error) {
+	var warnings []string
+
+	isToolboxImage, err := IsToolboxImage(image)
```
debarshiray: We are invoking `podman inspect --type image` several times while doing the checks.  First, to check the labels; second, to check `LD_PRELOAD`; and third, to check the `EntryPoint`.  Generally, each invocation of `podman(1)` is sufficiently expensive, because it does a lot of heavy initialization when it starts.  So, it's good to avoid them, if possible.  This is especially true for Toolbx commands like `enter` and `run` that are used very often, and a delay of one extra second can start to annoy the user over time.

For example, here are some past commits where we optimized `enter` and `run` by reducing the number of `podman(1)` invocations:
* commit [74d4fcf00c6ec3d1](https://github.com/containers/toolbox/commit/74d4fcf00c6ec3d1954a87da60b38fdc5eed7b2d) removed an `podman start` and `podman inspect`
* commit[4536e2c8c28f6c4f](https://github.com/containers/toolbox/commit/4536e2c8c28f6c4fed5e0346ea1094156e449f59) removed two `podman exec` calls

The `create` command is not that frequently used and already does enough disk and network I/O that people expect it to be slow, but we also shouldn't make it slower than it needs to be.  :)

We can reduce the number of `podman inspect` invocations if we parse the JSON into an `Image` object and then keep using the object repeatedly.

debarshiray (changes requested):
Thanks for working on this, @DaliborKr !  It's really good that you added so many tests to ensure that things keep working as expected.

After seeing your code changes, I realized that I might have underestimated the complications hidden in this effort.  I will try to explain inline.

src/pkg/podman/podman.go
```diff
@@ -352,17 +353,84 @@ func IsToolboxImage(image string) (bool, error) {
...
+	case []string:
+		for _, envVar := range envVars {
+			envVarTrimmed := strings.TrimSpace(envVar)
+			if strings.HasPrefix(envVarTrimmed, "LD_PRELOAD=") {
+				return true, nil
+			}
+		}
+	default:
+		return false, fmt.Errorf("unexpected type '%T' of environment variables in image %s", env, image)
+	}
```
debarshiray, line 304: Decoding the JSON in this way can be fragile if the structure of the JSON changes, and over time Podman has sometimes done that.  :(

See the comments in [src/pkg/podman/container.go](https://github.com/containers/toolbox/blob/main/src/pkg/podman/container.go), and the `UnmarshalJSON()` method of the [Image object](https://github.com/containers/toolbox/blob/main/src/pkg/podman/podman.go).

Fortunately, Go can decode JSON on a best effort basis, which can cover-up many of the superficial changes and needs a lot less code.  [Here's](https://go.dev/blog/json) an introductory article about how it works.  Sometimes, Go can't automatically handle the changes and we need to handle it ourselves, like in the above comments.

To take advantage of this, and to avoid invoking `podman inspect` repeatedly, we can decode the JSON into an `Image` object that implements the [json.Unmarshaler](https://pkg.go.dev/encoding/json/v2#Unmarshaler) interface.  It's implemented by the `containerInspect`, `containerPS` and `Image` objects above.  The interface will limit all manual interventions during decoding in one place.

Unfortunately, the JSON used by Podman to represent images differs between `podman images` and `podman inspect --type image`.  The `Image` object only handles the former, but not the latter.

We have the same problem for containers.  That's why we have a `Container` interface to hide the details, and it's implemented by the private `containerInspect` and `containerPS` objects that actually decode the JSON.

So, our first step would be to have a commit that introduces an `Image` interface, and renames the existing `Image` object to make it private and makes it implement the interface.  I don't know what would be the best name for the object, because `imageImages` looks awkward, but it's also a private type so it won't be seen in too many places.

Then, we have to add a second implementation of the `Image` interface that decodes the JSON from `podman inspect --type image`.

Here are some commits that did the same thing for containers that you can use for help:
* commit [e611969726ed4260](https://github.com/containers/toolbox/commit/e611969726ed4260fbbbe97396cf9fc656ea0862)
  `cmd, pkg/podman: Turn Container into an interface`
* commit [ec7eb59bb058c12d](https://github.com/containers/toolbox/commit/ec7eb59bb058c12d44413535445b2795af22814e)
  `cmd/run, pkg/podman: Make podman.InspectContainer() return a Container`

Then, the `IsToolboxImage()` function can turn into the `IsToolbx()` method of the `Image` interface:
* commit [d363eb26d10cd93c](https://github.com/containers/toolbox/commit/d363eb26d10cd93c0550412035de9a0972a8298a)
  `cmd, pkg/podman: Turn IsToolboxContainer() into Container.IsToolbx()`

debarshiray (changes requested):

debarshiray (commented):
Maybe it's time to rebase/rewrite this on top of https://github.com/containers/toolbox/pull/1724 , and the `Image` interface that's now in `main`?

DaliborKr:
The requested changes have been applied in a new approach based on the `Image` interface introduced in #1724 and #1779. The `imageInspect` type was extended with `entrypoint` and `env` fields. Also a `CheckImageCompatibility()` function was added to the `podman` package and it can be extended with other image checks in the future if necessary.

The compatibility check is currently only performed in `toolbox create`. I found it redundant to warn users every time they start a container via `toolbox run` or `toolbox enter`, since they have already accepted the warnings during container creation phase. If there is a need for checking and notifying users in `toolbox run` and `toolbox enter` as well, I can extend it.

DaliborKr:
Added the image compatibility check for `toolbox run` and `toolbox enter` as well.

src/cmd/create.go
```diff
@@ -241,6 +241,17 @@ func createContainer(container, image, release, authFile string, showCommandToEn
...
 
+	var imageUnknownLabel []string
+
+	hasWarnings, shouldContinue := checkAndWarnImageCompatibility(imageFull, true)
+	if !shouldContinue {
+		return nil
+	}
+
+	if hasWarnings {
+		imageUnknownLabel = []string{"--label", "com.github.containers.toolbox.image-unknown=true"}
```
debarshiray: Let's use `com.github.containers.toolbx.image-unknown=true` (ie., without the *o*).  That way we won't have to do a migration in the future, which is always painful.  :)

src/cmd/create.go
```diff
@@ -241,6 +241,17 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 		}
 	}
 
+	var imageUnknownLabel []string
+
+	hasWarnings, shouldContinue := checkAndWarnImageCompatibility(imageFull, true)
```
debarshiray: It will be good if we could somehow re-use the `Image` object from the earlier `InspectImage()` call in `GetFullyQualifiedImageFromRepoTags()`.

It's not a huge problem - just that one of the big advantages of your work on the `Image` interface is that we can carry the object around without repeatedly calling `InspectImage()`.  So, it will be good to take advantage of it.

I haven't fully tested this, so I could be wrong.  I am wondering if something like this will work, where we move some code out of the existing `GetFullyQualifiedImageFromRepoTags()` function:
```
imageObj, err := InspectImage(image)
if err != nil {
    return "", fmt.Errorf("failed to inspect image %s", image)
}

var imageFull string

if utils.ImageReferenceHasDomain(image) {
    imageFull = image
} else {
    imageFull, err := podman.GetFullyQualifiedImageFromRepoTags(imageObj)
}

var imageUnknownLabel []string

hasWarnings, shouldContinue := checkAndWarnImageCompatibility(imageObj, true)
if !shouldContinue {
    return nil
}
```

src/pkg/podman/image.go
```diff
@@ -296,6 +300,8 @@ func (image *imageInspect) UnmarshalJSON(data []byte) error {
 		image.created = utils.HumanDuration(int64(value))
 	}
 
+	image.entrypoint = raw.Config.Entrypoint
+	image.env = raw.Config.Env
```
debarshiray, line 304: We should restore the Go unit tests for entry point and environment variables that you had originally written for https://github.com/containers/toolbox/pull/1724

I have a backup of the older version of your code.  So, if you no longer have them, I can help with the restoration.  :)

debarshiray (changes requested):
Thanks for resurrecting this pull request so quickly, @DaliborKr !

Here are some high level comments after a quick look.  I know you are busy and don't have too much time, so no pressure.

src/cmd/create.go
DaliborKr: Alright, I have changed it in the latest commit.

src/cmd/create.go
DaliborKr: Yes, very good point! I updated it so it only requires one `InspectImage()` call. I did it similarly to what you proposed here. :)

src/pkg/podman/image.go
DaliborKr: Yes, if you still have that backup, I would really appreciate it if you could restore those Go unit tests. Thanks!

DaliborKr:
recheck

PR #1710, opened 2025-09-05
cmd/initContainer: Configure environment.d(5), su(1) and sudo(8) to set P11_KIT_SERVER_ADDRESS

debarshiray:


src/cmd/initContainer.go
```diff
@@ -623,36 +623,14 @@ func configurePKCS11(targetUser *user.User) error {
 		return fmt.Errorf("failed to configure PKCS #11 to read from the host: %w", err)
 	}
 
-	if err := configureSSHD(serverSocket); err != nil {
+	if err := configurePKCS11SSHD(serverSocket); err != nil {
```
juhp, line 626: Is something missing or is it me?
How does just renaming this function change anything?

I think two separate commits would also be easier to read. :)

debarshiray: Yes, you are right.  My apologies.  I should have marked this pull request as a draft.

I haven't pushed all the commits, yet.  I only pushed the ones that were done to let the CI run on them.

Issue #1706, opened 2025-08-12
failing to compile because it can't lookup undefined symbol nvmlGpuInstanceGetComputeInstanceProfileInfoV


drjoms:
meson compile -C builddir
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/sbin/ninja -C /home/dimko/compiled/toolbox/builddir
ninja: Entering directory `/home/dimko/compiled/toolbox/builddir'
[1/1] Generating src/zsh-completion with a custom command (wrapped by meson to capture output)
FAILED: src/_toolbox 
/usr/lib/python-exec/python3.13/meson --internal exe --capture src/_toolbox -- /usr/bin/python3.13 /home/dimko/compiled/toolbox/src/meson_generate_completions.py /home/dimko/compiled/toolbox/src zsh
--- stderr ---
/tmp/go-build2866398736/b001/exe/toolbox: symbol lookup error: /tmp/go-build2866398736/b001/exe/toolbox: undefined symbol: nvmlGpuInstanceGetComputeInstanceProfileInfoV
exit status 127
Traceback (most recent call last):
  File "/home/dimko/compiled/toolbox/src/meson_generate_completions.py", line 32, in <module>
    output = subprocess.run(['go', 'run', '.', 'completion', completion_type], check=True)
  File "/usr/lib/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['go', 'run', '.', 'completion', 'zsh']' returned non-zero exit status 1.

ninja: build stopped: subcommand failed.

**Expected behaviour**
No errors after compilation attempt

**Actual behaviour**
as per output above

**Toolbx package info (`rpm -q toolbox`)**
compiling from source

**Output of `podman version`**
e.g.,
```
Version:            1.9.2
RemoteAPI Version:  1
Go Version:         go1.14.2
OS/Arch:            linux/amd64
```

**Podman package info (`rpm -q podman`)**
[ebuild   R    ] app-containers/podman-5.3.2::gentoo  USE="seccomp systemd -apparmor -btrfs (-selinux) -wrapper" 0 KiB

**Info about your OS**
current one only and latest Gentoo.(ongoing release)

**Additional context**
Attempting to follow those instructions:
https://containertoolbx.org/install/

meson compile -C builddir
INFO: autodetecting backend as ninja
[...]

PR #1722, opened 2025-09-21
Replace build wrapper with native meson targets

PeakKS:
Tentative fix for #1706 

Replaces the go build wrapper and the completions wrapper with just doing the build right in meson. Essentially I am just tricking go into building as a static library, then linking that with itself. This prevents the weird go build cache issue I ran into in #1706.

Since `-z now` can not be used for this application, I am forcing `-z lazy` in the linker args. It seems like the reason this only popped up on gentoo is because we default to `-z now` for security reasons (see : https://packages.gentoo.org/useflags/default-znow).

Getting rid of some of the boilerplate is just a nice side benefit :)


PeakKS:
Not sure how CI is failing, logs seem to show the binary getting compiled then not existing for completions?

debarshiray (commented):
Thanks for working on this, @PeakKS !  I don't have time right now to take a very close look, but I didn't want to completely ignore your significant contribution.  So some fly-by comments:

First, I think what you are suggesting is a good idea.  The less shell scripts we use, the better it is!  :)

Second, do you think it will be possible to have tests to check if the `toolbox` binary was correctly built?  You have already seen what `src/go-build-wrapper` was doing, and you will see the background in the Git history of that file.  The `toolbox` binary needs to be linked in a very specific way for different Toolbx containers to work across different host OSes.

If we have tests to check the important attributes of the binary, then it becomes a lot easier to make changes to the build, and review those changes, etc..  It's one of the big reasons why we never revamped `src/go-build-wrapper` before.  Manually checking everything felt too nerve-wracking and stressful.  :)

debarshiray (commented):
Could you please use your full name and a real email address in your Git authorship information?

The `Signed-off-by` trailer has no meaning without a real full name and email address.

I could have made an exception for a one line change because it would be legally insignificant.  However, that's not the case with this pull request, because you have done a substantial amount of great work.  :)

We need a real full name and email address for tracking copyright holders in case a licensing issue crops up.  Note how projects as diverse as [GCC](https://gcc.gnu.org/dco.html), [GnuPG](https://www.gnupg.org/faq/HACKING.html), [Linux](https://www.kernel.org/doc/html/latest/process/1.Intro.html) and [Podman](https://github.com/containers/podman/blob/main/CONTRIBUTING.md) don't allow anonymous or pseudonymous contributions.

PeakKS:
> Could you please use your full name and a real email address in your Git authorship information?
> 
> The `Signed-off-by` trailer has no meaning without a real full name and email address.
> 
> I could have made an exception for a one line change because it would be legally insignificant. However, that's not the case with this pull request, because you have done a substantial amount of great work. :)
> 
> We need a real full name and email address for tracking copyright holders in case a licensing issue crops up. Note how projects as diverse as [GCC](https://gcc.gnu.org/dco.html), [GnuPG](https://www.gnupg.org/faq/HACKING.html), [Linux](https://www.kernel.org/doc/html/latest/process/1.Intro.html) and [Podman](https://github.com/containers/podman/blob/main/CONTRIBUTING.md) don't allow anonymous or pseudonymous contributions.

No problem, didn't even realize I used my username lol.

debarshiray:
Hey @PeakKS !  How is it going with this pull request.  I am asking because I see that the CI is failing.  Do you need some help?

PeakKS:
> Hey @PeakKS ! How is it going with this pull request. I am asking because I see that the CI is failing. Do you need some help?

I haven't worked on this in a while but I think it's correct besides the CI. I can rebase on main to see if that magically fixes the CI, otherwise I'll need someone to take a closer look.

PeakKS:
Okay seems like the same issue as before.

Links target? Yep: `[12/15] Linking target src/toolbox`
Correct dependencies in `build.ninja`? Seems so: `build src/_toolbox: CUSTOM_COMMAND  | src/toolbox`
Fails anyways? Yep: `FileNotFoundError: [Errno 2] No such file or directory: '/home/zuul-worker/src/github.com/containers/toolbox/builddir/src/toolbox'`

I also tried running my local build of toolbox to run a fedora42 container and try building there to replicate the CI build failure but it still builds fine inside.

@debarshiray could you take a look?

debarshiray:
This is how the CI fails on Ubuntu hosts:
```
...
ninja: Entering directory `/home/runner/work/toolbox/toolbox/containers/toolbox/builddir'
[1/15] /usr/bin/go-md2man -in ../doc/toolbox-enter.1.md -out doc/toolbox-enter.1
[2/15] /usr/bin/go-md2man -in ../doc/toolbox.1.md -out doc/toolbox.1
[3/15] /usr/bin/go-md2man -in ../doc/toolbox-create.1.md -out doc/toolbox-create.1
[4/15] /usr/bin/go-md2man -in ../doc/toolbox-help.1.md -out doc/toolbox-help.1
[5/15] /usr/bin/go-md2man -in ../doc/toolbox-init-container.1.md -out doc/toolbox-init-container.1
[6/15] /usr/bin/go-md2man -in ../doc/toolbox-list.1.md -out doc/toolbox-list.1
[7/15] /usr/bin/go-md2man -in ../doc/toolbox-rmi.1.md -out doc/toolbox-rmi.1
[8/15] /usr/bin/go-md2man -in ../doc/toolbox.conf.5.md -out doc/toolbox.conf.5
[9/15] /usr/bin/go-md2man -in ../doc/toolbox-run.1.md -out doc/toolbox-run.1
[10/15] /usr/bin/go-md2man -in ../doc/toolbox-rm.1.md -out doc/toolbox-rm.1
[11/15] /usr/lib/go-1.22/bin/go build -C /home/runner/work/toolbox/toolbox/containers/toolbox/src -trimpath -buildmode=c-archive -tags '' -ldflags '-X github.com/containers/toolbox/pkg/version.currentVersion=0.3' -o /home/runner/work/toolbox/toolbox/containers/toolbox/builddir/src/toolbox.a
[12/15] cc  -o src/toolbox  -Wl,--as-needed -Wl,--no-undefined -Wl,--whole-archive -Wl,--start-group src/toolbox.a -Wl,--no-whole-archive -Wl,-rpath,/run/host/usr/lib/x86_64-linux-gnu -Wl,-z,lazy -Wl,--export-dynamic -Wl,--unresolved-symbols=ignore-in-object-files -Wl,-dynamic-linker,/run/host/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2 -Wl,--end-group
[13/15] /usr/bin/meson --internal exe --capture src/toolbox.bash -- /home/runner/work/toolbox/toolbox/containers/toolbox/builddir/src/toolbox completion bash
FAILED: [code=1] src/toolbox.bash 
/usr/bin/meson --internal exe --capture src/toolbox.bash -- /home/runner/work/toolbox/toolbox/containers/toolbox/builddir/src/toolbox completion bash
Traceback (most recent call last):
  File "/usr/bin/meson", line 29, in <module>
    sys.exit(mesonmain.main())
  File "/usr/lib/python3/dist-packages/mesonbuild/mesonmain.py", line 257, in main
    return run(sys.argv[1:], launcher)
  File "/usr/lib/python3/dist-packages/mesonbuild/mesonmain.py", line 246, in run
    return run_script_command(args[1], args[2:])
  File "/usr/lib/python3/dist-packages/mesonbuild/mesonmain.py", line 194, in run_script_command
    return module.run(script_args)
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/meson_exe.py", line 122, in run
    return run_exe(exe)
  File "/usr/lib/python3/dist-packages/mesonbuild/scripts/meson_exe.py", line 66, in run_exe
    p = subprocess.Popen(cmd_args, env=child_env, cwd=exe.workdir,
  File "/usr/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1751, in _execute_child
    self._posix_spawn(args, executable, env, restore_signals,
  File "/usr/lib/python3.10/subprocess.py", line 1696, in _posix_spawn
    self.pid = os.posix_spawn(executable, args, env, **kwargs)
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/toolbox/toolbox/containers/toolbox/builddir/src/toolbox'
```

As far as I can make out, step 12 generates `src/toolbox` and step 13 tries to use it to generate the shell completions and fails with `FileNotFoundError`.

debarshiray (commented):
It's not immediately clear to me why the build breaks on the CI environment, because, like you, I can't reproduce it on my local system either.  Meanwhile, here's an idea on how to move this forward:

src/meson.build
```diff
@@ -22,123 +17,156 @@ sources = files(
...
+  output: 'toolbox.a',
+)
+
+cc_exe = find_program(cc.cmd_array().get(-1))
+readlink = find_program('readlink')
+dirname = find_program('dirname')
+
+libc_path = run_command(cc_exe, '--print-file-name=libc.so', capture: true, check: true).stdout()
+libc_canon = run_command(readlink, '--canonicalize', libc_path, capture: true, check: true).stdout()
+libc_dir = run_command(dirname, libc_canon, capture: true, check: true).stdout().strip()
```
debarshiray, line 72: I like this approach of using Meson's [run_command()](https://mesonbuild.com/Reference-manual_functions_run_command.html) to move things out of the `go-build-wrapper` shell script into Meson.

It seems to me that these parts can be split into a simpler commit that's less likely to break the CI.

src/meson.build
```diff
@@ -22,123 +17,156 @@ sources = files(
...
-  dynamic_linker = '/lib/ld64.so.1'
-elif cpu_family == 'x86' and endian == 'little'
-  dynamic_linker = '/lib/ld-linux.so.2'
-elif cpu_family == 'x86_64' and endian == 'little'
-  dynamic_linker = '/lib64/ld-linux-x86-64.so.2'
-elif cpu_family == 'riscv64' and endian == 'little'
-  dynamic_linker = '/lib/ld-linux-riscv64-lp64d.so.1'
-else
-  host_machine_description = cpu_family + ' (' + endian + ' endian)'
-  error('Please specify dynamic linker for:', host_machine_description)
```
debarshiray, line 69: I am curious about the switch from this to `objcopy(1)` to set the dynamic linker.  In the past we used `patchelf(1)` to do the same thing, and we [dropped](https://github.com/containers/toolbox/commit/8dd7ee47c5a76b7da7b29f3395dbcccaf16838d9) it because some downstream distributors decided against shipping `patchelf(1)`.

I see that `objcopy(1)` is part of GNU Binutils.  So, it's more likely to be present than `patchelf(1)`, but there's still benefit in the general principle of fewer build requirements within reason.

debarshiray:
> Second, do you think it will be possible to have tests to check if the `toolbox` binary was correctly built? You have already seen what `src/go-build-wrapper` was doing, and you will see the background in the Git history of that file. The `toolbox` binary needs to be linked in a very specific way for different Toolbx containers to work across different host OSes.
> 
> If we have tests to check the important attributes of the binary, then it becomes a lot easier to make changes to the build, and review those changes, etc.. It's one of the big reasons why we never revamped `src/go-build-wrapper` before. Manually checking everything felt too nerve-wracking and stressful. :)

I found some tests hidden in https://github.com/containers/toolbox/pull/1031  Let's see if I can rescue them.

PR #1724, opened 2025-10-07
cmd, pkg/podman: Introduce Image Interface and Unit Tests

DaliborKr:
Introduces an Image interface based on the conversation in PR #1707.

Similar to how container information is handled (introduced in commits e611969 and ec7eb59), this PR abstracts the representation of an image. This is necessary because `podman inspect --type image` and `podman images` return different JSON structures. The Image interface provides a unified way to access image data, with two distinct implementations to handle the different JSON formats.

Updates related functions:
- GetImages() now returns an Images iterator, which handles image flattening and sorting internally.
- InspectImage() now returns an Image interface instead of a generic map[string]interface{}.
- The IsToolboxImage() function has been replaced with an IsToolbx() method on the Image interface.

Adds unit tests: This PR includes unit tests for the new Image interface implementation, covering the output of both `podman inspect --type image` and `podman images`.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -25,6 +25,7 @@ sources = files(
   'pkg/podman/errors.go',
   'pkg/podman/image.go',
   'pkg/podman/imageImages_test.go',
+  'pkg/podman/imageInspect_test.go',
   'pkg/podman/podman.go',
   'pkg/podman/containerInspect_test.go',
   'pkg/shell/shell.go',
--- a/src/pkg/podman/image.go
+++ b/src/pkg/podman/image.go
@@ -29,6 +29,7 @@ type Image interface {
 	Labels() map[string]string
 	Name() string
 	Names() []string
+	RepoTags() []string
 }
 
 type Images struct {
@@ -37,10 +38,19 @@ type Images struct {
 }
 
 type imageImages struct {
-	created string
-	id      string
-	labels  map[string]string
-	names   []string
+	created  string
+	id       string
+	labels   map[string]string
+	names    []string
+	repoTags []string
+}
+
+type imageInspect struct {
+	created      string
+	id           string
+	labels       map[string]string
+	namesHistory []string
+	repoTags     []string
 }
 
 type imageSlice []imageImages
@@ -163,12 +173,24 @@ func (image *imageImages) Names() []string {
 	return ret
 }
 
+func (image *imageImages) RepoTags() []string {
+	if image.repoTags == nil {
+		return nil
+	}
+
+	repoTagsCount := len(image.repoTags)
+	ret := make([]string, repoTagsCount)
+	copy(ret, image.repoTags)
+	return ret
+}
+
 func (image *imageImages) UnmarshalJSON(data []byte) error {
 	var raw struct {
-		Created interface{}
-		ID      string
-		Labels  map[string]string
-		Names   []string
+		Created  interface{}
+		ID       string
+		Labels   map[string]string
+		Names    []string
+		RepoTags []string
 	}
 
 	if err := json.Unmarshal(data, &raw); err != nil {
@@ -188,6 +210,96 @@ func (image *imageImages) UnmarshalJSON(data []byte) error {
 	image.id = raw.ID
 	image.labels = raw.Labels
 	image.names = raw.Names
+	image.repoTags = raw.RepoTags
+	return nil
+}
+
+func (image *imageInspect) Created() string {
+	return image.created
+}
+
+func (image *imageInspect) ID() string {
+	return image.id
+}
+
+func (image *imageInspect) IsToolbx() bool {
+	if isToolbx(image.labels) {
+		return true
+	}
+
+	return false
+}
+
+func (image *imageInspect) Labels() map[string]string {
+	if image.labels == nil {
+		return nil
+	}
+
+	labelsCount := len(image.labels)
+	ret := make(map[string]string, labelsCount)
+	for label, value := range image.labels {
+		ret[label] = value
+	}
+
+	return ret
+}
+
+func (image *imageInspect) Name() string {
+	if len(image.namesHistory) == 0 {
+		panic("no name is available in Image name history")
+	}
+
+	return image.namesHistory[0]
+}
+
+func (image *imageInspect) Names() []string {
+	if image.namesHistory == nil {
+		return nil
+	}
+
+	namesHistoryCount := len(image.namesHistory)
+	ret := make([]string, namesHistoryCount)
+	copy(ret, image.namesHistory)
+	return ret
+}
+
+func (image *imageInspect) RepoTags() []string {
+	if image.repoTags == nil {
+		return nil
+	}
+
+	repoTagsCount := len(image.repoTags)
+	ret := make([]string, repoTagsCount)
+	copy(ret, image.repoTags)
+	return ret
+}
+
+func (image *imageInspect) UnmarshalJSON(data []byte) error {
+	var raw struct {
+		Created interface{}
+		ID      string
+		Config  struct {
+			Labels map[string]string
+		}
+		NamesHistory []string
+		RepoTags     []string
+	}
+
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	switch value := raw.Created.(type) {
+	case string:
+		image.created = value
+	case float64:
+		image.created = utils.HumanDuration(int64(value))
+	}
+
+	image.id = raw.ID
+	image.labels = raw.Config.Labels
+	image.namesHistory = raw.NamesHistory
+	image.repoTags = raw.RepoTags
 	return nil
 }
 
--- a/src/pkg/podman/imageImages_test.go
+++ b/src/pkg/podman/imageImages_test.go
@@ -30,6 +30,7 @@ func TestImageImages(t *testing.T) {
 		isToolbx bool
 		labels   map[string]string
 		names    []string
+		repoTags []string
 	}
 
 	testCases := []struct {
@@ -66,6 +67,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/f29/fedora-toolbox:29",
 					},
+					repoTags: nil,
 				},
 				{
 					id:       "8c5e0075ddaf4651b73e71dcf420bab922209f246e1165f306f223112a3061a4",
@@ -74,6 +76,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"localhost/fedora-toolbox-user:29",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -100,6 +103,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/f30/fedora-toolbox:30",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -154,6 +158,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:32",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -214,6 +219,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:35",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -272,6 +278,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:38",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -345,6 +352,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"localhost/fedora-toolbox:38",
 					},
+					repoTags: nil,
 				},
 				{
 					id:       "3eda970c1318e5ea564134338746c04d86ae9817c3a0defdc138b8ccdf787285",
@@ -358,6 +366,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora:38",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -426,7 +435,8 @@ func TestImageImages(t *testing.T) {
 						"vendor":                        "Fedora Project",
 						"version":                       "38",
 					},
-					names: nil,
+					names:    nil,
+					repoTags: nil,
 				},
 				{
 					id:       "3eda970c1318e5ea564134338746c04d86ae9817c3a0defdc138b8ccdf787285",
@@ -440,6 +450,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora:38",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -486,6 +497,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:39",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -544,6 +556,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:40",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -597,7 +610,8 @@ func TestImageImages(t *testing.T) {
 						"vendor":                           "Fedora Project",
 						"version":                          "40",
 					},
-					names: nil,
+					names:    nil,
+					repoTags: nil,
 				},
 			},
 		},
@@ -691,6 +705,7 @@ func TestImageImages(t *testing.T) {
 						"localhost/fedora-toolbox:40",
 						"registry.fedoraproject.org/fedora-toolbox:40",
 					},
+					repoTags: nil,
 				},
 				{
 					id:       "5c5b2e637806fccb644effa4affc4a5d08dc7e6140586ecb8c601c8739e12628",
@@ -712,6 +727,7 @@ func TestImageImages(t *testing.T) {
 						"localhost/fedora-toolbox:40",
 						"registry.fedoraproject.org/fedora-toolbox:40",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -805,6 +821,7 @@ func TestImageImages(t *testing.T) {
 						"registry.fedoraproject.org/fedora-toolbox:40-test",
 						"registry.fedoraproject.org/fedora-toolbox:40",
 					},
+					repoTags: nil,
 				},
 				{
 					id:       "5c5b2e637806fccb644effa4affc4a5d08dc7e6140586ecb8c601c8739e12628",
@@ -826,6 +843,7 @@ func TestImageImages(t *testing.T) {
 						"registry.fedoraproject.org/fedora-toolbox:40-test",
 						"registry.fedoraproject.org/fedora-toolbox:40",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -920,6 +938,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"docker.io/library/postgres:latest",
 					},
+					repoTags: nil,
 				},
 				{
 					id:       "5c5b2e637806fccb644effa4affc4a5d08dc7e6140586ecb8c601c8739e12628",
@@ -940,6 +959,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:40",
 					},
+					repoTags: nil,
 				},
 				{
 					id:       "49008798958db0411330d8b6a996c03bdab1d1b10d139e7df2b75ed3298f84a7",
@@ -960,6 +980,7 @@ func TestImageImages(t *testing.T) {
 					names: []string{
 						"registry.fedoraproject.org/fedora-toolbox:40-aarch64",
 					},
+					repoTags: nil,
 				},
 			},
 		},
@@ -1008,6 +1029,13 @@ func TestImageImages(t *testing.T) {
 					}
 
 				}
+
+				repoTags := image.RepoTags()
+				if len(repoTags) != 0 {
+					repoTags[0] = "foo/bar"
+				}
+
+				assert.Equal(t, expect.repoTags, image.RepoTags())
 			}
 		})
 	}
--- a/src/pkg/podman/podman.go
+++ b/src/pkg/podman/podman.go
@@ -191,31 +191,30 @@ func GetFullyQualifiedImageFromRepoTags(image string) (string, error) {
 	if utils.ImageReferenceHasDomain(image) {
 		imageFull = image
 	} else {
-		info, err := InspectImage(image)
+		imageObj, err := InspectImage(image)
 		if err != nil {
 			return "", fmt.Errorf("failed to inspect image %s", image)
 		}
 
-		if info["RepoTags"] == nil {
+		repoTags := imageObj.RepoTags()
+		if repoTags == nil {
 			return "", &ImageError{image, ErrImageRepoTagsMissing}
 		}
 
-		repoTags := info["RepoTags"].([]interface{})
 		if len(repoTags) == 0 {
 			return "", &ImageError{image, ErrImageRepoTagsEmpty}
 		}
 
 		for _, repoTag := range repoTags {
-			repoTagString := repoTag.(string)
-			tag := utils.ImageReferenceGetTag(repoTagString)
+			tag := utils.ImageReferenceGetTag(repoTag)
 			if tag != "latest" {
-				imageFull = repoTagString
+				imageFull = repoTag
 				break
 			}
 		}
 
 		if imageFull == "" {
-			imageFull = repoTags[0].(string)
+			imageFull = repoTags[0]
 		}
 	}
 
@@ -264,7 +263,7 @@ func InspectContainer(container string) (Container, error) {
 }
 
 // InspectImage is a wrapper around 'podman inspect --type image' command
-func InspectImage(image string) (map[string]interface{}, error) {
+func InspectImage(image string) (Image, error) {
 	var stdout bytes.Buffer
 
 	logLevelString := LogLevel.String()
@@ -275,31 +274,12 @@ func InspectImage(image string) (map[string]interface{}, error) {
 	}
 
 	output := stdout.Bytes()
-	var info []map[string]interface{}
-
-	if err := json.Unmarshal(output, &info); err != nil {
+	var images []imageInspect
+	if err := json.Unmarshal(output, &images); err != nil {
 		return nil, err
 	}
 
-	return info[0], nil
-}
-
-func IsToolboxImage(image string) (bool, error) {
-	info, err := InspectImage(image)
-	if err != nil {
-		return false, fmt.Errorf("failed to inspect image %s", image)
-	}
-
-	if info["Labels"] == nil {
-		return false, fmt.Errorf("%s is not a Toolbx image", image)
-	}
-
-	labels := info["Labels"].(map[string]interface{})
-	if labels["com.github.containers.toolbox"] != "true" && labels["com.github.debarshiray.toolbox"] != "true" {
-		return false, fmt.Errorf("%s is not a Toolbx image", image)
-	}
-
-	return true, nil
+	return &images[0], nil
 }
 
 func isToolbx(labels map[string]string) bool {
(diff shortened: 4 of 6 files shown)


src/pkg/podman/image.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return images.data[i].Names()[0] < images.data[j].Names()[0]
+}
+
+func (images Images) Swap(i, j int) {
+	images.data[i], images.data[j] = images.data[j], images.data[i]
+}
+
+// ALREADY EXISTS IN CONTAINER.GO
+//
```
debarshiray: I guess we need to take a decision about the code here in comments?  :)

debarshiray (changes requested):
Thanks for working on this, @DaliborKr !  My apologies for the delay.

I am still reading through the changes, but one thing popped out at me.

src/pkg/podman/podman.go
```diff
@@ -190,27 +106,49 @@ func GetContainers(args ...string) (*Containers, error) {
...
 // GetImages is a wrapper function around `podman images --format json` command.
 //
+// Parameter fillNameWithID is a boolean that indicates if the image names should be filled with the ID, when there are no names.
 // Parameter args accepts an array of strings to be passed to the wrapped command (eg. ["-a", "--filter", "123"]).
 //
 // Returned value is a slice of Images.
 //
 // If a problem happens during execution, first argument is nil and second argument holds the error message.
-func GetImages(args ...string) ([]Image, error) {
+func GetImages(fillNameWithID bool, sortByName bool, args ...string) (*Images, error) {
```
debarshiray: Isn't `sortByName` always set to `true` or did you have some future use in mind?  Or did I miss something?  :)

src/pkg/podman/image.go
```diff
@@ -0,0 +1,258 @@
...
+	image.envVars = raw.Config.Env
+	image.namesHistory = raw.NamesHistory
+	image.repoTags = raw.RepoTags
+	image.entrypoint = raw.Config.Entrypoint
+	return nil
+}
+
+func (images *Images) Get() Image {
+	if images.i < 1 {
+		panic("called Containers.Get() without calling Containers.Next()")
```
debarshiray: Typo alert: should be _Images_, not _Containers_.  :)

src/pkg/podman/image.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return available
+}
+
+func (images *Images) Reset() {
+	images.i = 0
+}
+
+func (images Images) Len() int {
+	return len(images.data)
```
debarshiray: It might be better to change these pre-existing methods from `ImageSlice` that earlier had value receivers to pointer receivers.

I am not an expert on value versus pointer receivers, and I get confused all the time.  It seems that the general [advice](https://go.dev/tour/methods/8) is to not mix pointer and value receivers for a given type.

In the case of the `Images` type, the `Reset()` method must have a pointer receiver because it modifies a field of the type.

In case you wonder how the `Swap()` method was working, just like I did, it's because in Go slices are [references](https://go.dev/tour/moretypes/8) to an underlying array and changing an element of a slice modifies the array and other slices see the same change.  The value receiver copies the slice and the underlying array stays the same.

src/cmd/list.go
```diff
@@ -154,38 +147,20 @@ func listHelp(cmd *cobra.Command, args []string) {
...
-
-		if isToolboxImage {
-			flattenedImages := image.FlattenNames(fillNameWithID)
-			toolboxImages = append(toolboxImages, flattenedImages...)
-		}
-
 	}
 
-	sort.Sort(podman.ImageSlice(toolboxImages))
 	return toolboxImages, nil
```
debarshiray: It looks like you are trying to get rid of this wrapper `getImages()` function, which is a good point.  It also applies to the `getContainers()` function.

Other than the changes you have already made, I wonder if we need the wrappers to filter Toolbx containers and images.  Do we ever use non-Toolbx containers and images?

As far as I can see, the wrappers originated when we were using [podman images --filter](https://github.com/containers/toolbox/commit/2369da5d31830e5cd3e1a13857a686365875ae61) and `podman ps --filter` to get the images instead of only `podman images` and `podman ps`.

I wonder if we can get rid of the wrappers completely, and if we can do it separately from this commit to keep the changes small.

debarshiray (changes requested):

src/pkg/podman/image.go
debarshiray: I think we can move the code common to `containers.go` and `images.go` to `podman.go`.

src/cmd/list.go
debarshiray: I made an attempt to split the part where you moved the flattening of the images from `getImages()` to `podman.GetImages()` in https://github.com/containers/toolbox/pull/1772  I hope I got it right.  Please take a look and let me know if I made some mistake.  :)

If it works out, then we can do the same for the rest of the code that you moved from `getImages()` to `podman.GetImages()`.  It help to reduce the size of this pull request, and give you more commits to brag about.  :)

src/pkg/podman/image.go
debarshiray: I took the liberty to do this in https://github.com/containers/toolbox/pull/1772

src/pkg/podman/podman.go
DaliborKr: Unfortunately, I don't really remember the intention behind this, since it's been a few months since the release. Maybe I'll figure it out once I focus on PR #1707 again, which is waiting for this to be merged, but we can get rid of this for now.

debarshiray: Okay, makes sense!

src/cmd/list.go
DaliborKr: Thanks @debarshiray for taking care of this! I've reviewed #1772, and everything seems right to me. :)

debarshiray: Thanks for the review!  Much appreciated.  :)

src/pkg/podman/podman.go
```diff
@@ -33,15 +33,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-type Image struct {
-	Created string
-	ID      string
-	Labels  map[string]string
-	Names   []string
-}
-
```
debarshiray: We might have to retain the `ImageSlice` type (which was renamed to `imageSlice` and made private in commit 2f72a0c4bda3da83390ec24f90c24aa77de4b4f0).  Detailed explanation below.

src/pkg/podman/podman.go
```diff
@@ -189,46 +104,51 @@ func GetContainers(args ...string) (*Containers, error) {
...
+
+		flattenedImages := image.flattenNames(fillNameWithID)
+		retImages = append(retImages, flattenedImages...)
 	}
 
-	sort.Sort(imageSlice(toolbxImages))
-	return toolbxImages, nil
+	ret := Images{retImages, 0}
+	if sortByName {
+		sort.Sort(ret)
```
debarshiray: I had originally introduced the `podman.Containers` type (in commit e611969726ed4260fbbbe97396cf9fc656ea0862), similar to the `podman.Images` type in this pull request, to be an immutable container whose contents can't be modified.

I was worried about passing around a slice because in Go [slices](https://go.dev/tour/moretypes/8) are references to a underlying arrays.  Changing the elements of a slice modifies the corresponding elements of its underlying array.  Other slices that share the same underlying array will see those changes.   So, it's theoretically possible that after copying and slicing a slice several times, a piece of code can unexpectedly change the elements to the surprise of a different piece of code.  I wanted to prevent this.

Following that train of thought, and the current realities of code, it doesn't look like we need to sort a collection of images outside the `podman` package.  So, it might be better if `podman.Images` didn't implement the [sort.Interface](https://pkg.go.dev/sort#Interface) interface, and a different private type (such as `imageSlice`) implemented it.

src/pkg/podman/podman.go
```diff
@@ -270,31 +190,30 @@ func GetFullyQualifiedImageFromRepoTags(image string) (string, error) {
...
 			return "", fmt.Errorf("failed to inspect image %s", image)
 		}
 
-		if info["RepoTags"] == nil {
+		if imageObj.RepoTags() == nil {
 			return "", &ImageError{image, ErrImageRepoTagsMissing}
 		}
 
-		repoTags := info["RepoTags"].([]interface{})
+		repoTags := imageObj.RepoTags()
```
debarshiray: Nitpick: we can do this a few lines earlier to prevent a second invocation of the `RepoTags()` method.

debarshiray (changes requested):
I noticed a few more relatively minor things while trying to rebase this against `main`.

debarshiray:
I rebased this branch on top of `main`, and squashed the two commits together so that the tests stay as close as possible to the code being tested.

src/pkg/podman/image.go
```diff
@@ -0,0 +1,258 @@
...
+
+func (image *imageImages) Labels() map[string]string {
+	return image.labels
+}
+
+func (image *imageImages) Names() []string {
+	return image.names
+}
+
+func (image *imageImages) setNames(names []string) {
```
debarshiray: Do we really need this private setter method?  We don't have them for the other fields, and it might be overkill because these values can't be set from outside the `podman` package.

src/meson.build
```diff
@@ -23,6 +23,7 @@ sources = files(
   'pkg/nvidia/nvidia.go',
   'pkg/podman/container.go',
   'pkg/podman/errors.go',
+  'pkg/podman/image.go',
```
debarshiray, line 26: We need to add `pkg/podman/imageImages_test.go` and `pkg/podman/imageInspect_test.go`.

src/pkg/podman/image.go
```diff
@@ -0,0 +1,258 @@
...
+
+	"github.com/containers/toolbox/pkg/utils"
+)
+
+type Image interface {
+	Created() string
+	ID() string
+	IsToolbx() bool
+	Labels() map[string]string
+	Names() []string
```
debarshiray, line 31: I am wondering if we can have a `Name()` variant (ie., singular) of the `Names()` method.  In practical terms, it would simplify the rest of the code because we almost always want `Names()[0]`, except in `imageImages.flattenNames`.

In philosophical terms, I am a bit worried about returning a slice because in Go [slices](https://go.dev/tour/moretypes/8) are references to a underlying arrays.  Changing the elements of a slice modifies the corresponding elements of its underlying array.  Other slices that share the same underlying array will see those changes.   So, it's theoretically possible that after copying and slicing a slice several times, a piece of code can unexpectedly change the elements to the surprise of a different piece of code.

One way to avoid that is to return a copy of the internal slice using [copy](https://pkg.go.dev/builtin#copy).

I had originally added the `Names()` method to the `podman.Containers` type (in commit e611969726ed4260fbbbe97396cf9fc656ea0862) purely for the sake of completeness.  So, maybe it's simpler to remove the method?  What do you think?

One inconvenient detail is that we have assertions that check the number of elements in `Names()` to ensure that the images were flattened.  Even if we retain the `Names()` method by returning a copy of the slice, it would be good to avoid the copies just for assertions.

I wonder if we can move the assertions to the `Name()` variant (ie., singular).  Now that `getImages()` is removed and everything happens in `podman.GetImages()`, it seems that all images should be already flattened before any code outside this function sees them, and `podman.GetImages()` doesn't use the names.  What do you think?

debarshiray (changes requested):

src/pkg/podman/podman.go
debarshiray: I took the liberty to do this.

src/pkg/podman/podman.go
debarshiray: I took the liberty to do this.

src/meson.build
debarshiray: I took the liberty to do this.

src/pkg/podman/image.go
DaliborKr: Yes, you are right. This setter function is really unnecessary here. In other functions in `image.go` (e.g., `UnmarshalJSON()`), fields are set directly so that we can do the same here.

src/pkg/podman/image.go
DaliborKr: This is a very good idea! We can have a method `Name()` like the example below, which also encapsulates the assertion, and use the `Names()` method only to flatten the names within the `podman` package, ensuring that only flattened images are returned from its functions. 

```
func (image *imageImages) Name() string {
    if len(image.names) != 1 {
        panic("cannot use Name() on unflattened Image")
    }
    return image.names[0]
}
```


Did you mean something like this?

src/pkg/podman/image.go
DaliborKr: Applied in commit d197ff438fb13a3cd9750ca5d71d044ebd56c51a

src/pkg/podman/image.go
DaliborKr: Applied in commit d197ff438fb13a3cd9750ca5d71d044ebd56c51a

src/pkg/podman/image.go
```diff
@@ -0,0 +1,261 @@
...
+		RepoTags []string
+	}
+
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	// Until Podman 2.0.x the field 'Created' held a human-readable string in
+	// format "5 minutes ago". Since Podman 2.1 the field holds an integer with
+	// Unix time. Go interprets numbers in JSON as float64.
```
debarshiray, line 204: On the surface, this comment about the field `Created` holding a human-readable string like *5 minutes ago* until Podman 2.0.x doesn't match the reality exposed by the tests.  :)

Speaking from memory, I think there are three cases.  Older than Podman 2.0, the Podman 2.0.x releases, and Podman 2.1 onwards.  We have to check this because I am not entirely sure.

Anyway, it's not the fault of this PR.  So there's no need to block on fixing this.  It's a note to ourselves to do something about this dissonance.  :)

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+			name: "podman 1.1.2, fedora-toolbox:29",
+			data: "" +
+				"[" +
+				"	{" +
+				"		\"id\": \"4a6adf1f2a96adf5ea0c02b61f9fa574306f77fc522f39c2ce6bb164daead882\"," +
+				"		\"names\": [" +
+				"			\"registry.fedoraproject.org/f29/fedora-toolbox:29\"" +
+				"		]," +
+				"		\"digest\": \"sha256:f324a546a5e894af041eea47f8f2392bf2a9e2d88ee77199b25a129174b7a0e1\"," +
+				"		\"created\": \"2019-11-06T11:36:10.586442Z\"," +
```
debarshiray: The `created` field in Podman older than 2.0.x holds an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formatted string, not something like *5 minutes ago*.

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+				"		\"Names\": [" +
+				"			\"registry.fedoraproject.org/fedora-toolbox:32\"" +
+				"		]," +
+				"		\"Digest\": \"sha256:dd65063ba3ee25991de8e3e13b6a44b67e4d1d4f4875153591c55f2f0cb78515\"," +
+				"		\"Digests\": [" +
+				"			\"sha256:dd65063ba3ee25991de8e3e13b6a44b67e4d1d4f4875153591c55f2f0cb78515\"," +
+				"			\"sha256:1d1077ce12b6d45990709ab81543884838d71a3b2014d023b82fd465a3ce4e23\"" +
+				"		]," +
+				"		\"Created\": 1619352159," +
+				"		\"CreatedAt\": \"2021-04-25T12:02:39Z\"" +
```
debarshiray: Then, `Created` started holding a UNIX timestamp, and the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formatted string moved to `CreatedAt`.

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+						"com.github.containers.toolbox":  "true",
+						"com.github.debarshiray.toolbox": "true",
+						"com.redhat.build-host":          "osbs-node02.iad2.fedoraproject.org",
+						"com.redhat.component":           "fedora-toolbox",
+						"distribution-scope":             "public",
+						"license":                        "MIT",
+						"maintainer":                     "Debarshi Ray \u003crishi@fedoraproject.org\u003e",
+						"name":                           "fedora-toolbox",
+						"release":                        "14",
+						"summary":                        "Base image for creating Fedora toolbox containers",
```
debarshiray: Nitpick: we can omit some of these unimportant labels that make the lines too wide.  :)

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+				},
+			},
+			imageCnt: 1,
+		},
+		{
+			name: "podman 3.4.7, fedora-toolbox:35",
+			data: "" +
+				"[" +
+				"	{" +
+				"		\"Id\": \"862705390e8b1678bbac66beb30547e0ef59abd65b18e23ea533f059ba069227\"," +
```
debarshiray: Nitpick: we can indent the JSON with 2 spaces instead of a tab to prevent the lines from becoming too wide.

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+				"		\"ParentId\": \"\"," +
+				"		\"RepoTags\": null," +
+				"		\"RepoDigests\": [" +
+				"			\"sha256:b5f419099423fae80421cda692b00e301894101575806f44558e6c9c911769e6\"," +
+				"			\"sha256:bdab10512284f81480235e89939babff43b2a8ca5af870dbebdeb127c6533568\"" +
```
debarshiray: Nitpick: we can drop the `RepoDigests` because we don't use them anywhere and they make the lines too wide.

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+					names: []string{
+						"registry.fedoraproject.org/fedora-toolbox:40",
+					},
+					repoTags: nil,
+				},
+			},
+			imageCnt: 1,
+		},
+		{
+			name: "podman 5.6.1, fedora-toolbox:41",
```
debarshiray: Covering Fedora 40, 41 and 42 seems a bit wasteful because the Podman versions are likely to be similar.  Or did you have a specific reason for covering them?

There might be more value if we spaced them out.  eg., Fedora 40, 43, etc..  It's not a big deal.  We can just stay with Fedora 40 in this PR, like we have for the `containerInspect` tests, and extend the list later in future as more Fedora versions come out.

Strictly speaking, there can be some value in covering Fedora 38, 39 and 40, as three consecutive releases, because the `fedora-toolbox` images went through some major changes in each of them.  Fedora 38 was the last release where the image was built from a [Container/Dockerfile](https://src.fedoraproject.org/container/fedora-toolbox).  In Fedora 39, it moved to [fedora-kickstarts](https://pagure.io/fedora-kickstarts/) and [pungi-fedora](https://pagure.io/pungi-fedora/) as part of the [ToolbxReleaseBlocker](https://fedoraproject.org/wiki/Changes/ToolbxReleaseBlocker) Change.  In Fedora 40, it moved to [fedora-kiwi-descriptions](https://pagure.io/fedora-kiwi-descriptions) as part of the [KiwiBuiltCloudImages](https://fedoraproject.org/wiki/Changes/KiwiBuiltCloudImages) Change.

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+					names: []string{
+						"registry.fedoraproject.org/fedora-toolbox:41",
+					},
+					repoTags: nil,
+				},
+			},
+			imageCnt: 1,
+		},
+		{
+			name: "podman 5.6.1, fedora-toolbox:41, no name specified",
```
debarshiray: How was the JSON for this was generated?  Just curious.

I can think of using either `podman untag ...` to remove the name from an existing image or `podman build ...` to build the image from a Containerfile.  In both cases, it seems like the `podman images` JSON has a `Dangling: true` field, and in the latter you would have the base `fedora` image showing up in `podman images`.

Or did you use some other trick?  :)

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+			name: "podman 5.6.1, fedora-toolbox:41, multiple names specified",
+			data: "" +
+				"[" +
+				"	{" +
+				"		\"Id\": \"b65ee158f921088db52e8b98a6f3267de75324c8a6a04afc9ff095338c40e59b\"," +
+				"		\"ParentId\": \"\"," +
+				"		\"RepoTags\": null," +
+				"		\"RepoDigests\": [" +
+				"			\"registry.fedoraproject.org/fedora-toolbox@sha256:0a51adc6bab55d49ff00da8aaad81ca1f02315511ed23d55ee5bbbe1a976a663\"," +
+				"			\"registry.fedoraproject.org/fedora-toolbox@sha256:8599f0c0d421c0dc01c4b7fb1c07b2780c0ab1931d0f22dd7b6da3b93ff6b77b\"" +
```
debarshiray: While playing with this, I learnt that the `RepoDigest` entries are duplicated if the second name is something like `localhost/fedora-toolbox:41`, but not if it's `registry.fedoraproject.org/fedora-toolbox:41-test` like in this test.  I didn't expect that.  :)

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+				"			\"org.opencontainers.image.url\": \"https://fedoraproject.org/\"," +
+				"			\"org.opencontainers.image.vendor\": \"Fedora Project\"," +
+				"			\"org.opencontainers.image.version\": \"41\"," +
+				"			\"vendor\": \"Fedora Project\"," +
+				"			\"version\": \"41\"" +
+				"		}," +
+				"		\"Containers\": 1," +
+				"		\"Digest\": \"sha256:0a51adc6bab55d49ff00da8aaad81ca1f02315511ed23d55ee5bbbe1a976a663\"," +
+				"		\"History\": [" +
+				"			\"registry.fedoraproject.org/fedora-toolbox:41\"" +
```
debarshiray: Shouldn't the `History` field also have `registry.fedoraproject.org/fedora-toolbox:41-test`?

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+			assert.NoError(t, err)
+			assert.Len(t, images, tc.imageCnt)
+
+			for i, expect := range tc.expects {
+				image := images[i]
+				assert.Equal(t, expect.id, image.ID())
+				assert.Equal(t, expect.isToolbx, image.IsToolbx())
+				assert.Equal(t, expect.labels, image.Labels())
+				assert.Equal(t, expect.names, image.Names())
+				assert.Equal(t, expect.repoTags, image.RepoTags())
```
debarshiray: I wonder if there's an easy way to test the `Created()` method.  It can be tricky because the relative time difference between the timestamp in the JSON and when the tests are run will keep increasing.  So, *5 minutes ago* will become *5 days ago*, etc.

I think we hit the same problem with tests for `containerInspect` and that's why it's not tested there either.  We can maybe simplify things a bit by using the same timestamp in the JSON for all the test cases.

It's not a big deal.  Just a thought.  I didn't investigate this much.  It's not a blocker for this PR.

debarshiray (commented):
I finally finished going through all the test cases for parsing the `podman images --format json` output.  The tests have revealed some mistakes elsewhere in the code, and I have one or two minor comments and questions about the changes here in this PR.

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+				"		\"History\": [" +
+				"			\"registry.fedoraproject.org/fedora-toolbox:41\"" +
+				"		]," +
+				"		\"Names\": [" +
+				"			\"registry.fedoraproject.org/fedora-toolbox:41-test\"," +
+				"			\"registry.fedoraproject.org/fedora-toolbox:41\"" +
+				"		]," +
+				"		\"Created\": 1759729685," +
+				"		\"CreatedAt\": \"2025-10-06T05:48:05Z\"" +
+				"	}" +
```
debarshiray: Finally, shouldn't there be two identical images in the `podman images` JSON when an image has two names?

src/pkg/podman/imageImages_test.go
```diff
@@ -0,0 +1,792 @@
...
+				"		\"created\": \"2020-03-03T10:32:20.503064Z\"," +
+				"		\"size\": 404259600," +
+				"		\"readonly\": false," +
+				"		\"history\": []" +
+				"	}" +
+				"]",
+			expects: []expect{
+				{
+					id:       "c49513deb6160607504d2c9abf9523e81c02f69ea479fd07572a7a32b50beab8",
+					isToolbx: false,
```
debarshiray, line 102: This made me realize that our claim of supporting Podman 1.6.4 is invalid.  :)

Due to commit [2369da5d31830e5c](https://github.com/containers/toolbox/commit/2369da5d31830e5cd3e1a13857a686365875ae61), Toolbx depends on the `podman images` JSON containing the labels, and this test case shows that they are not there with Podman 1.8.0.  :D

Anyway, it's not the fault of this PR. So there's no need to block on fixing this. It's a note to ourselves to do something about this dissonance. :)

debarshiray:
To keep things moving, I have split the changes introducing the `Image` interface and the `imageImages` concrete type, and the associated tests into https://github.com/containers/toolbox/pull/1779

src/pkg/podman/imageImages_test.go
debarshiray: I took the liberty to change this.

src/pkg/podman/imageImages_test.go
debarshiray: I took the liberty to change this.

src/pkg/podman/imageImages_test.go
debarshiray: I took the liberty to add a `Dangling: true` field.

src/pkg/podman/imageImages_test.go
debarshiray: I took the liberty to change the test cases using Fedora 41 and 42 to use Fedora 40, and added one with Fedora 39.

debarshiray:
> To keep things moving, I have split the changes introducing the `Image` interface and the `imageImages` concrete type, and the associated tests into #1779

I took the liberty to rebase this PR on top of https://github.com/containers/toolbox/pull/1779 , which has now been merged into `main`.

src/pkg/podman/image.go
```diff
@@ -37,10 +38,21 @@ type Images struct {
...
+	id       string
+	labels   map[string]string
+	names    []string
+	repoTags []string
+}
+
+type imageInspect struct {
+	created      string
+	entrypoint   []string
+	envVars      []string
```
debarshiray, line 51: It looks like `entrypoint` and `envVars` are not needed to parse the output of `podman inspect --type image` into an `Image` implementation.  I suppose you needed them for the changes in https://github.com/containers/toolbox/pull/1707 to address https://github.com/containers/toolbox/issues/1622 , right?  If so, it might be better to move these parts to the pull request that needs them.

More importantly, don't we need to add methods to the `Image` interface to cover these?

debarshiray (commented):
Now onto the, hopefully, final round involving the `imageInspect` concrete type.  :)

src/pkg/podman/image.go
DaliborKr: Yes, you are right. I prepared fields `entrypoint` and `envVars` in advance for #1707, so there is no use case for them right now. We can get rid of them now, and we can add them once we rewrite #1707. 

I am not sure about adding methods for them to the `Image` interface, since `imageImages`, which parses the `podman images` command introduced in #1779, does not provide `entrypoint` and `envVars` information.

DaliborKr: I checked it out to be sure, and it is true. The `podman images --format json` does not provide properties for `entrypoint` and `envVars`.

src/pkg/podman/image.go
```diff
@@ -37,10 +38,21 @@ type Images struct {
...
+	id           string
+	labels       map[string]string
+	namesHistory []string
+	repoTags     []string
```
debarshiray, line 53: While playing with the `podman inspect` JSON for images without names, it seems like the `RepoTags` field should be used for `Names()`, not `NamesHistory`.  eg., in such images `RepoTags` is always empty and `NamesHistory` always has some values.

Anyway, let's not block this pull request on this issue.  The existing code was ambiguous and it will be better to have this clarified before choosing to use `RepoTags` for `Names()`.

PR #1729, opened 2025-11-10
fix: ensure -zlazy for systems that default to -znow

xandris:
I see #1706 and #1722. this is more minimal change for systems that default to znow

### DIFF
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -116,7 +116,7 @@ dynamic_linker="/run/host$dynamic_linker_canonical_dirname/$dynamic_linker_basen
 go build \
         $tags \
         -trimpath \
-        -ldflags "-extldflags '-Wl,-dynamic-linker,$dynamic_linker -Wl,--export-dynamic -Wl,-rpath,/run/host$libc_dir_canonical_dirname -Wl,--unresolved-symbols,ignore-in-object-files' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$4" \
+        -ldflags "-extldflags '-Wl,-dynamic-linker,$dynamic_linker,--export-dynamic,-rpath,/run/host$libc_dir_canonical_dirname,--unresolved-symbols,ignore-in-object-files' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$4" \
         -o "$2/$3"
 
 exit "$?"
--- a/src/go-build-wrapper
+++ b/src/go-build-wrapper
@@ -116,7 +116,7 @@ dynamic_linker="/run/host$dynamic_linker_canonical_dirname/$dynamic_linker_basen
 go build \
         $tags \
         -trimpath \
-        -ldflags "-extldflags '-Wl,-dynamic-linker,$dynamic_linker,--export-dynamic,-rpath,/run/host$libc_dir_canonical_dirname,--unresolved-symbols,ignore-in-object-files' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$4" \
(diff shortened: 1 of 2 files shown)


silverhadch:
Ping.

src/go-build-wrapper
```diff
@@ -116,7 +116,7 @@ dynamic_linker="/run/host$dynamic_linker_canonical_dirname/$dynamic_linker_basen
 go build \
         $tags \
         -trimpath \
-        -ldflags "-extldflags '-Wl,-dynamic-linker,$dynamic_linker -Wl,-rpath,/run/host$libc_dir_canonical_dirname -Wl,--export-dynamic -Wl,--unresolved-symbols=ignore-in-object-files' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$4" \
+        -ldflags "-extldflags '-Wl,-z,lazy,-dynamic-linker,$dynamic_linker,-rpath,/run/host$libc_dir_canonical_dirname,--export-dynamic,--unresolved-symbols=ignore-in-object-files' -linkmode external -X github.com/containers/toolbox/pkg/version.currentVersion=$4" \
```
debarshiray: I suspect we have a typo here that explains the CI failures.

Shouldn't it be `-Wl,-dynamic-linker,$dynamic_linker -Wl,-z,lazy`?

debarshiray (changes requested):
Thanks for looking into this, @xandris , and my apologies for the delay!  I always shudder at this linker hackery.  :)

debarshiray:
recheck

debarshiray (commented):
It looks like this fixes https://github.com/containers/toolbox/issues/1706  Could you please link to the issue instead of the PR in the commit message?

src/go-build-wrapper
xandris: it's okay to stack up linker arguments in `-Wl`. the commas just separate successive argv elements that ld receives, so it's not necessary to have multiple `-Wl` flags... but it doesn't hurt either

i think the problem was actually a missing comma in my change to `src/meson_generate_completions.py`! that caused Python to concatenate the `lazy` with the `.` that was intended for `go run`. sorry about that

debarshiray: Oops!  Yes, you are right.  I shouldn't do fly-by reviews so late at night.

debarshiray:
Thanks for updating this PR, @xandris !  Let me try to rebase this against `main`.  I am temporarily marking this as a draft while I do so.

src/go-build-wrapper
debarshiray: I took the liberty to split the stacking up of external linker flags into a separate commit.  These build details are hairy and the changes are hard to read, so this will make it easier for our (mainly my) future selves.  :)

debarshiray:
I took the liberty to dump all the references you dug up for me in https://github.com/containers/toolbox/issues/1706 in the commit message for the sake of my future self.

xandris:
:eyes:

debarshiray:
> Build failed. https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/local/buildset/2c9465a0661544b89f4c3822984dc1fe
> 
> ❌ [system-test-fedora-rawhide-commands-options ](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/local/build/a5430751ccf9421b8e0ec44da70f4e63) **TIMED_OUT** in 2h 00m 26s
> ❌ [system-test-fedora-44-commands-options ](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/local/build/4895bed699e047b5bf406f78340286d4) **TIMED_OUT** in 1h 45m 29s

These are the same CI failures that showed up recently and is being debugged.  See https://github.com/containers/toolbox/pull/1802 and https://github.com/containers/toolbox/pull/1805

So, let's ignore them here.

debarshiray:
Thanks for all the insight and investigation, @xandris !

Issue #1348, opened 2023-08-14
Toolbox can't start if user has any group-only permissions for device nodes in folders


mtalexan:
**Describe the bug**
If a user is a member of any group that grants access to device nodes that have `750` permissions within folders, `toolbox` can no longer `enter`.

For example when VirtualBox is installed, the `/dev/vboxusb/*` device nodes are created with `750` permissions on both the folder and the files, and they're owned by the `vboxusers` group (this behavior is consistent with standards for device node permissions, and is like the `/dev/ttyUSB*` nodes on Fedora systems). Toolbox maps the entire `/dev` into the container, which uses the user permissions from outside the container to map all visible files into the container.  The `crun` mapping of sub-folders uses the permissions inside the container to create a matching folder however.  This causes an error because the folder has permissions granted only via an unmapped group.  

TL;DR  
Blanket mapping of `/dev` into the container without also mapping all the user's groups breaks in very common conditions.

**Steps how to reproduce the behaviour**
1. Create device node in a folder with `750` permissions, owned by a group the user is part of but not by the user directly.
    1. `sudo mkdir -Z -m 750 /dev/testFolder`
    2. `sudo mkdnod -Z -m 750 /dev/testFolder/testNode b 7 30` (loop device 30)
    3. `sudo chown -R root:dialout /dev/testFolder` (`dialout` used for convenience rather than 
creating a new group)
    4. `groups`, and confirm user is part of `dialout` group
    5. `ls /dev/testFolder`, confirm the user can see the files using the `dialout` group permissions
2. `toolbox create`
3. `toolbox enter`

**Expected behaviour**
Toolbox is entered successfully.

**Actual behaviour**
Error, unable to enter toolbox.
```
Error: unable to start container XXXX: crun: mkdir `/dev/testFolder`: Permission denied: OCI permission denied
```

**Screenshots**

N/A

**Output of `toolbox --version` (v0.0.90+)**
```
toolbox version 0.0.99.4
```

**Toolbox package info (`rpm -q toolbox`)**
```
toolbox version 0.0.99.4
```

**Output of `podman version`**
```
Client:       Podman Engine
Version:      4.6.0
API Version:  4.6.0
Go Version:   go1.20.6
Built:        Fri Jul 21 08:23:26 2023
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
```
podman-4.6.0-1.fc38.x86_64
```

**Info about your OS**
Fedora Kinoite 38

[...]

Issue #1297, opened 2023-05-14
toolbox enter failed with "Error: failed to start container fedora-toolbox-37"


coiby:
**Describe the bug**

`toolbox enter` failed with "Error: failed to start container fedora-toolbox-37".

**Steps how to reproduce the behaviour**
1.  dnf install VirtualBox
2. toolbox create
3. toolbox enter
 
**Expected behaviour**
`toolbox enter` should work.

**Actual behaviour**

`toolbox enter` failued.

**Screenshots**

**Output of `toolbox --version` (v0.0.90+)**
`toolbox version 0.0.99.4
`

**Toolbox package info (`rpm -q toolbox`)**
e.g., `toolbox-0.0.99.4-1.fc37.x86_64`

**Output of `podman version`**
```
Client:       Podman Engine
Version:      4.5.0
API Version:  4.5.0
Go Version:   go1.19.7
Built:        Fri Apr 14 23:42:56 2023
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
`podman-4.5.0-1.fc37.x86_64`

**Info about your OS**
e.g., Fedora 37

**Additional context**

1. journalctl shows the following logs with `toolbox enter -vv`
```
conmon[3469]: conmon d5bc67555f5777078a4d <nwarn>: runtime stderr: mkdir `/dev/vboxusb/003`: Permission denied
conmon[3469]: conmon d5bc67555f5777078a4d <error>: Failed to create container: exit status 1
```
2. `sudo chmod 755 /dev/vboxusb/003`  (suggested by [OCI Permission Denied when running on host with VirtualBox · Issue #995 · containers/crun](https://github.com/containers/crun/issues/995)) or remove current user from the vboxusers group could workaround this issue.
3. `podman --group-add keep-groups` seems to also work according to [`podman run` fails with Permission denied: OCI permission denied · Issue #14284 · containers/podman](https://github.com/containers/podman/issues/14284). But I don't know how to let toolbox call podman with ` --group-add keep-groups`.

mtalexan:
This prior issue seems to have been closed incorrectly, and is a duplicate: https://github.com/containers/toolbox/issues/247

debarshiray:
Duplicate of https://github.com/containers/toolbox/issues/1348

Issue #247, opened 2019-09-03
toolbox: failed to create container: open /dev/vboxusb: permission denied


zapateo:
When I run the command

```
$ toolbox create -c mycontainer
```

I get the output: 

```
toolbox: failed to create container mycontainer
```

Running `toolbox` with the `-v` option I get

```
$ toolbox -v create -c mycontainer
toolbox: resolved absolute path for /usr/bin/toolbox to /usr/bin/toolbox
toolbox: checking if /etc/subgid and /etc/subuid have entries for user gianluca
toolbox: TOOLBOX_PATH is /usr/bin/toolbox
toolbox: checking if 'podman system migrate' exists
toolbox: migration not needed: 1.5.1 is unchanged
toolbox: Fedora generational core is f30
toolbox: base image is fedora-toolbox:30
toolbox: container is mycontainer
toolbox: checking value /var/run/.heim_org.h5l.kcm-socket (Stream) of property Listen in sssd-kcm.socket
toolbox: parsing value /var/run/.heim_org.h5l.kcm-socket (Stream) of property Listen in sssd-kcm.socket
toolbox: checking if 'podman create' supports --dns=none and --no-hosts
toolbox: 'podman create' supports --dns=none and --no-hosts
toolbox: looking for image localhost/fedora-toolbox:30
toolbox: looking for image registry.fedoraproject.org/f30/fedora-toolbox:30
toolbox: base image fedora-toolbox:30 resolved to registry.fedoraproject.org/f30/fedora-toolbox:30
toolbox: checking if container mycontainer already exists
toolbox: /home/gianluca canonicalized to /home/gianluca
toolbox: checking if /home is a symbolic link to /var/home
toolbox: calling org.freedesktop.Flatpak.SessionHelper.RequestSession
toolbox: creating container mycontainer
Error: open /dev/vboxusb: permission denied
toolbox: failed to create container mycontainer
```

The problem seems to be in line: **Error: open /dev/vboxusb: permission denied**

**System informations**

- Toolbox version 0.0.12
- Podman version 1.5.1

HarryPhoon:
Hi @zapateo, thanks for the ticket. Did this issue appear after you updated your system? Also you are using classic Fedora Workstation, right?

zapateo:
`toolbox` used to work properly on my system; after that, I updated/installed/removed software many times before finding out that it stopped working, so I can't establish what broke it.

Yes, I'm using Fedora Workstation.

wesleysliao:
[...]

PR #1647, opened 2025-05-15
Pass supplemental groups when Podman is using crun

mwinters0:
This solves the common issue where `/dev/vboxusb/*` is owned by group `vboxusers`, causing volume mounts to fail when running `toolbox enter`. Same is true for any other mapped dirs with supplemental group owners.  This is a common issue, frequently reported by its rather cryptic symptoms.

Tested on Fedora Workstation 42.

Fixes https://github.com/containers/toolbox/issues/1589
Probably https://github.com/containers/toolbox/issues/1640

And maybe:
https://github.com/containers/toolbox/issues/1310
https://github.com/containers/toolbox/issues/1085
https://github.com/containers/toolbox/issues/569

---

**Note:**
Podman option `keep-groups` was created in https://github.com/containers/podman/commit/e356160f415b6111df09af214f0dea299e78ad04 and first released in [v3.2.0](https://github.com/containers/podman/releases/tag/v3.2.0) (2021-06-03).

PR #1732, opened 2025-12-10
cmd/create: keep host supplementary groups with --group-add keep-groups

Knogle:
Toolbx currently fails to create or start containers on hosts where the user has group-only access to device nodes under /dev, for example when VirtualBox creates /dev/vboxusb/* owned by root:vboxusers with 0750 permissions. In this situation a rootless Toolbx container inherits /dev from the host, the /dev/vboxusb directory is visible, but the container process has no group membership for vboxusers. When the OCI runtime walks /dev it eventually tries to open or create device nodes under /dev/vboxusb and fails with an OCI permission denied error, so `toolbox enter` aborts with "failed to start container".

This behaviour is described in #1348, which generalizes the problem to any directory under /dev with 0750 permissions where the user only has access via a supplementary group (eg. dialout, vboxusers, docker). It is also the root cause behind reports like #1297 and the older #247, and matches similar issues seen with rootless Podman and VirtualBox in other projects (distrobox, crun, Debian bug reports).

Rootless Podman already has a mechanism to address this class of problems: passing `--group-add keep-groups` to `podman create` or `podman run` tells Podman to keep the caller's supplementary groups instead of dropping them during container setup. The Podman documentation explicitly recommends this flag for cases where device access is granted only via group membership and notes that otherwise accessing such devices from rootless containers will fail with permission errors.

This patch wires that mechanism into Toolbx by adding `--group-add keep-groups` to the `podman create` invocation used by the `toolbox create` command when running as a non-root user and when the Podman version is >= 3.2.0, which is when the flag became available. The call is gated through the existing podman.CheckVersion helper, so older Podman versions are not affected and keep the previous behaviour.

With this change, the OCI runtime sees the same supplementary groups as the host process, so a user who is a member of vboxusers (or dialout, docker, etc.) can access devices that are only group readable from inside a rootless Toolbx container. On systems with VirtualBox installed this fixes errors of the form:

  crun: creating `/dev/vboxusb/00x/00y`: openat2 `dev/vboxusb`:
  Permission denied: OCI permission denied

when entering a freshly created Toolbx container.

This patch intentionally does not change how /dev is mounted into the container and does not implement the more invasive /dev filtering approach discussed in #1348. It is a minimal, backwards compatible improvement that fixes the common VirtualBox and other group-only device cases by leveraging Podman's existing `--group-add keep-groups` support.

Tests:

 * On Fedora 42 with VirtualBox installed and the user in the vboxusers group, verified that `toolbox create` logs the `podman create` command including `--group-add keep-groups`.
 * Confirmed that `toolbox enter` succeeds for a new container where it previously failed with an OCI permission denied on /dev/vboxusb.
 * Verified that behaviour is unchanged on a system without VirtualBox and that the flag is not added when running as root.

Fixes: #1348
Related: #1297, #247

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -441,8 +441,8 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 		"--dns", "none",
 	}
 
-        // Add '--group-add keep-groups' when available, so the container inherits
-        // the host user's supplementary groups (eg. vboxusers).
+	// Add '--group-add keep-groups' when available, so the container inherits
+	// the host user's supplementary groups (eg. vboxusers).
 	createArgs = append(createArgs, keepGroups...)
 	createArgs = append(createArgs, toolbxDelayEntryPointEnv...)
 	createArgs = append(createArgs, toolbxFailEntryPointEnv...)


debarshiray:
recheck

Knogle:
> recheck

I think it's gone again :D

petejohanson:
I've just tested this locally, on a new Silverblue install, by building within a container, then running the new executable from my host to create a new container with the extra flags, and I can confirm now I can access things like the UF2 volume mount from a connected RP2040 in bootloader mode without issue. Would be really nice to get this merged, and I'm happy to help with whatever is necessary to do so.

(edit: we'll see if I've got the juice to kick off a recheck?! Would be surprised if I did)

petejohanson:
recheck

petejohanson:
The main caveat that *may* apply, cribbed from https://www.redhat.com/en/blog/files-devices-podman:

> If you use the --group-add keep-groups call, you cannot set other groups within the container. Instead, the container can only inherit the parent's groups.

Not sure if this means we might want to make this somehow optional?

petejohanson:
Gotcha: In the new containers, *wheel* is no longer mapped properly and as a result I can't sudo.

e.g. in a container created with this branch:

````
⬢ [peter@toolbx toolbox]$ sudo id
peter is not in the sudoers file.
⬢ [peter@toolbx toolbox]$ groups
peter nobody
```

Versus current release:

```
⬢ [peter@toolbx metro_rp2040]$ groups
peter wheel
⬢ [peter@toolbx metro_rp2040]$ sudo id
uid=0(root) gid=0(root) groups=0(root)
```

Knogle:
> Gotcha: In the new containers, _wheel_ is no longer mapped properly and as a result I can't sudo.
> 
> e.g. in a container created with this branch:
> 
> ```
> ⬢ [peter@toolbx toolbox]$ sudo id
> peter is not in the sudoers file.
> ⬢ [peter@toolbx toolbox]$ groups
> peter nobody
> ```
> 
> Versus current release:
> 
> ```
> ⬢ [peter@toolbx metro_rp2040]$ groups
> peter wheel
> ⬢ [peter@toolbx metro_rp2040]$ sudo id
> uid=0(root) gid=0(root) groups=0(root)
> ```
> ```

Can you check this one as well?

https://github.com/containers/toolbox/pull/1647

'su' should work though

Knogle:
Check this out, should also work and i am now using a different approach.

https://github.com/containers/toolbox/pull/1778

PR #1733, opened 2025-12-17
.zuul: Drop testing on Fedora 41

debarshiray:
Fedora 41 reached End of Life on 15th December 2025:
https://docs.fedoraproject.org/en-US/releases/eol/


debarshiray:
This is marked as a draft because Fedora 41 hasn't yet been added to https://docs.fedoraproject.org/en-US/releases/eol/

debarshiray:
> This is marked as a draft because Fedora 41 hasn't yet been added to https://docs.fedoraproject.org/en-US/releases/eol/

This has been done now.

PR #1735, opened 2026-01-10
fix(cmd): ensure rm and rmi exit with errors on failure

kairosci:
This PR ensures that `toolbox rm` and `toolbox rmi` exit with a non-zero status code if any error occurs during the removal process.

Previously, these commands would print an error message but continue execution and exit with 0, which could be misleading for scripts and automation.


kairosci:
@debarshiray PTAL, thanks!

PR #1741, opened 2026-01-26
[release-0.0.99.5] Prepare stable branch (part 3)

debarshiray:
... for CVE-2025-65637 or GHSA-4f99-4q7p-p3gh.

https://github.com/containers/toolbox/security/dependabot/26

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -51,7 +51,7 @@
 - job:
     name: system-test-fedora-rawhide
     description: Run Toolbx's system tests in Fedora Rawhide
-    timeout: 7800
+    timeout: 10800
     nodeset:
       nodes:
         - name: fedora-rawhide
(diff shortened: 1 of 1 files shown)


debarshiray:
> My review has identified a critical issue where a helper function `cleanup_containers` is removed, but one of the test files (`test/system/101-create.bats`) still calls it in its `teardown` block, which will cause tests to fail.

Oops!  We need to include commit 55c0e6378657757c305334adaead804c487817b6 from the `main` branch.

> I've also found several typos in the job descriptions within the `.zuul.yaml` configuration file.

These are false positives about the missing *o* in Toolbx, because the project was renamed as Toolbx.

debarshiray:
recheck

debarshiray:
recheck

Issue #1615, opened 2025-01-30
failed to get the current user's default shell


jele84:
**Describe the bug**
Toolbox won't start 

**Steps how to reproduce the behaviour**
1. Use Fedora CoreOS 38
2. Enter toolbox create
3. See the error (Additional context)

**Expected behaviour**
successful launch

**Actual behaviour**
Error: failed to get the current user's default shell

**Screenshots**

**Output of `toolbox --version` (v0.0.90+)**
toolbox version 0.0.99.4

**Toolbx package info (`rpm -q toolbox`)**
toolbox-0.0.99.4-1.fc38.x86_64

**Output of `podman version`**
```
Client:       Podman Engine
Version:      4.5.1
API Version:  4.5.1
Go Version:   go1.20.4
Built:        Fri May 26 20:58:48 2023
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
podman-4.5.1-1.fc38.x86_64

**Info about your OS**
Fedora CoreOS 38.20230625.3.0

**Additional context**
Please help me figure out why toolbox does not start

# toolbox create --container test --log-level debug
DEBU Running as real user ID 0                    
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Migrating to newer Podman                    
DEBU Toolbox config directory is /root/.config/toolbox 
DEBU Current Podman version is 4.5.1              
DEBU Creating runtime directory /run/toolbox      
DEBU Old Podman version is 4.5.1                  
DEBU Migration not needed: Podman version 4.5.1 is unchanged 
DEBU Setting up configuration                     
DEBU Setting up configuration: file /root/.config/containers/toolbox.conf not found 
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-38'               
DEBU Image: 'fedora-toolbox:38'                   
DEBU Release: '38'                                
DEBU Resolving container and image names          
DEBU Container: 'test'                            
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): 'registry.fedoraproject.org/fedora-toolbox:38' 
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
[...]

PR #1760, opened 2026-02-19
cmd: fallback to parsing `getent passwd` output when $SHELL is not set

Rolv-Apneseth:
As suggested in #1615, this PR just adds a fallback if $SHELL is not set for whatever reason.

Not very comfortable with Go, but hopefully the code is alright. Happy to make any changes anyway. And let me know if you want some tests added, I just wasn't sure entirely where to add that.

### DIFF
--- a/src/cmd/utils.go
+++ b/src/cmd/utils.go
@@ -31,6 +31,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/containers/toolbox/pkg/shell"
 	"github.com/containers/toolbox/pkg/utils"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
@@ -381,6 +382,44 @@ func getCurrentUserHomeDir() string {
 	logrus.Debug("Using user.Current() instead")
 
 	return currentUser.HomeDir
+}
+
+func getCurrentUserShell() (string, error) {
+	if currentUser == nil {
+		panic("current user unknown")
+	}
+
+	if userShell := os.Getenv("SHELL"); userShell != "" {
+		return userShell, nil
+	}
+
+	logrus.Debug("Getting the current user's login shell: failed to read SHELL")
+	logrus.Debug("Using 'getent passwd' instead")
+
+	var stderr strings.Builder
+	var stdout strings.Builder
+
+	if err := shell.Run("getent", nil, &stdout, &stderr, "passwd", currentUser.Uid); err != nil {
+		errString := stderr.String()
+		logrus.Debugf("Getting the current user's login shell failed: %s", errString)
+		return "", fmt.Errorf("failed to get the current user's login shell: %w", err)
+	}
+
+	output := stdout.String()
+	passwdLine := strings.TrimSpace(output)
+	if len(passwdLine) == 0 {
+		return "", errors.New("failed to get the current user's login shell: no getent(1) output")
+	}
+
+	passwdLineParts := strings.Split(passwdLine, ":")
+	passwdLinePartsCount := len(passwdLineParts)
+	if passwdLinePartsCount != 7 {
+		logrus.Debugf("Getting the current user's login shell: failed to parse getent(1) output: %s",
+			passwdLine)
+		return "", errors.New("failed to get the current user's login shell: invalid getent(1) output")
+	}
+
+	return passwdLineParts[passwdLinePartsCount-1], nil
 }
 
 func getUsageForCommonCommands() string {
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -404,9 +404,9 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 
 	logLevelString := podman.LogLevel.String()
 
-	userShell := os.Getenv("SHELL")
-	if userShell == "" {
-		return errors.New("failed to get the current user's login shell")
+	userShell, err := getCurrentUserShell()
+	if err != nil {
+		return err
 	}
 
 	entryPoint := []string{
--- a/src/cmd/enter.go
+++ b/src/cmd/enter.go
@@ -114,9 +114,9 @@ func enter(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	userShell := os.Getenv("SHELL")
-	if userShell == "" {
-		return errors.New("failed to get the current user's login shell")
+	userShell, err := getCurrentUserShell()
+	if err != nil {
+		return err
 	}
 
 	command := []string{userShell, "-l"}
--- a/src/cmd/rootMigrationPath.go
+++ b/src/cmd/rootMigrationPath.go
@@ -61,9 +61,9 @@ func rootRunImpl(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	userShell := os.Getenv("SHELL")
-	if userShell == "" {
-		return errors.New("failed to get the current user's login shell")
+	userShell, err := getCurrentUserShell()
+	if err != nil {
+		return err
 	}
 
 	command := []string{userShell, "-l"}
--- a/test/system/101-create.bats
+++ b/test/system/101-create.bats
@@ -60,6 +60,35 @@ teardown() {
   assert_output "true"
 }
 
+@test "create: Smoke test with SHELL unset" {
+  local default_container
+  default_container="$(get_system_id)-toolbox-$(get_system_version)"
+
+  pull_default_image
+  unset SHELL
+
+  run --keep-empty-lines --separate-stderr "$TOOLBX" create
+
+  assert_success
+  assert_line --index 0 "Created container: $default_container"
+  assert_line --index 1 "Enter with: toolbox enter"
+  assert [ ${#lines[@]} -eq 2 ]
+  assert [ ${#stderr_lines[@]} -eq 0 ]
+
+  run podman ps --all
+
+  assert_success
+  assert_output --regexp "Created[[:blank:]]+$default_container"
+
+  run podman inspect \
+        --format '{{index .Config.Labels "com.github.containers.toolbox"}}' \
+        --type container \
+        "$default_container"
+
+  assert_success
+  assert_output "true"
+}
+
 @test "create: With a custom name (using option --container)" {
   pull_default_image
 


dustymabe:
man.. it sure would be nice if https://github.com/golang/go/issues/59121 were implemented :)

Rolv-Apneseth:
> man.. it sure would be nice if [golang/go#59121](https://github.com/golang/go/issues/59121) were implemented :)

Yeah that's a shame

src/cmd/utils.go
```diff
@@ -383,6 +383,43 @@ func getCurrentUserHomeDir() string {
...
 
+func getUserShell() (string, error) {
+	// $SHELL is supposed to get set during login
+	userShell := os.Getenv("SHELL")
+	if userShell != "" {
+		return userShell, nil
+	}
+
+	// However, fallback to parsing `getent passwd` output if it is not set for whatever reason
+	logrus.Debug("$SHELL environment variable was empty - falling back to getent passwd")
```
debarshiray: The comment looks superfluous, because the debug log says the same.  :)

src/cmd/utils.go
```diff
@@ -383,6 +383,43 @@ func getCurrentUserHomeDir() string {
...
+
+	if currentUser == nil {
+		return "", errors.New("failed to get the current user's default shell")
```
debarshiray: Did you actually hit any scenario where `currentUser` was `nil` inside this function?

I am asking because I expect `currentUser` to be set to a valid `*user.User` very early in the lifecycle of the process.  Look for the `setUpGlobals()` function.  It's called in the `init()` of the `cmd` package, which should be very early.  So, it would have to be a programmer error for `currentUser` to be `nil`.

If all that is true, then it will be better to use a `panic()` to express a mandatory pre-condition, because returning an error makes it look like a possible runtime scenario.  See the `getCurrentUserHomeDir()` function for an example.

src/cmd/utils.go
```diff
@@ -383,6 +383,43 @@ func getCurrentUserHomeDir() string {
...
+	}
+	cmd := exec.Command("getent", "passwd", currentUser.Uid)
```
debarshiray: Instead of directly using `os/exec`, it will be better to go through one of the `github.com/containers/toolbox/pkg/shell` wrappers because they hook up the standard error and output streams of the child process to `toolbox --verbose`.

Something like this should work:
```Go
import (
    ...
    "github.com/containers/toolbox/pkg/shell"
    ...
)
...
...
    var stderr strings.Builder
    var stdout strings.Builder
    if err := shell.Run("getent", nil, &stdout, &stderr, "passwd", currentUser.Uid); err != nil {
        errString := stderr.String()
        logrus.Debugf("Getting entry for user %s from the passwd database failed: %s", currentUser.Uid, errString)
        return "", fmt.Errorf("failed to get the current user's default shell: %w", err)
    }
...
```

src/cmd/utils.go
```diff
@@ -383,6 +383,43 @@ func getCurrentUserHomeDir() string {
 	return currentUser.HomeDir
 }
 
+func getUserShell() (string, error) {
+	// $SHELL is supposed to get set during login
+	userShell := os.Getenv("SHELL")
+	if userShell != "" {
```
debarshiray: I wonder if these two lines can be [combined](https://go.dev/tour/flowcontrol/6) for brevity and to restrict the scope of `userShell`.  Like:
```Go
    if userShell := os.Getenv("SHELL"); userShell != "" {
```

src/cmd/utils.go
```diff
@@ -383,6 +383,43 @@ func getCurrentUserHomeDir() string {
...
+		return "", errors.New("no output from getent")
+	}
+	split := strings.Split(trimmed, ":")
+	if len(split) < 7 {
+		return "", errors.New("could not parse shell from getent output: expected 7 fields")
+	}
+	userShell = split[len(split)-1]
+
+	// Ensure $SHELL is set moving forward
+	os.Setenv("SHELL", userShell)
```
debarshiray: Was it really necessary to set the `SHELL` environment variable?  It's a bit unexpected for a `getSomething()` function to have the side effect of setting something.  :)

If it's not really necessary, then let's not set it.  The `SHELL` environment variable is not used in any hot code path, and only once at the beginning for the `create` and `enter` commands.

debarshiray:
> The changes look mostly good to me, other than some of these details.
> 
> I have one question arising out of curiosity. Did you check that the `SHELL` environment variable is really missing from the CoreOS command line environment? I wanted to try it for myself but I never got around to it.
> 
> The only other time I saw it go missing was in the GitHub Actions workflow environment when trying to run the Toolbx tests on a Ubuntu 22.04 host. See: https://github.com/orgs/community/discussions/59413 [8c28dc2660d61025](https://github.com/containers/toolbox/commit/8c28dc2660d61025) [c22b09d095c7a8ac](https://github.com/containers/toolbox/commit/c22b09d095c7a8ac)
> 
> The log-in sequence jumps through various hoops, and our future selves will be grateful, if we documented these deviations from the usual. :)

I was beginning to wonder if `SHELL` disappears only on OpenShift pods created by `oc debug node`, but not when someone is directly logging into CoreOS without OpenShift.  However, the original issue at https://github.com/containers/toolbox/issues/1615 mentions Fedora CoreOS 38.  :)

src/cmd/utils.go
Rolv-Apneseth: No I suppose not, I just assumed it would make sense if $SHELL is also forwarded to the container itself as it might cause issues with other programs? Without that line, `env | grep SHELL` within the container has no output.

src/cmd/utils.go
Rolv-Apneseth: Sure, yes

src/cmd/utils.go
Rolv-Apneseth: Ah, didn't know about that, thank you

src/cmd/utils.go
Rolv-Apneseth: Ah, makes sense. I'll use a panic.

src/cmd/utils.go
Rolv-Apneseth: True, I'll remove it

Rolv-Apneseth:
> Let me first answer your question about how the tests work.

Thank you for the breakdown of the testing setup and writing a test! It looks good to me.

> I was beginning to wonder if `SHELL` disappears only on OpenShift pods created by `oc debug node`, but not when someone is directly logging into CoreOS without OpenShift. However, the original issue at #1615 mentions Fedora CoreOS 38. :)

I linked the other issue as it was related, and that's where your suggestion for a solution was, but I'm actually here from [this issue](https://issues.redhat.com/browse/OCPBUGS-76557), which is indeed an OpenShift debug pod. I haven't personally been able to reproduce the issue on CoreOS locally. Where would you like these occurrences documented?

And thank you for the thorough review. Hopefully I've addressed all your points but just let me know if more changes are required.

debarshiray:
> > I was beginning to wonder if `SHELL` disappears only on OpenShift pods created by `oc debug node`, but not when someone is directly logging into CoreOS without OpenShift. However, the original issue at #1615 mentions Fedora CoreOS 38. :)
> 
> I linked the other issue as it was related, and that's where your suggestion for a solution was, but I'm actually here from [this issue](https://issues.redhat.com/browse/OCPBUGS-76557), which is indeed an OpenShift debug pod.

Yes, I noticed that OpenShift issue.  I was hesitating to refer to it without confirming that it's public.  It looks like it's indeed public - I could access it from a public network without any credentials.  :)

Rolv-Apneseth:
> Thanks for the updates!  I got a bit stuck digging into how the `SHELL` environment variable gets set under different circumstances.
> 
> The changes look good to me.  I will merge it tomorrow after looking at them with a fresh pair of eyes, because it's late and I am not thinking straight.  :)

No worries, thanks for having a look

Rolv-Apneseth:
> Yes, I noticed that OpenShift issue.  I was hesitating to refer to hit without confirming that it's public.

Yes I checked first too haha

src/cmd/utils.go
```diff
@@ -383,6 +384,39 @@ func getCurrentUserHomeDir() string {
 	return currentUser.HomeDir
 }
 
+func getUserShell() (string, error) {
```
debarshiray: Nitpick: `getCurrentUserShell()` will be consistent with `getCurrentUserHomeDir()` above.

src/cmd/utils.go
```diff
@@ -383,6 +384,39 @@ func getCurrentUserHomeDir() string {
...
+	// $SHELL is supposed to get set during login
+	if userShell := os.Getenv("SHELL"); userShell != "" {
+		return userShell, nil
+	}
+
+	logrus.Debug("$SHELL environment variable was empty - falling back to getent passwd")
+
+	if currentUser == nil {
+		panic("current user unknown")
+	}
```
debarshiray: It will be better to move this to the top of the function, like `getCurrentUserHomeDir()`, so that we enforce the pre-condition under all circumstances.

Otherwise, I am worried that a programmer error can stay hidden in the less used fallback code, until it's hit in the wild or the CI (which is I/O heavy) is run.  Getting immediate feedback in the change code, compile and run cycle will be slightly better.

src/cmd/utils.go
debarshiray: I took the liberty to adjust this myself.

src/cmd/utils.go
debarshiray: Ditto.

debarshiray:
I took the liberty to rebase against `main`, removed the extra `Signed-off-by` trailer, dumped some references to various things (like GDM, systemd, etc.) that set the `SHELL` environment variable for our future selves, obsessed a bit over the debug logs and errors to make them consistent with the rest of the code base, and updated the branch.

Rolv-Apneseth:
> I took the liberty to rebase against `main`, removed the extra `Signed-off-by` trailer, dumped some references to various things (like GDM, systemd, etc.) that set the `SHELL` environment variable for our future selves, obsessed a bit over the debug logs and errors to make them consistent with the rest of the code base, and updated the branch.

All good, thanks for making the changes

debarshiray:
> > I took the liberty to rebase against `main`, removed the extra `Signed-off-by` trailer, dumped some references to various things (like GDM, systemd, etc.) that set the `SHELL` environment variable for our future selves, obsessed a bit over the debug logs and errors to make them consistent with the rest of the code base, and updated the branch.
> 
> All good, thanks for making the changes

Let's merge once the CI turns green.

PR #1779, opened 2026-04-09
cmd, pkg/podman: Turn Image into an interface

debarshiray:
A subsequent commit will switch to unmarshalling the JSON returned from
`podman inspect --format json --type image` directly inside
`podman.InspectImage()` to confine the details within the `podman` package
and make it easier to write unit tests for it.  eg., it requires
tracking changes to the JSON output across different Podman versions.

Unfortunately, the JSON from `podman inspect --format json --type image`
and `podman images --format json` are considerably different and it will
be awkward to use the same implementation of the `json.Unmarshaler`
interface [1] for both.  One option is to have two different concrete
types separately implement `json.Unmarshaler` to handle the differences in
the JSON, and then hiding these concrete types behind an `Image` interface
that provides access to the values parsed from the JSON.

[1] https://pkg.go.dev/encoding/json#Unmarshaler

https://github.com/containers/toolbox/pull/1724

### DIFF
--- a/src/cmd/completion.go
+++ b/src/cmd/completion.go
@@ -150,12 +150,10 @@ func completionImageNames(cmd *cobra.Command, _ []string, _ string) ([]string, c
 	if images, err := podman.GetImages(true); err != nil {
 		logrus.Debugf("Getting all images failed: %s", err)
 	} else {
-		for _, image := range images {
-			if len(image.Names) != 1 {
-				panic("cannot complete unflattened Image")
-			}
-
-			imageNames = append(imageNames, image.Names[0])
+		for images.Next() {
+			image := images.Get()
+			name := image.Name()
+			imageNames = append(imageNames, name)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


debarshiray:
> ## Code Review
>
> [...]
>
> ### src/pkg/podman/image.go (46-53)
> medium
> 
> The `Get()` method currently creates an unnecessary copy of the `imageImages` struct before returning its address. Since `imageImages` implements the `Image` interface via pointer receivers, you can return a pointer to the element in the slice directly. This avoids the overhead of copying the struct and the heap allocation for the local variable. Additionally, for consistency with `Len()`, it's better to include a nil check for the receiver.
> ```go
> func (images *Images) Get() Image {
> 	if images == nil {
> 		panic("called Images.Get() on a nil Images pointer")
> 	}
> 
> 	if images.i < 1 {
> 		panic("called Images.Get() without calling Images.Next()")
> 	}
> 
> 	return &images.data[images.i-1]
> }
> ```

Good point!  Fixed.

debarshiray:
> ## Code Review
> 
> [...]
>
> ### src/pkg/podman/image.go (63-70)
> 
> medium
> 
> The `Next()` method is not nil-safe, which is inconsistent with the `Len()` method at line 55. If `images` is `nil`, calling `Next()` will cause a panic. Adding a nil check ensures that the iterator behaves safely and consistently when used on a nil pointer (e.g., in a `for images.Next() { ... }` loop).
> 
> ```
> func (images *Images) Next() bool {
> 	if images == nil {
> 		return false
> 	}
> 
> 	available := images.i < len(images.data)
> 	if available {
> 		images.i++
> 	}
> 
> 	return available
> }
> ```
> 
> ### src/pkg/podman/image.go (72-74)
> 
> medium
> 
> The `Reset()` method should be made nil-safe for consistency with the `Len()` method. This prevents a panic if `Reset()` is called on a nil `*Images` pointer.
> 
> ```
> func (images *Images) Reset() {
> 	if images == nil {
> 		return
> 	}
> 
> 	images.i = 0
> }
> ```

Good point! Fixed.

PR #1780, opened 2026-04-22
pkg/shell: Preserve error types in shell command execution

DaliborKr:
This is PR 1/10 in a series adding cross-architecture container support using QEMU and binfmt_misc.

## Summary

- Add `RunContextWithExitCode2()` and `RunWithExitCode2()` to the shell package
- These variants preserve original error types (`exec.ErrNotFound`, `exec.ExitError`, `syscall.ENOEXEC`, etc.) instead of wrapping them in generic messages
- Allows its callers to use `errors.Is()` and `errors.As()` to handle specific failure modes
- Needed by cross-architecture code to distinguish "binary not found" from "binary failed"

These new functions are meant to replace its original versions in the future.

### DIFF
--- a/src/pkg/shell/shell.go
+++ b/src/pkg/shell/shell.go
@@ -81,8 +81,49 @@ func RunContextWithExitCode(ctx context.Context,
 	return 0, nil
 }
 
+func RunContextWithExitCode2(ctx context.Context,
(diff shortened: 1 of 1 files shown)


DaliborKr:
recheck

PR #1783, opened 2026-04-23
pkg/architecture: Add architecture support validation

DaliborKr:
This is PR 4/10 in a series adding cross-architecture container support using QEMU and binfmt_misc.

Depends on: #1782 (cmd/create, pkg/utils: Add spinner helpers and distro image matching)
Please review #1782 first. The new commit in this PR is:

- pkg/architecture: Add architecture support validation

## Summary

Before creating or initializing a cross-architecture container, the system must be verified to have the required QEMU emulator and `binfmt_misc` registration in place, and provide a user with an appropriate message. Without these checks, container creation would silently fail or produce broken containers.

Add:
- Host-side validation (`IsArchSupportedOnCreation()`) that locates a statically linked QEMU binary via `exec.LookPath` and verifies that a matching binfmt_misc registration exists. Returns the path to the QEMU binary for use during container initialization

- Container-side validation (`IsArchSupportedOnInitialization()`) that verifies QEMU support from inside the container using the interpreter path passed from the host, with fallback to the standard host-mounted locations under `/run/host/usr/bin/`

- ELF binary inspection (`isStaticallyLinkedELF()`) that confirms the QEMU binary is statically linked, because a dynamically linked one would cause the kernel to attempt to resolve its host-native shared libraries (such as `libc.so`) within the container, resulting in an immediate crash.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -486,19 +486,15 @@ func createContainer(container, image, release, authFile string, showCommandToEn
 		logrus.Debugf("%s", arg)
 	}
 
-	s := spinner.New(spinner.CharSets[9], 500*time.Millisecond, spinner.WithWriterFile(os.Stdout))
-	if logLevel := logrus.GetLevel(); logLevel < logrus.DebugLevel {
-		s.Prefix = fmt.Sprintf("Creating container %s: ", container)
-		s.Start()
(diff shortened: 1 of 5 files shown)


DaliborKr:
recheck

PR #1786, opened 2026-04-24
cmd/utils: Add architecture resolution and cross-arch name handling

DaliborKr:
This is PR 6/10 in a series adding cross-architecture container support
using QEMU and binfmt_misc.

Depends on: #1784 (pkg/skopeo: Extend Image struct and add cross-architecture operations)
Please review #1784 first. The new commits in this PR are:
- cmd/utils: cmd/utils: Add architecture resolution from --arch flag, image tags, and names
- cmd/utils: Update resolveContainerAndImageNames for cross-arch support

## Summary

The --arch flag value and architecture embedded in image tags (e.g., _fedora-toolbox:42-aarch64_) need to be resolved into an architecture ID that the rest of the create/enter/run commands can use. When both sources are present, they must not conflict. The architecture in tag detection applies only to images from distributions that Toolbx explicitly supports (see [1]), to avoid a false architecture approach on custom images where a dash-separated component might not represent an architecture, since there is no standard set regarding preserving architecture in the tag (see detailed explanation at [2]).

Cross-architecture images and containers also need distinct names to coexist with native ones. The naming convention appends the OCI architecture name as a suffix (e.g., image -> _fedora-toolbox:42-arm64_, container -> _fedora-toolbox-43-arm64_). The reasoning behind this convention is described in the Podman discussion at [2].

Add:
- `resolveArchitectureID()` that parses the --arch flag, detects architecture from image tags, and validates consistency if both are set. If none of them is set, it falls back to the host architecture, which represents a native behavior.

- `resolveImageNameWithArchitectureSuffix()` that appends an architecture suffix to supported distro image names when the target differs from the host

- Updated `resolveContainerAndImageNames()` with an archID parameter, auto-suffixing container names for non-native architectures

[1] https://containertoolbx.org/distros/
[2] https://github.com/containers/podman/discussions/27780#discussioncomment-16662213

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -175,7 +175,8 @@ func create(cmd *cobra.Command, args []string) error {
 		containerArg,
 		createFlags.distro,
 		createFlags.image,
-		createFlags.release)
+		createFlags.release,
+		architecture.HostArchID)
 
 	if err != nil {
 		return err
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/briandowns/spinner"
+	"github.com/containers/toolbox/pkg/architecture"
 	"github.com/containers/toolbox/pkg/podman"
 	"github.com/containers/toolbox/pkg/shell"
 	"github.com/containers/toolbox/pkg/skopeo"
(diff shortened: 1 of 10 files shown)


DaliborKr:
recheck

PR #1787, opened 2026-04-24
cmd/create: Integrate cross-architecture support into container creation

DaliborKr:
This is PR 7/10 in a series adding cross-architecture container support using QEMU and binfmt_misc.

Depends on: #1786 (cmd/utils: Add architecture resolution and cross-arch name handling)
Please review #1786 first. The new commits in this PR are:

- cmd/create: Integrate cross-architecture support into container creation
- cmd/create: Rework download prompt flow for cross-arch verification

## Summary

Add the `--arch` flag into the `create` command, connecting all the building blocks from the previous PRs. This allows users to create Toolbx containers for architectures different from the host (e.g., `toolbox create --arch arm64`).

- Add the `--arch` / `-a` flag to the `create` command and integrate the architecture resolution pipeline: use `resolveArchitectureID()` (added in [1]) to determine the target architecture, validate host support via `IsArchSupportedOnCreation()` (added in [2]), and pass the architecture ID to `resolveContainerAndImageNames()` (added in [1]) for architecture-suffixed container naming

- Update `pullImage()` to use `skopeo.CopyOverrideArch()` (added in [3]) instead of `podman.Pull()` for non-native architectures, since Podman does not support pulling foreign architecture images into locally addressable names (see [podman discussion](https://github.com/containers/podman/discussions/27780))

- Rework the image download prompt flow to fetch the full `skopeo.Image` struct instead of just the image size, enabling architecture verification via `VerifyArchitectureMatch()` (added in [3]) before pulling non-native images.

- Use `Image.GetSizeHuman()` (added in [3]) for image size display in the download prompt, replacing the local size computation

- Add a `toolbox-arch` label to created containers to record the target architecture

- Add `createErrorSkopeoNotFound()` and `createErrorImagePull()` helpers in utils.go for consistent error reporting when skopeo is missing or an image pull fails

- Update the `createContainer()` call in `run.go` to pass the default architecture config via `GetArchConfigDefault()`, maintaining existing native-architecture behavior

[1] #1786
[2] #1783
[3] #1784

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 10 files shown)


DaliborKr:
recheck

DaliborKr:
recheck

PR #1788, opened 2026-04-24
cmd/initContainer: Set up QEMU emulation for cross-arch containers

DaliborKr:
This is PR 8/10 in a series adding cross-architecture container support using QEMU and binfmt_misc.

Depends on: #1787 (cmd/create: Integrate cross-architecture support into container creation)
Please review #1787 first. The new commit in this PR is:

- cmd/initContainer: Set up QEMU emulation for cross-arch containers

## Summary

When a cross-architecture container is created, the init-container
entry point must re-register QEMU emulation inside a sandboxed
binfmt_misc mount to enable the C flag, which cannot be safely set
on the host's registration (as explained in [1]).

- Add `--arch` and `--arch-emulator-path` flags to the init-container command, passed from container creation phase.

- When the architecture differs from the host, set up emulation in four steps: 
   1. validate that the host's QEMU emulation is working by running `true` and detecting ENOEXEC via `RunWithExitCode2()` (added in [2]), because host emulation must be functional to perform the sandboxed re-registration in the following step

   2. mount a fresh binfmt_misc filesystem inside the container via `MountBinfmtMisc()` (added in [1])
   
   3. validate architecture support via `IsArchSupportedOnInitialization()` (added in [3])

   4.  register the QEMU interpreter with the C flag via `RegisterBinfmtMisc()` (added in [1])

[1] #1782
[2] #1780
[3] #1783

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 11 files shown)


DaliborKr:
I didn't realize that the architecture package variable `HostArchID`, initialized in `root.go`'s `setUpGlobals()`, doesn't get assigned a value before it is used as a default value for the new flag `--arch` in `initContainer.go`. 

This causes a bug for containers created with Toolbx before cross-architecture support was implemented, because the default value is always 0 (`NotSpecified`), which is evaluated as a non-native container when compared with the already initialized `HostArchID` later in the `initContainer()` function.

I moved the initialization of `HostArchID` directly to the `init()` function in `architecture.go`, which is always evaluated before its dependents (see commit e18be06 in PR #1782).

DaliborKr:
recheck

PR #1805, opened 2026-06-09
Debug ci timed out, do not merge

TristanCacqueray:


debarshiray:
This is where it gets stuck:
```
fedora-rawhide | ...
fedora-rawhide | $ _setup_docker_registry
fedora-rawhide | ...
fedora-rawhide | $$ run podman --root "${DOCKER_REG_ROOT}" run --detach --env REGISTRY_AUTH=htpasswd --env REGISTRY_AUTH_HTPASSWD_PATH="/auth/htpasswd" --env REGISTRY_AUTH_HTPASSWD_REALM="Registry Realm" --env REGISTRY_HTTP_TLS_CERTIFICATE=/certs/domain.crt --env REGISTRY_HTTP_TLS_KEY=/certs/domain.key --name "${DOCKER_REG_NAME}" --privileged --publish 50000:5000 --rm --volume "${DOCKER_REG_AUTH_DIR}":/auth --volume "${DOCKER_REG_CERTS_DIR}":/certs "${IMAGES[docker-reg]}"
fedora-rawhide | ...
fedora-rawhide | $$ run podman login --authfile "${BATS_SUITE_TMPDIR}/authfile.json" --username user --password user "${DOCKER_REG_URI}"
fedora-rawhide | ...
fedora-rawhide | $$ run skopeo copy --dest-authfile "${BATS_SUITE_TMPDIR}/authfile.json" dir:"${IMAGE_CACHE_DIR}"/fedora-toolbox-34 docker://"${DOCKER_REG_URI}"/fedora-toolbox:34
```

The test suite runs its own Docker registry using `podman ... run ...`, and populates it with an old `fedora-toolbox:34` image using `skopeo copy ...`.  The old image is already downloaded to disk in `$IMAGE_CACHE_DIR`, so it should be a local operation.

debarshiray (commented):
I added `--show-output-of-passing-tests --timing --trace` to `playbooks/system-test-runtime-environment-arch-fedora.yaml` and `playbooks/system-test-runtime-environment-ubuntu.yaml` to see why the other jobs aren't getting stuck.

Rolv-Apneseth:
Maybe this is a race where the push starts before the registry container is actually ready. WDYT about introducing some way to wait on the registry? Maybe something like:

```bash
local -i retries
for ((retries = 0; retries < 30; retries++)); do
    if curl -fsk https://localhost:50000/v2/ > /dev/null 2>&1; then
        break
    fi
    sleep 1
done
```

TristanCacqueray:
It looks like the job is hanging on a `skopeo` command, perhaps this could also be reported upstream?

Rolv-Apneseth:
> It looks like the job is hanging on a `skopeo` command, perhaps this could also be reported upstream?

Yeah probably would be a good idea.

These `skopeo` calls should probably use `--command-timeout` at the very least so they can't hang forever.

Rolv-Apneseth:
https://github.com/podman-container-tools/skopeo/issues/2878 is potentially related, with https://github.com/podman-container-tools/container-libs/pull/777 being the proposed solution there

debarshiray:
Sorry, I was away for a week at [DevConf.cz](https://www.devconf.info/cz/) and [Flock](https://fedoraproject.org/flock/2026/).

> These `skopeo` calls should probably use `--command-timeout` at the very least so they can't hang forever.

That's a great idea, @Rolv-Apneseth !  I didn't know about `skopeo --command-timeout ...`, and this Skopeo issue is a good opportunity to try it out.  Do you want to submit a pull request for it?

I see 4 instances of `skopeo copy ...`.  Of those, I think the one in `_pull_and_cache_distro_image()` downloads the image over the network, and the other 3 should be doing local I/O and a timeout definitely makes sense.

For the local copies, we could go with a timeout of 25 seconds.  We use it [elsewhere](https://github.com/containers/toolbox/commit/d3e0f3df06d3f5ac88ae4a5df293fc06820d8700) in Toolbx because it's the default for D-Bus method calls in [most](https://docs.gtk.org/gio/property.DBusProxy.g-default-timeout.html) [implementations](https://www.freedesktop.org/software/systemd/man/latest/sd_bus_set_method_call_timeout.html).

That said, I wonder if that will be enough for most storage hardware.  We recommend and [try](https://github.com/containers/toolbox/commit/571dc9780bdf9a9eac86d4ffa0b518b04155ca79) [not](https://github.com/containers/toolbox/commit/818a22bf80686f36719f79f635ed4e1340be70ed) to use tmpfs for `IMAGE_CACHE_DIR`, and the user's home directory is usually not tmpfs.

I don't know about the one that downloads the image over the network.  If we do use a timeout, then it should probably be longer.  The images are big enough that 25s won't be enough without very good hardware and network bandwidth.  :)

debarshiray:
> Maybe this is a race where the push starts before the registry container is actually ready.

Yes, once in a while on some rare occasions I have seen `_setup_docker_registry()` fail the CI with a complaint about port 50000 being unavailable.  I never figured out if it's the `podman run` that failed because the port was already taken on the CI host, or if (like you said) the `skopeo copy` started pushing the image too soon before the registry was ready to accept it.

> WDYT about introducing some way to wait on the registry? Maybe something like:
> 
> ```shell
> local -i retries
> for ((retries = 0; retries < 30; retries++)); do
>     if curl -fsk https://localhost:50000/v2/ > /dev/null 2>&1; then
>         break
>     fi
>     sleep 1
> done
> ```

Yes, it sounds like a good idea, @Rolv-Apneseth !  We already use a similar loop in `_pull_and_cache_distro_image()` when downloading the images.  Do you want to submit a pull request for it?

I am wondering if we can:

-  use `openssl(1)` for it instead of `curl(1)`, because we already use `openssl(1)` in the test suite and it can avoid the need to update the dependencies in a bunch of places upstream and downstream:
```
openssl s_client localhost:50000 </dev/null
```

- avoid the `curl --insecure`, because we already have the TLS certificate and key in `domain.crt` and `domain.key`

In any case, I slightly prefer the longer options because aliases like `-f` and `-s` are surprisingly difficult to grep for in the manual and elsewhere.  `curl --fail --insecure --silent` is easier for the reader.  :)

Rolv-Apneseth:
Thanks @debarshiray, I've opened #1811. Let's see how it goes.

debarshiray:
@TristanCacqueray It seems like the indefinite hangs of `skopeo copy` went away for some reason, because the `system-test-fedora-44-commands-options` job is now reliably succeeding.

Now, all the jobs using a Fedora Rawhide image seem to be hitting `RETRY_LIMIT`.  I wonder if those images are broken for some reason.  Do you know?  Otherwise, I will try to find out.

debarshiray:
recheck

debarshiray:
recheck

TristanCacqueray:
@debarshiray sorry for the slow reply, I just got back from PTO. In the executor logs we can see the following Ansible error:

```
TASK [add-build-sshkey : Enable access via build key on all nodes] *************
An exception occurred during task execution. To see the full traceback, use -vvv.
The error was: AttributeError: module 'ssl' has no attribute 'PROTOCOL_TLSv1'. Did you mean '.PROTOCOL_TLS' instead of '.PROTOCOL_TLSv1'?
```

Stacktrace is:

Traceback (most recent call last):
- File \\\\\"/tmp/ansible_authorized_key_payload_3zjcqakg/ansible_authorized_key_payload.zip/ansible_collections/ansible/posix/plugins/modules/authorized_key.py\\\\\", line 225, in <module>
- File \\\\\"/tmp/ansible_authorized_key_payload_3zjcqakg/ansible_authorized_key_payload.zip/ansible/module_utils/urls.py\\\\\", line 150, in <module>

Apparently the missing attribute has been removed in https://github.com/ansible/ansible/pull/81880 , so I think we need to update the Ansible running in the executor to support the rawhide python runtime.

TristanCacqueray:
Turns out it is possible to use a more recent ansible as-is: https://github.com/containers/toolbox/pull/1824

debarshiray:
Thanks for fixing and investigating this, @TristanCacqueray !  Much appreciated!  Let's close this in favour of https://github.com/containers/toolbox/pull/1824

debarshiray:
> @debarshiray sorry for the slow reply, I just got back from PTO. In the executor logs we can see the following Ansible error:
> 
> ```
> TASK [add-build-sshkey : Enable access via build key on all nodes] *************
> An exception occurred during task execution. To see the full traceback, use -vvv.
> The error was: AttributeError: module 'ssl' has no attribute 'PROTOCOL_TLSv1'. Did you mean '.PROTOCOL_TLS' instead of '.PROTOCOL_TLSv1'?
> ```
> 
> Stacktrace is:
> 
> Traceback (most recent call last):
> 
>     * File \\"/tmp/ansible_authorized_key_payload_3zjcqakg/ansible_authorized_key_payload.zip/ansible_collections/ansible/posix/plugins/modules/authorized_key.py\\", line 225, in
> 
>     * File \\"/tmp/ansible_authorized_key_payload_3zjcqakg/ansible_authorized_key_payload.zip/ansible/module_utils/urls.py\\", line 150, in

Was it possible for me to see this error?  It might have given me enough clue to make some progress without bothering you.  I wonder if I didn't look closely enough.  :)

TristanCacqueray:
I'm not sure, you'll have to grab the running build from the status page to get its log, but perhaps the stacktraces are not reported. We are looking for a solution to expose more of the ansible logs, but there is always a risk of leaking secrets.

debarshiray:
> I'm not sure, you'll have to grab the running build from the status page to get its log, but perhaps the stacktraces are not reported. We are looking for a solution to expose more of the ansible logs, but there is always a risk of leaking secrets.

Okay, got it!  Thanks.

PR #1811, opened 2026-06-25
Attempt to fix CI timeouts

Rolv-Apneseth:
Attempt to fix CI timeouts with suggestions from #1805.

1. Add `--command-timeout` to `skopeo copy` calls so they can't hang forever
2. When setting up the local Docker registry, add a step to wait and ensure it's ready before proceeding

@debarshiray I chose to go with 10m timeout for the one over the network and 60s for the ones happening locally (in case there's some really slow IO). Did also take the `openssl s_client` approach like you suggested too. WDYT?

### DIFF
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -116,7 +116,9 @@ function _pull_and_cache_distro_image() {
   local -i ret_val
 
   for ((j = 0; j < num_of_retries; j++)); do
-    error_message="$( (skopeo copy --dest-compress \
+    error_message="$( (skopeo copy \
+                          --command-timeout 10m \
+                          --dest-compress \
                           "docker://${image}" \
                           "dir:${IMAGE_CACHE_DIR}/${image_archive}" >/dev/null) 2>&1)"
     ret_val="$?"
@@ -211,7 +213,9 @@ function _setup_docker_registry() {
   assert_success
 
   # Add fedora-toolbox:34 image to the registry
-  run skopeo copy --dest-authfile "${BATS_SUITE_TMPDIR}/authfile.json" \
+  run skopeo copy \
+    --command-timeout 60s \
+    --dest-authfile "${BATS_SUITE_TMPDIR}/authfile.json" \
     dir:"${IMAGE_CACHE_DIR}"/fedora-toolbox-34 \
     docker://"${DOCKER_REG_URI}"/fedora-toolbox:34
   assert_success
@@ -331,7 +335,10 @@ function pull_distro_image() {
   fi
 
   # https://github.com/containers/skopeo/issues/547 for the options for containers-storage
-  run skopeo copy "dir:${IMAGE_CACHE_DIR}/${image_archive}" "containers-storage:[overlay@$TOOLBX_ROOTLESS_STORAGE_PATH+$TOOLBX_ROOTLESS_STORAGE_PATH]${image}"
+  run skopeo copy \
+      --command-timeout 60s \
+      "dir:${IMAGE_CACHE_DIR}/${image_archive}" \
+      "containers-storage:[overlay@$TOOLBX_ROOTLESS_STORAGE_PATH+$TOOLBX_ROOTLESS_STORAGE_PATH]${image}"
 
   # shellcheck disable=SC2154
   if [ "$status" -ne 0 ]; then
@@ -364,6 +371,7 @@ function pull_default_image_and_copy() {
 
   # https://github.com/containers/skopeo/issues/547 for the options for containers-storage
   run skopeo copy \
+      --command-timeout 60s \
       "containers-storage:[overlay@$TOOLBX_ROOTLESS_STORAGE_PATH+$TOOLBX_ROOTLESS_STORAGE_PATH]$image" \
       "containers-storage:[overlay@$TOOLBX_ROOTLESS_STORAGE_PATH+$TOOLBX_ROOTLESS_STORAGE_PATH]$image-copy"
 
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -205,6 +205,23 @@ function _setup_docker_registry() {
     "${IMAGES[docker-reg]}"
   assert_success
 
+  # Ensure the registry is ready
+  local registry_ready=false
+  local -i retries
+  for ((retries = 0; retries < 30; retries++)); do
+    if timeout 5 openssl s_client \
+         -connect "${DOCKER_REG_URI}" \
+         -CAfile "${DOCKER_REG_CERTS_DIR}/domain.crt" \
+         </dev/null >/dev/null 2>&1; then
+      registry_ready=true
+      break
+    fi
+    sleep 1
+  done
+  if ! $registry_ready; then
+    fail "Docker registry at ${DOCKER_REG_URI} did not reach a ready state"
+  fi
+
   run podman login \
     --authfile "${BATS_SUITE_TMPDIR}/authfile.json" \
     --username user \
--- a/test/system/libs/helpers.bash
+++ b/test/system/libs/helpers.bash
@@ -42,7 +42,7 @@ readonly TOOLBX_TEST_SYSTEM_TAGS="${TOOLBX_TEST_SYSTEM_TAGS:-$TOOLBX_TEST_SYSTEM
 # Images
 declare -Ag IMAGES=([arch]="quay.io/toolbx/arch-toolbox" \
                    [busybox]="quay.io/toolbox_tests/busybox" \
-                   [docker-reg]="quay.io/toolbox_tests/registry" \
+                   [docker-reg]="ghcr.io/distribution/distribution:3" \
                    [fedora]="registry.fedoraproject.org/fedora-toolbox" \
                    [rhel]="registry.access.redhat.com/ubi8/toolbox" \
                    [ubuntu]="quay.io/toolbx/ubuntu-toolbox")


test/system/libs/helpers.bash
```diff
@@ -203,6 +205,23 @@ function _setup_docker_registry() {
...
   assert_success
 
+  # Ensure the registry is ready
+  local registry_ready=false
+  local -i retries
+  for ((retries = 0; retries < 30; retries++)); do
+    if openssl s_client \
+         -connect "${DOCKER_REG_URI}" \
+         -CAfile "${DOCKER_REG_CERTS_DIR}/domain.crt" \
+         </dev/null >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
Rolv-Apneseth: Sure, can't hurt

Rolv-Apneseth:
It does seem like an issue on the registry or maybe the node side of things. `skopeo` shouldn't hang, but it seems that the registry only actually finishes the response when `skopeo` hits its timeout.

Updating the registry image to `ghcr.io/distribution/distribution:3` gets us past the setup failure but as you can see from the latest runs, one of the tests hangs on `toolbox create` with the same `fedora-toolbox:34` image.

Test passes fine locally:

```sh
$ bats --filter "custom image that needs an authentication file" test/system/101-create.bats
   test suite: Set up
101-create.bats
 ✓ create: With a custom image that needs an authentication file
   test suite: Tear down

1 test, 0 failures
```

**EDIT:** seems like the test was only failing due to a short timeout, removed that commit and looks like we're good now. I guess I need your thoughts on the registry image, and maybe you know who to contact about the rawhide node failures? @debarshiray

test/system/libs/helpers.bash
```diff
@@ -211,7 +230,9 @@ function _setup_docker_registry() {
   assert_success
 
   # Add fedora-toolbox:34 image to the registry
-  run skopeo copy --dest-authfile "${BATS_SUITE_TMPDIR}/authfile.json" \
+  run skopeo copy \
+    --command-timeout 60s \
```
debarshiray, line 234: Nitpick: the `--command-timeout` [option](https://github.com/podman-container-tools/skopeo/blob/main/docs/skopeo.1.md) is a global `skopeo(1)` option, not an option of the `copy` command.

It doesn't matter when invoking the command in practice, but it can avoid confusing someone reading the code if the option is specified before `copy`.

test/system/libs/helpers.bash
```diff
@@ -331,7 +352,10 @@ function pull_distro_image() {
   fi
 
   # https://github.com/containers/skopeo/issues/547 for the options for containers-storage
-  run skopeo copy "dir:${IMAGE_CACHE_DIR}/${image_archive}" "containers-storage:[overlay@$TOOLBX_ROOTLESS_STORAGE_PATH+$TOOLBX_ROOTLESS_STORAGE_PATH]${image}"
+  run skopeo copy \
```
debarshiray, line 355: Thanks for wrapping that extra-wide line.  It always bothered me.  :)

debarshiray (commented):
Thanks for working on this @Rolv-Apneseth !

It looks like the underlying problem with `skopeo copy` hanging indefinitely went away because `system-test-fedora-44-commands-options` is no longer timing out, and we now have a problem with Fedora Rawhide that's potentially due to broken composes or some other bug because Software Factory is failing to set up the jobs.  Anyway, I think your changes are still worth pursuing.

test/system/libs/helpers.bash
debarshiray: I took the liberty to make this adjustment, and merged the `--command-timeout` commit through https://github.com/containers/toolbox/pull/1815  It gave me the opportunity to trigger the CI a few times to understand the current state.

test/system/libs/helpers.bash
```diff
@@ -42,7 +42,7 @@ readonly TOOLBX_TEST_SYSTEM_TAGS="${TOOLBX_TEST_SYSTEM_TAGS:-$TOOLBX_TEST_SYSTEM
 # Images
 declare -Ag IMAGES=([arch]="quay.io/toolbx/arch-toolbox" \
                    [busybox]="quay.io/toolbox_tests/busybox" \
-                   [docker-reg]="quay.io/toolbox_tests/registry" \
+                   [docker-reg]="ghcr.io/distribution/distribution:3" \
```
debarshiray, line 45: Do you think we should update the image?  We probably should.

As far as I understand, we started using `quay.io/toolbox_tests` because Docker Hub has relatively restrictive rate limits, and we were occasionally [hitting](https://github.com/containers/toolbox/commit/09fb2377272b196f6422f85f54ea2c60be474c22) [them](https://github.com/containers/toolbox/pull/858).  @martymichal will correct me if I am wrong.

If GitHub's Container Registry is generous enough, then maybe we can switch to it?

Does the GitHub Container Registry have a front-end or catalogue, like Docker Hub does?  Something like [hub.docker.com/_/registry](https://hub.docker.com/_/registry) for `docker.io/library/registry`.  I am looking for a canonical source of documentation, metadata and the sources for the image.

I know this sounds stupid, and I just can't figure it out.  If I type `ghcr.io` in Firefox, it takes me to [github.com/features/actions](https://github.com/features/actions).  That page has a lot of things but nothing about the OCI images.

debarshiray:
> It looks like the underlying problem with `skopeo copy` hanging indefinitely went away because `system-test-fedora-44-commands-options` is no longer timing out

The `skopeo` package was last [updated](https://src.fedoraproject.org/rpms/skopeo/commits/f44) for Fedora 44 on the 14th of April.  That doesn't explain why `skopeo copy` was working on the 28th of May and starting hanging indefinitely from the 4th of June.  :(

test/system/libs/helpers.bash
```diff
@@ -203,6 +205,23 @@ function _setup_docker_registry() {
...
   assert_success
 
+  # Ensure the registry is ready
+  local registry_ready=false
+  local -i retries
+  for ((retries = 0; retries < 30; retries++)); do
+    if timeout 5 openssl s_client \
+         -connect "${DOCKER_REG_URI}" \
+         -CAfile "${DOCKER_REG_CERTS_DIR}/domain.crt" \
+         </dev/null >/dev/null 2>&1; then
```
debarshiray, line 215: I was wondering if we could retain the standard error stream from `openssl(1)` and the exit code, because they could be illuminating when dealing with weird failures and [timeout(1)](https://man7.org/linux/man-pages/man1/timeout.1.html) seems to take the trouble to have different exit codes.  However, it gets funky because Bats runs everything with [set -e](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html).

After teaching myself enough about `set -e`, I was able to tweak your change to do that.  I put it up at https://github.com/containers/toolbox/pull/1817 to not clobber your original version and to run the CI on it.  It needs more testing.  Let me know if I missed something.

test/system/libs/helpers.bash
```diff
@@ -203,6 +205,23 @@ function _setup_docker_registry() {
...
+      registry_ready=true
+      break
+    fi
+    sleep 1
+  done
+  if ! $registry_ready; then
+    fail "Docker registry at ${DOCKER_REG_URI} did not reach a ready state"
```
debarshiray, line 222: `fail` is a [wrapper](https://github.com/bats-core/bats-support/blob/main/src/error.bash) [around](https://github.com/bats-core/bats-support/blob/main/src/output.bash):
```ShellSession
echo "..." >&2
return 1
```

... and it loses the exit code from `openssl(1)` and `timeout(1)`.

More subjectively, I have been leaning towards using less of these simpler non-core wrappers from Bats.  Shell scripting has enough weird surprises and adding another layer from Bats adds another layer of potential surprises.  eg., until [recently](https://github.com/bats-core/bats-core/commit/33ffbedd2760e51c74040bb9a45b970a463615da) `run --separate-stderr` didn't work or had undefined behaviour inside `setup_suite()` because [BATS_TEST_TMPDIR](https://bats-core.readthedocs.io/en/stable/writing-tests.html) isn't defined, and we have been blissfully using `run()` in `setup_suite()` all this time.  :)

I like Bats - it's great, and many of its wrappers (both core and non-core) are very useful.  It's just less widely tested and used than Bash itself, so the simpler non-core ones stress me out more than they are worth.  :)

test/system/libs/helpers.bash
Rolv-Apneseth: Ah, yeah I was debating where to put it exactly. Thanks for fixing and merging

test/system/libs/helpers.bash
Rolv-Apneseth: Sure yeah, seemed out of place with the other similar calls

test/system/libs/helpers.bash
Rolv-Apneseth: Well, I thought it was the updated image that had resolved the issue, but I guess it was just lucky timing haha. Maybe we don't need to update it at all.

But those are good questions about the github container registry. As far as I know they don't have a single front-end/catalogue for all the images they host. But, following the "Packages" linked on the [repo](https://github.com/distribution/distribution), I did find this: https://github.com/distribution/distribution/pkgs/container/distribution

As for the rate limits, I don't see anything mentioned [here](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry). I can't find anything else about it online either though.

test/system/libs/helpers.bash
Rolv-Apneseth: Sure, we'll continue that over there then. Maybe this PR is OK to be closed?

test/system/libs/helpers.bash
Rolv-Apneseth: That makes sense yeah. Some of this stuff gets very tricky, may as well keep it simple whenever possible.

test/system/libs/helpers.bash
debarshiray: I see.  Let me see if I can update our copy of the image on `quay.io/toolbox_tests`.

Rolv-Apneseth:
This work was continued (and improved) by @debarshiray in https://github.com/containers/toolbox/pull/1815 and https://github.com/containers/toolbox/pull/1817

Issue #1617, opened 2025-03-07
Permission denied on /dev/usb/lp0


woprandi:
**Describe the bug**
I'm trying to write to `/dev/usb/lp0` from a toolbox container to print but I get `permission denied` (even with root). I tried add myself to `lp` group but it does not help. It works on host. I tried disable SELinux but no changes.

```
⬢ [william@toolbx Documents]$ ls -l /dev/usb
total 0
crw-------. 1 nobody nobody 180, 96  7 mars  09:32 hiddev0
crw-rw----. 1 nobody nobody 180,  0  7 mars  10:05 lp0
```

```
⬢ [william@toolbx Documents]$ echo "test" > /dev/usb/lp0
bash: /dev/usb/lp0: Permission non accordée
⬢ [william@toolbx Documents]$ sudo echo "test" > /dev/usb/lp0
bash: /dev/usb/lp0: Permission non accordée
```

```
⬢ [william@toolbx Documents]$ groups
william lp wheel
```

**Expected behaviour**
I can send bytes to printer file

**Actual behaviour**
Always get `Permission denied`

**Output of `toolbox --version` (v0.0.90+)**
```
❯ toolbox --version
toolbox version 0.1.1
```

**Toolbx package info (`rpm -q toolbox`)**
```
❯ rpm -q toolbox
toolbox-0.1.1-1.fc41.x86_64
```

**Output of `podman version`**
```
❯ podman version
Client:       Podman Engine
Version:      5.3.2
API Version:  5.3.2
Go Version:   go1.23.4
Built:        Wed Jan 22 01:00:00 2025
OS/Arch:      linux/amd64
```

**Podman package info (`rpm -q podman`)**
```
❯ rpm -q podman
podman-5.3.2-1.fc41.x86_64
```

**Info about your OS**
Fedora Silverblue 41

**Additional context**

woprandi:
`--group-add keep-groups` when creating container seems to fix my issue. This flag is not set with `toolbox create`

Issue #1589, opened 2024-11-10
Error: failed to start container fedora-toolbox-41


Kolous:
**Describe the bug**
Toolbox created but is not running. Default configuration Fedora 41. Refering to missing libnvidia-ml.so but I don't have Nvidia in my laptop.

```
DEBU Running as real user ID 1000                 
DEBU Resolved absolute path to the executable as /usr/bin/toolbox 
DEBU Running on a cgroups v2 host                 
DEBU Looking up sub-GID and sub-UID ranges for user fedora 
DEBU TOOLBX_DELAY_ENTRY_POINT is                  
DEBU TOOLBX_FAIL_ENTRY_POINT is                   
DEBU TOOLBOX_PATH is /usr/bin/toolbox             
DEBU Migrating to newer Podman                    
DEBU Toolbx config directory is /home/fedora/.config/toolbox 
DEBU Current Podman version is 5.2.5              
DEBU Creating runtime directory /run/user/1000/toolbox 
DEBU Old Podman version is 5.2.5                  
DEBU Migration not needed: Podman version 5.2.5 is unchanged 
DEBU Setting up configuration                     
DEBU Setting up configuration: file /home/fedora/.config/containers/toolbox.conf not found 
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-41'               
DEBU Image: 'fedora-toolbox:41'                   
DEBU Release: '41'                                
DEBU Resolving container and image names          
DEBU Container: ''                                
DEBU Distribution (CLI): ''                       
DEBU Image (CLI): ''                              
DEBU Release (CLI): ''                            
DEBU Resolved container and image names           
DEBU Container: 'fedora-toolbox-41'               
DEBU Image: 'fedora-toolbox:41'                   
DEBU Release: '41'                                
DEBU Checking if container fedora-toolbox-41 exists 
DEBU Inspecting container fedora-toolbox-41       
DEBU Entry point of container fedora-toolbox-41 is toolbox (PID=0) 
DEBU Inspecting mounts of container fedora-toolbox-41 
DEBU Generating Container Device Interface for NVIDIA 
DEBU Generating Container Device Interface for NVIDIA: Management Library not found: could not load NVML library: libnvidia-ml.so.1: cannot open shared object file: No such file or directory 
[...]

PR #1834, opened 2026-08-06
Add targeted supplementary GID mappings

maxbeckmann:
This change lets a Toolbox container receive selected supplementary host groups through explicit GID mappings. A mapping can be supplied when the container is created:

```console
$ toolbox create --gid-map 971:100000
```

The option is repeatable. The same policy can be made persistent in `toolbox.conf`, including the host-wide `/etc/containers/toolbox.conf`:

```toml
[general]
gid_maps = ["971:100000"]
```

Command-line mappings replace the configured list. If neither interface is used, container creation keeps its existing `--userns keep-id` behaviour.

Each entry maps `HOST_GID` to `CONTAINER_GID` and adds the latter to the container's supplementary groups. Toolbox checks that the host GID is one of the calling user's supplementary groups and rejects duplicate or conflicting mappings. The host administrator must also delegate the host GID through `subgid(5)`; for example:

```console
# usermod --add-subgids 971-971 user
# podman system migrate
```

Mappings are fixed when a container is created, so existing containers must be recreated after changing them.

## Motivation

This addresses the still-open Toolbox supplementary-group problem tracked by [issue #1348](https://github.com/containers/toolbox/issues/1348), including the [VirtualBox failure in #1589](https://github.com/containers/toolbox/issues/1589) and related device-access reports such as the [USB printer in #1617](https://github.com/containers/toolbox/issues/1617). Toolbox recursively bind-mounts the host's `/dev`, but the runtime normally clears the user's host supplementary groups while setting up the container. This makes some devices visible but unusable. Even more severely, it can prevent the container from starting at all when the runtime cannot traverse a group-restricted device tree.

VirtualBox USB devices are one example. They are normally owned by `root:vboxusers`, with `0750` directories and `0660` device nodes below `/dev/vboxusb`. Membership in `vboxusers` grants access on the host but is dropped by crun when starting the container unless instructed otherwise. 

As far as I understood, there is no general Podman-side fix foreseeable. Podman treats this as an authorization and namespace-policy decision, not a group it can transparently infer and expose. Membership in a host group permits a process to retain its existing credential, but it does not authorize namespace root to manufacture that credential later. A real mapping therefore requires prior administrator delegation through `/etc/subgid`.

Podman maintainers have improved the explicit mapping primitives in [PR #18713](https://github.com/containers/podman/pull/18713), but have not adopted automatic mapping of all current supplementary groups. The general [rootless group-mapping issue #13090](https://github.com/containers/podman/issues/13090) remains open, while the exact [VirtualBox report #14284](https://github.com/containers/podman/issues/14284) was closed without a general solution. A more recent attempt to warn and recommend mappings, [PR #29208](https://github.com/containers/podman/pull/29208), is still a draft with changes requested; its review again directs the simple case to crun's `keep-groups`.

Podman therefore already provides the mechanism needed by this change, while Toolbox must provide the opt-in integration and the host administrator must provide the authorization. This proposal replaces `--userns keep-id` with equivalent explicit mappings for the calling user's primary UID and GID, adds only the requested supplementary GID mappings, and passes their container-side GIDs through `--group-add`.

This is deliberately narrower than crun's `--group-add keep-groups`. Retaining all host supplementary credentials would change how the runtime establishes the container's group set and would hence displace container-native groups such as `wheel`, affecting `sudo` and login behaviour. Explicit mappings preserve normal OCI group handling, allow container-native groups to coexist, and make each additional host authority an administrator-approved opt-in through `/etc/subgid`.

## Relationship to existing Toolbox proposals

No existing Toolbox pull request implements targeted host-to-container GID mappings. This change is distinct from the three current proposals around `/dev` and supplementary groups:

- [PR #1647](https://github.com/containers/toolbox/pull/1647) conditionally adds crun's `--group-add keep-groups`. It is still open and stalled after its broader host group set invalidated existing group expectations.
- [PR #1732](https://github.com/containers/toolbox/pull/1732) also uses `keep-groups`, without checking for crun. It has changes requested, testing demonstrated loss of `wheel`/`sudo`, and its author subsequently pointed to PR #1778 instead.
- [PR #1778](https://github.com/containers/toolbox/pull/1778) stops mounting the whole host `/dev` during OCI setup and projects selected entries later. This improves startup robustness and preserves container groups, but deliberately does not grant access to group-restricted host devices and relies on a hard-coded exclusion policy.

The whole-host `/dev` bind originated in [PR #119](https://github.com/containers/toolbox/pull/119) to make dynamic host devices visible. This change retains that visibility model and addresses its missing authorization half. It neither preserves every host group nor creates a new policy for which `/dev` entries Toolbox should expose: it maps only the groups explicitly selected and delegated by the host administrator.

## Why not adjust the device permissions with udev?

A host udev rule is a workable local fix. For `/dev/vboxusb`, for example, it can add traversal permission to the generated directories and grant a named host user access to each device node with an ACL.

That approach changes the permission policy of every offending device tree. It requires device- and vendor-specific rules, per-user ACLs, and careful handling of creation events and rule ordering. Each new group-restricted device family needs another exception.

A host-wide Toolbox configuration that pins mappings for device-controlling groups is a better policy boundary for this use case, in my opinion. It leaves the ownership and modes chosen by the device provider intact, authorizes the mapping once for the controlling group, and applies to every path governed by that group. In other words, it carries the host's existing group-based authorization into selected Toolbox containers instead of bending each device's permissions around the container runtime.


maxbeckmann:
If this approach seems useful, I’d be happy to explore further directions, like automatically forwarding supplementary GIDs delegated to the user creating a toolbox, automatically resolving collisions, and preserving host group names.

PR #1835, opened 2026-08-12
README: Fix the broken star history chart

Dessalines39394:
The star history chart in the README is currently broken and doesn't
render, because the data source it relied on is no longer available.

This points the chart and its surrounding link to a working replacement
so that the README renders correctly again.


debarshiray (commented):
Thanks for noticing the problem and working on it, @Dessalines39394 !

However, it seems that the existing data source is working again when I look at the [README](https://github.com/containers/toolbox) or go to the [link](https://star-history.com/#containers/toolbox&Date).  Am I missing something?  Is `star-history.com` getting phased out in favour of star-history.dera.page, and it's working by chance or not expected to work in future?

debarshiray:
Ping @Dessalines39394

PR #1837, opened 2026-08-24
cmd/list, doc, test/system: Add 'ls' as an alias for 'list'

### LINKED ISSUE #1836: `ls` as an alias of `list`
I'm frustrated having to use `toolbox list` instead of `toolbox ls`

An alias to list - `ls`

**Describe alternatives you've considered**
NA

**Additional context**
NA.

NewtonChutney:
## Summary

Adds `ls` as an alias for the `toolbox list` command, since many users
reach for `ls` out of habit from everyday Unix usage.

- `src/cmd/list.go`: add `ls` to `listCmd.Aliases` (Cobra's built-in
  alias mechanism), so flags (`--containers`/`-c`, `--images`/`-i`),
  error handling and shell completion (`GenBashCompletionV2`,
  `GenFishCompletion`, `GenZshCompletion`) all keep working exactly as
  they do for `list`, with no other code changes needed.
- `doc/toolbox-list.1.md`, `doc/toolbox.1.md`: document the alias.
- `test/system/102-list.bats`, `test/system/002-help.bats`: add smoke
  and unknown-flag tests for `toolbox ls`, mirroring existing `list`
  tests.

Fixes #1836


src/cmd/list.go
```diff
@@ -38,6 +38,7 @@ var (
 
 var listCmd = &cobra.Command{
 	Use:               "list",
+	Aliases:           []string{"ls"},
```
debarshiray, line 41: The `ls` alias is not expected to show up in the shell completion, right?  ie., `toolbox l<TAB><TAB>` will result in `toolbox list `.  If so, then perfect, because it would be slightly annoying to choose between `list` and `ls`.

I am always nervous when looking at the shell completions because of the complex machinery involved.

debarshiray (changes requested):
Thanks for working on this, @NewtonChutney !  I have been thinking about this myself, because `ls(1)` is such a common command, and `podman image ls` is an alias to `podman image list` and `podman images`.

At first glance, the code changes look good to me.

Could you please correct your Git authorship information with your full name and a real email address?  For what it's worth, projects as diverse as [GCC](https://gcc.gnu.org/dco.html), [GnuPG](https://www.gnupg.org/faq/HACKING.html), [Linux](https://www.kernel.org/doc/html/latest/process/1.Intro.html), [Moby](https://github.com/moby/moby/blob/master/CONTRIBUTING.md) and [Podman](https://github.com/containers/podman/blob/main/CONTRIBUTING.md) don't allow anonymous or pseudonymous contributions.

src/cmd/list.go
NewtonChutney: Ahh, let me test and confirm. And if `ls` does show up, if it is similar to how `podman image ls` is an auto completion, it shouldn't be an issue? 
And I think I corrected my authorship? Lemme verify

doc/toolbox-list.1.md
```diff
@@ -1,11 +1,13 @@
 % toolbox-list 1
 
 ## NAME
-toolbox\-list - List existing Toolbx containers and images
+toolbox\-list, toolbox\-ls - List existing Toolbx containers and images
```
debarshiray, line 4: I wouldn't add `toolbox-ls` here.  There's limited space here in the `NAME` section because it's supposed to have a brief one-line description.  If a command has a longer name or has more aliases, then it won't fit.  eg., the manual for `podman image ls` doesn't include the alias here.

We need to find a way to make this work:
```ShellSession
$ man toolbox-ls
```

One way of doing that is to use the `.so` [request](https://www.gnu.org/software/groff/manual/groff.html.node/Host-System-Service-Access.html) like [Podman](https://github.com/podman-container-tools/podman/blob/main/docs/source/markdown/links/podman-image-ls.1) does.

debarshiray (changes requested):

src/cmd/list.go
debarshiray: As far as I can make out, `podman image l<TAB><TAB>` includes only `list`, not `ls`.

NewtonChutney: Ahh.. Okay then

So, it's with other tools and short flags that there's a shortcut

debarshiray:
Ping @NewtonChutney

src/cmd/list.go
NewtonChutney: Confirmed with the updated binary: `toolbox __complete l` offers only `list`, not `ls`. Once the alias is typed, `toolbox __complete ls --c` correctly offers `--containers`. So the alias does not add a duplicate top-level completion choice.

doc/toolbox-list.1.md
NewtonChutney: Updated as suggested. The `NAME` section now contains only `toolbox-list`; `doc/toolbox-ls.1` contains `.so man1/toolbox-list.1` and is installed as a raw man page through `doc/meson.build`. I verified a temporary Meson install and `man toolbox-ls` renders the `toolbox-list(1)` page.

NewtonChutney:
@debarshiray The requested authorship and man-page changes are now pushed, and I replied to both review threads with verification details. Please do review now.
I've reviewed the changes (done with LLMs), and find them acceptable.

